PR #40, opened 2016-12-30
chore(dependencies): Removes unnecessary optional @types/graphql dependency

alexgorbatchev:
I don't believe `optionalDependency` is necessary. It installs `@types/graphql` which apparently conflicts with `graphql` typings causing issues like so:

```
$ tsc -p .
...
node_modules/@types/graphql/index.d.ts(126,9): error TS2484: Export declaration conflicts with exported declaration of 'IntrospectionNonNullTypeRef'
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,9 +105,6 @@
     "tslint": "^4.0.2",
     "typescript": "^2.1.4"
   },
-  "optionalDependencies": {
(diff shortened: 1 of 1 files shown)


dotansimha:
Great. Thanks @alexgorbatchev !

PR #81, opened 2017-03-21
Fix for multiple extends in typescript generator

Anthonyzou:
Hi,

This is a fix for the typescript generator when an entity extends multiple other entities, a comma wasn't added previously, this PR fixes that.


generators/typescript-multiple-files/model.handlebars
```diff
@@ -1,3 +1,4 @@
+/* tslint:disable */
```
dotansimha: Why do we need to add this line? the linter should not be run on `handlebars` file

Anthonyzou: Its just so the resulting typescript file doesn't get linted not the handlebars file. It would probably be better if there was an option to add this line through the CLI.

dotansimha: Sure. makes sense.

dotansimha:
Thanks @Anthonyzou !! 
I will release a new version soon.

Anthonyzou:
When will the update be included? Thanks!

PR #109, opened 2017-06-24
update dev-test type files

zephraph:
The dev-test files were out of date. They're a good example of the types generated by the library so I think it's important to keep them updated. 

I'd suggest adding a script to do so automatically. I just manually ran the commands for now. 

BTW this library is GREAT! It's the best one I've found so far so awesome job!


dotansimha:
Thanks @zephraph !

PR #113, opened 2017-06-25
Prevent multi-file outputs from having duplicate imports

### LINKED ISSUE #112: Multi-file type files include duplicate imports
image

It looks like it currently just iterates through the special imports and adds an entry for every item that occurs in the body. Just need to dedupe the list.

zephraph:
This resolves #112.


dotansimha:
Great! Thanks @zephraph

PR #116, opened 2017-06-28
Fixes #111: Prevent primitives from being output like other models

### LINKED ISSUE #111: Multi-file outputs make a file and export for built in graphql types
image

There's two empty files to go along with that.

zephraph:
I'm not actually sure this is the best way to go about doing this. Maybe the logic should've went in the handleType method? Regardless, this gets the desired result. 

By the way, I think some of these methods really should be broken down to be more testable.


dotansimha:
Makes sense. 
Thanks @zephraph

Issue #96, opened 2017-05-13
Mutation with only a string return.


DJWassink:
Heya, awesome work on this project!

I really love getting typescript defenitions for my queries but I am running into a issue with a single mutation which only returns a `String!`

My mutations schema look like this:
```js
type Mutation {
    authenticate(email: String!, password: String!): String!
    register(email: String!, userName: String!, password: String!, password2: String!): User!
}
```

And in my react component:
```js
//in register.tsx
const RegisterMutation = gql`
    mutation register($email: String!, $userName: String!, $password: String!, $password2: String!) {
        register(email: $email, userName: $userName, password: $password, password2: $password2) {
            id
        }
    }
`;

//in login.tsx
const AuthenticateMutation = gql`
    mutation authenticate($email: String!, $password: String!) {
        authenticate(email: $email, password: $password)
    }
`;
```

The `RegisterMutation` is returning a user object where I am selecting the `id` only and the defenitions gets generated correctly but for the `AuthenticateMutation` I am not selecting anything because it only returns a string.

running the command below will result in a crash
```
 gql-gen --file schema.json --template typescript --out ./typings/b.ts ./src/account/login.tsx
```

```
TypeError: Cannot read property 'fields' of undefined
    at /home/dirkjan/.config/yarn/global/node_modules/graphql-code-generator/dist/handlers/inner-models-builer.js:67:29
    at Array.forEach (native)
    at Object.exports.buildInnerModelsArray (/home/dirkjan/.config/yarn/global/node_modules/graphql-code-generator/dist/handlers/inner-models-builer.js:11:47)
    at Object.exports.handleOperation (/home/dirkjan/.config/yarn/global/node_modules/graphql-code-generator/dist/handlers/operation-handler.js:51:49)
    at /home/dirkjan/.config/yarn/global/node_modules/graphql-code-generator/dist/engine/codegen.js:20:56
    at Array.forEach (native)
    at Object.exports.prepareCodegen (/home/dirkjan/.config/yarn/global/node_modules/graphql-code-generator/dist/engine/codegen.js:17:30)
    at Transform (/home/dirkjan/.config/yarn/global/node_modules/graphql-code-generator/dist/engine/transform-engine.js:27:29)
Error: TypeError: Cannot read property 'fields' of undefined
```

I guess it has something to do with the fact that it returns a string instead of a object?

DJWassink:
For now I have solved the issue by returning a type only containing a single string, so in my schema I now have added:

[...]

Issue #84, opened 2017-03-23
input types are not emitted (typescript, noSchema===true)


thunder7553:
given a query:

```graphql
query getUsers($pager: PagerInput) {
    users(pager: $pager) {
        docReferenceId
        docType
        displayName
    }
}
```

the output is 

```javascript
   export type DOCTYPE = "user" | "subject";
   export type ROLE = "admin" | "subjectcoordinator" | "author" | "superuser" | "evaluator" | "translator";
   export namespace GetUsersQuery {
      export type Variables = {
         pager: PagerInput | null;
      }
     export type Result = {
        users: Array<Users>;
     } 
     export type Users = {
       docReferenceId: string;
       docType: DOCTYPE;
      displayName: string;
    } 
   }
```

The PagerInput is missing.

dotansimha:
@thunder7553 
Done in 0.8.0 :)

PR #117, opened 2017-06-28
WIP: Refactor package

dotansimha:
Progress: 99.999%

The goals are: 
- [x] Move codegen core to a package and don't use `fs` 
- [x] Move generators to separate package
- [x] Refactor directories structure and use Lerna to manage package's dependencies 
- [x] Refactor GraphQL Schema template context builder and object
- [x] Handle fields and types in a better way  (related to https://github.com/dotansimha/graphql-code-generator/issues/96 and https://github.com/dotansimha/graphql-code-generator/issues/84) - my idea is to use expose the schema template context the same way GraphQL treat it's types (separate arrays for enum, type, input, union, interface, scalar)
- [x] Treat Query and Mutation just like any other type
- [x] Fix CI
- [x] Refactor operations builder with flatten feature
- [x] Reimplement `typescript-single-file`
- [x] Reimplement imports builder for multiple files 
- [x] Reimplement `typescript-multi-file`
- [x] Move and refactor CLI util to a separate package (use fs only here)
- [x] Add support for input/output of a whole project (to generate apps)
- [x]  Fix lerna dependencies 
- [x] Add support for magic comments to easier integration in existing project (`/*gqlgen {{#each types }} */`
- [x] Move generators to separate packages with compilation package 
- [x] Add support for directives (https://github.com/graphql/graphql-js/pull/746)
- [ ] Update README
 
After merging to master:
- [ ] Reimplement `flow-single-file`
- [ ] Reimplement `swift-single-file`
- [ ]  100% code coverage for core & compiler packages
- [ ] Blog post

Ideas:
- [x] Consider implementing the flatten types map in a better way - maybe just add a reference inside the template context to the actual type along the type name.


dotansimha:
@DxCx  @zephraph any thoughts?

zephraph:
Whew... this is a huge change. Okay, I'll comment as I can go through it...

dotansimha:
@zephraph my goal is to simplify the schema handler first. 
I am almost done, and you can find it here: https://github.com/dotansimha/graphql-code-generator/blob/refactor/all/packages/graphql-codegen-core/src/schema/schema-to-template-context.ts
Also, I changed the schema template context object to be: https://github.com/dotansimha/graphql-code-generator/blob/refactor/all/packages/graphql-codegen-core/src/types.ts#L55-L62

This is much simpler implementation for the Schema handling, the difference is that I'm creating an array for each GraphQL base type (scalar/enum/type/input...), instead of a single array for all of them together. 

I also separated it to 3 packages: 
- core (GraphQLSchema and operations string => template context object)
- generators (combines template and template context, also in charge of converting primitives to language primitives, eg. `String` => `string` and `Int` to `number` in case of TS)
- cli (Find, read, introspect and execute generator + core)

zephraph:
It might be a good idea to generate a copy of dev-test from master (on all supported variations) and have a test just for comparing final generated outputs.

dotansimha:
@zephraph I'm done with the scheme generation for typescript, and it looks like the result file is the same as the current implementation - so we are good so far :)

zephraph:
Cool! I'll run through the updates. There's a few things I want to test.

dotansimha:
@zephraph do you have good use cases for complicated operations (fragment/mutation/query)? I want to use it to write tests before implementing.

zephraph:
I don't at the moment but we can put one together.

zephraph:
It's really starting to come together!

dotansimha:
@zephraph almost there!! my next goal is to add the ability to generate a whole project with custom templates, something like this:
```
my-project/
    - index.js
    - router.js
    - graphql/
             - feed.graphql
    - wrappers/
              - operation.js.template
    - data-tables/
              - table.js.type.template
```

and you have a `type` called `MyType` on your GraphQLSchema.

The output will be:
```
    - index.js
    - router.js
    - graphql/
             - feed.graphql
    - wrappers/
              - operation.js.template
              - feed.query.js
    - data-tables/
              - table.js.type.template
              - mytype.table.js
```

This way, users can create a project and run the generator over a project's directory and generate anything they need that related to GraphQL (generate UI, HOC wrappers for React, ready-to-use `gql` wrappers or anything else :) )

What do you think?

PR #142, opened 2017-08-25
feat(typescript): emit nullable fields as optional

### LINKED ISSUE #138: Nullable fields with typescript
I'm having a recurring issue with how nullable fields are emitted in typescript. The current way is to use `field: <type> | null` which I suppose is theoretically correct, but in practice it doesn't always play well since typescript will complain that some nullable field is missing (because it's undefined and not actually null). The more typescriptish way would be to make it an optional field like `field?: <type>`, but I'm not sure if that would have other implications.

altschuler:
Fixes #138


altschuler:
Couldn't get tests to run locally, will fix and patch PR

altschuler:
Should be good now, I don't think the code duplication tests failing are regressions?

Aleksion:
Anything blocking this from being merged?

altschuler:
@dotansimha would you be able to have a look at this PR? :)

dotansimha:
Looks good. Thank you @altschuler !

PR #161, opened 2017-11-08
[Do not merge] Initial flow support

blackxored:
This is very barebones initial Flow support, mostly copied from the TS single file generator. 
I was able to generate a file without Flow complains for the 3 schemas in dev-test and 2 custom ones. Haven't been able to test the client schema generation part yet. 

I would very much like some help getting a better test setup to ensure all of this is correct, currently is really hard to tests (i.e either write all the tests by hand or hack the cli, in addition, provided schemas are not complex enough to ensure correctness it seems).

- [x] Add single file to list of generators
- [ ] Add multiple file
- [x] Type support
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/graphql-codegen-generators/src/flow-single-file/config.ts
+++ b/packages/graphql-codegen-generators/src/flow-single-file/config.ts
@@ -1,0 +1,28 @@
+import * as index from './template.handlebars';
+import * as type from './type.handlebars';
+import * as schema from './schema.handlebars';
+import * as documents from './documents.handlebars';
+import * as selectionSet from './selection-set.handlebars';
+import { EInputType, GeneratorConfig } from '../types';
+
+const config: GeneratorConfig = {
+  inputType: EInputType.SINGLE_FILE,
+  templates: {
+    index,
+    type,
+    schema,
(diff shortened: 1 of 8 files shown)


dotansimha:
That's great, @blackxored !! Thank you! 
I will review is soon :)

dotansimha:
Looks good @blackxored . 

How was the experience of writing Handlebars template? Because I'm adding a new feature that will allow to add custom output processors (per generator) that will allow us to write custom JS code to generate the template (instead of Handlebars).

Also, do you think that there are missing Handlebars helpers that could make it simpler? 

Thank you!

blackxored:
I think definitely being able to write your code in JS for some advanced scenarios is gonna be needed at some point. However, I do like the simplicity of Handlebars templates, since there's basically such a minimal API so people can get started really easy. 
The scenarios I have in mind (generating UI, backend, etc) would probably need that more likely than not.

dotansimha:
@blackxored any chance we can complete this? 😊

dotansimha:
Closing for now.

PR #173, opened 2017-12-12
Feature/add require option

### LINKED ISSUE #171: Allow the --export option to accept a typescript file by adding --require option
If `gql-gen` had an option for `--require` that it would forward to `node` then the `--export` option should be able to directly handle `.ts` files by requiring [ts-node](https://github.com/TypeStrong/ts-node)'s register module.

It would be something like this:

```
gql-gen --require ts-node/register --export ./schema-object.ts
```

To get more info about the `--register` option, check `node --help`. Also check [here](https://github.com/TypeStrong/ts-node#mocha) for an example of using ts-node with mocha.

jonaskello:
This pull request adds a `--require` option similar to the same option in `node`. It will enable the `--export` option to accept a `.ts` file by doing `--require ts-node/register`.

It also adds a test case for the `--export` option (I could not find any test for it). To make this work I needed to have `graphql` only as peer dependency so all packages uses the same version. Before it was specified both as peer dependency and devDependency in `graphql-codegen-core`.

I tried very hard to make a test case for ts-node/register but in the end it was not just possible because of the way `jest` loads modules. Since `ts-node/register` patches `require` you need to use the real patched `require` to test it. But `jest` always gives all files its own fake `require` and there is no way to turn it off. See [here](https://stackoverflow.com/questions/46433678/specify-code-to-run-before-any-jest-setup-happens) for some more info.

Fixes #171.

### DIFF
--- a/dev-test/githunt/schema.js
+++ b/dev-test/githunt/schema.js
@@ -1,0 +1,3 @@
+const introspectionSchema = require("./schema.json");
+const GraphQL = require("graphql");
+module.exports.default = GraphQL.buildClientSchema(introspectionSchema);
--- a/packages/graphql-codegen-cli/package.json
+++ b/packages/graphql-codegen-cli/package.json
@@ -78,6 +78,9 @@
     "tslint": "^5.4.3",
     "typescript": "^2.4.1"
   },
+  "peerDependencies": {
+    "graphql": "^0.8.0 || ^0.9.0 || ^0.10.0"
+  },
   "dependencies": {
     "commander": "^2.11.0",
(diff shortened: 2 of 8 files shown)


jonaskello:
The build breaks with: `src/utils/get-directives.ts(2,36): error TS2307: Cannot find module 'graphql/execution/values'.`. There are no typings for that file. Not sure if the build actually worked when I forked master?

dotansimha:
@jonaskello Thanks! I'll merge and resolve the typings issue.

Amazing work!

PR #175, opened 2017-12-14
Update enumeration handling to emit Typescript string-backed enums

darthtrevino:
Recent versions of Typescript allow for string-values based enumerations. This allows enums that looks like this: 

```ts
enum Color {
  RED='red',
  GREEN='green',
  BLUE='blue',
}
```

The benefit of this in client code is that you can reference enumeration values as symbols instead of just relying on raw string values.

e.g. 

```ts
const color: Color  = 'red'
// becomes
const color = Color.RED
```


dotansimha:
That's great! Thanks @darthtrevino !

PR #177, opened 2017-12-27
Accept more formats for the --export flag

### LINKED ISSUE #170: How does the --export flag work?
I cannot find docs about exactly how to use it.

I tried with a file that exports a schema like this:

**schema.ts**

```ts
import gql from "graphql-tag";

export const schema = gql`
  type User {
    id: ID!
    first_name: String!
    last_name: String!
  }
`
```

And I run this:

```
gql-gen --export schema.js --template typescript --out ./types.ts ./src/**/*.ts
```

And I get this error:

```
Loading GraphQL Introspection from JavaScript ES6 export: schema.js...
TypeError: schema.getTypeMap is not a function
```

jonaskello:
This is a PR to be able to accept more formats for the `--export` flag. It will now accept all of these formats:

1. GraphQLSchema object
2. Text (the GraphQL schema language textual format)
3. AST (the GraphQL schema language textual format parsed into an AST)
4. Introspection JSON (result returned by introspection query)

The PR adds tests for all of the above cases.

In order for the tests to work there can only be one instance of the `graphql` package imported or the `instanceof` checks [here](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-core/src/schema/schema-to-template-context.ts#L61) will fail. They will fail even if the exact same version of the `graphql` package is imported by all packages. In order for the `instanceof` operator to work, all packages must import the exact same classes from the exact same files (this is how the `instanceof` operator works). In order to make this work this PR adds the `--hoist` flag to `lerna bootstrap` (read [docs here](https://github.com/lerna/lerna/blob/master/doc/hoist.md)). This will cause all common packages to be installed in the top-level node_modules, which will make sure all packages import the same `graphql` package. It will also speed up the bootstrapping, and save diskspace as common dependencies are only installed once. (It works similar to yarn workspaces).

Fixes #170.

### DIFF
--- a/dev-test/test-schema/schema-ast.js
+++ b/dev-test/test-schema/schema-ast.js
@@ -1,0 +1,2 @@
+const schemaFormats = require("./schema-formats");
+module.exports.schema = schemaFormats.schemaAst;
--- a/dev-test/test-schema/schema-formats.js
+++ b/dev-test/test-schema/schema-formats.js
@@ -1,0 +1,22 @@
+/**
+ * There are four formats the schema can be in:
+ * 1. GraphQLSchema object
+ * 2. Text (the GraphQL schema language textual format)
+ * 3. AST (the GraphQL schema language textual format parsed into an AST)
+ * 4. Introspection JSON (result returned by introspection query)
+ * 
+ * This file imports the textual and introspection json files and
+ * exports all four formats to be used in tests.
+ */
+
+const GraphQL = require("graphql");
+const fs = require('fs');
+const path = require('path');
+const schemaText = fs.readFileSync(path.join(__dirname, "schema.graphql"), 'utf8');
+const schemaObject = GraphQL.buildSchema(schemaText);
+const schemaAst = GraphQL.parse(schemaText);
+const schemaJson = require("./schema.json");
+module.exports.schemaText = schemaText;
+module.exports.schemaAst = schemaAst;
+module.exports.schemaObject = schemaObject;
+module.exports.schemaJson = schemaJson;
--- a/dev-test/test-schema/schema-json.js
+++ b/dev-test/test-schema/schema-json.js
@@ -1,0 +1,2 @@
+const schemaFormats = require("./schema-formats");
+module.exports.schema = schemaFormats.schemaJson;
--- a/dev-test/test-schema/schema-object.js
+++ b/dev-test/test-schema/schema-object.js
@@ -1,0 +1,2 @@
+const schemaFormats = require("./schema-formats");
+module.exports.schema = schemaFormats.schemaObject;
--- a/dev-test/test-schema/schema-text.js
+++ b/dev-test/test-schema/schema-text.js
@@ -1,0 +1,2 @@
+const schemaFormats = require("./schema-formats");
+module.exports.schema = schemaFormats.schemaText;
--- a/packages/graphql-codegen-cli/src/loaders/schema-from-export.ts
+++ b/packages/graphql-codegen-cli/src/loaders/schema-from-export.ts
@@ -1,9 +1,18 @@
 import * as fs from 'fs';
-import { GraphQLSchema } from 'graphql-codegen-core';
 import * as path from 'path';
+import {
+  GraphQLSchema,
+  DocumentNode,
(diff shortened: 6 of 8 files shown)


jonaskello:
The travis build is failing with:

```
lerna info version 2.5.1
lerna ERR! --hoist is not supported with --npm-client=yarn, use yarn workspaces instead
```

However, "it works on my machine" :-). I did some checking and locally lerna is installed as 2.0.0 for me locally, but in travis as 2.5.1. I installed the root package.json using `yarn install` on my machine, but the travis build uses `npm install`, this seems to be the reason why I get a different lerna version. Using `npm install` locally I also get 2.5.1. I guess the best path forward is to setup workspaces as is suggested by the error message. I have worked with yarn workspaces, and lerna, but not used them together. I'll give it a try and update the PR if it is successful.

jonaskello:
I switched to using yarn for the travis build. When using workspaces it is enough to just run `yarn install`once to install for all packages.

Also I switched to running the tests directly with yarn instead of lerna because the output was hard to read when using lerna. Using lerna is nicer though since it automatically runs all scripts in all packages.

It seems the tests fail because there is a whitespace difference between the expected and generated code. Not sure why this is happening, the tests that are failing are not related to the --export flag.

jonaskello:
The changes for this feature (adding formats to --export) is quite simple, the problem is that for the tests to work there needs to be only one copy of the graphql package at runtime due to how the instance of operator works.

Here is a [comment](https://github.com/graphql/graphiql/issues/58#issuecomment-226661181) that mentions the problem. And [another](https://github.com/graphql/graphiql/issues/58#issuecomment-303228898) that suggests using --hoist with lerna.

Since --hoist does not work with yarn as npmClient in lerna, the only solution I can think of is using yarn workspaces like I try to do in this PR. Not sure if this is the right way to proceed though. I'll put this on hold until someone has time to review and suggest a path forward.

jonaskello:
Ran across the multiple `graphql` package problem again today. There is actually a pretty clear error message that says what I tried to explain above:

```
 |   Error: Cannot use GraphQLSchema "[object Object]" from another module or realm.
 |
 |   Ensure that there is only one instance of "graphql" in the node_modules
 |   directory. If different versions of "graphql" are the dependencies of other
 |   relied on modules, use "resolutions" to ensure only one version is installed.
 |
 |   https://yarnpkg.com/en/docs/selective-version-resolutions
 |
 |   Duplicate "graphql" modules cannot be used at the same time since different
 |   versions may have different capabilities and behavior. The data from one
 |   version used in the function from another could produce confusing and
 |   spurious results.
```

dotansimha:
@jonaskello I replace Lerna hoist with Yarn Workspaces. does this makes it easier now?

jonaskello:
I'll have a look at it soon (maybe tonight). Perhaps I'll close this one and do a new PR since some of the changes here are about yarn workspaces.

dotansimha:
Awesome! Thank you @jonaskello !

jonaskello:
The yarn workspaces changes were mostly in separate commits so I managed to drop those commits and then do a rebase. The changes on master regarding yarn workspaces really helped, and I think this PR now is now ready for review again :-).

dotansimha:
@jonaskello Looks good :D 
I'm thinking about merging `--url`, `--file` and `--export` into a single flag called `--schema`. 
What do you think about that?

jonaskello:
Yes I think that makes sense. A flag such as `--export` does not really relate to loading a schema so it is not very intuitive. How would you propose to distinguish between the different cases?

```
--schema http://myserver/graphql // introspection from server
--schema my-local-introspection.json // local introspection json
--schema my-export.js // from export, either GraphQLSchema, Text, AST, or introspection JSON
```
I have not thought about it a lot but here is some ideas:

* For the url, we can just check if it is a valid URL.
* For the local files we can either:
  * Try to parse as JSON, and fallback to loading as a JS/TS ES6 export.
  * Use the file-extension and require that JSON file has .json extension.

dotansimha:
@jonaskello yeah exactly. So let's merge this one and later I'll add `--schema`. 

Thank you!

Issue #169, opened 2017-12-08
Handle lists and non-null properly


michi-covalent:
Hello,

Right now gql-gen doesn't check if `!` (non-nullable) applies to a list or elements inside the list:

https://github.com/dotansimha/graphql-code-generator/blob/f3f6e0aa4d482bf407e507267424e363d1d32758/packages/graphql-codegen-core/src/schema/resolve-type.ts#L10

As a result, the list is considered required even when it isn't:

GraphQL type | Expected Typescript Type | Actual Generated Type
------------ | ------------- | ------------- 
`[String]` | `(string \| null)[] \| null` | `string[] \| null`
`[String]!` | `(string \| null)[]` | `string[]`
`[String!]` | `string[] \| null` | `string[]`
`[String!]!`| `string[]` | `string[]`

More generally, GraphQL allows arbitrary nesting of list and non-null modifiers according to http://graphql.org/learn/schema/#lists-and-non-null .

shinyaohira:
I ran into the same problem.

FredyC:
Yea just encountered this too. Should be pretty simple fix seeing that code, something like `lastIndexOf(']!')` should be sufficient in my opinion.

Unfortunately, it feels like @dotansimha is rather busy lately, he is not responding to issues much. Looks like custom fork will be necessary, but that's going to be a pain considering monorepo structure.

FredyC:
I made the PR. It's a working solution, I have a rather big schema and nothing was broken by that change.

shinyaohira:
@FredyC Hi, I'd like to try your PR. Could you tell me a little bit about how to use it in my project.

package.json
```
"graphql-code-generator": "git://github.com/FredyC/graphql-code-generator.git#FredyC-patch-1",
```

I executed `yarn install`, but gql-gen command was not created in `./node_modules/.bin/`.

FredyC:
@shinyaohira I don't think you can do that because of a monorepo structure which includes multiple packages that need to be built. However, looks like that PR will be merged soon, just be a little patient :)

dotansimha:
Fixed in 0.8.16. Thanks @FredyC !

dotansimha:
Also, @shinyaohira, @FredyC is right, you can't use it from git. But you can compile it locally and try to use it this way...

shinyaohira:
@FredyC @dotansimha Thank you 🙌

PR #186, opened 2018-03-12
Handle lists and non-null properly

FredyC:
Quick and dirty fix for #169. I will try to publish custom fork to NPM soon unless @dotansimha will respond here. Anyone feeling for adding some tests for this?


FredyC:
Looks like there is actually wrong test written. In this case, a field should not be required. Should I change that or am I missing something here?

https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-core/tests/transform-fields.spec.ts#L131

dotansimha:
@FredyC Thank you! 
I think you are right, the test is incorrect. can you please change and, and then I'll merge and release a new version?

dotansimha:
Thanks @FredyC ! Great!

PR #194, opened 2018-03-13
Add prettier

jonaskello:
This PR adds automatic formatting with the [prettier](https://prettier.io) code-style. Some of the changes includes:

* Add prettier as a devDependency of the root workspace. Since this is a monorepo, IMO it is better to maintain one version of dev tools at the root workspace (by using `yarn add --dev -W <package>`). This is also recommended by the owner of lerna (see [here](https://github.com/lerna/lerna/issues/1168#issuecomment-369279946)). This PR only adds the new dev tools to the root but IMO typescript and other dev deps should also be moved later.

* Add lint-staged so that staged files will be linted with tslint and formatted with prettier before being committed. This ensures there is no files that does not pass linting or formatting committed.

* Move tslint.json to the top level. All packages/*/tslint.json were identical and since this is a monorepo a big win is to maintain only one config of dev tools such as tslint.

After this is merged, I can make a separate PR to update the formatting of all existing files. This can be done by running this commmand: `yarn prettier packages/*/src/**/*.{ts,tsx} --write`. It only has to be done once since lint-staged will ensure formatting of all future changes.


jonaskello:
To be more specific about the referenced [comment](https://github.com/lerna/lerna/issues/1168#issuecomment-369279946) of @evocateur (owner of lerna) about moving dev deps to the root, this is the part I wanted to highlight:

> All of your devDependencies should be moved to the root package.json, regardless of whether or not you hoist the dependencies (lerna bootstrap --hoist).

jonaskello:
In some cases when adding prettier there can be conflict between existing tslint rules and how prettier does formatting. I tried to config prettier to respect the existing tslint rules (single quote to be specific). Also I did a trial run where I reformatted all files and then ran the lint script on all packages. So I am pretty confident there is no conflict in rules between tslint and prettier.

jonaskello:
A brief explaination of how the "automatic" formatting and linting of staged files work in this PR:

* The package `husky` will add the ability to have git-hooks as regular package.json scripts. It will look for scripts with certain names, such as `precommit` and run them at the corresponding point during git operations. 
* The package `lint-staged` adds the ability to run commands only for files that are staged for git commit.
* This PR adds a package.json script `precommit` that will be run by husky before committing. This script runs the `lint-staged` command, which will look in the `lint-staged` section of package.json for commands to run. There `tslint` and `prettier` are run. So this is how commands are executed only for staged files right before they are committed.

package.json
```diff
@@ -9,12 +9,24 @@
...
+    "precommit": "lint-staged"
   },
   "dependencies": {
     "lerna": "2.9.0"
   },
-  "workspaces": [
-    "packages/*"
-  ]
+  "workspaces": ["packages/*"],
+  "devDependencies": {
```
evocateur: It strikes me that you're missing `devDependencies` on `prettier` and `tslint`, both called from the `lint-staged` config? Otherwise, you're either depending on globally-installed modules, or the implicit provision via yarn workspace-hoisted dependencies.

jonaskello:
@evocateur That's a very good catch, I did not think of that. I wanted to move as few devDeps as possible to the root in this PR in order to keep it small, but of course whatever is used from `lint-staged` would have to be moved too. Thank you for pointing that out, I will update the PR.

@dotansimha Do you have a preference for devdeps in the root or not? If you want I could go ahead and move all devdeps to the root in this PR. Or do you prefer I only move what is needed in this PR? I initially thought this PR would be small and simple and just add prettier, but as usual things quickly get more complicated :-).

@evocateur This repo uses typescript and I am unsure how to thandle the `@types/package` devdeps. I guess they should not be moved to the root since they are related to reguar deps within each package? In my own typescript monorepos I have moved every devdep but `@types` packages to the root.

jonaskello:
Btw, in yarn v1.5.1 there was a feature added so you can now cd into a sub-package directory and do something like `yarn tslint` there and yarn will then look upwards to the root of the monorepo and find tslint there. So even if the devdeps are installed in the root, it is still possible to run them when your cwd is a sub-package. This does not work with previous versions of yarn.

dotansimha:
Thank you @jonaskello @evocateur !

Regrading the devDependencies - it sounds good, I think we can move all devDependencies to the root. 
I think that depending on the fact that Yarn 1.5.1 looks upward is okay, but let's also add a constraint in `engines` field to makes it easier to understand for other developers.
 
Regarding `@types` packages - I think TypeScript also looks upwards so maybe it's best to do the same (move to root). 

@jonaskello I think you can format the entire repo and I'll make sure that this PR goes in first to avoid though merge. 

🎉

jonaskello:
I think this is ready now, @dotansimha could you please review and see if you can resolve the package.json conflict?

jonaskello:
I was able to resolve the conflict myself :-).

jonaskello:
Seems travis does not have the latest yarn version, it fails becuase of the engines restriction added by this PR. I will try to fix it by installing a newer version. There is some info [here](https://yarnpkg.com/en/docs/install-ci#travis-tab).

jonaskello:
Seems like travis is having problems connecting to canonical to downoad packages. Someone in [this issue](https://github.com/travis-ci/travis-ci/issues/8507#issuecomment-366629982) mentioned that travis is probably being throttled by cannocial.

I think I will simply remove the ^1.5.1 constraint for yarn. It is not required unless you want to run package.json scripts when you are currently in a specific sub-package folder. If you run scripts in the root or if you use `lerna run` then any version of yarn will work.

jonaskello:
Travis seems happy now but I [noticed this](https://travis-ci.org/dotansimha/graphql-code-generator/jobs/353249197#L489) in the travis build log. Seems like there is a problem with webpack-node-externals.

PR #199, opened 2018-03-14
Export file as .ts instead of .d.ts

guilhermehubner:
`enum` on `typescript` represents a value therefore it can't be imported from `d.ts` (type definition file);


dotansimha:
Thank you @guilhermehubner !

guilhermehubner:
@dotansimha it's already published?

dotansimha:
Released 0.8.21 :) 

@guilhermehubner thanks

Issue #209, opened 2018-03-17
using custom templates via --project ignores --out


micimize:
running `gql-gen --file types/schema.json --project types/templates/ gql-gen --out types/generated` ignores `--out` and dumps output in `--project`

dotansimha:
Lets continue to discuss it here: https://github.com/dotansimha/graphql-code-generator/issues/209

PR #210, opened 2018-03-17
Output custom template output to --out

micimize:
Address #209, as well as:
* preserve directory structure to `--out` 
* make `ts.{ schema, documents }.handlebars` work without a prefix
  (e.g. `ts.schema.handlebars` => `schema.ts`)
* automatically `registerPartial` custom templates without breaking normal partial registration (`registerPartial(filename.split('.').reverse()[1], template)`)
* mark comments safe to avoid url escaping
* fixes some links in the docs and add relevant docs for the included changes

### DIFF
--- a/packages/graphql-codegen-cli/src/loaders/templates-scanner.ts
+++ b/packages/graphql-codegen-cli/src/loaders/templates-scanner.ts
@@ -13,7 +13,7 @@ export function scanForTemplatesInPath(dirPath: string, fileExtensions: string[]
   debugLog(`[scanForTemplatesInPath] Got results from glob: `, results);
 
   return results.reduce((prev, filePath) => {
-    prev[filePath] = fs.readFileSync(filePath).toString();
+    prev[filePath.substr(absolutePath.length + 1)] = fs.readFileSync(filePath).toString();
 
     return prev;
   }, {});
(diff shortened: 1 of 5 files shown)


dotansimha:
@micimize 
It sounds good. Only issue is that we need to address this scenario:
```
- a/
    ts.schema.handlebars
- b/
    ts.type.handlebars
```

What will `--out` should do? because there are multiple directories with templates...  The existing implementation will output the file to the same directory as the template

micimize:
@dotansimha great, I'll get this PR up to scuff so it can be merged and doesn't break everything.
For the directory structure - I've now started on that locally, and have the files outputting to the same relative directory structure. Am currently trying to solve the `#eachImport` module resolution problem, as it isn't directory structure aware

micimize:
It looks like it would require breaking template name parsing and handling into separate batch processes, and then passing a resolution map. Might be a good idea, but seems better done as a separate PR. Will update the custom templates docs on how to handle import paths

micimize:
@dotansimha fixed the parts that were breaking the CI / `yarn test` and updated the description & docs, so it's good to go

minheq:
any update on this?

dotansimha:
Great work @micimize !

PR #216, opened 2018-03-20
Use snapshots for tests

FredyC:
So it turns out that tests for document-finder that were simply checking for the existence of some text were not sufficient. Instead, I've added snapshot testing so every small change in the output matters. Then I've manually tweaked the snapshot to a form that will pass through the parser.

There is some sort of breaking change in 0.8.19 (works in 0.8.18) which includes brackets in the output while those are not part of `gql` literal string and parsing fails afterward.

```graphql
query GPlayerList {
    allPlayers {
      id
      tier
      name
    }
  }
) # this bracket should not be here
```

I am not strong with regexp, so looking at the actual code, I am not sure how to fix this. However, at least now there is failing test.

### DIFF
--- a/packages/graphql-codegen-cli/tests/__snapshots__/document-finder.spec.ts.snap
+++ b/packages/graphql-codegen-cli/tests/__snapshots__/document-finder.spec.ts.snap
@@ -1,0 +1,55 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`extractDocumentStringFromCodeFile file with commented code 1`] = `
+"query myQuery2 {
+    data {
+      field1
+      field2
+    }
+  }"
+`;
+
+exports[`extractDocumentStringFromCodeFile file with gql and string and function use 1`] = `
+"query myQuery {
+    data {
+      field1
+      field2
+    }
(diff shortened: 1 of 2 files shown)


dotansimha:
@FredyC thank you! that's a great idea!! 

I'll merge it as soon as the ci is done

FredyC:
Well, do not merge, it's broken, that's why I made this PR to show it :)

dotansimha:
Yeah just noticed the change from 0.8.18->19, i'll fix it soon and then we can merge it :D 

thank you!!

FredyC:
@dotansimha Any luck on figuring this out? Not that's blocking me, the 0.8.18 works just fine so far. It's a friendly bump :)

dotansimha:
@FredyC found the issue :) and I added better tests (parse the results with GraphQL to make sure it's a valid document) 

Will commit soon

dotansimha:
@FredyC, can you please rebase and update the tests? it should work now :)

FredyC:
Alright, it's passing. Thanks :)

dotansimha:
Awesome!! Thanks @FredyC !

Issue #155, opened 2017-10-26
TS and Nullable types, again


rasdaniil:
Hi!

I know it was discussed and closed here https://github.com/dotansimha/graphql-code-generator/issues/138
However I think the current implementation is not totally correct

> By default, every type is nullable - it's legitimate to return null as any of the scalar types. 
from http://graphql.org/graphql-js/basic-types/

Now when generating interfaces it marks the field optional but the value remains non-nullable.
Optional field is good for convinience (as discussed in issue 138), but non-nullable value leaves us to return either value or undefined.  And we can't return null, which is against the spec.

```
// schema.gql
extend type Query {
   nutrient(_id: String): Nutrient
}

// types.generated.ts
export interface Query {
  nutrient?: Nutrient; 
}

// resolver.ts
const Query: TypedResolver<Types.Query> = {

  nutrient: async (root, ctx) => {
    const result = await db.nutrients.findOne({ _id: new ObjectID(ctx._id) }); **// Promise(Nutruent | null)**
    return result || undefined; **// ugly trick**
  }

};

```

IMHO, the solution would be to generate types like this:

```
// types.generated.ts
export interface Query {
  nutrient?: Nutrient | null; 
}
```

Thanks!

dotansimha:
It makes sense. Fixing it now!

dotansimha:
@rasdaniil @altschuler fixed in 0.8.14 :)

rasdaniil:
Thanks @dotansimha  !

PR #229, opened 2018-03-27
Remove '?' (optional properties) in tsgeneration to match apollo-codegen

gaberudy:
I think the apollo client-side codegen off queries is correct, in that the typescript optional (?) annotation on keys should not be used for generated schema code.

Most GQL resolver frameworks require keys to be defined and set to null, not undefined (as the ? allows). 

This is now the only thing preventing typescript types from exactly matching between my backend and front-end codegen, which would enable some code-sharing even as they use different code generators.

### DIFF
--- a/packages/graphql-codegen-generators/src/typescript-multi-file/operation.handlebars
+++ b/packages/graphql-codegen-generators/src/typescript-multi-file/operation.handlebars
@@ -4,7 +4,7 @@ import { {{ name }} } from './{{ file }}';
 export namespace {{ toPascalCase name }} {
   export type Variables = {
 {{#each variables}}
-    {{ name }}{{#unless isRequired}}?{{/unless}}: {{ toPrimitive type }}{{#if isArray}}[]{{/if}}{{#unless isRequired}} | null{{/unless}};
+    {{ name }}: {{ toPrimitive type }}{{#if isArray}}[]{{/if}}{{#unless isRequired}} | null{{/unless}};
 {{/each}}
   }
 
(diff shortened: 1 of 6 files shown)


dotansimha:
@gaberudy thank you for creating this PR! 

Few steps before we can continue with it:
- Have you seen this discussion? https://github.com/dotansimha/graphql-code-generator/issues/155 
- Can you please check the tests and CI?

gaberudy:
@dotansimha I would take a crack at fixing the tests if you think you would accept this PR.

It looks like #155 was about making the type of the properties be nullable (adding ` | null`), which is great and needed, but I'm arguing the better default would be to make the properties not optional (allowed to be undefined, assigned to undefined).

It may be a change some people like myself appreciate, others don't care, and some don't like. But it would match the spec of GQL better (IMHO) and the use case I am running into.

dotansimha:
@gaberudy I think the best solution is to use the generator-specific config to decide if to use `?` or not, and by default we should keep the current state.

how to set it: https://github.com/dotansimha/graphql-code-generator/tree/master/packages/graphql-codegen-cli#generator-specific-config 
how to use it in the template: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/templates/typescript/src/template.handlebars#L2-L4

dotansimha:
@gaberudy I implemented it in this PR: https://github.com/dotansimha/graphql-code-generator/pull/293 
Thank you for your help and this PR! It's a great idea!!

PR #246, opened 2018-04-09
Update exported enums to use 'export declare enum'

darthtrevino:
We've seen some errors with enum output not being usable downstream. It looks like enums require 'export declare' syntax in .d.ts files.

I'm looking into figuring out a way to test this to ensure that Typescript doesn't complain downstream


darthtrevino:
This may be some goofiness on our end; researching further. I'll reopen if I find anything

PR #255, opened 2018-04-14
CLI: fix export short alias

eritikass:
# What does this MR do?

This MR is fixing export (`--export`) short alias as it's not as shown in project readme (`-e`), but `-u` what is used by `--url`.

export himself seems to be working nicely.


dotansimha:
Great! Thanks @eritikass !

PR #290, opened 2018-05-06
Use npx to run the boilerplate generator

dobesv:
Previously it said to use yarn to run this script, which actually doesn't work.

### DIFF
--- a/packages/scripts/handlebars-templates-scripts/README.md
+++ b/packages/scripts/handlebars-templates-scripts/README.md
@@ -4,26 +4,16 @@ Create GraphQL Codegen templates easily with this Handlebars.
 
 This package exposes a CLI util called `codegen-handlebars-templates-scripts` which allow you to easily create a GraphQL Codegen Template project, and built it without dealing with external utils.
 
-To get started with this package, install it globally, or as local dependency inside an existing directory:
-
-    $ yarn global add codegen-handlebars-templates-scripts
-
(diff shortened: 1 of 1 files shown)


dotansimha:
Thank you @dobesv !

I think the issue with the docs in caused by the difference between installing the codegen as global dependency and as a local dependency.

If you install it as global dependency (`yarn global add ...`), it should be: `codegen-handlebars-templates-scripts init`. It you install it as a local dependency (`yarn add ...`), then it should be with Yarn prefix: `yarn codegen-handlebars-templates-scripts init`.

I think the docs should note that, and maybe `npx` should be an alternative for NPM users that installs it locally.

Do you mind updating it? :)

dobesv:
I don't think `yarn codegen-handlebars-templates-scripts init` will work unless you added it to your `scripts` section in `package.json` but I could be wrong.  However, the whole point of `npx` is that if you have it installed locally it'll use the one that's installed, otherwise it'll get a fresh one and use that.  So using `npx` should work in either case.

dotansimha:
@dobesv Yarn has a built-in mechanism that works like `npx`.
From what I know, `npx` is an alternative for running local scripts if you are using `npm` instead of `yarn`... I prefer Yarn, that's why I think the docs should still use Yarn :)

dobesv:
Oh interesting I didn't know yarn had the feature.  Neat!

PR #327, opened 2018-05-24
Switch console to logger

jycouet:
Hey @dotansimha 
What do you think about having a bit nicer logs? Especially to catch problems fast.


dotansimha:
Awesome!! Thanks @jycouet !!!

PR #341, opened 2018-05-31
use own strip-comments

### LINKED ISSUE #318: Improve .tsx support
```typescript
import * as React from "react";

export default class extends React.Component<{}, {}> {
  public render() {
    return <div />;
  }
}

// export const pageQuery = graphql`
//   query IndexQuery {
//     site {
//       siteMetadata {
//         title
//       }
//     }
//   }
// `;
```

Trying to execute the program with this code will fail.
The problem starts because https://github.com/jonschlinkert/strip-comments doesn't play nicely with ts.
If you have jsx the parser will get confused and will throw. Then, it won't remove the comments. Later, the [regex](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/utils/document-finder.ts#L22) will capture the query with the comments inside it and the graph-ql parser will throw. 
But we can pass strip-comments custom babylon plugins: https://github.com/jonschlinkert/strip-comments/pull/40

The problem is that https://github.com/jonschlinkert/babel-extract-comments which is used by strip comments is using an older version of babylon which doesn't support typescript.

When both these dependencies are updated the options for typescript and jsx should be passed here: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/utils/document-finder.ts#L13

I'll submit PRs for them and will comment here later.

degroote22:
It fixes #318.


dotansimha:
Looks good! Thanks @degroote22 !

PR #345, opened 2018-06-01
refactor strip-comments

degroote22:
The old code wasn't so pretty, my bad :(

### DIFF
--- a/packages/graphql-codegen-cli/src/utils/strip-comments.ts
+++ b/packages/graphql-codegen-cli/src/utils/strip-comments.ts
@@ -1,4 +1,5 @@
 import * as babylon from 'babylon';
+import { CommentBlock, CommentLine, Comment } from 'babel-types';
 
 /*!
  * strip-comments <https://github.com/jonschlinkert/strip-comments>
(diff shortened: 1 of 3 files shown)


packages/graphql-codegen-cli/src/utils/strip-comments.ts
```diff
@@ -1,41 +1,46 @@
...
+const removeComment = (
+  str: string,
+  comment: CommentBlock | CommentLine | Comment,
+  pos: { start: number; end: number; removed: number }
+) => {
+  let nl = '';
+
+  // the casting is safe because if type does not exist
+  // we will be comparing undefined against a string
+  if ((comment as CommentLine).type === 'CommentLine') {
```
dotansimha: You can use `is` function and avoid it:
```
function isCommentLine(obj: any): obj is CommentLine {
   return obj && obj.type && obj.type === 'CommentLine';
}
```

degroote22: Will do :)
I heavily appreciate this review, thanks.

dotansimha:
Looks good. Thanks @degroote22 !

PR #354, opened 2018-06-03
workaround bom in json

### LINKED ISSUE #353: Accept json schema files with Byte Order Mask
I'm using AWS AppSync and the files they provide me for the schema come with this [BOM](https://en.wikipedia.org/wiki/Byte_order_mark).
I guess it wouldn't be a problem [to work around this](https://github.com/npm/npm/blob/latest/lib/utils/parse-json.js).
Will submit a PR soon :)

degroote22:
Related #353


dotansimha:
Thanks @degroote22 !

PR #371, opened 2018-06-13
Introspection template

kamilkisiela:
@dotansimha 

- [x] tests
- [x] write about it in README.md

`graphql-codegen-introspection-template`


packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -40,7 +42,7 @@ export interface CLIOptions {
   skipDocuments?: any;
   config?: string;
   require?: string[];
-  overwrite: boolean;
+  overwrite?: boolean;
```
kamilkisiela: TypeScript was showing a warning about this

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -29,6 +30,7 @@ process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
 export interface CLIOptions {
   file?: string;
   url?: string;
+  introspect?: boolean;
```
kamilkisiela: there's a new flag that overwrites others

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -125,7 +129,7 @@ export const validateCliOptions = (options: CLIOptions) => {
     logger.warn(`--export is deprecated, use --schema instead.`);
   }
 
-  if (!template && !project) {
+  if (!template && !project && !introspect) {
```
kamilkisiela: avoid throwing an error since we don't need `template` or `project` while dumping a schema

kamilkisiela:
@dotansimha I added few comments

kamilkisiela:
I assumed that `gql-gen` with `--introspect` overwrites other logic, like generating templates, files etc but I can make it work side by side in this PR or in near future.

kamilkisiela:
@dotansimha I used template now

packages/templates/introspection/src/config.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as index from './template.handlebars';
+import { toJSON } from './helpers/to-json';
+
+export const config: GeneratorConfig = {
+  inputType: EInputType.SINGLE_FILE,
+  flattenTypes: true,
+  templates: {
+    index
+  },
+  primitives: {
```
kamilkisiela: TypeScript says it's necessary to define them but I don't think this template even use it

packages/templates/introspection/src/config.ts
```diff
@@ -0,0 +1,22 @@
+import { EInputType, GeneratorConfig } from 'graphql-codegen-core';
+import * as index from './template.handlebars';
+import { toJSON } from './helpers/to-json';
+
+export const config: GeneratorConfig = {
+  inputType: EInputType.SINGLE_FILE,
+  flattenTypes: true,
```
kamilkisiela: I don't know if I should skip that or leave it as is

packages/templates/introspection/src/config.ts
dotansimha: Primitives? it uses them in the template helpers

packages/templates/introspection/src/config.ts
dotansimha: Just leave it as it

dotansimha:
Thanks @kamilkisiela !

PR #381, opened 2018-06-16
Implement watch mode

### LINKED ISSUE #379: Watch mode using watchman?
I must say it becomes a bit of a pain over time to always have to run the command whenever I make changes to my queries. I am thinking if it would be possible to utilize watchman for this. I might even take a stab at it if it's something you consider a good idea?

FredyC:
Alright, here goes very basic and crude implementation. I have checked it against my project and it's working very nicely. 

I should probably add some tests, but I am not entirely sure about it. Would need to mock filesystem most likely and I can imagine it would be failing on CI anyway since Watchman is usually not installed there. What do you think @dotansimha?

PS: I added `semi` option to prettier config here. I hope you don't mind, it was driving me crazy to write those semicolons manually everywhere :)

Fixes #379

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,12 +36,12 @@ If you develop a **client-side with TypeScript, Angular and GraphQL**, you can u
 
 ## Available Templates:
 
-| Language                  | Purpose                                                                                 | Package Name & Docs                                                                        |
-| ------------------------- | --------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------ |
-| TypeScript                | Generate server-side TypeScript types, and client-side typings                          | [`graphql-codegen-typescript-template`](./packages/templates/typescript)                   |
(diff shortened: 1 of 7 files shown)


dotansimha:
Looks good @FredyC! I think we should also watch for the schema file (in case of a local or introspection file)

FredyC:
@dotansimha Yea, I think I agree, it's just I have never used this portion of the tool and never really understood what is it good for :)

dotansimha:
@FredyC You can just use `--schema` value, and use code similar to that to make sure it's a local file: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/codegen.ts#L159-L159 , and then just watch this file as well.

dotansimha:
Thanks @FredyC !

PR #386, opened 2018-06-19
typescript: generate types for resolvers

### LINKED ISSUE #319: Generate resolver types?
Currently using [graphql-schema-typescript](https://github.com/dangcuuson/graphql-schema-typescript#type-resolvers) to generate resolver typings, forcing me to use different types on client and server. Would love to see this added to graphql-code-generator.

kamilkisiela:
Closes #319

I changed a bit how `expect().toBeSimilarStringTo` works, it was comparing a template from start to end now it checks if a template contains a part we're looking for.

- [x] some tests are breaking other tests
- [x] add more tests
- [x] issue with `graphql-tools` ([show it to me!](https://github.com/kamilkisiela/gql-gen-with-resolvers/blob/7cc59d6c1020ec1e98ee189e80cb737c3d81d898/src/index.ts#L5-L10) / [show an example](https://stackblitz.com/edit/typescript-mpc1c7?file=index.ts))
- [x] maybe a flag that enables / disables generating types for resolvers?
- [x] maybe a different template that does that

Example: https://github.com/kamilkisiela/gql-gen-with-resolvers

@dotansimha

### DIFF
--- a/packages/graphql-codegen-cli/tests/cli.spec.ts
+++ b/packages/graphql-codegen-cli/tests/cli.spec.ts
@@ -35,7 +35,7 @@ describe('executeWithOptions', () => {
       config: '../../dev-test/config/gql-gen.json'
     });
 
-    expect(result[0].content).toContain('Generated in');
+    expect(result[0].content).toMatch('Generated in');
   });
 
   it('execute the correct results when using schema export as object', async () => {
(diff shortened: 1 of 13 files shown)


dotansimha:
> maybe a flag that enables / disables generating types for resolvers? 

You can use schema-specific config, for example: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/templates/typescript/src/enum.handlebars#L2 

> maybe a different template that does that 

I think it's okay that it's part of this template. 

> issue with `graphql-tools`

It's an issue with TypeScript: https://github.com/Microsoft/TypeScript/issues/15300 , I think we can safely use `as any` for now, because there is not other option :/

dotansimha:
Awesome! Thanks @kamilkisiela !

fabien0102:
I really need to try this!!! Thanks @kamilkisiela :tada:

PR #405, opened 2018-07-02
fix(introspection-template): write IntrospectionQuery to file

kamilkisiela:


kamilkisiela:
CI is green but it fails locally. I also deployed a preview of it `@kamilkisiela/u6mmkqhibkg` (introspection template).

The thing is `graphql` screams that the object we provide is not a `GraphQLSchema` object, it checks it like this: `schema instanceof GraphQLSchema`. I would say it means we have multiple `graphql` modules but we don't and that's super weird.

@dotansimha Could you give it a try? Maybe a different pair of eyes might help and maybe I'm missing something here.

dotansimha:
It usually happens when you have more than one instance of `graphql` in your project.
You can try to make sure you have only one with `yarn why graphql`, and you can try to resolve it by making sure that all dependencies are using the same version, or with `resolutions` (yarn).

I might also happen if you are trying to generate client-side using a file from a server-side (and it has it's own `node_modules/graphql`).

kamilkisiela:
I did all of that before release a pr

Issue #390, opened 2018-06-24
Missing moment.js deps in core package


brikou:
moment.js is missing in core package... https://github.com/dotansimha/graphql-code-generator/blob/v0.9.3/packages/graphql-codegen-core/src/utils/logger.ts#L1

so graphql-codegen-core cannot be used as standalone package

PR #408, opened 2018-07-03
add graphql to typescript template dependencies

degroote22:
It won't compile as a stand-alone package if it isn't here.
It complains about missing types.
The same thing as #390


dotansimha:
@degroote22 just pushed another bugfix, can you please rebase? :) 

Thank you!

dotansimha:
Thanks @degroote22 !

PR #441, opened 2018-07-24
Add 'noNamespaces' option to typescript template

### LINKED ISSUE #285: Generate without namespaces
The babel typescript transformer doesn't support namespaces, so we can't use the generated types with that package.

ardatan:
Fixes #285


dotansimha:
Thanks @ardatan ! looks good! Can you please add some unit tests for the generated output?

ardatan:
Added unit tests for `fragments` and `resolvers` with `noNamespaces` option.

PR #444, opened 2018-07-27
`interfacePrefix`, `scalars` and `prepend` options, non-pascalcase handling, add generic types for context and parent parameters

### LINKED ISSUE #153: Custom Scalars?
In Typescript, custom scalar types are emitted like:

```ts
export type Foo = any;
```

How can I make the emitted Foo refer to my existing types?

For example, if somewhere else in my code base, I have:
```ts
class Foo {
  bar: string

  someMethod() {
    ...
  }
}
```

I know this supports templating, but even with templating, I'm not sure how to make Foo in the generated types refer to my project's Foo.  It seems like I'd need to import existing scalar definitions into my schema template?  Or I'd need to define all my custom scalars in templates?  Both of these sound pretty unmaintainable.  Is there some convenient way to do this?  Or is there an example for solving this sort of problem?

PS. Thank you for this package!

### LINKED ISSUE #362: Prefix Typescript interfaces?
Hello - I'm using the generator and it's working great (great work!) - I was wondering if it'd be possible to add prefixes to the TS interfaces (User -> IUser)?

### LINKED ISSUE #410: generate type for the root object resolver
The `graphql-codegen-typescript-template` generates types as follows:

```
// schema.gql
#import "./types.gql"
type Query {
  me: User!
}

type schema {
  query: Query
}

type User {
  Id: String!
  UserName: String!
}

// generated types
/* tslint:disable */
import { GraphQLResolveInfo } from 'graphql'

type Resolver<Result, Args = any> = (
  parent: any,
  args: Args,
  context: any,
  info: GraphQLResolveInfo
) => Promise<Result> | Result

export interface Query {
  me: User
}

export interface User {
  Id: string
  UserName: string
}

export interface schema {
  query?: Query | null
}

export namespace QueryResolvers {
  export interface Resolvers {
    me?: MeResolver
  }

  export type MeResolver = Resolver<User>
}

export namespace UserResolvers {
  export interface Resolvers {
    Id?: IdResolver
    UserName?: UserNameResolver
  }

  export type IdResolver = Resolver<string>
  export type UserNameResolver = Resolver<string>
}

export namespace schemaResolvers {
  export interface Resolvers {
    query?: QueryResolver
  }

  export type QueryResolver = Resolver<Query | null>
}

```

Is it possible to generate single type that contains all the resolvers nestedly, instead of spreading them into namespaces? Just an example of what I'm thinking...

This somehow follow the shape of the `graphql-tools` for the resolvers.

```
interface IResolvers<TContext> {
  Query: {
    me: MeResolver<TContext>,
    you: YouResolver<TContext>
  },
  Mutation: {
    add: AddResolver<TContext>
  }
}
```

### LINKED ISSUE #414: Do not emit Resolver type when not used
Definitely an interesting idea I will use in other projects, however, there is a slight issue that this is also generated when `skipSchema` is set ultimately causing `'Resolver' is declared but never used.` and unable to compile the code.

cc @kamilkisiela

ardatan:
This PR adds fixes and some features explained below;

* Handle custom scalars
Fixes #153 
Add `prepend` option  for custom strings to be added on top of file.
```json
 prepend: [
        "// My Comment",
        "// My Other Comment"
    ]
```
Add `scalar` option for custom scalar mappings
```json
    scalars: {
        "Date": "Date"
    }
```

`prepend` and `scalars` options can be combined to map custom types to scalars.

```json
    prepend: [
        "import { CustomScalarType } from './custom-types';"
        "import { AnotherCustomScalarType } from './another-custom-types';"
    ],
    scalars: {
        "CustomScalar": "CustomScalarType",
        "AnotherCustomScalar": "AnotherCustomScalarType"
    }
```

* Add Typings for Parent object. So, parent parameter of resolver function has type of its parent object.
Parameters of resolver function type are now optional, because sometimes graphql resolver may not have parent, args, context or info parameters.
* Add `interfacePrefix` option
Fixes #362 
* Add generic types for context and parent parameters
Fixes #410 
* Do not emit resolver type when skip schema
Fixes #414


dotansimha:
Thanks @ardatan ! that's looks awesome!!!!! 👍

packages/templates/typescript/src/schema.handlebars
```diff
@@ -1,18 +1,20 @@
...
+{{#if types}}
 import { GraphQLResolveInfo } from 'graphql';
 
-export type Resolver<Result, Args = any> = (
-  parent: any,
-  args: Args,
-  context: any,
-  info: GraphQLResolveInfo
+export type Resolver<Result, Parent = any, Context = any, Args = any> = (
+  parent?: Parent,
```
matej21: hi, why is everything optional now? it is a bc break, all resolvers now have an incompatible signature

ardatan: Resolvers can work same without defining each parameter (parent, args, context, info). If you're using all of them, this shouldn't break your code.

PR #465, opened 2018-08-06
Parent of resolver should always be any; Use Resolver or SubscriptionResolver directly on field

### LINKED ISSUE #462: Parent in *Resolver types should be `any`
All parents should be `any` since we don't know them, we may expect how they may look like but we can't be sure.

```ts
export type NameResolver<
    R = string | null,
    Parent = Task,
    Context = any
  > = Resolver<R, Parent, Context>;
```

`Parent = Task` but it's not true because maybe we pass an object that has a different interface than `Task` and instead of `Task.name` it has `Task.title`. Because of that, it TypeScript will show it as an error.

`Parent` generic type should be optional and `any` by default.

### LINKED ISSUE #466: Missing signature call in Resolver type
```ts
export type Resolver<Result, Parent = any, Context = any, Args = any> = 
    QueryResolver<Result, Parent, Context, Args>
  | SubscriptionResolver<Result, Parent, Context, Args>;
```

Because `QueryResolver` and `SubscriptionResolver` has the same arguments TypeScript doesn't know if an object of type `Resolver` is a function or an object with `subscribe` and `resolve` methods and it screams about it 😢 

I guess we need to set each field as either QueryResolver or SubscriptionResolver directly.

kamilkisiela:
If end user decide to put something there it should be able to do it

Fixes #462 
Fixes #466
https://github.com/dotansimha/graphql-code-generator/issues/153#issuecomment-410647014
https://github.com/dotansimha/graphql-code-generator/pull/444#discussion_r207878190

- [x] check the output in ✈️ 🇫🇷 
- [x] check the output in 🎰 🇩🇪

### DIFF
--- a/dev-test/githunt/types.avoidOptionals.ts
+++ b/dev-test/githunt/types.avoidOptionals.ts
@@ -1,7 +1,7 @@
 /* tslint:disable */
 import { GraphQLResolveInfo } from 'graphql';
 
-export type QueryResolver<Result, Parent = any, Context = any, Args = any> = (
+export type Resolver<Result, Parent = any, Context = any, Args = any> = (
   parent?: Parent,
   args?: Args,
   context?: Context,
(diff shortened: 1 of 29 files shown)


kamilkisiela:
@dotansimha don't release it yet, there's an issue with `Resolver = QueryResolver | SubscriptionResolver`

PR #467, opened 2018-08-07
Make types pascal cased

arvind-chandrashekar-ck:
The template doesn't currently Pascal case union types; this PR updates the schema handlebar to fix this.


dotansimha:

kamilkisiela:
@dotansimha same as my change except there's no tests but we can merge one after another and have tests too

arvind-chandrashekar-ck:
@kamilkisiela I've updated the test spec with a test for the update I've made; let me know if any issues. Also thanks to you and @dotansimha for the timely feedback!

kamilkisiela:
@arvind-chandrashekar-ck thank you!

PR #480, opened 2018-08-09
Support for the "# import" syntax

davidyaha:
[GraphQL Yoga](https://github.com/prismagraphql/graphql-yoga) uses [`graphql-import`](https://github.com/prismagraphql/graphql-import) that allows for using the following syntax: 

```graphql
# import B from "b.graphql"

type A {
  # test 1
  first: String
  second: Float
  b: B
}
```

This PR adds support for that syntax.

Thanks for this amazing package @dotansimha 👍

### DIFF
--- a/dev-test/test-schema/schema-with-imports.graphql
+++ b/dev-test/test-schema/schema-with-imports.graphql
@@ -1,0 +1,9 @@
+# import User from './schema.graphql'
+
+type Query {
+  allUsers: [User]!
+  userById(id: Int!): User
+
+  # Generates a new answer for the guessing game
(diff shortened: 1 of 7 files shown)


dotansimha:
Looks good. Thanks @davidyaha !

webberwa:
The import doesn't seem to be working for the graphql-codegen-apollo-angular-template, I will reproduce when I get a chance

webberwang:
The import doesn't seem to be working for the graphql-codegen-apollo-angular-template, I will reproduce when I get a chance

PR #487, opened 2018-08-10
Avoid circular type aliases for scalars (dotansimha/graphql-code-gene…

### LINKED ISSUE #485: [typescript] Example for `scalars` produces illegal code
Using the example for the `scalars` generator config from `packages/templates/typescript/README.md` produces the following invalid TypeScript code in the output:

```ts
export type Date = Date;
```

TypeScript here complains that `[ts] Type alias 'Date' circularly references itself.`

I suggest the following fix:

Whenever the `typescript` generator detects that a scalar has been set to the same type name, simply omit the default line that gets issued for unknown scalars, i.e. instead of outputting `export type Date = any;`, do not output anything at all for that scalar.

---

In the meantime, the following generator config can be used as a work-around (pending #484):

```ts
"prepend": [
    "type $Date = Date;"
],
"scalars": {
    "Date": "$Date"
}
```

sgoll:
Fixes #485.

### DIFF
--- a/packages/templates/typescript/src/helpers/get-scalar-type.ts
+++ b/packages/templates/typescript/src/helpers/get-scalar-type.ts
@@ -1,7 +1,10 @@
 export function getScalarType(type, options) {
   const config = options.data.root.config || {};
   if (config.scalars && type in config.scalars) {
-    return config.scalars[type];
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @sgoll !!

PR #499, opened 2018-08-12
Fixed relative import to the core package

davidyaha:
Found that issue by trying to publish this package to my own scope so I could test my #480 PR on my own project. 
Seems to me like a good change nonetheless..

Thanks ahead!

### DIFF
--- a/packages/graphql-codegen-cli/src/loaders/schema/schema-from-export.ts
+++ b/packages/graphql-codegen-cli/src/loaders/schema/schema-from-export.ts
@@ -1,9 +1,9 @@
-import { SchemaLoader } from './schema-loader';
 import { existsSync } from 'fs';
 import { extname, isAbsolute, resolve as resolvePath } from 'path';
 import * as isValidPath from 'is-valid-path';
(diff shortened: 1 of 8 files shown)


packages/graphql-codegen-cli/src/loaders/schema/schema-from-export.ts
```diff
@@ -1,9 +1,9 @@
-import { SchemaLoader } from './schema-loader';
 import { existsSync } from 'fs';
 import { extname, isAbsolute, resolve as resolvePath } from 'path';
 import * as isValidPath from 'is-valid-path';
 import { buildASTSchema, buildClientSchema, DocumentNode, GraphQLSchema, IntrospectionQuery, parse } from 'graphql';
-import { debugLog, logger } from '../../../../graphql-codegen-core/dist';
+import { debugLog, logger } from 'graphql-codegen-core/dist';
```
dotansimha: I think it should be without `/dist`

davidyaha: fixed. thanks!

dotansimha:
Thanks @davidyaha !

PR #509, opened 2018-08-17
fix(typescript): non optional arguments in Resolver

### LINKED ISSUE #489: [typescript] Why is parent/args hardcoded as optional?
Hi there, awesome project to make server-side graphql even more enjoyable.

I just generated some typescript typings

```
gql-gen --schema ./schema.js --template graphql-codegen-typescript-template --out ./typings/
```

It generated typings/types.ts as expected:

```
...
export type Resolver<Result, Parent = any, Context = any, Args = any> = (
  parent?: Parent,
  args?: Args,
  context?: Context,
  info?: GraphQLResolveInfo
) => Promise<Result> | Result;
...
```

"parent" and "args" are optional, this is hardcoded in:

https://github.com/dotansimha/graphql-code-generator/blob/master/packages/templates/typescript/src/schema.handlebars#L5-L7

AFAIK the "parent" field is only optional for root types Query/Mutation/Subscription.

The "args" field should only be optional if the corresponding graphql field arguments are optional.

Am I doing something wrong?

kamilkisiela:
Fixes #489

### DIFF
--- a/packages/templates/typescript/src/schema.handlebars
+++ b/packages/templates/typescript/src/schema.handlebars
@@ -3,24 +3,24 @@
 import { GraphQLResolveInfo } from 'graphql';
 
 export type Resolver<Result, Parent = any, Context = any, Args = any> = (
-  parent?: Parent,
-  args?: Args,
-  context?: Context,
-  info?: GraphQLResolveInfo
+  parent: Parent,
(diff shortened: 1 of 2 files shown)


kamilkisiela:
@dotansimha @ardatan

Issue #356, opened 2018-06-04
Throw error if user passes invalid query


degroote22:
I've found this bug where if you mistype queries then the type generation process will emit an invalid typescript file and won't throw any errors.
[There is a simple repository which shows what can cause the bug.](https://github.com/degroote22/gql-gen-bug) (run "yarn gen" to reproduce)

The point is that [at this line](https://github.com/degroote22/gql-gen-bug/blob/531d8ef4eb36174a8871e876b8f8212e7b10ba93/src/a.tsx#L5) the query is wrong.
[Then, this invalid file is created](https://github.com/degroote22/gql-gen-bug/blob/master/src/typings/types.ts#L15)
I'm [using this schema](https://github.com/degroote22/gql-gen-bug/blob/master/schema.graphql).

I'll be happy to help fix this issue but I don't know much about the internals of this project.
If one could point me at to the files that handle this part I'll try to fix it.

dotansimha:
@degroote22 Thanks for reporting it! 

I guess that we need to run GraphQL validation on the documents before using the queries. I look into it.

Jontem:
I have the same problem. Have this bug been introduced in later versions?  I've upgraded to the latest version some days ago and have started to see this problem since then.

dotansimha:
Done! 0.11 is available now. @degroote22 @Jontem

PR #513, opened 2018-08-20
Validate queries against the graphql schema

Jontem:
Hi!

This is quick implementation of a fix for: #356. I don't know if this is the right place for the validation but more of starting point of a discussion how this could be implemented. Please let me know what you think.


dotansimha:
Thanks @Jontem !

PR #524, opened 2018-08-22
Fix validating queries against schema bug

Jontem:
Hi! PR #513 was a WIP but got merged :) Here is a fix for the error that arises when loading documents sources without a gql tag. I also filtered out the validation rule `NoUnusedFragments` because as of now it seems that we don't compile the template literals which will not include external fragments declared in other variables


dotansimha:
Great :) Thanks @Jontem

PR #545, opened 2018-08-28
[cli] fix handling gql-gen config

### LINKED ISSUE #543: Overriding default scalars
I have a custom resolver for GraphQL's `ID` scalar, which sends strings to the client but resolves to integers for the server.

I have tried specifying a custom type inside `gql-gen.json`, however I still get `string` inside the generated types.

Is there a way to override the types for default scalars such as `ID`?

```js
// gql-gen.json
{
  "generatorConfig": {
    "scalars": {
      "ID": "number"
    }
  }
}
```

```gql
# schema.graphql
type MyType {
    id: ID!
}
```

```ts
// types.ts
export interface MyType {
 id: string;
}
```

ardatan:
CLI dismisses `primitives` and `flattenTypes` option in `gql-gen.json`.
This PR fixes this issue and also refactors by using `primitives` everywhere instead of `primitivesMap`.
Fixes #543


dotansimha:
Thanks @ardatan !

PR #560, opened 2018-09-01
enhance react-apollo template

ardatan:


dotansimha:
@ardatan how is it going with this PR?

PR #561, opened 2018-09-02
Client Schema

kamilkisiela:
Three cases:

1. user provides `--schema` - nothing changes
1. user provides `--schema` `--clientSchema` - we merge both
1. user provides  `--clientSchema` only - now we throw an error that there's no `--schema` (maybe we shouldn't?)

@dotansimha @ardatan what do you think?


emmanuelbuah:
@kamilkisiela Per my earlier comment, it think its missing the case where `--schema` is provided,  `--clientSchema` is not provided but they exact a query with `@client`. Currently, the type generator throughs an error. It should skip the generation of queries with `@client` directive if a graph query schema isn't present.

dotansimha:
@emmanuelbuah , skipping the fields with `@client` is one option, but then you won't get the typings for it... I think the implementing in this PR will be more useful.
Maybe we can add something like `--ignored-directives` for your use case... what do you think? 

@kamilkisiela please rebase :D

emmanuelbuah:
@dotansimha You are right that you won't get the typings but that's ok because it won't be able to generate the typings anyways since there isn't a matching schema definition. `--ignored-directives` will only be useful if **all** your queries using that directive don't have a schema definition but that might not be the case. Using the example below

clientside_schema.js
```
type StateA {
   foo: String
   bar: String
}
```

clientside_queries.js
```
query GetStateA {
   StateA @client  {
     foo
     bar
  }
}

query GetStateB {
   StateB @client  {
     boo
     doo
  }
}
```
Typings for `type StateA` & `query GetStateA` will be generated by that of `query GetStateB` will throw an error. In Apollo's world, this is perfectly valid; for you can write to and query apollo cache/local state (via @client) without schema type definition.

DonaldWebb:
Thinking out loud when there is no client schema for the element we could fall back to any. The GetStateB could be typed:
```typescript
export type StateB = {
    boo?: any;
    doo?: any;
};
```

This way we get as much type information as possible. If a client schema is later created for StateB then the generated definition can take advantage of it.

PR #564, opened 2018-09-03
specify return type of generate()

### LINKED ISSUE #563: need return type definition of `generate()` function
# Summary

The `generate` function from `graphql-code-generator` now has a return type of `Promise<{}>`, which is not suitable for programmatic usage with TypeScript.

# Context

I'm using `graphql-code-generator` with a TypeScript script to generate schema files to use.

While I'm calling:

```typescript
import { generate } from 'graphql-code-generator'

async function generateSchema() {
  const fileOutputs = await generate({
    schema: INTROSPECTION_FILE,
    out: GENERATED_INTROSPECTION_FILE,
    template: 'typescrypt',
  }, false)
}
```

I get a `fileOutputs` with type `{}`. I can't directly iterate on `fileOutputs` because ts complaints about improper typing.

However, with the above arguments, `fileOutputs` should be `FileOuput[]` as logic defined in [`graphql-codegen-cli/src/generate.ts`](https://github.com/dotansimha/graphql-code-generator/commit/574150263dd6e69a95ebd9d5265f351f723ee781#diff-1d8570fda74c9db06a551d23bb14291a).

# Reason

After `watch` option is added to graphql-code-generator at https://github.com/dotansimha/graphql-code-generator/commit/574150263dd6e69a95ebd9d5265f351f723ee781#diff-1d8570fda74c9db06a551d23bb14291a , the return type of `generate` is changed from type `any[]` to `any[] | Promise<{}>` (i.e. `Promise<ReturnType<createWatcher>> | Promise<RetureType<writeOutput>>`), and causes type error when trying to iterate on return value of `generate()`

# Expection

`generate()` should have correct :ok_man: typing.

# Environment

- graphql-code-generator: 0.11.0
- TypeScript: 3.0.3
- Node: 10.9.0

frantic1048:
resolve #563


frantic1048:
Opps, just noticed [latest documentation](https://github.com/dotansimha/graphql-code-generator/blob/576bcbfbae2931867a6af950a5ceb7a4dac56b61/README.md) of `generate()` says it returns `Promise<FileOutput[]>`.

We may keep the return type unchanged.

frantic1048:
> I think it's a good idea to have the type also in the generated .d.ts and not only in the docs.
Thanks for the reminding, I will be aware of them :)

Issue #556, opened 2018-08-31
Regression error using graphql v14


emmanuelbuah:
There's a regression error that surfaces using graphql v14

```
TypeError: Cannot read property 'slice' of undefined
    at buildDirective (../node_modules/graphql/utilities/buildClientSchema.js:286:51)
    at Array.map (<anonymous>)
    at Object.buildClientSchema (../node_modules/graphql/utilities/buildClientSchema.js:302:84)
    at Object.introspectionToGraphQLSchema (../node_modules/graphql-codegen-core/dist/utils/introspection-to-schema.js:11:22)
    at Request._callback (../node_modules/graphql-code-generator/dist/loaders/schema/introspection-from-url.js:59:48)
    at Request.self.callback (../node_modules/graphql-code-generator/node_modules/request/request.js:185:22)
    at Request.emit (events.js:182:13)
    at Request.<anonymous> (../node_modules/graphql-code-generator/node_modules/request/request.js:1161:10)
    at Request.emit (events.js:182:13)
    at IncomingMessage.<anonymous> (../node_modules/graphql-code-generator/node_modules/request/request.js:1083:12)
```

This looks like its due to the breaking change by https://github.com/graphql/graphql-js/pull/1429/. More on changes in graphql v14 at https://github.com/graphql/graphql-js/issues/1383

IvanGoncharov:
@emmanuelbuah It's because of this change https://github.com/graphql/graphql-js/pull/1429
This introspection format was deprecated for last few years so we removed it in `14.0.0` release.
What library do you use to implement GraphQL on your backend?

emmanuelbuah:
@IvanGoncharov  I'm using apollo express server  v1.4

IvanGoncharov:
@emmanuelbuah Looks like everything good with your server and the bug happening because of this line:
https://github.com/dotansimha/graphql-code-generator/blob/8c071a8076a0ce5f58ae14347ae1512eab02deb6/packages/graphql-codegen-cli/src/loaders/schema/introspection-from-url.ts#L52
I'm not sure why this library needs to remove directive locations.

It can't be fixed in 'graphql-js' but I will add explicit check and better error message on the next release.

dotansimha:
Fixed in 0.12.0

PR #587, opened 2018-09-11
Restore `locations` field in introspection query

shmax:
Fix for #556 

Not sure where things were left in the discussion, but so far I haven't found any explanation for the `replace` call that removes `locations`, and as things stand I'm unable to use this library. Furthermore, neither of the other `introspection-from-` files seem to have a similar transformation.


dotansimha:
Thanks @shmax :)

Issue #528, opened 2018-08-22
Error when fragments are separated


alexk111:
Currently when fragments are stored separately from queries and mutations, the codegen produces a "Unknown fragment ..." error.

Example:

_fragment.ts_
```ts
import gql from 'graphql-tag';

export const userInfo = gql`
  fragment UserInfo on User {
    id
    name
    email
  }
`;
```

_query.ts_
```ts
import gql from 'graphql-tag';

import { userInfo } from './fragment.ts';

export const userQuery = gql`
  query UserQuery($id: ID!) {
    user(id: $id) {
      ...UserInfo
    }
  }

  ${userInfo}
`;
```

dotansimha:
For now, we released an alpha version with a fix for this:
```
Successfully published:
 - graphql-code-generator@0.12.0-alpha.821f12df
 - graphql-codegen-compiler@0.12.0-alpha.821f12df
 - graphql-codegen-core@0.12.0-alpha.821f12df
 - codegen-templates-scripts@0.12.0-alpha.821f12df
 - graphql-codegen-introspection-template@0.12.0-alpha.821f12df
 - graphql-codegen-apollo-angular-template@0.12.0-alpha.821f12df
 - graphql-codegen-typescript-mongodb-template@0.12.0-alpha.821f12df
 - graphql-codegen-typescript-react-apollo-template@0.12.0-alpha.821f12df
 - graphql-codegen-typescript-template@0.12.0-alpha.821f12df
```
@ganeshpms

emmanuelbuah:
Run into this as well but the alpha builds resolve it. Thanks.

kamilkisiela:
#559 should fix it for apollo-angular template

ganeshpms:
Thanks for the Alpha versions. 
Error: Property 'definitions' does not exist on type '{ new (): Document; prototype: Document; }' with Alpha version on the generated code. Which version of typescript is recommended?

ganeshpms:
![fragment](https://user-images.githubusercontent.com/5632736/45218440-3a0e5f80-b26d-11e8-818c-9c081a4f0542.jpg)
As seen in the attachment the `Document` is conflicting with the typescript's dom Document type. Please add the `*namepace*.Docuemnt.definition`. Also, the generated code adds the same fragment multiple time as  I have used the fragment 'Measurement' in multiple places in my query. And finally `,` is missing when there are two are more fragments.

RadAcademy:
[...]

PR #591, opened 2018-09-12
fix mutation component props, and use getter function to get document

### LINKED ISSUE #588: Generated mutation components can not be compiled
I use react-apollo 2.1.11
I have generated react apollo component for my OnLogin mutation operation. It looks like this:
```typescript
/* tslint:disable */

import gql from 'graphql-tag';

import * as ReactApollo from 'react-apollo';
import * as React from 'react';

export interface LoginUserInput {
    id: string;
    token: string;
    email: string;
    displayName: string;
    photoURL?: string | null;
    isEmailVerified: boolean;
}

export namespace Login {
    export type Variables = {
        user: LoginUserInput;
    };

    export type Mutation = {
        __typename?: 'Mutation';
        login: boolean;
    };
}

export namespace Login {
    export interface ComponentProps {
        mutation?: any;
        variables?: Variables;
        children: (
            result: ReactApollo.QueryResult<Mutation, Variables>
        ) => React.ReactNode;
    }
    export class Component extends React.Component<ComponentProps> {
        render() {
            return (
                <ReactApollo.Mutation<Mutation, Variables>
                    mutation={gql`
                mutation login($user: LoginUserInput!) {
                login(user: $user)
                }
            `}
                    {...this.props}
                />
            );
        }
    }
}
```

The generated code fails typescript compilation with the following error:
```
[ts]
Type '{ children: ((result: QueryResult<Mutation, Variables>) => ReactNode) | (string & ((result: Query...' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Mutation<Mutation, Variables>> & Readonly<{ childr...'.
  Type '{ children: ((result: QueryResult<Mutation, Variables>) => ReactNode) | (string & ((result: Query...' is not assignable to type 'Readonly<MutationProps<Mutation, Variables>>'.
    Types of property 'children' are incompatible.
      Type '((result: QueryResult<Mutation, Variables>) => ReactNode) | (string & ((result: QueryResult<Mutat...' is not assignable to type '(mutateFn: MutationFn<Mutation, Variables>, result: MutationResult<Mutation>) => ReactNode'.
        Type '(result: QueryResult<Mutation, Variables>) => ReactNode' is not assignable to type '(mutateFn: MutationFn<Mutation, Variables>, result: MutationResult<Mutation>) => ReactNode'.
          Types of parameters 'result' and 'mutateFn' are incompatible.
            Type 'MutationFn<Mutation, Variables>' is not assignable to type 'QueryResult<Mutation, Variables>'.
              Property 'client' is missing in type 'MutationFn<Mutation, Variables>'.
(alias) class Mutation<TData = any, TVariables = OperationVariables>
export Mutation
```

Could you please help me to figure out what I am doing incorrectly?

ardatan:
This PR fixes #588 the issue about component props for mutation component 
and it also fixes another issue for nested fragments, because the order of namespaces definitions matter, and causes undefined error ( uses getDocument function instead of constant Document value )
I also think this can solve potential conflicts between DOM Document and our Document field mentioned in this issue #528
Any suggestions welcomed :)

Then, I updated the example by using based on an example @kamilkisiela created on Stackblitz.

I removed the location replace stuff from introspection url loader because the remote API fails like this PR does #587. If you're ok with that, we can merge #587 then this one.
What do you think? @kamilkisiela @dotansimha


packages/templates/typescript-react-apollo/src/helpers/generate-fragment.ts
```diff
@@ -12,12 +12,12 @@ export function generateFragment(fragment: any, options: any): string {
...
     if (config.noNamespaces) {
       return `
-      const ${fragment.name}Document = ${gql(fragment, options)};
+      export const get${fragment.name}Document = () => ${gql(fragment, options)};
       `;
     } else {
       return `
         export namespace ${fragment.name} {
-          export const Document = ${gql(fragment, options)};
+          export const getDocument = () => ${gql(fragment, options)};
```
kamilkisiela: Why a getter?

packages/graphql-codegen-cli/src/loaders/schema/introspection-from-url.ts
```diff
@@ -49,7 +49,7 @@ export class IntrospectionFromUrlLoader implements SchemaLoader {
         {
           url: url,
           json: {
-            query: introspectionQuery.replace('locations', '')
```
kamilkisiela: it fixes one issue I guess, good

packages/templates/typescript-react-apollo/src/helpers/generate-fragment.ts
dotansimha: Because `Document` sometimes conflicts with the DOM `Document`

ardatan: Also it sometimes throws undefined due to the order of fragment definitions ( if PostFragment has AuthorFragment and AuthorFragment is defined after PostFragment, it throws undefined for AuthorFragment )

PR #594, opened 2018-09-13
Fix typo in a debug log.

mkantor:
Just noticed this while I was looking through the code. Thanks for the useful tool!


dotansimha:
Thanks @mkantor !

Issue #573, opened 2018-09-05
Error is not helpful when you forget to install template


jsphweid:
In my case, I forgot to install `graphql-codegen-typescript-template` as specified in the README. But the error I got was not helpful.

```bash
11:39:25.970 - error:
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @ refresh:types: `gql-gen --require ts-node/register --template typescript --schema packages/graphql-endpoint-lambda/src/schema.ts --out packages/graphql-endpoint-lambda/__generatedTypes__/types.ts`
npm ERR! Exit status 1
npm ERR!
```

It would be nice to see something like "template not installed" or "you are trying to use graphql-codegen-typescript-template but it appears you forgot to install it."

dotansimha:
Fixed in 0.12.0

PR #598, opened 2018-09-13
fix the issue with empty error when template is not installed

dotansimha:


dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/573

Issue #605, opened 2018-09-14
Underscores gets converted to PascalCase graphql-codegen-typescript-template


Jontem:
I don't know which version this is introduced in  but today i upgraded(both codegen and template) from version ^0.9.0 to ^0.12.0 and every type or fragment that has an underscore gets convertet to PascalCase. 

ex:
before: __typename: "My_Type" created an interface named My_Type
now: __typename: "My_Type" creates an interface named MyType

After looking further its not just underscores:
__typename: "MY_type" gets converted to an interface named MYType and when referenced in other places converted to MyType

kamilkisiela:
I don't think it's a bug. We decided to PascalCase everything to have naming strategy consistent.

dotansimha:
@kamilkisiela Yeah you are right, I'm just not sure if we need to keep and handle `_` in a special case.
In pascal-case, `_` is not valid, but it usually has a special meaning (private member), so I'm not sure what we should do with it.

kamilkisiela:
I mean, the library we use to PascalCase things, doesn't emit `_` so we don't have them

kamilkisiela:
Let's handle that after the weekend, I'm going offline :)

Jontem:
But my generated types.ts file doesn't even compile ? I don't care what the interfaces are named

dotansimha:
@Jontem can you please provide an example? why it's not compiling? do you have an error?

Jontem:
Yes i'll make an repro

Jontem:
Repro: https://github.com/Jontem/graphql-codegen-repro

Background for using strange names is that we're auto generating our graphql schema server side from  a sql database tables which has some historic names

dotansimha:
I can see it now, added it as failing test here: https://github.com/dotansimha/graphql-code-generator/pull/607 , not it will be easier to work on a fix for it :) 

Thanks @Jontem .

dotansimha:
@Jontem the issue is fixed and merge in #607, I'll release a new version very soon :)

dotansimha:
@Jontem fixed and release 0.12.2, can you please test it?

Jontem:
@dotansimha I've manage to reproduce it. I've updated the repro.

Jontem:
@dotansimha The problem is that the generated file doesn't compile. In the repro the type "_User" is declared twice

dotansimha:
[...]

PR #607, opened 2018-09-14
Fix for Issue #605

dotansimha:
https://github.com/dotansimha/graphql-code-generator/issues/605


ardatan:
We can add option for that issue. What do you think?
Something like below;
```
"compound": "snakecase"
```

PR #617, opened 2018-09-17
apollo-angular: allow to lazy load services

kamilkisiela:
Right now, all GQL services are defined as below which means they use root injector and are included in the main bundle. 

```ts
@Injectable({
  providedIn: 'root'
})
```

What we want is to allow to specify a module in which service should be included.

```graphql
query Feed {
  feed @client @NgModule(module:"./app/feed.module#FeedModule") {
    id
  }
}
```

This will produce:

```ts
import { FeedModule } from './app/feed.module';

@Injectable({
  providedIn: FeedModule
})
export class FeedGQL
```

Because we can't set `@NgModule` on an operation but on a field it means it doesn't matter on which field we used the directive. It must be defined once per operation, otherwise we throw an error.

@dotansimha @darkbasic  It will help us in both ✈️ and 🇩🇪🎰apps.


packages/templates/apollo-angular/src/helpers/ngmodule-directive.ts
```diff
@@ -0,0 +1,35 @@
+const R_DEF = /\s\@NgModule\([^)]+\)/gm; // matches: @NgModule(...)
+const R_MOD = /module\:\s*"([^"]+)"/; // matches: module: "..."
+
+// checks if operation has NgModule directive
+export function operationHasNgModule(operation) {
```
dotansimha: I'm not sure that looking for it that way and then removing it from the string is the best way to deal with it.
I was thinking about letting templates the option to "inject" types/directives into the schema, and this way templates that has directives will provide it, and the codegen will add it to the schema.
This way you'll be able to check the existence of the NgModule directive using GraphQL AST. 

What do you think? @kamilkisiela

PR #622, opened 2018-09-18
Print file path and number of errors when validating queries against schema

jonaskello:
This is an attempt to solve #621.

I have not been able to test it. Is there some easy way to test?

Fixes #621.

PR #624, opened 2018-09-18
Add failing test for fragment validation

jonaskello:
This shows a bug in the code added by PR #622. If we have a fragment thats lives in a different file, the validation will not work as it does not find the fragment.

### DIFF
--- a/packages/graphql-codegen-cli/tests/load-documents-sources.spec.ts
+++ b/packages/graphql-codegen-cli/tests/load-documents-sources.spec.ts
@@ -39,4 +39,11 @@ describe('loadDocumentsSources', () => {
     expect(errors[0].errors[0] instanceof GraphQLError).toBeTruthy();
     expect(errors[0].errors[0].message).toContain('Cannot query field "fieldD" on type "Query"');
   });
+
+  it.only('should not return an error array when one file references fragment in other file', () => {
+    const documentPath1 = join(__dirname, './test-documents/my-fragment.ts');
+    const documentPath2 = join(__dirname, './test-documents/query-with-my-fragment.ts');
+    const result = loadDocumentsSources(schema, [documentPath1, documentPath2]);
+    expect(Array.isArray(result)).toBeFalsy();
+  });
 });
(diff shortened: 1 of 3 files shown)


jonaskello:
One way to solve this might be to add empty dummy fragments to the ast for each file before doing the validation.

dotansimha:
Hmm it might be a good solution. 
The basic problem is that we are scanning the documents file as text files, and not as code files. 
We could try to require the doc file but then it becomes harder to find the correct export... 

In the angular/react templates, @kamilkisiela and @ardatan did something that might help us prevant these issues: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/templates/typescript-react-apollo/src/helpers/gql.ts  (not sure, let's wait for them to respond)

kamilkisiela:
@dotansimha I need to check what is the case here but the thing you referred to gathers everything and outputs one file. So we've got all the fragments there.

dotansimha:
@kamilkisiela Basically, when validating a document that loaded as string, we might failed if the documents uses a fragment that been imported in code. 
I think the you added a solution for loading it and concatenating all fragments into the documents that loaded it, right?

kamilkisiela:
We could validate documents after we've got all of them or I'm missing something? When we've got all, then we can merge it into one huge schema.

kamilkisiela:
@dotansimha yes I did exactly that

kamilkisiela:
Okay, I see now what the issue is. The thing I did with fragments should fix it.

dotansimha:
Maybe we can move this logic to the documents loader? then all templates will enjoy this feature

jonaskello:
If a file exports an invalid fragment I guess an error will be printed both for that file when it is validated and the same error for every file that imports it. This may not be a problem though. I would rather have double errors with file names than errors without file names :-).

kamilkisiela:
I guess loading everything as a string at first, then creating dependencies between documents and fragments, after that validating fragments, then validating documents (with fragments included) seems fine.

jonaskello:
And resolving imported fragments would of course have the benefit of finding errors where the fragment name is misspelled in the spread of the importing file.

jonaskello:
The test in this PR is included in #628 which was merged. So this PR is not needed.

Issue #619, opened 2018-09-17
error: Syntax Error: Expected Name, found $


furier:
**Describe the bug**
when running the `gql-gen` tool on the `schema.json` file below I get the error: `error: Syntax Error: Expected Name, found $` and nothing else, all other tools iv tried out does not give me any errors, so I am clueless on how to find the actual error here... a better error message could maybe help?

I have used GraphQL Playground and Apollo Engine to try and look for errors in the scema, but no luck...

**To Reproduce**
1. Use [schema.json](https://gist.github.com/furier/b892fae6091311b3a7e56ba26e4711c0#file-schema-json) as Schema
3. Run the following command: `gql-gen --schema schema.json --template graphql-codegen-typescript-template --out types.ts`

**Expected behavior**
A `types.ts` file that compiles.

**Environment:**
 - OS: Windows 10
 - Codegen: 0.10.7
 - Node: 10.0.0

**DEBUG Output**
[debug.log](https://gist.github.com/furier/b892fae6091311b3a7e56ba26e4711c0#file-debug-log)

kamilkisiela:
> error: Syntax Error: Expected Name, found $

It's an error that might come from your documents, your GraphQL operations. You probably try to put a variables directly and they should be named. An example:

```graphql
query feed($first: Int) {
  feed($first) {
    id
    name
  }
}
```

Or something similar. Maybe you use variables directly in the fields without defining them in the operation.

Try to find queries / mutations etc where you've got arguments so we can help you.

---

This error comes from GraphQL library, it doesn't tell much but @dotansimha  maybe we could catch it and re-throw with name of the operation or something, or a whole document (in debug mode).

kamilkisiela:
In terms of displaying an error, yes.

dotansimha:
@furier I'm not sure if that's something with GitHub, but the `schema.json` file you uploaded is not valid JSON, it's breaking in the middle of it:
![image](https://user-images.githubusercontent.com/3680083/45677152-c9502880-bb33-11e8-80d0-1c94216634b7.png)

Can you please try to upload it to a different place?

dotansimha:
Also, @furier, according to your command: `gql-gen --schema schema.json --template graphql-codegen-typescript-template --out types.ts`, you are generating typings for the schema, without GraphQL documents, right?

dotansimha:
[...]

PR #625, opened 2018-09-18
Catch an error while parsing

kamilkisiela:
Addresses #619

@dotansimha

### DIFF
--- a/packages/graphql-codegen-core/src/utils/parse.ts
+++ b/packages/graphql-codegen-core/src/utils/parse.ts
@@ -1,0 +1,17 @@
+import { parse as coreParse, Source } from 'graphql';
+
+export function parse(source: Source | string) {
+  try {
+    return coreParse(source);
+  } catch (e) {
+    let error: Error;
+
+    if (typeof source === 'string') {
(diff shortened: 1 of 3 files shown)


packages/graphql-codegen-core/src/utils/parse.ts
```diff
@@ -0,0 +1,17 @@
+import { parse as coreParse, Source } from 'graphql';
+
+export function parse(source: Source | string) {
+  try {
+    return coreParse(source);
+  } catch (e) {
+    let error: Error;
+
+    if (typeof source === 'string') {
+      error = new Error(`Parsing '${source.trim().substr(0, 25)}...' failed with: ${e.message}`);
```
kamilkisiela: I don't think caught error's stack is worth to attach to new error because it's mostly internal stack of `graphql`

kamilkisiela: @dotansimha

dotansimha: Yeah it's fine

PR #628, opened 2018-09-18
Proposed quickfix for unknown fragment

jonaskello:
This is a quickfix for #624 to allow users to continue working. 

It will simply ignore errors with message "Unknown fragment". This way we can still get the filename of every other error.  I think only the case with misspelled fragment names in spreads where the fragment is imported from another file will be not be validated with this fix.

The proper fix would be to actually resolve all the fragments.

### DIFF
--- a/packages/graphql-codegen-cli/src/loaders/documents/document-loader.ts
+++ b/packages/graphql-codegen-cli/src/loaders/documents/document-loader.ts
@@ -42,7 +42,7 @@ export const loadDocumentsSources = (
   const errors: ReadonlyArray<LoadDocumentError> = loadResults
     .map(result => ({
       filePath: result.filePath,
-      errors: validate(schema, result.content, effectiveRules)
+      errors: validate(schema, result.content, effectiveRules).filter(e => e.message.indexOf('Unknown fragment') === -1)
     }))
     .filter(r => r.errors.length > 0);
   return errors.length > 0 ? errors : concatAST(loadResults.map(r => r.content));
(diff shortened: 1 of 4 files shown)


dotansimha:
Looks good. Thanks @jonaskello !

dotansimha:
@furier We have CI that publishes alpha version. for now, you can use it with these versions:
```
Successfully published:
 - graphql-code-generator@0.13.0-alpha.106f2555
 - graphql-codegen-compiler@0.13.0-alpha.106f2555
 - graphql-codegen-core@0.13.0-alpha.106f2555
 - codegen-templates-scripts@0.13.0-alpha.106f2555
 - graphql-codegen-apollo-angular-template@0.13.0-alpha.106f2555
 - graphql-codegen-introspection-template@0.13.0-alpha.106f2555
 - graphql-codegen-typescript-mongodb-template@0.13.0-alpha.106f2555
 - graphql-codegen-typescript-react-apollo-template@0.13.0-alpha.106f2555
 - graphql-codegen-typescript-template@0.13.0-alpha.106f2555
```

jonaskello:
Turns out this quickfix does not cover the case when a query has an argument that is only used in a fragment. The validation will correctly give a `Error: Variable "$foo" is never used in operation "FooQuery".` message. This case should be somewhat rare but I have it in my codebase...

jonaskello:
@dotansimha Should I could try to make a new PR with a proper fix, or is someone else already working on that?

furier:
Do you have an example?
man. 24. sep. 2018 kl. 11:09 skrev Jonas Kello <notifications@github.com>:

jonaskello:
@furier Do you mean example where the unused variable problem will show up?

furier:
yes

man. 24. sep. 2018 kl. 16:51 skrev Jonas Kello <notifications@github.com>:

furier:
I have never made fragments for Queries or Mutations before only for return
types, so was interesting to see...

man. 24. sep. 2018 kl. 17:27 skrev Jonas Kello <notifications@github.com>:

PR #650, opened 2018-09-27
Add a repository to package.jsons that are missing it

reconbot:
This makes it a lot easier to find the code when wanting to contribute to the project.


dotansimha:
Thanks @reconbot !

PR #651, opened 2018-09-27
suppress warnings from apollo-tools when making schema from typedefs

reconbot:
I get a ton of warnings like this due to using a lot of graphql interfaces. This warning doesn't apply here because we don't deal with resolvers. This patch turns it off.

```
Type "Node" is missing a "resolveType" resolver. Pass false into "resolverValidationOptions.requireResolversForResolveType" to disable this warning.
```


dotansimha:
Thanks @reconbot !

obedm503:
This is happening again in version `1.10.0`

PR #655, opened 2018-10-01
Fix some logging messages not properly displaying the expected information

alvarogzp:
Hello! While debugging gql-gen with a new template I was building, I found that some log messages were displaying empty strings instead of the values they were supposed to display. I fixed it locally in order to understand what was going on.
Please, consider if it could be appropriate to be merged upstream.


dotansimha:
Awesome! Thanks @alvarogzp

Issue #643, opened 2018-09-24
Types are not generated for introspection query


FredyC:
**Describe the bug**
I have a query like this which reads introspection in runtime to figure user permissions. Unfortunately, the generated types do not include the `GUserPermissions` type at all. But if I add something else to that query, it appears there, but without the introspection part.

```js
export const UserPermissionsQuery = gql`
  query GUserPermissions {
    __schema {
      queryType { fields { name } }
      mutationType { fields { name } }
    }
  }
`
```

**To Reproduce**
I am using API approach and call it like this. Full introspection is generated in JSON file. I've checked and it includes types for `__schema`, so that should be a problem.

```js
  generate({
    schema: jsonPath,
    out: './src/graph/types.ts',
    template: 'graphql-codegen-typescript-template',
    args: ['./src/**/*graphql.ts'],
    skipSchema: true,
    overwrite: true,
  })
```

**Expected behavior**
To have TypeScript types based on introspection query.

**Environment:**
 - OS: Windows 10 x64
 - Codegen: 0.9.2
 - Node: 10.11.0

dotansimha:
Interesting. Thanks @FredyC ! 

I think we might ignore fields that starts with `__`. I'll check.

dotansimha:
@FredyC We are ignoring the internal fields of GraphQL because otherwise it will bloat the results (see https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-core/src/schema/schema-to-template-context.ts#L27) .

I can make this filtering an option, and then it will solve your issue.

patricknazar:
I've noticed that when I use simply a `__typename` in fragment, I get similar issues:

    fragment MyFragment on MyType {
        ... on TypeOne {
            __typename
        }
    }

gives me:


    export namespace MyFragment {

        export type TypeOneInlineFragment =

    }

Notice the trailing `=`. Will this resolve this problem too? It's not an introspection query really is it? It is an internal field, right?

Jonatthu:
@dotansimha I am having the same issues on the latest version, I really need to do some internal types queries

kamilkisiela:
I tried to fix it in #1380, could you guys check if that works in your projects? A canary release under `0.18.0-alpha.d365409c`, make sure you bump all packages from codegen group.

dotansimha:
Fixed in 0.18.0 🎉

Jonatthu:
[...]

PR #660, opened 2018-10-02
WIP: include internal introspection types

dotansimha:
WIP to solve this issue: https://github.com/dotansimha/graphql-code-generator/issues/643 

So there are 2 issues here:
- [X] We need to keep types starts with `__` 
- [ ] We need to make the introspection types available.


dotansimha:
So far, I managed to keep the introspection types in the schema, and added a cli flag for it. 

My problem now is with matching the `__schema` type inside `Query` type.
When the codegen builds the schema context, in this line: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-core/src/utils/get-field-def.ts .
When we are running `parentType.getFields`, it returns only the regular `Query` fields, without the introspection fields, so when we are trying to do `parentType.getFields()[name]`, it resolves to `undefined`.

@FredyC @kamilkisiela @ardatan - any ideas? :)

Jonatthu:
@dotansimha Is this still going? How can I generate this with a workaround?

PR #674, opened 2018-10-08
readme: gql-gen.json

tomconroy:
this readme is not correct, the default config file is `gql-gen.json`


dotansimha:
Thanks @tomconroy !

PR #684, opened 2018-10-11
Added CLI flag to shallow error on schema validation

valerybugakov:
This PR adds optional CLI flag `--shallow-error` useful for `watch` mode. 
At the moment codegen will terminate the watch process if the error will be found on schema validation and the developer needs to restart watcher if he made a mistake in the schema or some query/mutation.

With `--shallow-error` flag process will output the error but also, will keep watcher alive giving an ability to fix a mistake without restarting the process.

### DIFF
--- a/packages/graphql-codegen-cli/src/codegen.ts
+++ b/packages/graphql-codegen-cli/src/codegen.ts
@@ -88,7 +88,7 @@ export const initCLI = (args: any): CLIOptions => {
   return (commander as any) as CLIOptions;
 };
 
-export const cliError = (err: any) => {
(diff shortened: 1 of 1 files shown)


ardatan:
Thank you @valerybugakov ! 
I think we should always keep watcher alive by default. Could you update your code like that?

valerybugakov:
@ardatan sounds great, I've updated the PR, please take a look!

ardatan:
Thank you @valerybugakov ! LGTM!
What do you think? @dotansimha @kamilkisiela

dotansimha:
Thanks @valerybugakov !

PR #690, opened 2018-10-14
Allow js files to be required for template config

tgriesser:
Found this to be useful if you want to import the enum mappings from external files:

`gql-gen.js`:
```ts
import enumMappings from '../../constants/enumMappings';

export default {
  generatorConfig: {
    printTime: true,
    resolvers: false,
    enums: enumMappings
  }
};
```

### DIFF
--- a/dev-test/config/gql-gen.js
+++ b/dev-test/config/gql-gen.js
@@ -1,0 +1,5 @@
+module.exports = {
+  generatorConfig: {
+    printTime: true
+  }
+};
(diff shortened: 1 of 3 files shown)


tgriesser:
@dotansimha any idea why the CI wouldn't be running here? I just rebased against master & re-pushed and they still weren't triggered...

tgriesser:
Ehh nevermind just ended up using the programmatic usage instead, worked awesome. Thanks for such a great tool!

PR #709, opened 2018-10-18
[TypeScript] Added support for const enums

dotansimha:
As described here: 
https://www.typescriptlang.org/docs/handbook/enums.html#const-enums


ardatan:
I think const enums are not supported by babel/typescript-preset like namespaces :/
https://babeljs.io/docs/en/babel-plugin-transform-typescript

Issue #700, opened 2018-10-17
__typename should not be optional when used in query


frantic1048:
**Describe the bug**
When use `__typename` in a query, it's type should not be optional in corresponding query namespace.

**To Reproduce**

*schema*:

```gql
# schema.graphql

type Person {
  name: String!
}

type Query {
  person: Person
}
```

*document*:

```gql
query search {
  person {
    __typename
    name
  }
}
```

*command*: `gql-gen --template graphql-codegen-typescript-template --schema schema.graphql document.graphql --out schema.ts`

**Expected behavior**

The namespace of `query search` in generated `schema.ts`:

```ts
export namespace Search {
  export type Variables = {};

  export type Query = {
    __typename?: "Query";
    person?: Person | null;
  };

  export type Person = {
    __typename?: "Person"; // <- this should not be optional because we used it in query
    name: string;
  };
}
```

`Search.Person.__typename` is optional, while it should not be.

**Environment:**
- OS: Arch Linux (latest)
- Codegen: 0.12.6
- Node: 10.11.0
- graphql: 14.0.2

frantic1048:
ehh, I found `__typename` is directly droped in [`get-field-def.ts`](https://github.com/dotansimha/graphql-code-generator/blob/c5ee0f70690190bb74d7d97bda545779323c9644/packages/graphql-codegen-core/src/utils/get-field-def.ts#L6). It looks many places need to be changed to solve this..

dotansimha:
Thanks @frantic1048 . It's a bug and the way to fix it is to check if the selection set has `__typename` and generate the `?` according to it.

dotansimha:
Fixed in 0.15.2 :)

mfbx9da4:
So this is supposedly fixed but I don't see any docs of how to use the query version where `__typename` is required any guidance?

saschb2b:
@mfbx9da4 `nonOptionalTypename: true` made it required for me

PR #712, opened 2018-10-19
[DO NOT MERGE]fix(core): make __typename required when used

frantic1048:
Yet another hotfix (ﾟ∀。)

This PR is for fixing #700

The goal for this PR is to make `__typename` required when actually used in a document.

Summary of this PR:

1. `__typename` field in a trivial Field will be required when a selection has `__typename` field, otherwise, no `__typename` field will be generated in the selection field.
2. `__typename` on a union field, will not generate `__typename` on the union, but will *always* generate on each inline fragment
3. `__typename` on spread fragment is now *always* required.

The way to achieve above behavior is somewhat inappropriate I think. Thus I put `[DO NOT MERGE]` on the title and call this `hotfix`. I just want things working at first.

About `2.` and `3.`, they do not really check `__typename` is used in the query or not. Because I found it is hard to do such check. I do so because it is a conventional use case for union and fragments. We should find a proper way to handle this.

### DIFF
--- a/packages/templates/typescript/src/documents.handlebars
+++ b/packages/templates/typescript/src/documents.handlebars
@@ -3,27 +3,24 @@
 {{#unless @root.config.noNamespaces}}
 export namespace {{ toPascalCase name }} {
 {{/unless}}
-  export type {{#if @root.config.noNamespaces}}{{ toPascalCase name }}{{/if}}Variables = {
-  {{#each variables}}
-    {{#if @root.config.immutableTypes }}readonly {{/if}}{{ name }}{{ getOptionals this }}: {{ convertedType this }};
-  {{/each}}
-  }
+export type {{#if @root.config.noNamespaces}}{{ toPascalCase name }}{{/if}}Variables = {
+{{#each variables}}
+{{#if @root.config.immutableTypes }}readonly {{/if}}{{ name }}{{ getOptionals this }}: {{ convertedType this }};
+{{/each}}
+}
 
-  export type {{#if @root.config.noNamespaces}}{{ toPascalCase name }}{{/if}}{{ toPascalCase operationType }} ={{#if hasFields}} {
-    {{#if @root.config.immutableTypes }}readonly {{/if}}__typename?: "{{ toPascalCase operationType }}";
-    {{> selectionSet fields=fields prefix=name }}
-  }{{/if}}{{> fragments this=this prefix=name }}
-  {{#each innerModels }}
+export type {{#if @root.config.noNamespaces}}{{ toPascalCase name }}{{/if}}{{ toPascalCase operationType }} ={{#if
+hasFields}} {
+{{#if @root.config.immutableTypes }}readonly {{/if}}__typename?: "{{ toPascalCase operationType }}";
+{{> selectionSet fields=fields prefix=name }}
+}{{/if}}{{> fragments this=this prefix=name }}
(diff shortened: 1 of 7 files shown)


packages/templates/typescript/src/documents.handlebars
```diff
@@ -3,27 +3,24 @@
 {{#unless @root.config.noNamespaces}}
 export namespace {{ toPascalCase name }} {
 {{/unless}}
-  export type {{#if @root.config.noNamespaces}}{{ toPascalCase name }}{{/if}}Variables = {
-  {{#each variables}}
-    {{#if @root.config.immutableTypes }}readonly {{/if}}{{ name }}{{ getOptionals this }}: {{ convertedType this }};
-  {{/each}}
-  }
+export type {{#if @root.config.noNamespaces}}{{ toPascalCase name }}{{/if}}Variables = {
```
dotansimha: Can you please fix it? your IDE changed the indentation :X

dotansimha:
Thanks @frantic1048 ! 
Regarding 2,3 - that's right. 

You can also write new Handlebars helpers and it will make it easier to develop and write custom logics. 
In general, it looks good. Just fix the indentation on the template, and can you please add some basic tests? specifically to use case 1, but also for 2+3.

dotansimha:
Thanks @frantic1048 ! 
Regarding 2,3 - that's right. 

You can also write new Handlebars helpers and it will make it easier to develop and write custom logics. 
In general, it looks good. Just fix the indentation on the template, and can you please add some basic tests? specifically to use case 1, but also for 2+3.

dotansimha:
Thanks @frantic1048 ! 
Regarding 2,3 - that's right. 

You can also write new Handlebars helpers and it will make it easier to develop and write custom logics. 
In general, it looks good. Just fix the indentation on the template, and can you please add some basic tests? specifically to use case 1, but also for 2+3.

packages/templates/typescript/src/documents.handlebars
frantic1048: Sorry, I should turn off the auto-formatting of VSCode. ._.

frantic1048:
Thanks for the kind reply.

I'd fix the indents. Of course, tests will be added, I want to firstly show the basic thought about this PR last week.

raysuelzer:
I noticed after switching to this package from using graphql-tag that typename was no longer being added for me. Does this fix that?

frantic1048:
> I noticed after switching to this package from using graphql-tag that typename was no longer being added for me. Does this fix that?

@raysuelzer  Have you explicitly queried `__typename` field in gql tag?

raysuelzer:
> > I noticed after switching to this package from using graphql-tag that typename was no longer being added for me. Does this fix that?
> 
> @raysuelzer Have you explicitly queried `__typename` field in gql tag?

@frantic1048  No, it's not necessary, it is added automatically by `graph-tag`, not sure why that is not the behavior on this code generator.

dotansimha:
@raysuelzer Hi!
You are right, `graphql-tag` adds `__typename`, but `graphql-code-generator` does not use `graphql-tag` for loading the GraphQL documents, it just reads the query as string from the code file, and then uses `parse` from `graphql` to convert it to AST.

dotansimha:
Closing for now because this PR is outdated. Feel free to re-open later

PR #714, opened 2018-10-19
Fix getLogger(...).warning is not a function

NickBolles:
Running `gql-gen -w` would result in the following crash.
```
error: graphql_codegen_core_1.getLogger(...).warning is not a function
```
This issue is from this line: 

https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/utils/watcher.ts#L36

The warning I was getting was an Access denied error, which won't happen for everyone, which might be why this is kinda a hard issue to find. Simple fix though!


dotansimha:
Good catch 😄 thanks @NickBolles !

PR #719, opened 2018-10-22
Allow subscription resolver to be a function

### LINKED ISSUE #698: Cannot use generated Subscription types with NestJS resolver
**Describe the bug**

I cannot find a way to use the generated subscription resolver type in a NestJS app 

Given the following schema :
```graphql
type Mutation {
  signIn(email: String!, password: String!): String
}

type Subscription {
  userSignedIn: User
}
```

I get the following generated types :
```typescript
export interface SignInMutationArgs {
  email: string;
  password: string;
}

export namespace MutationResolvers {
  export interface Resolvers<Context = any> {
    signIn?: SignInResolver<string | null, any, Context>;
  }

  export type SignInResolver<
    R = string | null,
    Parent = any,
    Context = any
  > = Resolver<R, Parent, Context, SignInArgs>;
}

export namespace SubscriptionResolvers {
  export interface Resolvers<Context = any> {
    userSignedIn?: UserSignedInResolver<User | null, any, Context>;
  }

  export type UserSignedInResolver<
    R = User | null,
    Parent = any,
    Context = any
  > = Resolver<R, Parent, Context>;
}
```

Then I can create a handy union type of all resolvers :
```typescript
export interface MyAppResolvers extends MutationResolvers.Resolvers, SubscriptionResolvers.Resolvers {};
```

Finally, my resolvers look like this :
```typescript
import { Args, Mutation, Resolver, Subscription } from '@nestjs/graphql';
import { PubSub } from 'apollo-server-express';

import * as GQL from 'src/generated.ts';

const pubSub = new PubSub();

@Resolver()
export class AuthResolver implements GQL.MyAppResolvers {
  @Mutation()
  async signIn(@Args() args: GQL.SignInMutationArgs) {
    // ...
  }

  @Subscription()
  userSignedIn() {
    return {
      subscribe: () => pubSub.asyncIterator('userSignedIn'),
    };
  }
}
```

Then I get a typescript error because `userSignedIn()` resolver does not return a `User` but a function with a `subscribe` method :
```
error TS2416: Property 'userSignedIn' in type 'AuthResolver' is not assignable to the same property in base type 'AppResolver'.
  Type '() => Promise<{ subscribe: () => AsyncIterator<{}>; }>' is not assignable to type 'Resolver<User, any, any, any>'.
    Type 'Promise<{ subscribe: () => AsyncIterator<{}>; }>' is not assignable to type 'User | Promise<User>'.
      Type 'Promise<{ subscribe: () => AsyncIterator<{}>; }>' is not assignable to type 'Promise<User>'.
        Type '{ subscribe: () => AsyncIterator<{}>; }' is not assignable to type 'User'.
          Property 'id' is missing in type '{ subscribe: () => AsyncIterator<{}>; }'.

   async userSignedIn() {
           ~~~~~~~~~~~~
```

I'm quite new with GraphQL here, maybe I should not use a union type of resolvers ?
If so, how can I strongly type my resolver class ?

kamilkisiela:
Fixes #698

### DIFF
--- a/packages/templates/typescript/src/schema.handlebars
+++ b/packages/templates/typescript/src/schema.handlebars
@@ -9,7 +9,7 @@ export type Resolver<Result, Parent = any, Context = any, Args = any> = (
   info: GraphQLResolveInfo
 ) => Promise<Result> | Result;
 
-export type SubscriptionResolver<Result, Parent = any, Context = any, Args = any> = {
+export interface ISubscriptionResolverObject<Result, Parent, Context, Args> {
   subscribe<R = Result, P = Parent>(
     parent: P,
     args: Args,
@@ -23,6 +23,10 @@ export type SubscriptionResolver<Result, Parent = any, Context = any, Args = any
     info: GraphQLResolveInfo
   ): R | Result | Promise<R | Result>;
 }
+
+export type SubscriptionResolver<Result, Parent = any, Context = any, Args = any> =
+  | ((...args: any[]) => ISubscriptionResolverObject<Result, Parent, Context, Args>)
+  | ISubscriptionResolverObject<Result, Parent, Context, Args>;
 
 {{/if}}
 {{/ifCond}}
--- a/packages/templates/typescript/tests/resolvers.spec.ts
+++ b/packages/templates/typescript/tests/resolvers.spec.ts
@@ -44,7 +44,7 @@ describe('Resolvers', () => {
       info: GraphQLResolveInfo
     ) => Promise<Result> | Result;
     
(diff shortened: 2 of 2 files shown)


packages/templates/typescript/src/schema.handlebars
```diff
@@ -9,7 +9,7 @@ export type Resolver<Result, Parent = any, Context = any, Args = any> = (
   info: GraphQLResolveInfo
 ) => Promise<Result> | Result;
```
jon301: I think this line should be :
```
) => Promise<Result> | Result | SubscriptionResolver<Result>;
```
@dotansimha @kamilkisiela

PR #721, opened 2018-10-23
[New Template] Generate `module` definition for `.graphql` files

dotansimha:
The goal is to generate something like that:

Given a file `./src/my-query.graphql`, it will generate:

```
declare module './src/my-query.graphql' {
  const doc: DocumentNode;
  export default doc;
}
```

And support all use cases of [`graphql-tag/loader`](https://github.com/apollographql/graphql-tag#webpack-preprocessing-with-graphql-tagloader), to make sure that imports from `.graphql` files on client-side are strongly-typed now

Tasks:
- [x] Separate document error validations from handling.
- [x] Added `originalFile` to the each found document.
- [x] Removing gql documents concatation.
- [x] Add `documentsFiles` to `context` to get list of files, and then iterate list of documents from that file.
- [x] Add Handlebars helper that parses the `originalFile` into easy-to-use variables (such as, the relative path to `cwd`, absolute path, file extension...).
- [x] Create new template
- [x] Add README to template
- [x] Update root README

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,14 +38,15 @@ If you develop a **client-side with TypeScript, Angular and GraphQL**, you can u
 
 ## Available Templates:
 
-| Language                  | Purpose                                                        | Package Name & Docs                                                                                |
-| ------------------------- | -------------------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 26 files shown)


dotansimha:
@moshen1 - I think it might be relevant for you. 👍

PR #745, opened 2018-10-26
Spinner

kamilkisiela:
codegen-spinner

It logs every step of the process and when something fails it shows the last log on red.


dotansimha:
@kamilkisiela , ready to merge?

PR #756, opened 2018-10-29
resolvers: generic type for parent

kamilkisiela:
Uses `never` on root types
Otherwise uses parent's type as default


dotansimha:
Looks great!

PR #757, opened 2018-10-29
Mappers (in typescript-resolvers template)

kamilkisiela:
```js
{
  mappers: {
    User: './interfaces#UserParent',
    Post: 'string' // or even 'Post'
  }
}
```

I think tests show exactly what we want to achieve.

### DIFF
--- a/packages/templates/typescript-mongodb/package.json
+++ b/packages/templates/typescript-mongodb/package.json
@@ -16,7 +16,8 @@
     "graphql-codegen-core": "0.12.6"
   },
   "dependencies": {
-    "lodash": "4.17.11"
+    "lodash": "4.17.11",
+    "graphql-codegen-typescript-template": "0.12.6"
   },
   "main": "./dist/index.js",
   "typings": "dist/index.d.ts",
--- a/packages/templates/typescript-mongodb/src/helpers/entity-fields.ts
+++ b/packages/templates/typescript-mongodb/src/helpers/entity-fields.ts
@@ -1,6 +1,6 @@
 import { Field, Interface, Type } from 'graphql-codegen-core';
 import { set } from 'lodash';
-import { getResultType } from '../../../typescript/src/utils/get-result-type';
+import { convertedType } from 'graphql-codegen-typescript-template';
 
 // Directives fields
 const ID_DIRECTIVE = 'id';
@@ -27,7 +27,7 @@ function appendField(obj: object, field: string, value: string, mapDirectiveValu
 type FieldsResult = { [name: string]: string | FieldsResult };
 
 function buildFieldDef(type: string, field: Field, options: Handlebars.HelperOptions): string {
-  return getResultType(
+  return convertedType(
     {
       ...field,
       type
--- a/packages/templates/typescript-resolvers/package.json
+++ b/packages/templates/typescript-resolvers/package.json
@@ -9,8 +9,10 @@
     "pretest": "yarn build",
     "test": "codegen-templates-scripts test"
   },
+  "dependencies": {
+    "graphql-codegen-typescript-template": "0.12.6"
+  },
   "devDependencies": {
-    "graphql-codegen-typescript-template": "0.12.6",
     "codegen-templates-scripts": "0.12.6",
     "graphql-codegen-core": "0.12.6",
     "graphql-codegen-compiler": "0.12.6",
--- a/packages/templates/typescript-resolvers/src/helpers/import-mappers.ts
+++ b/packages/templates/typescript-resolvers/src/helpers/import-mappers.ts
@@ -1,0 +1,41 @@
+import { Type } from 'graphql-codegen-core';
+import { pickMapper } from './mappers';
+
+interface Modules {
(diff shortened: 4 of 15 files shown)


packages/templates/typescript-mongodb/src/helpers/entity-fields.ts
```diff
@@ -1,6 +1,6 @@
 import { Field, Interface, Type } from 'graphql-codegen-core';
 import { set } from 'lodash';
-import { getResultType } from '../../../typescript/src/utils/get-result-type';
```
kamilkisiela: I noticed we use relative path instead of a package

packages/templates/typescript-mongodb/package.json
```diff
@@ -16,7 +16,8 @@
     "graphql-codegen-core": "0.12.6"
   },
   "dependencies": {
-    "lodash": "4.17.11"
+    "lodash": "4.17.11",
+    "graphql-codegen-typescript-template": "0.12.6"
```
kamilkisiela: I added it as a dependency because mongodb template depends on it

packages/templates/typescript-resolvers/package.json
```diff
@@ -9,8 +9,10 @@
     "pretest": "yarn build",
     "test": "codegen-templates-scripts test"
   },
+  "dependencies": {
+    "graphql-codegen-typescript-template": "0.12.6"
```
kamilkisiela: same here, we depend on it

packages/templates/typescript-resolvers/src/helpers/import-mappers.ts
```diff
@@ -0,0 +1,41 @@
+import { Type } from 'graphql-codegen-core';
+import { pickMapper } from './mappers';
+
+interface Modules {
+  [path: string]: string[];
+}
+
+export function importMappers(types: Type[], options: Handlebars.HelperOptions) {
```
kamilkisiela: It doesn't emit unused types

packages/templates/typescript-resolvers/src/helpers/mappers.ts
```diff
@@ -0,0 +1,39 @@
...
+  [key: string]: string;
+}
+
+export interface Mapper {
+  isExternal: boolean;
+  type: string;
+  source?: string;
+}
+
+function isExternal(value: string) {
```
kamilkisiela: this says if Mapper comes from external module

packages/templates/typescript/src/utils/get-result-type.ts
```diff
@@ -26,16 +20,28 @@ export function getResultType(type: Field, options: Handlebars.HelperOptions, sk
...
-    if (type.isRequired) {
+    if (field.isRequired) {
       return realType;
     } else {
       return [realType, 'null'].join(' | ');
     }
   }
 }
+
+export function convertedType(type: Field, options: Handlebars.HelperOptions, skipPascalCase = false) {
```
kamilkisiela: The resolvers template needs to share that logic

packages/templates/typescript/src/utils/get-result-type.ts
```diff
@@ -1,22 +1,16 @@
 import { pascalCase } from 'change-case';
 import { Field } from 'graphql-codegen-core';
 
-export function getResultType(type: Field, options: Handlebars.HelperOptions, skipPascalCase = false) {
-  const baseType = type.type;
-  const underscorePrefix = type.type.match(/^[\_]+/) || '';
+export function getFieldType(field: Field, realType: string, options: Handlebars.HelperOptions) {
```
kamilkisiela: I extracted that because we need that logic in resolvers-template but without prefixes and stuff (things that convertedType has)

packages/templates/typescript-resolvers/tests/resolvers.spec.ts
```diff
@@ -424,4 +424,148 @@ describe('Resolvers', () => {
       export type UserNameResolver<R = string | null, Parent = User, Context = any> = Resolver<R, Parent, Context>;
     `);
   });
+
+  it('should accept a map of parent types', async () => {
```
kamilkisiela: Here it shows that fields and types get a mapper both in Result and Parent generics

packages/templates/typescript-resolvers/tests/resolvers.spec.ts
```diff
@@ -424,4 +424,148 @@ describe('Resolvers', () => {
...
+        }
+
+        export type IdResolver<R = string | null, Parent = UserParent, Context = any> = Resolver<R, Parent, Context>;
+        export type NameResolver<R = string | null, Parent = UserParent, Context = any> = Resolver<R, Parent, Context>;
+        export type PostResolver<R = PostParent | null, Parent = UserParent, Context = any> = Resolver<R, Parent, Context>;
+      }
+    `);
+  });
+
+  it('should accept mappers that reuse generated types', async () => {
```
kamilkisiela: Here I show that we can reuse generated types and even use primitives

Issue #688, opened 2018-10-12
typescript template can't override the ID type


reconbot:
**Describe the bug**
I have a scaylr

```graphql
"Scalar type to support our Graph DB IDs, which are integers capped at 4294967296"
scalar ID
```

It can be a string or a number

I want the types to be a string or a number.

```
type Post {
  id: ID!
}
``

Becomes
```typescript
interface Post {
  readonly id: string
}
```

**To Reproduce**
1. Use '...' as schema
```
scalar ID
type Post {
  id: ID!
}
```

2. Use '...' as documents 
(disabled)

3. Run the following command: `gql-gen ...`
```
await generate(
    {
      schema: 'graphql/**/*.graphql',
      template: 'graphql-codegen-typescript-template',
      skipDocuments: true,
      config: 'generators/graphqlCodeGeneratorConfig.json',
      overwrite: true,
      out: 'graphql/types.d.ts',
    },
    true
  )
```

```
{
  "generatorConfig": {
    "scalars": {
      "ID": "string | number"
    },
    "immutableTypes": true,
    "resolvers": false,
    "enumsAsTypes": true,
    "schemaNamespace": "GraphQL"
  }
}

```

**Expected behavior**
I'd get a typescript def similar to 

```typescript
interface Post {
  readonly id: string | number
}
```

**Schema/Documents**
If possible and relevant, please provide your GraphQL schema and GraphQL documents.

**Environment:**
 - OS: osx, linux
 - Node: 8

```
│ ├─┬ graphql-codegen-compiler@0.12.5
│ │ ├── graphql-codegen-core@0.12.5 deduped
│ ├─┬ graphql-codegen-core@0.12.5
├── graphql-codegen-typescript-template@0.12.6
│ ├─┬ escodegen@1.11.0
```
**Additional context**

I've tried using a type I've got called `GUID` instead of the union of string and number, still no dice.

Sorry I don't have the cli, it was never an option for my project so I don't know how to use it.

Thanks!

reconbot:
I figured it out it's considered a primitive not a scalar. So if you're brave you can hack it like so.

```
// hack the global require cache for not exposed config options
const template = require('graphql-codegen-typescript-template')
template.default.primitives.ID = 'GUID'
```

That's obviously brittle so I'd like to request we can configure the primitive types.

ardatan:
Did you try to use "primitives" field of gqlgen instead of scalars; because ID is not scalar AFAIK?

[...]

Issue #732, opened 2018-10-25
Remove `MULTIPLE_FILES` from code


dotansimha:
It was already deprecated a few month ago, it's time to remove it and clean some parts of the code.

dotansimha:
@desbrowne 
So without `MULTIPLE_FILES` you can get the same functionality with a custom template. The return value of the custom template function should return an array of `FileOutput`, so you should be able to iterate all `documents.operations` and create a matching file.

Here are things that might help you:
- [Example for custom template as package](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/templates/introspection/src/index.ts)
- As you can see in the first link, you should follow the `CustomProcessingFunction` interface for you function, and [this is it's signature](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-core/src/types.ts#L307-L311).
- You should be able to get the name of the file from `mergedDocuments`, it includes `operations` and `fragments`. So you can get from there the queries and the filenames.
- Your custom template function can be async, and should return `FileOutput[]` to write to the fs.

dotansimha:
Done in 0.14 🎉

Issue #731, opened 2018-10-25
Deprecate `PROJECT` generation


dotansimha:
`PROJECT` generation was an experiment we did, that intended to generate an existing project directly. 
It has a very messy API, and has a lot of bugs. Also, now that everyone can write his own template (with our without Handlebars), it makes it much easier to do it.

robck:
The --project option works well for me so far. Is having an internal/nested package the only other option?

dotansimha:
@robck we decided to deprecate it in favor of custom templates (which already available now). IT should be more powerful, and easier to customize.

dotansimha:
Done in 0.14 🎉

Issue #725, opened 2018-10-23
Plugin system instead of multiple templates


kamilkisiela:
_more soon_

dotansimha:
YES! 
You mean something like that? use typescript template, add "resolvers" plugins (which is just another template), add "react" plugin and output it to the same file.

kamilkisiela:
Yes, Codegen gets types, interfaces and stuff and you decide what to output.

Typescript gets a type and outputs a TS interface with some naming convention.
Apollo Angular knows that is using typescript and `import { typeToName } from 'graphql-codegen-typescript-plugin';`

kamilkisiela:
So they could interact and we no longer have to have one template

kamilkisiela:
I think it will make codegen waay more interesting (BUT IT IS INTERESTING HUUUUUGLY NOW FOR SURE)

kamilkisiela:
```typescript
function typeToName(type) { return pascalCase(type.name); }

class Typescript {
  visitType() { return `interface ${typeToName} { ... }`; }
}
```

You get the idea

dotansimha:
Hmm wait, I think there are multiple things here.

When you say plugins, I'm referring to plugins adding more template to a base template (or, actually, just a list of templates you want to merge together).

The `typeToName` is something else - it means implementing the codegen as visitor pattern, and it something I want to do for a long time, but it might break some stuff, and we'll have to do a lot of refactoring to the core package and templates

kamilkisiela:
Yes, that's why wrote _But that might be for a different issue_

For now having multiple templates seems ok but it would be super nice to use AST or something like this and not handlebars so we could share logic between packages.

dotansimha:
Done in 0.14 🎉

PR #764, opened 2018-10-30
New API & Website

dotansimha:
## Overview

At the moment, we have some issues with the current API: 
- Usage becomes complicated because of tons of CLI flags
- To use multiple templates you need to run the codegen more than once (all schema and documents are loaded twice)
- Difficult to pass configuration to templates
- Templates becomes huge and complicated

## New API Suggestion
- YML based configuration, that allow to create “flows”
- Each execution of the codegen creates a single file, with multiple pieces of templates inside it - “plugins”.
- Developer declare the output file, and then the contents of it.
- Get rid of the “local schema” and “remote schema” concept - just multiple schemas can be loaded to the codegen
- Multiple templates can be generated easily
- Custom template are easy to use (just npm package)
- Handlebars will be a dependency of the templates only
- Easy to configure
- Some parts of the existing templates configuration could change to be a standalone plugin (For example, “prepend” could be a plugin that get configuration).

**We hope that all of the changes mentioned above won't effect the existing usage of the codegen, the existing API will remain as-is.**

### TypeScript example for server side:

```yml
schemas:
   - http://localhost:3000/graphql
documents:
   - "**/*.graphql"
generates:
   - src/models.ts:
        - typescript-common
        - typescript-server
```

### TypeScript example for client-side and server side with multiple “plugins”:

```yml
schemas:
   - http://localhost:3000/graphql
documents:
   - "**/*.graphql"
generates:
   - server/src/models.ts:
        - typescript-common
        - typescript-resolvers
        - typescript-server
        - typescript-mongo

   - client/src/models.ts:
        - typescript-common
        - typescript-client
        - documents-import-modules
        - react-components
```

### With cool plugins and configuration:

```yml
schemas:
   - http://localhost:3000/graphql
documents:
   - "**/*.graphql"
generates:
   - server/src/models.ts:
        - time
        - add:
             - "import { MyContext } from "./context.ts"
        - typescript-common:
             constEnums: true
        - typescript-resolvers:
             contextType: "MyContext"
        - typescript-server
        - typescript-mongo

   - client/src/models.ts:
        - print-generation-time
        - typescript-common
        - typescript-client:
              immutable: true
        - documents-import-modules
        - react-components
```

### With per-output configuration (could add schemas and add documents to the base):

```yml
schemas:
   - http://localhost:3000/graphql

generates:
   - server/src/models.ts:
        - time
        - add:
               - "import { MyContext } from "./context.ts"
        - typescript-common:
               constEnums: true
        - typescript-resolvers:
               contextType: "MyContext"
        - typescript-server
        - typescript-mongo

   - client/src/models.ts:
        schemas:
          "client/local-schema.graphql"
        documents: 
          "client/**/*.graphql"
        plugins:
           - time
           - typescript-common
           - typescript-client:
                  immutable: true
           - documents-import-modules
           - react-components
```

## Tasks
- [x] Create TypeScript interfaces for the config file, and read it from json/yaml (https://github.com/dotansimha/graphql-code-generator/issues/722)
- [x] Plugins System (https://github.com/dotansimha/graphql-code-generator/issues/725)
- [x] Create "new codegen" executor based on config file (with normalized yml file)
- [x] Remove "multiple" (https://github.com/dotansimha/graphql-code-generator/issues/732) and "project"  (https://github.com/dotansimha/graphql-code-generator/issues/731)
- [x] Remove `graphql-codegen-compiler` (important stuff moved to helpers packages)
- [x] Change existing templates to `plugins`
- [x] Separate TypeScript template to `common` (only enums and input types) / `client` / `server`
- [x] Add `add`, `time`, `schema-ast` (https://github.com/dotansimha/graphql-code-generator/issues/696) plugins.
- [x] Allow to override base types (also fixes https://github.com/dotansimha/graphql-code-generator/issues/688)
- [x] Allow to override the pascalCase naming with config. 
- [x] Fix basic `core` tests 
- [x] Rebase to latest @kamilkisiela 's changes 
- [x] Plugins Tests
- [x] Adde `CodegenPlugin` interface with `{ plugin, addToSchema?, validate? }` signature
- [x] Add prettier as per-file feature
- [x] Use listr (https://github.com/dotansimha/graphql-code-generator/pull/797)
- [x] Fix watch mode (https://github.com/dotansimha/graphql-code-generator/pull/797)
- [x] Allow `string` for documents and schema (easier to test and to use with programatic usage)
- [x] Refactor document loaders (required for https://github.com/dotansimha/graphql-code-generator/issues/534)
- [ ] New Codegen tests
- [x] Add support for custom document loaders (https://github.com/dotansimha/graphql-code-generator/issues/534)
- [x] Add support for custom schema loaders (https://github.com/dotansimha/graphql-code-generator/issues/534)
- [x] Fix issue with Mongo template (https://github.com/dotansimha/graphql-code-generator/issues/827)
- [x] See if we can make the codegen agnostic again (to run it in browser 😎 )
- [x] ~~Webpack plugin~~
- [x] Backward compatibility
- [x] Website
- [ ] README


packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -1,587 +1,483 @@
...
-    config = JSON.parse(fs.readFileSync(configPath).toString()) as GqlGenConfig;
-    debugLog(`[executeWithOptions] Got project config JSON: ${JSON.stringify(config)}`);
-  }
+    /* Normalize root "schema" field */
+    schemas = normalizeInstanceOrArray<Types.Schema>(config.schema);
 
-  if (project && project !== '') {
-    getSpinner().log(`Using project: ${project}`);
-    if (config === null) {
+    if (schemas.length === 0) {
```
kamilkisiela: @dotansimha Why do we expect `codegen.yml` to have global schemas? In 🇫🇷 🇩🇪 🔌 project, for example, we won't have one schema for all modules but one for each.

kamilkisiela: #832

PR #788, opened 2018-11-06
Website

DAB0mB:


dotansimha:
@DAB0mB I think you should create this PR directly to v2 docs, because it includes the new API (instead of master)

PR #791, opened 2018-11-06
Implement --config flag and use yml files to generate all examples

kamilkisiela:
It checks if there's a flag `--config` or just `-c` and if it has a string as the flag's value and if it has nothing more than that. 

If so, it checks if the file exists and loads it as a config.

If not, it takes `codegen.yml`, and `codegen.json` after that to check if one of them exists.

If both are not defined, it parses `process.argv`.

```bash
$ gql-gen --config ./configs/codegen.yml
```


dotansimha:
LOVE IT

PR #792, opened 2018-11-06
Obey .prettierignore file

meshulam:
It was a little surprising when my team started experimenting with prettier and our generated typefiles changed. It would be nice to have some way to disable prettier, and rather than add another config option, I thought we could just use the existing `.prettierignore` file that prettier CLI supports.

This assumes that `.prettierignore` is in the current directory when executing gql-gen, but that's the same behavior as the prettier CLI, so it seemed OK to me.

I adapted this from Gitlab: https://gitlab.com/gitlab-org/gitlab-ce/blob/master/scripts/frontend/prettier.js#L90


dotansimha:
Thanks @meshulam !

PR #797, opened 2018-11-07
Listr

kamilkisiela:
**Single run:**
render1541582257883

**Watch mode:**
render1541592791006


packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -65,87 +68,211 @@ async function mergeSchemas(schemas: GraphQLSchema[]): Promise<GraphQLSchema> {
...
-  const result = [];
+  const result: FileOutput[] = [];
+  const commonListrOptions = {
+    exitOnError: true
+  };
+  const verboseOptions = {
+    ...commonListrOptions,
+    renderer: 'verbose',
+    nonTTYRenderer: 'verbose'
+  };
```
kamilkisiela: In verbose mode, it won't show every frame, only `task started`, `task completed` etc.

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -65,87 +68,211 @@ async function mergeSchemas(schemas: GraphQLSchema[]): Promise<GraphQLSchema> {
 }
 
 export async function executeCodegen(config: Types.Config): Promise<FileOutput[]> {
-  const result = [];
+  const result: FileOutput[] = [];
+  const commonListrOptions = {
+    exitOnError: true
```
kamilkisiela: It doesn't mean `process.exit()`, it just stops other tasks when something failed.

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -65,87 +68,211 @@ async function mergeSchemas(schemas: GraphQLSchema[]): Promise<GraphQLSchema> {
...
+
+  listr.add({
+    title: 'Load documents',
+    enabled: ctx => ctx.hasDocuments,
+    task: async () => {
+      try {
+        await loadRootDocuments();
+      } catch (error) {
+        if (!config.watch) {
+          throw error;
```
kamilkisiela: if it's not in watch, we pass it to listr

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -65,87 +68,211 @@ async function mergeSchemas(schemas: GraphQLSchema[]): Promise<GraphQLSchema> {
...
+        } else {
+          cliError(error, true);
```
kamilkisiela: I think we should have here something different in watch mode, not process.exit but some info

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -65,87 +68,211 @@ async function mergeSchemas(schemas: GraphQLSchema[]): Promise<GraphQLSchema> {
...
+              {
+                // it stops when one of tasks failed
+                exitOnError: true
+              }
+            );
+          }
+        })),
+        {
+          // it doesn't stop when one of tasks failed, to finish at least some of outputs
+          exitOnError: false,
```
kamilkisiela: it won't stop when some task failed

packages/graphql-codegen-cli/src/loaders/documents/document-loader.ts
```diff
@@ -21,12 +21,15 @@ export const loadFileContent = (filePath: string): DocumentNode | null => {
       return null;
     }
   } else {
-    throw new DetailedError(`
+    throw new DetailedError(
+      'Failed to load a document',
```
kamilkisiela: with that, we gain shorter error messages that are displayed next to the task in listr

packages/graphql-codegen-cli/src/loaders/documents/document-loader.ts
```diff
@@ -21,12 +21,15 @@ export const loadFileContent = (filePath: string): DocumentNode | null => {
...
+      `
 
       Failed to load a document.
       Document file ${filePath} does not exists.
```
kamilkisiela: this long message, is being displayed at the end

packages/graphql-codegen-cli/src/utils/listr-renderer.ts
```diff
@@ -0,0 +1,62 @@
...
+import * as identString from 'indent-string';
+import * as UpdateRenderer from 'listr-update-renderer';
+import { DetailedError, isDetailedError } from '../errors';
+import { ListrTask } from 'listr';
+
+export class Renderer {
+  private updateRenderer: any;
+
+  constructor(tasks: ListrTask, options: any) {
+    this.updateRenderer = new UpdateRenderer(tasks, options);
```
kamilkisiela: I could extend Renderer with UpdateRenderer but it throws an error... that's why I did it this way.

PR #798, opened 2018-11-07
Use pascalcase for enums

### LINKED ISSUE #796: Make Enums keys a PascalCase by default
If you take a look at typescript docs, you can see that Enums are PascalCase.
But here when you have enum generated, they are taken as is from the schema.
What do you think to generate left side as PascalCase while right side to be the actual value from the schema enum definition?

We have a huge codebase and many enums are already PascalCase having them mixed is rather a pain in the ass.

**Schema:**
```graphql
enum SomeEnum {
  COOL
  NOT_COOL
}
```
**Current:**
```typescript
enum SomeEnum {
  COOL = "COOL",
  NOT_COOL = "NOT_COOL"
}
```
**Proposed:**
```typescript
enum SomeEnum {
  Cool = "COOL",
  NotCool = "NOT_COOL"
}
```

What do you think?
Maybe an option?

ardatan:
Fixes #796

### DIFF
--- a/dev-test/githunt/types.avoidOptionals.ts
+++ b/dev-test/githunt/types.avoidOptionals.ts
@@ -142,15 +142,15 @@ export interface CommentAddedSubscriptionArgs {
 
 /** A list of options for the sort order of the feed */
 export enum FeedType {
-  HOT = 'HOT',
-  NEW = 'NEW',
-  TOP = 'TOP'
+  Hot = 'HOT',
+  New = 'NEW',
+  Top = 'TOP'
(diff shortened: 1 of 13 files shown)


dotansimha:
@ardatan, this is great, and answers @RIP21 needs, but I might be a breaking change for developers that depend on the current generated names.

We are going to introduce some breaking changes in v2 api, so it's probably better to include this feature only from v2 and not now. 
 
@RIP21, you can use this feature now using the published alpha version.

PR #800, opened 2018-11-07
Allow subscription resolver to return a promise

dbuezas:
Explicitly to match this use:

https://github.com/graphql-boilerplates/typescript-graphql-server/blob/master/advanced/src/resolvers/Subscription.ts#L6

which is here defined to return a promise:
https://github.com/graphql-boilerplates/typescript-graphql-server/blob/master/advanced/src/generated/prisma.ts#L32

### DIFF
--- a/dev-test/test-schema/resolvers-types.ts
+++ b/dev-test/test-schema/resolvers-types.ts
@@ -14,7 +14,7 @@ export interface ISubscriptionResolverObject<Result, Parent, Context, Args> {
     args: Args,
     context: Context,
     info: GraphQLResolveInfo
-  ): AsyncIterator<R | Result>;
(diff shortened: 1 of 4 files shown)


dbuezas:
in case somebody else needs an easy "fix", this does the trick until this is published:


```
import { SubscriptionResolvers } from '../generated/resolvers-types'

type hacked = unknown // this https://github.com/dotansimha/graphql-code-generator/pull/800 will fix type error

export default {
  newLink: {
    subscribe: (_parent, _args, context, info) => {
      return context.db.subscription.link({ where: { mutation_in: ['CREATED'] } }, info) as hacked
    },
  } as SubscriptionResolvers.NewLinkResolver,
}
```

dotansimha:
Awesome! Thanks @dbuezas !

PR #806, opened 2018-11-07
Second generic added to HoCs to be able to specify return type of `props` callback

### LINKED ISSUE #803: Second generic should be added to HoCs to be able to specify return type of `props` callback.
Right now there is only one generic that gives the ability to specify `ownProps` but there is no optional second generic that will give the ability to specify `props` callback return type.

So the result will be like that:
```typescript
export function GetMessagesHOC<TProps = any, TChildProps = any>(
  operationOptions:
    | ReactApollo.OperationOption<
        TProps,
        GetMessagesQuery,
        GetMessagesVariables,
        Partial<ReactApollo.DataProps<GetMessagesQuery, GetMessagesVariables>> & TChildProps
      >
    | undefined,
) {
  return ReactApollo.graphql<
    TProps,
    GetMessagesQuery,
    GetMessagesVariables,
    Partial<ReactApollo.DataProps<GetMessagesQuery, GetMessagesVariables>> & TChildProps
  >(GetMessagesDocument, operationOptions)
}
```

So effectively `PropsCallback` optional generic added. And then `&` with `Partial` in both places.

RIP21:
Fixes #803


RIP21:
I didn't run `dev-test` since it causes a lot of changes. But it effectively generates what is needed. (like in issue mentioned)

PR #810, opened 2018-11-08
v2: Fix dependencies in plugins

kamilkisiela:


packages/plugins/add/package.json
```diff
@@ -9,11 +9,13 @@
     "build": "codegen-templates-scripts build",
     "pretest": "yarn build"
   },
-  "dependencies": {},
+  "peerDependencies": {
+    "graphql-codegen-core": "^0.13.0"
```
kamilkisiela: @dotansimha peer or direct? When direct dep we might end up like `apollo-link`, instances with the same name but of different versions. We might avoid that because we probably doesn't check anything by `instanceof`.

With peer, we require graphql-codegen-core to be installed, probably with `graphql-code-generator` that has it. I'm not sure it would work though, need to check.

So either direct or peer.

packages/plugins/time/package.json
```diff
@@ -9,9 +9,10 @@
     "build": "codegen-templates-scripts build",
     "pretest": "yarn build"
   },
-  "dependencies": {},
+  "dependencies": {
+    "moment": "2.22.2"
```
kamilkisiela: I added it to dependencies and also did set it in webpack config as external module.

packages/plugins/typescript-apollo-angular/package.json
```diff
@@ -10,13 +10,16 @@
     "test": "codegen-templates-scripts test",
     "pretest": "yarn build"
   },
+  "peerDependencies": {
+    "graphql-tag": "^2.0.0"
```
kamilkisiela: since we use it

packages/plugins/typescript-apollo-angular/package.json
```diff
@@ -10,13 +10,16 @@
...
+  },
   "dependencies": {
     "dependency-graph": "0.7.2",
+    "graphql-codegen-plugin-helpers": "0.13.0",
```
kamilkisiela: this plugin is missing in all plugins because webpack treats it as an external module but it was a devDep, so not included in the project.

packages/plugins/typescript-mongodb/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { TypeScriptCommonConfig, initCommonTemplate } from 'graphql-codegen-typescript-common';
-import { PluginFunction, DocumentFile, schemaToTemplateContext } from 'graphql-codegen-core';
+import { PluginFunction, DocumentFile } from 'graphql-codegen-core';
```
kamilkisiela: unused

packages/plugins/typescript-server/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { TypeScriptCommonConfig, initCommonTemplate } from 'graphql-codegen-typescript-common';
-import { PluginFunction, DocumentFile, schemaToTemplateContext } from 'graphql-codegen-core';
+import { PluginFunction, DocumentFile } from 'graphql-codegen-core';
```
kamilkisiela: unused

packages/utils/codegen-templates-scripts/webpack.config.js
```diff
@@ -19,7 +19,8 @@ module.exports = {
     'graphql-codegen-core',
     'graphql-tag',
     'lodash',
-    'graphql-codegen-plugin-helpers'
+    'graphql-codegen-plugin-helpers',
+    'moment'
```
kamilkisiela: As I said, now it's treated as an external module, not included in a bundle generated by webpack.

packages/plugins/add/package.json
dotansimha: Hmm, `peer` ! @kamilkisiela

PR #814, opened 2018-11-08
Export props for HOC

ardatan:


RIP21:
Take a look at #803 This props should have generic to be able to extend them using the second generic argument to HoC.
So if you can merge those two PRs together it will be perfect.

PR #821, opened 2018-11-09
chore(template): mention its autogenerated in template

maapteh:
Make new people aware its autogenerated and should not be edited. Its a little more explicit like yarn lock :)


dotansimha:
Awesome! Thanks @maapteh.

We are currently refactoring the API, and we also adding a custom plugin that allow developers to add custom content to each generated file (see https://github.com/dotansimha/graphql-code-generator/pull/764). The plugin is called `add`. 
So I think after merging v2, we won't need this as part of the template, but we probably document how to use this plugin, and this is a great example.

PR #824, opened 2018-11-09
Fix documentation links

Notuom:
Links to `codegen-templates-scripts` were incorrectly `handlebars-templates-scripts`.

Also, a link to an external repo was pointing to a file that has since been deleted.


dotansimha:
Thanks @Notuom !

PR #839, opened 2018-11-13
V2: importable context type. {} as default context and also instead of every `never`

kamilkisiela:
`never` is hard to work with when you mock things...

`{}` as default context makes sense because with `any` we won't get any ts warnings, it's more strict (breaking).
`{}` for empty arguments also makes sense since GraphQL provides an empty object when there's no arguments.


dotansimha:
@kamilkisiela Maybe `mappers` can only can a mapping between a GraphQL `type` and a TypeScript type? and then in order to add the imports, developers can use `add` plugin. 
What do you think?

PR #884, opened 2018-11-19
Add failing test for problematic parsing of files.

Jontem:
Related to #885


dotansimha:
Thanks @Jontem , it was fixed and will be released soon!

PR #907, opened 2018-11-22
Fix gql template literals not being plucked correctly

DAB0mB:
Now you can get nuts with `graphql-tag` and it will most likely work.


packages/graphql-codegen-cli/src/utils/strip-comments.ts
```diff
@@ -1,72 +0,0 @@
...
-  }
-
-  for (const comment of comments) {
-    input = removeComment(input, comment, pos);
-  }
-
-  return input;
-};
-
-export default stripComments;
```
DAB0mB: This module is not necessary anymore, so I removed it and everything related to it.

PR #910, opened 2018-11-22
Resolvers for unions, interfaces and directives

kamilkisiela:
- [x] union
- [x] interface
- [x] directive for graphql-tools (old)
- [x] directive for graphql-tools (new) (there's no point of doing it, people can use *Args and Context)
- [x] scalar (no idea what we could generate here that would be helpful)

### DIFF
--- a/packages/plugins/typescript-resolvers/src/directive.handlebars
+++ b/packages/plugins/typescript-resolvers/src/directive.handlebars
@@ -1,0 +1,13 @@
+{{ toComment description }}
+export type {{ convert name }}DirectiveResolver<Result> = DirectiveResolverFn<Result, {{#if hasArguments}}{{ convert name }}DirectiveArgs{{else}}{}{{/if}}, {{{ getContext }}}>;
+
+{{~# if hasArguments }}
+
+export interface {{ convert name }}DirectiveArgs {
+{{#each arguments}}
+  {{ toComment description }}
+  {{ name }}{{ getOptionals this }}: {{ convertedType this }};
+{{/each}}
+}
+
+{{/if}}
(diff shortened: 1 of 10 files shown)


dotansimha:
Scalars can just have an implementation of `GraphQLScalarType` for now

kamilkisiela:
Yeah but how we can predict the type there? We can't tell if it's a string of number so what we can generate is a regular `GraphQLScalarType`, something that could be imported from `graphql`

kamilkisiela:
Unions and Interfaces:

```graphql
```

```ts
export type TypeResolveFn<Types, Parent = {}, Context = {}> = (
  parent: Parent,
  context: Context,
  info: GraphQLResolveInfo
) => Maybe<Types>;

export namespace EntryResolvers {
  export interface Resolvers {
    __resolveType: ResolveType;
  }

  export type ResolveType<
    R = 'Post' | 'Comment',
    Parent = Post | Comment, 
    Context = {}
  > = TypeResolveFn<R, Parent, Context>;
}
```

Directives:

```graphql
directive @modify(limit: Int) on FIELD_DEFINITION
```

```ts
export type NextResolverFn<T> = () => Promise<T>;

export type DirectiveResolverFn<TResult, TArgs = {}, TContext = {}> = (
  next: NextResolverFn,
  source: any,
  args: TArgs,
  context: TContext,
  info: GraphQLResolveInfo,
) => TResult | Promise<TResult>;

export type ModifyDirectiveResolver<Result> = DirectiveResolverFn<Result, ModifyDirectiveArgs, {}>;

export interface ModifyDirectiveArgs {
  limit?: number | null;
}
```

dotansimha:
Looks great!

Issue #713, opened 2018-10-19
Suggestion: Use Pick<T, K> to generate query types


fetis:
**Is your feature request related to a problem? Please describe.**
Currently, we just have a copy of original type for a query

**Describe the solution you'd like**
I think a more clear approach would be to use Pick<T, K> operator to make a connection to the original type. 

So this would be like this
```
export interface TypeA {
  id: string;
  name: string;
  age: number;
}

export namespace MyQuery {
  export type TypeA = Pick<TypeA, 'id' | 'name'>
}
```

See it's in action here https://medium.com/@curtistatewilkinson/typescript-2-1-and-the-power-of-pick-ff433f1e6fb

ardatan:
Thank you for your suggestion!
That looks a good idea, but how about nested types.
‘’’gql
query SomeQuery {
   foo
   bar
   something {
               x
                y
    }
}
‘’’

fetis:
@ardatan I think it should work the same `Pick<SomeQuery, 'foo', 'something'>`

fetis:
[...]

PR #917, opened 2018-11-24
Flow plugins

dotansimha:
The idea behind this PR is to create a Flow plugin. We will keep the idea of separating it to multiple plugins.
One of the reason behind these plugins is to test new methods to write Codegen plugins.

We are testing the `visit` method of `graphql` to see if we can use this to transform the AST into an output. Still not sure it's perfect and matches all our needs.

We would also like to test the option to use `Pick` instead of generating new types for the client side, and this way to create types aliases to the original GraphQL types. If it will work as expected, we'll also need to change the packages to be: `flow-types` (merge server and common, because now client will also need the original typings) and `flow-client`. (Suggested in https://github.com/dotansimha/graphql-code-generator/issues/713)

### `flow`
- [x] input type
- [x] enum
- [x] scalar
- [x] type
- [x] interface
- [x] union
- [x] arguments
- [x] Support custom naming convention
- [x] Custom Prefix

### `flow-documents`
- [x] basic export for query/mutation/subscription
- [x] Selection Set
- [x] Operation variables
- [x] Fragment definition
- [x] Support interfaces
- [x] Support unions
- [x] Inline fragment in selection set
- [x] Fix selections set arrays 
- [x] Fragment spread in selection set
- [x] Support `__typename`
- [x] Support custom naming convention
- [x] Custom Prefix

### `flow-resolvers`
- [x] Resolvers signature with custom mapping

### Documentation
- [x] README pointing to the website
- [x] Website docs 
- [x] Update live demo


dotansimha:
We can use `$Pick` utility to build the selection, that also means that we don't really need to separate it to 3 packages. We can use `flow` (all schema parts) and `flow-documents` (documents, based on the types declared in `flow` plugin).

PR #922, opened 2018-11-26
Add plugin's addition to the schema before executing plugins

### LINKED ISSUE #919: [0.14.2] `addToSchema` need to run before all plugins
Instead of applying it per-plugin, it should be applied per-file, and only then to execute it.

How to reproduce this?
Use `typescript-common` and `typescript-mongo` together, and use a directive. 
The schema that `typescript-common` gets does not contain the directives `typescript-mongo` adds. 

It requires some refactoring, because that means that we need to load the plugin before executing it, collect all `addToSchema`, built a schema, and then trigger all plugins with the same shared schema.

kamilkisiela:
Fixes #919


kamilkisiela:
Available at `v0.15.0-alpha.67cbcb66`

PR #949, opened 2018-11-29
Implement general resolvers interface

ardatan:


packages/plugins/typescript-resolvers/src/root.handlebars
```diff
@@ -65,4 +65,25 @@ export type DirectiveResolverFn<TResult, TArgs = {}, TContext = {}> = (
...
+    {{ convert name }}?: {{ convert name }}Resolvers{{#unless @root.config.noNamespaces}}.{{/unless}}Resolvers;
+  {{/each}}
+  {{#each interfaces}}
+    {{ convert name }}?: {{ convert name }}Resolvers{{#unless @root.config.noNamespaces}}.{{/unless}}Resolvers;
+  {{/each}}
+  {{#each unions}}
+    {{ convert name }}?: {{ convert name }}Resolvers{{#unless @root.config.noNamespaces}}.{{/unless}}Resolvers;
+  {{/each}}
+  {{#each scalars}}
+    {{ convert name }}?: GraphQLScalarType;
```
kamilkisiela: @ardatan scalars have now their own resolvers too, maybe we could update that line, what do you think?

ardatan: We have ScalarTypeConfig interface, but I couldn't find out the way to use it here, because resolvers take ScalarType which doesn't have any generic.

kamilkisiela: @ardatan I did some work for scalars recently: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript-resolvers/src/scalar.handlebars

PR #952, opened 2018-11-29
Fix multiline comments

### LINKED ISSUE #950: Multiline comments are squashed without whitespace
**Describe the bug**
The following gql:

```gql
  # Decides the order a module will run in. Modules with the same
  # level, will run in parallel.
  level: Int!
```

turns into this in typescript

```ts
  /** Decides the order a module will run in. Modules with the samelevel, will run in parallel. */
  level: number;
```

Notice that `same` and `level` get's merged to `samelevel`.

**Expected behavior**

I expected the comment to either be multiline, or be

```ts
/** Decides the order a module will run in. Modules with the same level, will run in parallel. */
```

**Environment:**
 - OS: OSX
 - Codegen: v0.14.4
 - Node: v11.2.0

**Additional context**
I've looked into the code a bit, and in all likelihood, changing `oneLineTrim` to `oneLine` will fix this issue:

https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-handlebars-helpers/src/index.ts#L2

Alxandr:
Resolves #950

### DIFF
--- a/dev-test/test-schema/schema.graphql
+++ b/dev-test/test-schema/schema.graphql
@@ -8,6 +8,7 @@ type Query {
   allUsers: [User]!
   userById(id: Int!): User
 
-  # Generates a new answer for the guessing game
+  # Generates a new answer for the
+  # guessing game
   answer: [Int!]!
 }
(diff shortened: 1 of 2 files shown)


kamilkisiela:
@Alxandr Thank you very much! 
Could you add a unit test for that or give me an access to make changes (there'a checkbox on the right)?

Alxandr:
Checkbox is already checked. Also, I figured splitting up one of the existing comments and having it still pass the tests would cover the change?

kamilkisiela:
Yeah, you can copy paste some test and change the description, just so we can tell what's the purpose of it.

dotansimha:
Thanks @Alxandr ! Merged :)

PR #954, opened 2018-11-29
Init Wizard

kamilkisiela:


kamilkisiela:
@dotansimha 👍 or 👎

dotansimha:
Looks good :) rebase please ;)

kamilkisiela:
@Urigo @dotansimha Now we can guess the target, if it's an angular/react app and typescript is used. I guess we can improve it later.

kamilkisiela:
It's not perfect, we might want to improve it a bit when flow related plugins are merged so user can pick the language.

PR #969, opened 2018-12-03
Enum source casing

speakerwiggin:
- added the new `disableEnumNamingConvention` option to allow existing case to be carried over from the SDL enum to TypeScript in a non-breaking way
- added a test for the new option
- added the new option to the markdown file

It looks like some VS Code whitespace auto-trimming made it into one of the commits as well. Sorry about that!

### DIFF
--- a/docs/plugins/typescript-common.md
+++ b/docs/plugins/typescript-common.md
@@ -27,6 +27,10 @@ You can override the naming conversion by pointing to a specific function in a m
 
 This will effect all the name conversions of the module.
 
+#### `disableEnumNamingConvention` (default value: `false`)
+
+Set this option to `true` to make enum values follow the naming convention of the source, as opposed to being converted to the `namingConvention` specified above.
+
 #### `scalars`
 
 Will map scalars to the predefined types.
--- a/packages/plugins/typescript-common/src/enum.handlebars
+++ b/packages/plugins/typescript-common/src/enum.handlebars
@@ -4,7 +4,7 @@ export type {{@root.config.interfacePrefix}}{{ convert name }} = {{#each values 
 {{else}}
 export {{#if @root.config.constEnums }}const {{/if}}enum {{@root.config.interfacePrefix}}{{ convert name }} {
 {{#each values }}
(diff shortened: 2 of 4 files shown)


kamilkisiela:
LGTML

@ardatan @dotansimha what do you think about the API?

speakerwiggin:
I originally wrote this up as `enumValueNamingConvention?: string` instead of `disableEnumNamingConvention?: boolean` so the user could specify a naming convention for the values in the same way they do for `namingConvention`.

I switched to the current design since I figured that a user would likely already be using whatever naming convention they wanted in the source schema, and since the library is already copying naming convention over for all other fields, this falls more in line with that pattern.

ardatan:
LGTM as well but I am not sure which one should be default? With or without convention?

dotansimha:
Hi @speakerwiggin ! Thank you so much for opening this PR! 

We would like to add a more generic way for naming configuration: https://github.com/dotansimha/graphql-code-generator/issues/977 

I know that the breaking change that we introduced is annoying and the enum naming is different now, but implementing naming in separated configuration will make it much easier to configure. 

I'm closing this PR for now, I [pushed your branch into the repo](https://github.com/dotansimha/graphql-code-generator/tree/speakerwiggin-enum-source-casing) so it will run through CI, so if you need to use your implementation at the moment (until we'll implement #977), using this alpha version: `0.15.0-alpha.3c84ac73`

speakerwiggin:
@dotansimha any idea when you'll get around to doing this? I'm only asking because this issue didn't exist before I submitted this pull request, and it seems silly to me to reject code that offers a fix in the meantime... especially if you're planning on making a breaking change later.

dotansimha:
@speakerwiggin very soon, I guess a day or two. The issue was not existing before, but we got a lot of complaints similar to yours (about the breaking change of the enum values casing). 
The reason I don't want to merge your changes as a temporary solution is because it will create another breaking change in the future when we'll implement https://github.com/dotansimha/graphql-code-generator/issues/977. The idea behind https://github.com/dotansimha/graphql-code-generator/issues/977 it not to introduce new breaking changes, just to allow more flexibility.

PR #1011, opened 2018-12-11
Fragment Matcher Plugin

kamilkisiela:


packages/graphql-codegen-cli/src/execute-plugin.ts
```diff
@@ -104,5 +104,7 @@ export async function executePlugin(options: ExecutePluginOptions, pluginPackage
     }
   }
 
-  return pluginPackage.plugin(outputSchema, options.documents, options.config);
+  return pluginPackage.plugin(outputSchema, options.documents, options.config, {
+    outputFile: options.outputFilename
+  });
```
kamilkisiela: @dotansimha I added 4th argument to PluginFunction so plugin can tell what is the output format. Some folks might want to use introspection result as a `.json` file, others directly in TypeScript without having another build step to move the `.json`.

packages/graphql-codegen-core/src/yml-config-types.ts
```diff
@@ -54,7 +54,10 @@ export namespace Types {
 export type PluginFunction<T = any> = (
   schema: GraphQLSchema,
   documents: DocumentFile[],
-  config: T
+  config: T,
+  info: {
+    outputFile: string;
+  }
```
kamilkisiela: actually it's here :)

PR #1047, opened 2018-12-14
Fix directive arguments transforms in flow-resolvers plugin

adelsz:
The visitors in flow-resovlers plugin weren't implementing the `convertName` method of the `BasicFlowVisitor` interface. Fixed that.

There was also a bug in `OperationVariablesToObject`:
When `typeof variable.type === 'string'` `variable.type` won't match the proper key from `DEFAULT_SCALARS` (it would be `string` instead of `String` or `number` instead of `Int`).
This happened only on directives parsing, in all other cases `typeof variable.type === 'object'`.
Fixed this by adding "mapped" default scalars `string, number, boolean` to `DEFAULT_SCALARS`.

This fixes the currently (master at https://github.com/dotansimha/graphql-code-generator/commit/7e3df340ff525f91e49d7ae349f0f1c982791b7f) broken tests in `flow-resolvers`. All tests in the monorepo should be green after merging.

### DIFF
--- a/packages/plugins/flow-resolvers/src/visitor.ts
+++ b/packages/plugins/flow-resolvers/src/visitor.ts
@@ -51,7 +51,7 @@ export class FlowResolversVisitor implements BasicFlowVisitor {
     return this._parsedConfig.scalars;
   }
 
-  private _convertName(name: any, addPrefix = true): string {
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @adelsz !

PR #1074, opened 2018-12-19
Make sure it can load a schema as a string

kamilkisiela:
`isValidPath` says that pretty much any string is a valid path... So I decided to check whether it's a glob path or has an alphanumeric extension.

### DIFF
--- a/packages/graphql-codegen-cli/src/loaders/schema/schema-from-string.ts
+++ b/packages/graphql-codegen-cli/src/loaders/schema/schema-from-string.ts
@@ -1,11 +1,11 @@
 import { Types } from 'graphql-codegen-core';
 import { SchemaLoader } from './schema-loader';
-import isValidPath = require('is-valid-path');
 import { parse, DocumentNode } from 'graphql';
+import isGlob = require('is-glob');
 
 export class SchemaFromString implements SchemaLoader {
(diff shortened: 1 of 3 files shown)


packages/graphql-codegen-cli/src/loaders/schema/schema-from-string.ts
```diff
@@ -1,11 +1,10 @@
 import { Types } from 'graphql-codegen-core';
 import { SchemaLoader } from './schema-loader';
-import isValidPath = require('is-valid-path');
 import { parse, DocumentNode } from 'graphql';
 
 export class SchemaFromString implements SchemaLoader {
   canHandle(str: string): boolean {
-    if (isValidPath(str)) {
+    if (/\.[a-z0-9]+$/i.test(str)) {
```
Urigo: should it be changed on other `is-valid-path` usages?
maybe also adding a comment about the regex?

kamilkisiela: @Urigo Now it has a comment.

About `is-valid-path` makes sense for other loaders because it still makes sure the path is valid, same with `is-glob`. They both just don't work well with SDL  (`scalar Date` passes because it might be a file / directory).

PR #1088, opened 2018-12-24
Flow plugin improvements  

dotansimha:
- [x] fixed flow imports (use `graphql` instead of `graphql/type`)
- [x] added root resolver
- [x] change resolvers arguments to be optional
- [ ] fix parent type of `Query`, `Mutation` and `Subscription` resolvers (should be `any`) 
- [ ] implement `react-apollo` flow plugin

### DIFF
--- a/packages/plugins/flow-resolvers/src/index.ts
+++ b/packages/plugins/flow-resolvers/src/index.ts
@@ -26,7 +26,7 @@ export const plugin: PluginFunction<FlowResolversPluginConfig> = (
   }
 
   const result = `
-import { ${imports.join(', ')} } from 'graphql/type';
+import { ${imports.join(', ')} } from 'graphql';
 
 export type Resolver<Result, Parent = {}, Context = {}, Args = {}> = (
   parent?: Parent,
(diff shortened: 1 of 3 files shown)


emmenko:
Hi, I have a question about the optional resolver arguments and found this PR.

Could you explain the reason for making the arguments optional? I couldn’t find any explanation about it and I’m wondering what’s the motivation behind it.
As far as I can tell, I would expect the resolver arguments to be always defined.

Thanks in advance!

dotansimha:
@emmenko because not all resolvers need to use arguments, it we'll force it - you'll have to add it in your resolvers declaration, even if you are not using it.

emmenko:
Thanks for the feedback.

However, given I expect my arguments to be defined, how would you recommend to use the resolver types? At the moment I have to force the compiler to treat the arguments as non optional by doing some if-else checking and it feels a bit wrong.

Would it make sense eg to generate 2 types of resolvers, one with required arguments and one with optional arguments?

dotansimha:
@emmenko does it still relevant? Could you please try to use the latest version, and open a new issue if that's still relevant?

PR #1091, opened 2018-12-27
Introduce stencil-apollo plugin

ardatan:


dotansimha:
@ardatan still WIP?

ardatan:
@dotansimha It's ready!

packages/plugins/typescript-stencil-apollo/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules
```
dotansimha: Please add `example` here

dotansimha: Oh sorry, I meant to add it to `.npmignore` because we don't want to publish it :) @ardatan

ardatan: Ok I see :) But I think you missed this one
https://github.com/dotansimha/graphql-code-generator/blob/stencil-apollo/packages/plugins/typescript-stencil-apollo/.npmignore

Issue #1097, opened 2018-12-28
[Flow] Wrong optional parameter  on Arg with default value


xanf:
**Describe the bug**

**To Reproduce**
Schema
```
schema {
  query: Query
}

type Query {
  foo(field: Int = 10): Int
}
```

config
```
schema:
  - /tmp/demo.gql
generates:
  /tmp/demo.js:
    plugins:
      - flow
```

output
```
/* @flow */

export type Query = {
  foo: ?number
};

export type QueryFooArgs = {
  field?: ?number
};
```
**Expected behavior**
field on `QueryFooArgs` will not have an optional sign, since there will be always this field due to default value
```
field: number
```

Am I correct in my expectations?

xanf:
No worry, I'll have a fork with fixes, will submit a PR soon :)

dotansimha:
Awesome!! Thanks @xanf !!

dotansimha:
Available in 0.16.0

PR #1104, opened 2019-01-01
[Flow] Fix optional sign for args with default values

xanf:
Related: #1097


dotansimha:
Great! Thanks @xanf !

PR #1118, opened 2019-01-04
[Flow Resolvers] fix type imports

xanf:
Both `GraphQLResolveInfo` and `GraphQLScalarTypeConfig` are types in `graphql` package flow import and should be imported via `import { type ... }` or flow lints will complain.

This pull request fixes this behaviour. I've intentionally put `type` specificator to both of them, in case we will need import some real entry from `graphql` package once.


dotansimha:
Thanks @xanf !!

PR #1092, opened 2018-12-27
Provide custom naming convention

ardatan:
Related to #977

- [ ] check if add new kinds
- [ ] check if all kinds are correctly used

PR #1136, opened 2019-01-07
Custom naming convention (cont. #1092)

marshall007:
Introduces a minor fix on top of @ardatan's work in #1092. This fixes the `typescript-common` and `typescript-client` tests (which is all my team needs), but some are still failing in (at least) `typescript-react-apollo` and `typescript-apollo-angular`.

The issue appears to be that some plugins attempt to pass objects as the second parameter to the `convert` utility function in the form `{ name: "convert", hash: {}, data: { ... } }`. I was unable to track down where in the templates or helper functions this is taking place.

We need the capability to preserve enum names in the generated output (as in #969, but `0.15.0-alpha.3c84ac73` had other problems when we tried to use it) ASAP, so just let me know what needs work here.


dotansimha:
Thanks @marshall007 ! 
@kamilkisiela took your changes and applied some of them to the original PR.

Issue #1134, opened 2019-01-07
Allow enum resolvers as part of `IResolvers`


mariusmarais:
(My issue is with the generated Typescript, but probably affects other templates too.)

I would like enums to also be included in the generated `IResolvers` definition, in my case so that I can re-map GraphQL enums to internal DB enums (see [ApolloServer docs](https://www.apollographql.com/docs/apollo-server/features/scalars-enums.html#internal-values)), automatically and with compiler assistance.

## Details
For the `schema.graphql`:
```graphql
scalar Date

enum State {
  SET
  UNSET
}

input SomeUpdate {
  state: State
}

type Thing {
  state: State
  updatedAt: Date
}

type Query {
  getTheThing(id: ID!): Thing!
}

type Mutation {
  updateIt(id: ID!, update: SomeUpdate): Thing!
}
```

the generated `IResolvers` type currently looks like (comments added):

```ts
export interface IResolvers {
  // query definition present
  Query?: QueryResolvers.Resolvers;

  // type definitions present
  Thing?: ThingResolvers.Resolvers;

  // mutation definition present
  Mutation?: MutationResolvers.Resolvers;

  // scalar definitions present
  Date?: GraphQLScalarType;

  // enum definitions not present :'(
}
```

I would like the above schema to generate something like the following:
```ts
export interface IResolvers {
  Query?: QueryResolvers.Resolvers;
  Thing?: ThingResolvers.Resolvers;
  Mutation?: MutationResolvers.Resolvers;
  Date?: GraphQLScalarType;

  // enum definitions present
  State?: { [key in State]: any };
}
```

which I could then use as follows:
```ts
const resolvers: IResolvers = {
  // ... all my normal resolvers
  State: {
    [State.Set]: '1',
    [State.Unset]: '0'
  }
};
```

I.e. the compiler would help me remember to declare all `State` values.

## Alternative / my current workaround
```ts
import { IResolvers } from './generated-types';
import { IResolvers as IApolloResolvers } from 'graphql-tools';

const coreResolvers: IResolver = {
  Query: ...,
  Mutation: ...,
  // ...etc, very nicely typed :)
};

export const resolvers: IApolloResolvers = {
  // see https://github.com/dotansimha/graphql-code-generator/issues/1133
  ...(coreResolvers as IApolloResolvers),

  State: {
    [State.Set]: '1',
    [State.Unset]: '0'
  }
};
```

## Other considerations
[...]

PR #678, opened 2018-10-09
Add support for internal enum values in typescript template

dobesv:
Fixes https://github.com/dotansimha/graphql-code-generator/issues/576

PR #1139, opened 2019-01-07
Support for reusing (importing) enums

mariusmarais:
After logging #1134 I realized that it would be very useful to me to import existing enums in my codebase instead of generating new ones with custom, but similar values.  (See #678 for some problems with that approach.)

This PR allows specifying a string like `path/to/enums#MyCustomEnum` as the value for an enum definition in the config, which will then import the enum instead of generating another.  In order to still map to internal values (with Apollo or otherwise) a `AValueMap` type is also generated which assists in doing so.

### Example
```graphql
# schema.graphql
enum A {
  ONE
  TWO
}
```

```yaml
# codegen.yaml
generates:
  path/to/generated.ts:
    plugins:
      - typescript-common
    config:
      enums:
        A: path/to/enums#MyCustomEnum
```
```ts
// path/to/generated.ts
import { MyCustomEnum as A } from "path/to/enums";

export type AValueMap = {
  ONE: A,
  TWO: A
};
```
```ts
// your/resolvers.ts
import { MyCustomEnum } from "path/to/enums";
import { AValueMap } from "path/to/generated";

const aMap: AValueMap = {
  ONE: MyCustomEnum.First,
  TWO: MyCustomEnum.Second
};

export const resolvers = {
  Query: { ... },
  Mutation: { ... },
  A: aMap
};
```

Edit: Additionally, generation for a specific enum can be skipped entirely. This allows manual definition or import when combined with the `add` plugin.

### Example
```yaml
# codegen.yaml
generates:
  path/to/generated.ts:
    plugins:
      - typescript-common
    config:
      enums:
        A: #empty
```

### DIFF
--- a/docs/plugins/typescript-common.md
+++ b/docs/plugins/typescript-common.md
@@ -54,7 +54,9 @@ Will generate the declared enums as TypeScript types. This is useful if you can'
 
 #### `enums` (default value: `{}`)
 
-Use this feature to set custom values for your GraphQL enums.
+Use this feature to set custom values for your GraphQL enums, or to reuse an existing enum.
+
+##### Custom values:
 
 ```yaml
 # ...
@@ -67,6 +69,25 @@ generates:
         MyEnum:
           A: 'foo'
 ```
+
+##### Existing enum:
+
+```yaml
+# ...
+generates:
+  path/to/file.ts:
+    plugins:
+      - typescript-common
+    config:
+      enums:
+        MyEnum: ./path/to/enum#MyEnum
+```
+
+Note:
+
+- The file path is relative to the generated output file.
+- Another type named `MyEnumValueMap` will also be generated which allows you to map the GraphQL API enum values to internal values if needed.
+- This option supercedes `enumsAsTypes`
 
 #### `immutableTypes` (default value: `false`)
 
--- a/docs/plugins/typescript-common.md
+++ b/docs/plugins/typescript-common.md
@@ -54,7 +54,7 @@ Will generate the declared enums as TypeScript types. This is useful if you can'
 
 #### `enums` (default value: `{}`)
 
-Use this feature to set custom values for your GraphQL enums, or to reuse an existing enum.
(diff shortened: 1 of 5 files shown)


dotansimha:
Hi @mariusmarais ! Thank you for your contribution! 
Maybe the API should specify `MyCustomEnum` only, and then in order to add the import, we can use `add` plugin? What do you think?

mariusmarais:
I looked at quite a few options, and that was one of them :)

Specifying just the name `MyCustomEnum` would be more difficult to implement since the import allows us to alias the name easily. If we don't import (automatically or manually), we need to update the name everywhere, including for other plugins.  I'm out of my depth for that one!

I considered a `null` option, which would then not generate an import, nor an enum for that type:
```yaml
# codegen.yaml
generates:
  path/to/generated.ts:
    plugins:
      - typescript-common
    config:
      enums:
        A: # empty
```

That could bypass the generation entirely and then you can use the `add` plugin to do the import by hand if you need to. It should work fine, but then you would need to chain 2x plugins to use the simple case, so I would prefer to keep the path in there, similar to the way it works [here](https://github.com/dotansimha/graphql-code-generator/blob/master/docs/plugins/typescript-resolvers.md#custom-context-type).

Either way, I'm happy to add some by of bypassing generation, but would really prefer to keep the `path#type` so that the simple case is very easy to use.

dotansimha:
Got it, I think both ways are fine. 
Can you just change back `enums` to `enumValues`? I would like to avoid this breaking change :)

mariusmarais:
I actually only updated the docs... Seems like it was always `enums` per [this](https://github.com/dotansimha/graphql-code-generator/blame/master/packages/plugins/typescript-common/src/helpers.ts#L27)?

Will add the `null` option ASAP.

dotansimha:
Oh that's right :D thanks!

mariusmarais:
@dotansimha Done \o/

dotansimha:
That's awesome!! Thanks @mariusmarais !!

Issue #1142, opened 2019-01-09
[Flow] Too restrictive type on optional fields in mutation responses


xanf:
**Describe the bug**
Consider following schema:
```
schema {
  mutation: Mutation
}

type MutationResponse {
  success: Boolean
  errors: [String!]
}
  
type Mutation {
  update(id: ID!): MutationResponse
}
```
and obviously config
```
generates:
  demo.js:
    - flow
```
generated type will be
```
export type MutationResponse = {
  success: ?boolean
  errors: ?Array<string>
}
```
which means I always need to return `success` and `errors` field even if they are null/undefined. All known graphql server implementations tolerate that, so I believe it should be:
```
export type MutationResponse = {
  success?: ?boolean
  errors?: ?Array<string>
}
```
which allows not to have field in response.

xanf:
Actually it's trickier than I originally thought
Let's asume some type
```
type Demo = { foo: String }
```
is used both in input and output position. For that type in input position I definitely want to have `type Demo = { foo: ?string }` because otherwise I will have to do "useless" checks that foo is present in variable of type demo.

But when I have it in output position, I want it as `{ foo?: ?string }` because I do not want to explicitly mention foo in my result, since it's optional.

(From strict typing system we should have `{ foo?: ?string }` in input position also, but this is wrong for developer experience, I assume)

xanf:
Actually after researching how for example how apollo works, I have a strong belief we need to swtich from "foo: ?string" to "foo?: string" both for input & normal types. For example when we have optional field in input graphql-js which is considered to be "reference implementation" will not add this optional field in response (tested by "hasOwnProperty")

Also I see no caveats from developer perspective in this change

adelsz:
@xanf `graphql-js` allows for both `null` and `undefined` as values of an optional mutation argument.
Typing an optional field as `foo?: string` is incorrect, because it doesnt allow passing `foo: null`.
Have a look at this [section](https://facebook.github.io/graphql/June2018/#sec-Null-Value) of the graphql specification.

dotansimha:
Fixed in 0.16.0. Thanks @xanf

PR #1158, opened 2019-01-12
Allow non-null fields to be missing for types (see #1142)

xanf:
As @adelsz mentioned in #1142 it is [normal behavior](https://facebook.github.io/graphql/June2018/#sec-Null-Value) not to provide value at all for field value. This update brings generated types in sync with specification allowing both "missing" value and specifying null


dotansimha:
Thanks @xanf !

PR #1166, opened 2019-01-15
Import variable from mapper, not entire identifier

blakeembrey:
Apologies if the tests fail, I couldn't get them passing locally.

This fixes an issue with mappers using nested data. E.g. in my application I want to use `Team#Team.Model` because I have an `index.ts` that does `import * as Team from './team'; export { Team }`.


dotansimha:
Awesome :) Thanks @blakeembrey !

PR #1178, opened 2019-01-17
Format and align errors

kamilkisiela:
## Before
Before

## After
After


kamilkisiela:
I also added `debugLog` (process.env.DEBUG) in couple of places and changed its behavior. Now we log stuff but it's being displayed after `Listr` finishes his job, so at the end :) I can revert it if that's something unwanted.

zrzut ekranu 2019-01-17 o 18 38 54

PR #1198, opened 2019-01-21
Support `strict` code output

blakeembrey:
I want my resolvers to be type-safe, this will make it a little extra work to write them out completely but at least it won't fail at runtime. FWIW, I have a way that would work to automatically reduce the types when the property is of the correct type for GraphQL to automatically use as a resolver (example below, could be tidied up and easily generated based on the import).

```ts
interface Bar {
    test: string
    another: number
}

interface Foo <T> {
    test: T extends Record<'test', string>
    ? never
    : () => string
}

export type KnownKeys<T> = ({ [K in keyof T]-?: T[K] extends never ? never : K })[keyof T];
export type Strip<T> = Pick<T, KnownKeys<T>>;

const foo: Strip<Foo<Bar>> = {}
```

Edit: What is the existing reasoning for having all the properties be optional?

Edit 2: Working in the playground but not in my editor, trying to refine the subtraction type now.

Edit 3: I have it working with these types!

```ts
export type ResolverCheck <T, K extends string | number | symbol, R extends Resolver<any, any, any, any>> = T extends Record<K, R | ReturnType<R>> ? never : R;
export type KnownKeys<T> = ({ [K in keyof T]-?: T[K] extends never ? never : K })[keyof T];
export type Strip<T> = Pick<T, KnownKeys<T>>;
```

Just need to use `ResolverCheck<TypeParent, 'prop', Resolver>` and it can automatically be stripped from the type later using `Strip`.

### DIFF
--- a/packages/plugins/typescript-resolvers/src/index.ts
+++ b/packages/plugins/typescript-resolvers/src/index.ts
@@ -13,6 +13,7 @@ import { importContext, getContext } from './context';
 import { getParentType, getParentTypes } from './parent-type';
 
 export interface TypeScriptServerResolversConfig extends TypeScriptCommonConfig {
+  strict?: boolean;
   noNamespaces?: boolean;
   contextType?: string;
   mappers?: { [name: string]: string };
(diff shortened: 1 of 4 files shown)


dotansimha:
Thank you @blakeembrey !

PR #1263, opened 2019-02-03
WIP: Refactor/typescript client pick

swcarlosrj:
**⚠️Don't merge it!**

**Issue related**: #713
**Status**: Working on it!
**Tests**: 11/16

@dotansimha I have some questions about the differents use cases, especially with `Maybe`, `!?` & others. In order to finish this off, I'd like to ask you **if you could check the tests cases and fix/complete them if it's required.**

The solution proposed its a recursive algorithm iterating over the `fields` where
if `hasFields` then `{ $name: recursiveAlgorithm(fields[index].fields) }`
else `Pick<$name, '$field'>`

In case of contains fragments, these are parsed and included on its respective fields before the execution (it'd probably change modifying the current `flattenDocuments` return object, I have to check if it's worth it)

It also could be improved because actually for the following example:
```graphql
query myFeed {
  feed {
    id
    commentCount
    repository {
      full_name
      html_url
      owner {
        avatar_url
      }
    }
  }
}
```

It'll return:
```typescript
export namespace MyFeed {
      export type Variables = {
      }
      export type Query =  {
        __typename?: "Query";
        feed: Maybe<(Pick<Entry,'id'> & Pick<Entry,'commentCount'>
        & { repository: Pick<Repository,'full_name'> & Pick<Repository,'html_url'>
             & { owner: Pick<User,'avatar_url'> }
        })[]>
      }

}
```
Instead of:
```typescript
export namespace MyFeed {
      export type Variables = {
      }
      export type Query =  {
        __typename?: "Query";
        feed: Maybe<(Pick<Entry, 'id', 'commentCount'>
        & { repository: Pick<Repository, 'full_name', 'html_url'>
              & { owner: Pick<User,'avatar_url'> }
        })[]>
      }

}
```

I just have just time work for the base case but it'd could be easily modified to support that syntax although the current implementation work as well!

[Here](https://goo.gl/KAFRvy) you might wanna check the types generated!

Let me know what you think about the proposed solution and how I could improve it 😄


dotansimha:
Hi @swcarlosrj ! Thank you so much! I'll have some time later today and I'll review it.

swcarlosrj:
Great @dotansimha, as you know better the corner cases as well as the core functionalities, I guess you could predict the possible issues with this proposal and thus, speed this up. With the new implementation, I'm not sure if all fragments have to be removed (as they could be automatically included on the gql documents using _Pick_), if that's the case, I could modify the `flattenTypes` function which is only used on the client related plugins:
-  typescript-client
-  typescript-react-apollo
-  typescript-apollo-angular
-  typescript-stencil-apollo

And then get the `flatenDocument` object composed also with fragments inside of each field (which is basically what I'm doing [here](https://github.com/dotansimha/graphql-code-generator/pull/1263/files#diff-2f9530056244af66d82cfec5054f12afR170))

Having your feedback about this and checking/adding the appropriate test regarding this implementation should be enough to finish it off!

Let me know if you'd have any questions about 🤗

dotansimha:
Thank you @swcarlosrj. I'm closing at the moment, we are working on it here: https://github.com/dotansimha/graphql-code-generator/pull/1353 . We decided to go for a much bigger refactor, and it's a chance to rewrite the plugins from scratch, and share the logic with the `flow` plugins. 
Thank you so much for opening it, you gave me a huge inspiration for the implementation!!

swcarlosrj:
Thanks @dotansimha, always a pleasure to help, looking forward to try the new version 😄

PR #1279, opened 2019-02-06
Rename Context to TContext

dusan-dragon:
Hi, this Pr just rename `Context` to `TContext` so that way if we want to specify our custom context, we can still name it `Context` instead of `MyContext`


dusan-dragon:
I am on it.

dusan-dragon:
@dotansimha fixed

dotansimha:
Great! Thanks @dusan-dragon !

PR #1282, opened 2019-02-06
Add import for interfaces mappings

dusan-dragon:
Currently the mappers does not generate import statement for interfaces. This PR fixes it.
For example take a look at added test case.

The interfaces are good if you want to extend some other interface and use it as Type...


dotansimha:
Awesome!!! thank you @dusan-dragon !

Pruxis:
Has this been published already? My mappers do not seem to be imported at this point :/

Pruxis:
Even on version ^0.17.0-alpha.79293e3b it's still not working 😭

dotansimha:
It's available in 0.18.0 :) @Pruxis

Pruxis:
I noticed, I forgot about this issue I should've closed it

Issue #1298, opened 2019-02-10
Support Iterable Interface for Resolvers


blakeembrey:
**Is your feature request related to a problem? Please describe.**

GraphQL supports the iterator protocol via `iterall`. It only supports sync iterables right now for regular returns though, but it'd be nice if the generated TypeScript interfaces also accepted iterator returns.

**Describe the solution you'd like**

Changing all arrays generated by GraphQL to TypeScript resolvers into iterables.

dotansimha:
Implement by @kamilkisiela in the recent refactor, it will be available soon.

blakeembrey:
I think I found the change but an array is iterable, it seems redundant to specify array or iterable.

dotansimha:
@kamilkisiela ^ ?

kamilkisiela:
@blakeembrey So if Array is Iterable then why to use Iterable in a first place?

blakeembrey:
Because Array implements Iterable, Iterable does not implement Array.

dotansimha:
So basically, @blakeembrey we can use just `Iterable`? are you sure it's compatible with all TypeScript configurations? 

@kamilkisiela

dotansimha:
Available in 1.0.0 👍

slosd:
I was looking for this exact feature, but discovered that the changes (https://github.com/dotansimha/graphql-code-generator/pull/1471) were reverted only a few days later (for the release of 1.0.3 it seems) with https://github.com/dotansimha/graphql-code-generator/pull/1534:
> Remove ArrayOrIterable at the moment, until we'll figure out if it's really needed.

Can this issue be reopened?

laurisvan:
I recently stumbled upon this issue when trying to speed up our subscriptions where we try to avoid excess memcopies by returning Iterables instead of Arrays.

Is there a way to explicitly make one resolver return e.g. Iterable, so that we could gradually transition from Arrays to Iterables?

PR #1299, opened 2019-02-11
Support iterable for TypeScript resolver returns

blakeembrey:
Closes https://github.com/dotansimha/graphql-code-generator/issues/1298.

### DIFF
--- a/packages/plugins/typescript-common/src/helpers.ts
+++ b/packages/plugins/typescript-common/src/helpers.ts
@@ -56,7 +56,7 @@ export function getEnumValue(type: string, name: string, options: Handlebars.Hel
   }
 }
 
-export function getFieldType(field: Field, realType: string, options: Handlebars.HelperOptions) {
+export function getFieldType(field: Field, realType: string, options: Handlebars.HelperOptions, isReturn = false) {
   const config = options.data.root.config || {};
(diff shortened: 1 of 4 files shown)


dotansimha:
Thank you @blakeembrey ! 
Can you please explain your implementation? I think I'm missing something there. 
`isReturn` is a way to tell if you are generating the return type of a resolver?
Also, do you think it's relevant only for resolvers package? Maybe stating `Iterable` might be useful for the type itself (and then the resolvers package will also have it).

kamilkisiela:
I don't know if `graphql` supports `Iterable` in every resolver and not only in subscription related ones. If that's something you're interested in, we can add Iterable to the `SubscriptionResolver` union so it applies to all subscription resolvers without having to pass a boolean anywhere.

blakeembrey:
@kamilkisiela It does, like I said in my linked issue GraphQL uses `iterall` which means it supports all iterables. You might be confusing async iterators with regular iterators here.

> isReturn is a way to tell if you are generating the return type of a resolver?

Yes.

> Also, do you think it's relevant only for resolvers package? 

I believe so, since it's only valid for the return type in resolvers for GraphQL JS.

dotansimha:
Thanks @blakeembrey , we implemented the your changes in the recent refactor :)

Issue #1109, opened 2019-01-02
[TypeScript] Default value with 0 shoud not be optional


phoenix741:
**Describe the bug**

If the default value is different of 0, the parameter isn't optional. If the default value is 0, the parameter is optional (and shouldn't be).

**To Reproduce**

1. Use this schema as schema

```
type Toto {
    myFeed(skip: Int = 0, limit: Int = 25): MyFeed!
}
```

2. Run the following command: `gql-gen ...`

The result is 
```
export interface MyFeedQueryArgs {
  skip?: Maybe<number>

  limit?: number
}
```

**Expected behavior**
I would like to have : 
```
export interface MyFeedQueryArgs {
  skip: number

  limit: number
}
```

**Environment:**
 - OS: Linux
 - Codegen: 0.15.2
 - Node: 10

kamilkisiela:
@phoenix741 I don't think that's a correct usage. How do you use `MyFeedQueryArgs`, next to a component or Apollo? If so then you need them to be optional, otherwise TypeScript will warn you that some variables are missing (if we change it to non-optional).

For example:

```ts
function useQuery<R, V>(query: DocumentNode, variables: V): Promise<ExecutionResult<R>> { ... }

// args are optional
useQuery<MyFeedQuery, MyFeedQueryArgs>(yourQuery, {}) // TS says "GREEN!"

// args are non-optional so required
useQuery<MyFeedQuery, MyFeedQueryArgs>(yourQuery, {}) // TS says "RED!" - skip and limit are not provided
```

Could you correct me?

phoenix741:
@kamilkisiela I use it in the server side to define interface of my resolver. 
As there is a default value, i expect that the value is not undefined, else i must redefine a default value in the resolver.

dotansimha:
Fixed in 0.18.0 🎉

PR #1335, opened 2019-02-22
Support falsy default values in args correctly

leighman:


leighman:
Independent but I see that this would close https://github.com/dotansimha/graphql-code-generator/issues/1109

dotansimha:
Awesome! Thank you @leighman !

PR #1339, opened 2019-02-22
Remove subscription hook import

FredyC:
Since `useSubscription` is [merged and released](https://github.com/trojanowski/react-apollo-hooks/pull/37), there is no need for a specific import.

It's also for consideration if `withSubscriptionHooks` is necessary. I don't feel that Subscriptions needs any special treatment really.

### DIFF
--- a/docs/plugins/typescript-react-apollo.md
+++ b/docs/plugins/typescript-react-apollo.md
@@ -73,21 +73,4 @@ Or if you are using `noNamespaces` option:
 
 #### `withSubscriptionHooks` (default value: `false`)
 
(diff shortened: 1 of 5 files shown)


dotansimha:
Great! Thank you @FredyC !

ardatan:
I added a comment in the code as a review. @dotansimha Did you miss that? :)

FredyC:
@ardatan Well, it seems you haven't submitted your review, otherwise it would have shown in here. What was it?

Anyway, I agree with removing the withSubscriptionHooks. Do you want to do that? :)

ardatan:
@FredyC I think I did something wrong :) It was related to the same thing.
I will create a PR in an hour

PR #1344, opened 2019-02-23
Add gqlImportFrom option

FredyC:
Related #1342 

Once again I am not sure if there is a better way to handle the default config.

Edit: On the contrary to a drafted solution, I had to opt for configuring whole import line because `graphql.macro` does not have a default export.

PR #1347, opened 2019-02-24
Some enhancements for hooks

ardatan:
- Remove `withSubscriptionHooks`
- Use the same behaviour that @FredyC added in here but in a different way, I think we can uae same `withHooks` option to provide custom module for hooks.
- Fix missing `ReactApollo` import if `noComponents` is true, but `noHOC` is false.
- Change `noComponents` and `noHOC` to `withComponents` and `withHOC` which look more consistent with `withHooks`.
- Fixed invalid import syntax
- `gqlImport` is more consistent with other import syntaxes in codegen. `gqlImport: graphql.macro#gql`

### DIFF
--- a/docs/plugins/typescript-react-apollo.md
+++ b/docs/plugins/typescript-react-apollo.md
@@ -52,13 +52,13 @@ Or if you prefer:
 Customize from where will be `gql` imported. This is useful if you want to use eg. `graphql.macro` for inlining documents.
 Note that you need to write whole import line, eg. `import { gql } from 'graphql.macro'`.
 
-#### `noHOC` (default value: `false`)
+#### `withHOC` (default value: `true`)
 
-This will disable the higher order components generation.
+This will disable the higher order components generation by setting this option to `false`.
 
-#### `noComponents` (default value: `false`)
+#### `withComponents` (default value: `true`)
 
-This will cause the code generator to _omit_ React **Components**. So, in case you are just using _hooks_, you can disable the Components setting this option to `true`.
+This will cause the code generator to _omit_ React **Components**. So, in case you are just using _hooks_, you can disable the Components setting this option to `false`.
 
 #### `withHooks` (default value: `false`)
 
@@ -76,6 +76,4 @@ Or if you are using `noNamespaces` option:
   const { data, loading, error } = useTest(...);
 ```
 
-#### `withSubscriptionHooks` (default value: `false`)
-
-This will cause the codegen to add React **Hooks** even for _Subscriptions_. 
+You can specify alternative module that is exports `useQuery` `useMutation` and `useSubscription`. This is useful for further abstraction of some common tasks (eg. error handling). Filepath relative to generated file can be also specified. But if you provide only `true`, it will use `react-apollo-hooks` by default.
--- a/packages/plugins/typescript-react-apollo/src/helpers.ts
+++ b/packages/plugins/typescript-react-apollo/src/helpers.ts
@@ -117,12 +117,21 @@ export const toFragmentName = convert => (fragmentName: string, options: Handleb
   }
 };
 
-export const shouldOutputHook = (operationType: string, options: Handlebars.HelperOptions): boolean => {
-  const config = options.data.root.config || {};
-  return operationType !== 'subscription' || config.withSubscriptionHooks;
-};
-
 export const gqlImport = (operationType: string, options: Handlebars.HelperOptions): string => {
   const config = options.data.root.config || {};
-  return config.gqlImport || 'import gql from \'graphql-tag\'';
+  return config.gqlImport || "import gql from 'graphql-tag'";
+};
+
+export const getImports = (options: Handlebars.HelperOptions) => {
+  let imports = '';
+  const config = options.data.root.config || {};
+  if (config.withComponents) {
+    imports += `import * as React from 'react';`;
+  }
+  if (config.withComponents || config.withHOC) {
+    imports += `import * as ReactApollo from 'react-apollo';`;
+  }
+  if (config.withHooks) {
+    imports += typeof config.withHooks === 'string' ? config.withHooks : 'react-apollo-hooks';
+  }
 };
--- a/packages/plugins/typescript-react-apollo/src/helpers.ts
+++ b/packages/plugins/typescript-react-apollo/src/helpers.ts
@@ -83,7 +83,7 @@ export const gql = convert => (operation: Operation, options: Handlebars.HelperO
     ${includeFragments(transformFragments(convert)(operation.document, options))}
   `;
 
-  return config.noGraphqlTag ? JSON.stringify(gqlTag(doc)) : 'gql`' + doc + '`';
+  return config.gqlImport ? JSON.stringify(gqlTag(doc)) : 'gql`' + doc + '`';
 };
 
 function includeFragments(fragments: string[]): string {
--- a/packages/plugins/typescript-react-apollo/src/helpers.ts
+++ b/packages/plugins/typescript-react-apollo/src/helpers.ts
@@ -117,22 +117,32 @@ export const toFragmentName = convert => (fragmentName: string, options: Handleb
   }
 };
 
-export const gqlImport = (operationType: string, options: Handlebars.HelperOptions): string => {
-  const config = options.data.root.config || {};
-  return config.gqlImport || 'import gql from \'graphql-tag\'';
+export const parseImport = (importStr: string) => {
+  const [moduleName, propName] = importStr.split('#');
+  return {
+    moduleName,
+    propName
+  };
 };
 
-export const getImports = (options: Handlebars.HelperOptions) => {
-  let imports = '';
+export const getImports = (operationType: string, options: Handlebars.HelperOptions) => {
   const config = options.data.root.config || {};
+  const gqlImport = parseImport(config.gqlImport || 'graphql-tag');
+  let imports = `
+    import ${
+      gqlImport.propName ? `{ ${gqlImport.propName === 'gql' ? 'gql' : `${gqlImport.propName} as gql`} }` : 'gql'
+    } from '${gqlImport.moduleName}';
+  `;
   if (config.withComponents) {
-    imports += `import * as React from 'react';`;
+    imports += `import * as React from 'react';\n`;
   }
   if (config.withComponents || config.withHOC) {
-    imports += `import * as ReactApollo from 'react-apollo';`;
+    imports += `import * as ReactApollo from 'react-apollo';\n`;
   }
   if (config.withHooks) {
-    imports += typeof config.withHooks === 'string' ? config.withHooks : 'react-apollo-hooks';
+    imports += `import * as ReactApolloHooks from '${
+      typeof config.withHooks === 'string' ? config.withHooks : 'react-apollo-hooks'
+    }';\n`;
   }
   return imports;
 };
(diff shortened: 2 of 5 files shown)


docs/plugins/typescript-react-apollo.md
```diff
@@ -52,13 +52,13 @@ Or if you prefer:
 Customize from where will be `gql` imported. This is useful if you want to use eg. `graphql.macro` for inlining documents.
 Note that you need to write whole import line, eg. `import { gql } from 'graphql.macro'`.
 
-#### `noHOC` (default value: `false`)
+#### `withHOC` (default value: `true`)
```
FredyC: Feels like unnecessary breaking change just for switching logic around. If anything, it should still support old flags, just deprecate them and convert into these.

docs/plugins/typescript-react-apollo.md
```diff
@@ -76,6 +76,4 @@ Or if you are using `noNamespaces` option:
   const { data, loading, error } = useTest(...);
 ```
 
-#### `withSubscriptionHooks` (default value: `false`)
-
-This will cause the codegen to add React **Hooks** even for _Subscriptions_. 
+You can specify alternative module that is exports `useQuery` `useMutation` and `useSubscription`. This is useful for further abstraction of some common tasks (eg. error handling). Filepath relative to generated file can be also specified. But if you provide only `true`, it will use `react-apollo-hooks` by default.
```
FredyC: It doesn't feel right to abuse the boolean flag to also have a string value. I would keep it separate, nothing wrong about.

packages/plugins/typescript-react-apollo/src/helpers.ts
```diff
@@ -117,12 +117,22 @@ export const toFragmentName = convert => (fragmentName: string, options: Handleb
   }
 };
 
-export const shouldOutputHook = (operationType: string, options: Handlebars.HelperOptions): boolean => {
+export const gqlImport = (operationType: string, options: Handlebars.HelperOptions): string => {
   const config = options.data.root.config || {};
-  return operationType !== 'subscription' || config.withSubscriptionHooks;
+  return config.gqlImport || "import gql from 'graphql-tag'";
```
FredyC: Why not include this in the `getImports` as well?

packages/plugins/typescript-react-apollo/src/index.ts
```diff
@@ -4,31 +4,31 @@ import { flattenTypes } from 'graphql-codegen-plugin-helpers';
...
   gqlImport?: string;
 }
 
 export const plugin: PluginFunction<TypeScriptReactApolloConfig> = async (
   schema: GraphQLSchema,
   documents: DocumentFile[],
-  config: TypeScriptReactApolloConfig
+  config: TypeScriptReactApolloConfig = {}
 ): Promise<string> => {
+  config.withHOC = 'withHOC' in config ? config.withHOC : true;
```
FredyC: Just a personal preference, probably not a big different
```suggestion
  config.withHOC = config.withHOC !== undefined ? config.withHOC : true;
```

packages/plugins/typescript-react-apollo/src/root.handlebars
```diff
@@ -1,11 +1,5 @@
 {{#ifCond operations.length '!==' 0}}
-{{#unless @root.config.noComponents}}
-import * as ReactApollo from 'react-apollo';
-import * as React from 'react';
-{{/unless}}
-{{#if @root.config.withHooks}}
-import * as ReactApolloHooks from 'react-apollo-hooks';
-{{/if}}
+{{{getImports}}}
```
FredyC: Got burned by this as well :)
```suggestion
{{{getImports this}}}
```

docs/plugins/typescript-react-apollo.md
dotansimha: I agree with @FredyC . I think it's better to change it back @ardatan :)

ardatan:
I also think there is another inconsistency between `gqlImport` and `importHooks`; because while `importHooks` takes the name of the module, `gqlImport` takes all the `import` expression.

FredyC:
Yea well, if you have a better idea... I've explained in #1344

FredyC:
Btw, that's also why there is a different name `gqlImport` for a full import an `hooksImportFrom` to clearly denote a difference. That's also why the `withHooks` should not be used for supplying the module name. It's confusing when someone looks at it without diving into docs.

ardatan:
We use `typeModule#ContextType` syntax in mappers, so I think this can be used here. The other thing we should consider `noGraphqlTag` option as well. If this is set to `true` with `gqlImport`, TS compiler can give some errors if you set `noUnusedImports` in your `tsconfig.json`.
After my change in this PR, if you set `gqlImport` to `false`, codegen will render `gql` tags without any putting `gql` tags before document strings. I think using two different configuration fields can make inconsistency. If we use two different options there; the code can get messy.
I think same for `withHooks` and `importHooks` stuff; because I think we should keep number of config fields as less as possible.

FredyC:
Ok, it's an interesting approach, I suppose it will work, but it requires a better documentation as it's not the usual approach :)

> The other thing we should consider `noGraphqlTag` option as well. If this is set to `true` with `gqlImport`

Yea, this config option is not even documented, so I doubt someone was using it :) What do you think @dotansimha?

> After my change in this PR, if you set `gqlImport` to `false`, codegen will render `gql` tags without any putting `gql` tags before document strings.

I guess it makes sense, not sure who would actually use that, but 🙈 

> I think same for `withHooks` and `importHooks` stuff; because I think we should keep number of config fields as less as possible.

I don't think it makes a difference to have one or two extra configs. It's a DEV tool after all. It's more about clarity so it's apparent from the first view what each config does. You shouldn't be forced look into docs every time you need to tweak something.

leonardfactory:
I'd cast a vote for different options (like I originally implemented them). More options, with clean docs, allows us to output clear errors if plugin is configured badly. I may understand something like *we accepts a list of foos or just a single foo*, but switching boolean with string seems not so useful in this case. However, enough bikeshedding for today! 🎉

sandiiarov:
> We use `typeModule#ContextType` syntax in mappers, so I think this can be used here. The other thing we should consider `noGraphqlTag` option as well. If this is set to `true` with `gqlImport`, TS compiler can give some errors if you set `noUnusedImports` in your `tsconfig.json`.
> After my change in this PR, if you set `gqlImport` to `false`, codegen will render `gql` tags without any putting `gql` tags before document strings. I think using two different configuration fields can make inconsistency. If we use two different options there; the code can get messy.
> I think same for `withHooks` and `importHooks` stuff; because I think we should keep number of config fields as less as possible.

But, what if I want to use `gql` from `graphql-tag.macro`?
When I add in the config file `gqlImport: graphql-tag.macro` a generated file has `import gql from "graphql-tag.macro";` and also has parsed GraphQL queries.

Example with `gqlImport: graphql-tag.macro`
```js
import gql from "graphql-tag.macro";
import * as ReactApolloHooks from "react-apollo-hooks";

// ====================================================
// Fragments
// ====================================================

export const WitnessInfoFragmentDoc = {
  kind: "Document",
  definitions: [
    {
      kind: "FragmentDefinition",
      name: { kind: "Name", value: "WitnessInfo" },
      typeCondition: {
        kind: "NamedType",
        name: { kind: "Name", value: "Witness" }
      },
      directives: [],
      selectionSet: {
        kind: "SelectionSet",
        selections: [
          {
            kind: "Field",
            name: { kind: "Name", value: "id" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "firstName" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "lastName" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "croppedImage" },
            arguments: [],
            directives: []
          }
        ]
      }
    }
  ],
  loc: { start: 0, end: 103 }
};
```

How can I stop parsing the GraphQL queries and use `graphql-tag.macro` with this PR?

sandiiarov:
> We use `typeModule#ContextType` syntax in mappers, so I think this can be used here. The other thing we should consider `noGraphqlTag` option as well. If this is set to `true` with `gqlImport`, TS compiler can give some errors if you set `noUnusedImports` in your `tsconfig.json`.
> After my change in this PR, if you set `gqlImport` to `false`, codegen will render `gql` tags without any putting `gql` tags before document strings. I think using two different configuration fields can make inconsistency. If we use two different options there; the code can get messy.
> I think same for `withHooks` and `importHooks` stuff; because I think we should keep number of config fields as less as possible.

@ardatan @FredyC But, what if I want to use `gql` from `graphql-tag.macro`?
When I use `gqlImport: graphql-tag.macro` in the generated file has `import gql from "graphql-tag.macro";` and also has parsed GraphQL queries.

Example with `gqlImport: graphql-tag.macro`
```js
import gql from "graphql-tag.macro";
import * as ReactApolloHooks from "react-apollo-hooks";

// ====================================================
// Fragments
// ====================================================

export const WitnessInfoFragmentDoc = {
  kind: "Document",
  definitions: [
    {
      kind: "FragmentDefinition",
      name: { kind: "Name", value: "WitnessInfo" },
      typeCondition: {
        kind: "NamedType",
        name: { kind: "Name", value: "Witness" }
      },
      directives: [],
      selectionSet: {
        kind: "SelectionSet",
        selections: [
          {
            kind: "Field",
            name: { kind: "Name", value: "id" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "firstName" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "lastName" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "croppedImage" },
            arguments: [],
            directives: []
          }
        ]
      }
    }
  ],
  loc: { start: 0, end: 103 }
};
```

How can I stop parsing the GraphQL queries and use `graphql-tag.macro` with this PR?

FredyC:
@sandiiarov Do you realize that's what is the macro about? To inline parsed document so it doesn't need to be done on runtime :)

sandiiarov:
@FredyC Yes, I do. Do you understand what I am trying to ask? Do you understand how `JSON.stringify(gqlTag(doc))` is working?

sandiiarov:
When I use `gqlImport: graphql-tag.macro` I want to get

```js
import gql from "graphql-tag.macro";

export const WitnessInfoFragmentDoc = gql`
  fragment WitnessInfo on Witness {
    id
    firstName
    lastName
  }
`;
```

but getting this
```js
import gql from "graphql-tag.macro";

export const WitnessInfoFragmentDoc = {
  kind: "Document",
  definitions: [
    {
      kind: "FragmentDefinition",
      name: { kind: "Name", value: "WitnessInfo" },
      typeCondition: {
        kind: "NamedType",
        name: { kind: "Name", value: "Witness" }
      },
      directives: [],
      selectionSet: {
        kind: "SelectionSet",
        selections: [
          {
            kind: "Field",
            name: { kind: "Name", value: "id" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "firstName" },
            arguments: [],
            directives: []
          },
          {
            kind: "Field",
            name: { kind: "Name", value: "lastName" },
            arguments: [],
            directives: []
          }
        ]
      }
    }
  ],
  loc: { start: 0, end: 88 }
};
```

Make sense?

FredyC:
Ah, I see what you mean now. Looks like that @DAB0mB introduced that mistake when merging some of @ardatan changes in https://github.com/dotansimha/graphql-code-generator/commit/859de36d6f4ba5794eebae963cca74bf90766717. Originally there was different option (which wasn't really documented).

```js
config.noGraphqlTag ? JSON.stringify(gqlTag(doc)) : 'gql`' + doc + '`';
```

The logic should be reversed with `gqlImport` option.

dotansimha:
@FredyC you are right, we fixed it in the refactor :)

PR #1352, opened 2019-02-25
Support es2015 tree shaking

ardatan:
- Add `es2015` bundle
- Change target to `es2018` because NodeJS already supports `es2018`, no need to transpile code.
- Use `tslib` to reduce the code


ardatan:
@dotansimha It was due to the examples, so they're fixed now :)

Issue #1337, opened 2019-02-22
Do not enforce tsx if only hooks used


FredyC:
**Is your feature request related to a problem? Please describe.**
Since Hooks don't use the JSX, there is no need to enforce the extension.

**Describe the solution you'd like**
If I specify configuration like this, it should be possible to use `*.ts` file.

```yml
noHOC: true,
noComponents: true,
withHooks: true,
```

dotansimha:
Fixed in 0.18.0 🎉

PR #1356, opened 2019-02-25
Do not enforce tsx if only hooks used

ardatan:
Related to  #1337

### DIFF
--- a/packages/plugins/typescript-react-apollo/src/index.ts
+++ b/packages/plugins/typescript-react-apollo/src/index.ts
@@ -46,12 +46,12 @@ export const validate: PluginValidateFn<any> = async (
   outputFile: string
 ) => {
   if (config.noComponents) {
-    if (extname(outputFile) !== '.ts') {
-      throw new Error(`Plugin "react-apollo" requires extension to be ".ts"!`);
+    if (extname(outputFile) !== '.ts' && extname(outputFile) !== '.tsx') {
+      throw new Error(`Plugin "react-apollo" with "noComponents" requires extension to be ".ts" or ".tsx"!`);
     }
   } else {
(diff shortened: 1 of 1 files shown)


packages/plugins/typescript-react-apollo/src/index.ts
```diff
@@ -42,10 +42,16 @@ export const plugin: PluginFunction<TypeScriptReactApolloConfig> = async (
...
   schema: GraphQLSchema,
   documents: DocumentFile[],
-  config: any,
+  config: TypeScriptReactApolloConfig,
   outputFile: string
 ) => {
-  if (extname(outputFile) !== '.tsx') {
-    throw new Error(`Plugin "react-apollo" requires extension to be ".tsx"!`);
+  if (config.noComponents) {
+    if (extname(outputFile) !== '.ts') {
```
leonardfactory: Mh... why should we force `ts` instead of `tsx`? I know that in a ts file you can't use pragma directives, but banning it seems like.. too much, don't you think so?
Thansk!

ardatan: You're right fixed :)

SimenB:
Is this broken in v1?

```yaml
generates:
  ../protobank-frontend/src/gqltypes.ts:
    schema:
      - '../protobank-schemas/bank.graphql'
      - '../protobank-frontend/client-schema.graphql'
    documents:
      - ../protobank-frontend/src/**/*.graphql
      - ../protobank-common/src/**/*.graphql
    config:
      noHOC: true
      noComponents: true
      withHooks: true
    plugins:
      - typescript
      - typescript-react-apollo
      - typescript-operations
```

`Plugin "react-apollo" requires extension to be ".tsx"!`

Issue #925, opened 2018-11-26
More verbose error messages.


wesselvdv:
**Is your feature request related to a problem? Please describe.**
Currently it's quite difficult to discern what exactly is going wrong if there's an error. Error messages are simply along the lines of: `An error has occurred in one of your documents.`

**Describe the solution you'd like**
I'd prefer the ability to either set a flag that makes it more verbose, or simply enable it by default to include the exact file and lines that are erroneous.

thekevinbrown:
I'm having trouble with the download of a remote schema failing, and I don't know exactly what it's trying to download, nor do I know what the server is saying back. It'd be awesome if there was a --debug flag or something so I could see what it's doing to help me troubleshoot.

dotansimha:
In https://github.com/dotansimha/graphql-code-generator/issues/953, @kroeder mentioned that it works fine with PowerShell instead of CMD, so I guess this issue is related only to Windows. 
@wesselvdv @thekevinbrown do you use Windows as well?

kroeder:
*it works with power shell, not with git bash, using it as webstorm terminal (but on windows)

Just do be 100% exact :)

thekevinbrown:
@dotansimha No, I'm on OS X.

The request is for a way to see what the tool is doing under the hood in the form of logs via some kind of `--debug` flag though, so I don't know what about that need is platform dependent.

wesselvdv:
@dotansimha Nope, i'm on macOS Mojave, and using PHPStorm terminal. I had to manually place some console.log statements to figure out where stuff was going awry. I almost seems as if the logging is cutoff at a newline, because only the start of an error message is shown.

dotansimha:
@thekevinbrown makes sense, we had it before v0.14 and I think we need to add it back. 

@wesselvdv @thekevinbrown which shell do you use? I'm also using Mac OS Mojave and I can see the errors using https://github.com/robbyrussell/oh-my-zsh shell.

wesselvdv:
@dotansimha I’m using bash

colinfindlay-nz:
I'm on Windows - and the most I can get out of it is: `Found 2 errors in your documents`
I can see that the validator throws a `DetailedError` which the actual detail - but I can't gql-gen to actually log it. (unless I hack the source)

dotansimha:
Fixed in 0.15. Please re-open if the issue persist.

valerybugakov:
[...]

PR #1367, opened 2019-02-27
More verbose error logging for document validation

jeroenvisser101:
Should fix part of #925, where it would normally just display "Found N error(s) in your documents", now shows file, line and column, along with the error message.

### DIFF
--- a/packages/graphql-codegen-cli/src/execute-plugin.ts
+++ b/packages/graphql-codegen-cli/src/execute-plugin.ts
@@ -3,7 +3,7 @@ import { DocumentNode, GraphQLSchema } from 'graphql';
 import { resolve } from 'path';
 import { DetailedError } from './errors';
 import { buildSchema } from './merge-schemas';
-import { validateGraphQlDocuments, checkValidationErrors } from 'graphql-toolkit';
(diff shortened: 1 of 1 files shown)


jeroenvisser101:
@dotansimha I just now see that you mentioned @DAB0mB working on something like this. I was working on this yesterday but only just now created the PR. Let me know what you want to do.

ardatan:
Thank you @jeroenvisser101 !
I think we should move make this change in `checkValidationErrors` instead of removing it.
https://github.com/ardatan/graphql-toolkit/blob/8858a0a1116bc7236a6f29e7c1c3723cd5be4720/src/utils/validate-documents.ts#L20

jeroenvisser101:
@ardatan I actually also started on that, but I actually want to throw a DetailedError, and I believe that comes from graphql-code-generator. That's why I chose to re-implement the method (although altered) in the core.

dotansimha:
Thank you @jeroenvisser101 ! 
@DAB0mB is working on adding unit tests to reproduce it and fixing this issue in `graphql-toolkit`, and he takes inspiration from your solution.

jeroenvisser101:
@dotansimha thank you! :)

Issue #1302, opened 2019-02-12
Use environment variables in codegen.yml


n44ps:
**Describe the bug**
I can't use an environment variables in the codegen.yml file.
**To Reproduce**
1. Try using : `schema: ${API_ENDPOINT}`

And you'll get : `Unable to find matching files for glob: ${API_ENDPOINT} in directory`

**Expected behavior**
I would like to use the environment variables instead of raw string

**Environment:**
 - OS: MacOS Mojave
 - Codegen: ^0.16.0
 - Node: v10.15.0

dotansimha:
@n44ps thank you for reporting it.
The codegen does not support it at the moment, but that's definitely a valid use case.
As a workaround, I can suggest you to run the [codegen from code, using the programmatic API](https://graphql-code-generator.com/docs/getting-started/programmatic-usage), and use your environment variables.

n44ps:
This is what I've ended up with, thank you @dotansimha !

dotansimha:
Available in 0.18.0 🎉

steebchen:
It would be awesome if you could specify a default value, like you can usually do with yaml files such as in docker:

```yaml
version: '3'

services:
  lb:
    image: nginx:latest
    environment:
      PASSWORD: "${PASS:-secret}"
```

dotansimha:
@steebchen can you please open a new issue for that? It will be easier to track.

steebchen:
See #1812

PR #1368, opened 2019-02-27
Support env variables in yml

ardatan:
#1302

### DIFF
--- a/packages/graphql-codegen-cli/src/yml.ts
+++ b/packages/graphql-codegen-cli/src/yml.ts
@@ -1,7 +1,19 @@
 import { Types, debugLog } from 'graphql-codegen-core';
 import { safeLoad } from 'js-yaml';
 
+export function interpolate<T extends { [key: string]: string }>(str: string, context: T) {
+  // tslint:disable-next-line:forin
+  for (let key in context) {
+    str = str.replace(new RegExp('\\$\\{' + key + '\\}', 'g'), context[key]);
+  }
+  return str;
+}
+
 export function parseConfigFile(ymlString: string): Types.Config {
+  if (typeof process !== 'undefined' && 'env' in process) {
+    debugLog(`[CLI] Interpolation of Environmental Variables`);
+    ymlString = interpolate(ymlString, process.env);
+  }
   debugLog(`[CLI] Parsing YAML file`);
   return safeLoad(ymlString) as Types.Config;
 }
--- a/packages/graphql-codegen-cli/src/yml.ts
+++ b/packages/graphql-codegen-cli/src/yml.ts
@@ -1,18 +1,10 @@
 import { Types, debugLog } from 'graphql-codegen-core';
 import { safeLoad } from 'js-yaml';
 
-export function interpolate<T extends { [key: string]: string }>(str: string, context: T) {
-  // tslint:disable-next-line:forin
-  for (let key in context) {
-    str = str.replace(new RegExp('\\$\\{' + key + '\\}', 'g'), context[key]);
-  }
-  return str;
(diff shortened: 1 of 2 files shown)


packages/graphql-codegen-cli/tests/cli-flags.spec.ts
```diff
@@ -151,4 +151,17 @@ describe.only('CLI Flags', () => {
     const config = createConfig(args);
     expect(config.overwrite).toBeTruthy();
   });
+
+  it('Should interpolate environmental variables in YML', () => {
+    process.env['SCHEMA_PATH'] = 'schema-env.graphql';
+    mockConfig(`
+        schema: \${SCHEMA_PATH}
```
dotansimha: @ardatan why not just `$SOME_VAR`?

ardatan: In the issue, the interpolated string has `{}.
`${API_ENDPOINT}`

dotansimha: I know, but I think `$API_ENDPOINT` is better API (similar to usage in terminal), and shouldn't conflict with other syntax.

packages/graphql-codegen-cli/src/yml.ts
```diff
@@ -1,7 +1,19 @@
 import { Types, debugLog } from 'graphql-codegen-core';
 import { safeLoad } from 'js-yaml';
 
+export function interpolate<T extends { [key: string]: string }>(str: string, context: T) {
+  // tslint:disable-next-line:forin
+  for (let key in context) {
```
kamilkisiela: Maybe we should replace all variables outside of `context` with `undefined`?

Something like: `/($([a-z0-9\-_]))/g` and replace `$ASD` with `process.env.ASD` which will give an undefined if value is not available.

kamilkisiela: Now we iterate through all properties of `process.env` and if there's a variable that is not a part of that we end up with a string, `$ASD` for example.

packages/graphql-codegen-cli/tests/cli-flags.spec.ts
ardatan: But I think the standard is the other one.

packages/graphql-codegen-cli/src/yml.ts
ardatan: Done :)

kamilkisiela: I don't trust `.*` but if it works (as test shows) then I'm okay with it :))

jordmccord:
The following codegen.yml does not work, just returns an empty string I've tried with .env and also with the command `graphql-codegen -r dotenv/config`. Any ideas why?
```
schema:
  - ${SCHEMA_URL}:
      headers:
        Authorization: ${BEARER_TOKEN}
  - "src/client/client-schema.graphql"
documents:
  - "src/mutations/*.ts"
  - "src/queries/*.ts"
  - "src/subscriptions/*.ts"
generates:
  ./src/types/apolloTypes.tsx:
    plugins:
      - typescript
      - typescript-operations
      - typescript-react-apollo
    config:
      withMutationFn: false
      withComponent: false
      withHOC: false
      withHooks: true
      scalars:
        DateTime: string
  ./src/client/fragmentTypes.json:
    plugins:
      - fragment-matcher

```

ardatan:
@jordmccord See here https://github.com/dotansimha/graphql-code-generator/issues/2722#issuecomment-542172410

Issue #1365, opened 2019-02-27
fieldnames can give incorrect generated types


maapteh:
**Describe the bug**
fieldnames can give incorrect generated types.

**To Reproduce**
When i have a 'mutation' field in the type 'BasketMutationResult' i get the following result:

```
export namespace AddBasketItem {
    export type Variables = {
        item: BasketMutation;
    };

    export type Mutation = {
        __typename?: 'Mutation';

        addBasketItem: Maybe<AddBasketItem>;
    };

    export type AddBasketItem = {
        __typename?: 'BasketMutationResult';

        mutation: Maybe<Mutation>;
    };

    export type Mutation =
        | BasketMutationType.Fragment
        | BasketDeleteType.Fragment;
}
```
It adds type Mutation twice while the last is not correct.

**Expected behavior**
Not do this :)

**Environment:**
 - OS: OSX
 - Codegen: 0.17.0
 - Node: 10

kamilkisiela:
#1369 should fix it

maapteh:
super!

dotansimha:
Fixed in 0.18.0 🎉

PR #1369, opened 2019-02-27
Duplicated type when fields are named the same as operation types

kamilkisiela:
Related: #1365

### DIFF
--- a/packages/utils/plugins-helpers/src/flatten-types.ts
+++ b/packages/utils/plugins-helpers/src/flatten-types.ts
@@ -86,7 +86,6 @@ export function flattenSelectionSet(selectionSet: SelectionSetItem[], result: Fl
   selectionSet.forEach((item: SelectionSetItem) => {
     if (isFieldNode(item)) {
       if (item.selectionSet.length > 0) {
-        // here
         const model = buildModelFromField(item, result);
         item.type = model.modelType;
         result.push(model);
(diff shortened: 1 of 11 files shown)


packages/utils/plugins-helpers/src/flatten-types.ts
```diff
@@ -84,13 +86,15 @@ export function flattenSelectionSet(selectionSet: SelectionSetItem[], result: Fl
   selectionSet.forEach((item: SelectionSetItem) => {
     if (isFieldNode(item)) {
       if (item.selectionSet.length > 0) {
+        // here
```
dotansimha: here?

packages/utils/plugins-helpers/src/flatten-types.ts
```diff
@@ -84,13 +86,15 @@ export function flattenSelectionSet(selectionSet: SelectionSetItem[], result: Fl
...
         const model = buildModelFromField(item, result);
         item.type = model.modelType;
         result.push(model);
 
         flattenSelectionSet(item.selectionSet, result);
       }
     } else if (isInlineFragmentNode(item)) {
+      // here
```
dotansimha: here?

kamilkisiela: oopsie

packages/utils/plugins-helpers/src/flatten-types.ts
kamilkisiela: oopsie

kamilkisiela:
```suggestion
```

packages/utils/plugins-helpers/src/flatten-types.ts
kamilkisiela:
```suggestion
```

PR #1372, opened 2019-02-28
Migrate changes with graphql-toolkit@0.1.1 (WIP)

DAB0mB:
Waiting for `graphql-toolkit` version `0.1.1` to be released

### DIFF
--- a/packages/graphql-codegen-cli/src/merge-schemas.ts
+++ b/packages/graphql-codegen-cli/src/merge-schemas.ts
@@ -1,14 +1,14 @@
 import { GraphQLSchema, DocumentNode } from 'graphql';
-import { mergeGraphQLSchemas } from 'graphql-toolkit';
+import { mergeTypeDefs } from 'graphql-toolkit';
 import { makeExecutableSchema } from 'graphql-tools';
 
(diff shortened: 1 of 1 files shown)


DAB0mB:
There's a test which gives me the following error:

```
    Cannot use GraphQLNonNull "Int!" from another module or realm.

    Ensure that there is only one instance of "graphql" in the node_modules
    directory. If different versions of "graphql" are the dependencies of other
    relied on modules, use "resolutions" to ensure only one version is installed.

    https://yarnpkg.com/en/docs/selective-version-resolutions

    Duplicate "graphql" modules cannot be used at the same time since different
    versions may have different capabilities and behavior. The data from one
    version used in the function from another could produce confusing and
    spurious results.
```

I think it's caused because of how `yarn link` works, and if we will release a new version it should work on CircleCI. Not sure how to verify it though...

dotansimha:
@DAB0mB I think it's something with you local env. Try to `yarn lerna clean --yes` and then `rm -rf node_modules yarn.lock` and then install with `yarn` from the root dir. 
`graphql-toolkit@0.1.1` is available now.

PR #1374, opened 2019-02-28
Update README.md

AJ-7885:
install the tool in global NPM


dotansimha:
Hi @AJ-7885 
Thank you for your contribution. 
We prefer to install the codegen util as local dependency, because of the need to have the same copy of `graphql` as your project. Also, it's a good idea to generate and build your project in CI, in order to make sure nothing breaks. And installing it globally will not make sure that your have all deps you need for CI.

PR #1389, opened 2019-03-01
Fix IResolvers index signature

ardatan:
- Make `IResolvers` and `IDirectiveResolvers` are extendable with generic resolvers instead of restricting new resolvers by `never`.


dotansimha:
@ardatan please fix ci :)

ardatan:
@dotansimha Fixed!

SimenB:
A patch release with this would be awesome! 🙏

PR #1406, opened 2019-03-05
#1400 Include used introspection types in based on documents

kamilkisiela:


dotansimha:
@kamilkisiela can you please add a test for a document in `typescript-documents` and then use the introspection types? it should compile ts without issues.

kamilkisiela:
Mr @dotansimha, done 👍

packages/plugins/typescript-documents/tests/ts-documents.spec.ts
```diff
@@ -846,6 +846,64 @@ describe('TypeScript Documents Plugin', async () => {
...
+        export enum __TypeKind {
+          Scalar = 'SCALAR', 
+          Object = 'OBJECT', 
+          Interface = 'INTERFACE', 
+          Union = 'UNION', 
+          Enum = 'ENUM', 
+          Input_Object = 'INPUT_OBJECT', 
+          List = 'LIST', 
+          Non_Null = 'NON_NULL'
+        };
```
dotansimha: You need to run `validate` function here

packages/plugins/typescript/src/index.ts
```diff
@@ -22,6 +36,32 @@ export const plugin: PluginFunction<TypeScriptPluginConfig> = (
...
+      }
+    }
+  });
+
+  documents.forEach(doc => visit(doc.content, documentsVisitor));
+
+  const visitor = new TsIntrospectionVisitor(typesToInclude, new TsVisitor(config));
+  const result: DocumentNode = visit(parse(printIntrospectionSchema(schema)), { leave: visitor });
+
+  return result.definitions as any;
```
dotansimha: Isn't it an array? why `any?

kamilkisiela: It should be `string[]` but since we convert bunch of things then `DocumentNode` is no longer valid because `result.definitions` is `DefinitionNode[]` which is no longer true.

packages/plugins/typescript-documents/tests/ts-documents.spec.ts
kamilkisiela: yep

packages/plugins/typescript/src/index.ts
kamilkisiela: so either `result: any` or `result.definitions any`

kamilkisiela:
```suggestion
  return result.definitions as any[];
```

kamilkisiela:
MERGE MERGE MERGE @dotansimha

PR #1407, opened 2019-03-05
Naming convention + fixed double prefix thing

kamilkisiela:


packages/plugins/visitor-plugin-common/src/naming.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  if (['EnumValueDefinition', 'EnumValue'].includes(node.kind)) {
+    return 'enumValues';
+  }
+
+  return 'typeNames';
+}
+
+function getName(node: ASTNode | string): string | undefined {
```
dotansimha: nice, so you are passing the entire node, and then based on the type of the node you are converting the name? 
It's much better than we had before (because we needed to pass names each time).

packages/plugins/typescript/src/visitor.ts
```diff
@@ -81,15 +81,15 @@ export class TsVisitor extends BaseTypesVisitor<TypeScriptPluginConfig, TypeScri
       return new DeclarationBlock(this._declarationBlockConfig)
         .export()
         .asKind('type')
-        .withName(this.convertName(node.name))
+        .withName(this.convertName(node))
```
dotansimha: Please make sure that we don't need to make similar changes for `flow` and all other plugins

packages/plugins/visitor-plugin-common/src/types.ts
```diff
@@ -1,3 +1,18 @@
+import { ASTNode } from 'graphql';
+
 export type ScalarsMap = { [name: string]: string };
 export type EnumValuesMap = { [key: string]: string };
-export type ConvertNameFn = (name: any, addPrefix: boolean) => string;
+export type ConvertNameFn = ConvertFn;
+
+export interface ConvertOptions {
+  addPrefix?: boolean;
```
dotansimha: maybe just `{ prefix: string | null, suffix: string | null }`?

packages/plugins/visitor-plugin-common/src/naming.ts
kamilkisiela: YES! And later on we can add more rules to it and we get name conversion in one place.

packages/plugins/visitor-plugin-common/src/naming.ts
```diff
@@ -0,0 +1,95 @@
...
+      return getName(node.variable);
+    }
+  }
+
+  return undefined;
+}
+
+export function convertFactory(config: { namingConvention?: NamingConvention }): ConvertFn {
+  function resolveConventionName(type: keyof NamingConventionMap): (str: string) => string {
+    if (!config.namingConvention) {
```
dotansimha: Allow it to be `function` as well, because we can't run `resolveExternalModuleAndFn` when we are running in a browser.

packages/plugins/visitor-plugin-common/src/naming.ts
```diff
@@ -76,9 +76,6 @@ export function convertFactory(config: { namingConvention?: NamingConvention }):
     return config.namingConvention[type] as any;
   }
 
-  const convertTypeName = resolveConventionName('typeNames');
-  const convertEnumValues = resolveConventionName('enumValues');
-
   return (node, opts) => {
```
kamilkisiela: > Allow it to be function as well, because we can't run resolveExternalModuleAndFn when we are running in a browser.

@dotansimha is it okay now?

PR #1353, opened 2019-02-25
WIP: Refactor TypeScript plugins

dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/1350 

Planned in this PR:
- [x] Move all common code from `flow` to a common package.
- [x] Make sure `flow` plugin works.
- [x] Add the latest changes of `flow` plugin.
- [x] Create `typescript` package with the same features as `flow` package.
- [x] Write tests for `typescript` package.
 
- [x] Move all common code from `flow-documents` to a common package.
- [x] Make sure `flow-documents` plugin works.
- [x] Add the latest changes of `flow-documents` plugin.
- [x] Create `typescript-documents` package with the same features as `flow-documents` package.
- [x] Write tests for `typescript-documents` package.
- [x] Move all common code from `flow-resolvers` to a common package.
- [x] Make sure `flow-resolvers` plugin works.
- [x] Add the latest changes of `flow-resolvers` plugin.
- [x] Create `typescript-resolvers` package with the same features as `flow-resolvers` package.
- [x] Write tests for `typescript-resolvers` package.
- [x] Add missing features from previous `typescript` package.
- [x] Add missing features from previous `typescript-documents` package.
- [x] Add missing features from previous `typescript-resolvers` package.
- [x] Add introspection types @kamilkisiela  https://github.com/dotansimha/graphql-code-generator/pull/1406
- [x] Implement `namingConvention` again @kamilkisiela  https://github.com/dotansimha/graphql-code-generator/pull/1407
- [x] Fix issue with prefix @kamilkisiela https://github.com/dotansimha/graphql-code-generator/pull/1407
- [x] Apply this: https://github.com/dotansimha/graphql-code-generator/pull/1357 @ardatan 
- [x] Refactor `react-apollo` @dotansimha 
- [x] Refactor `apollo-angular` @dotansimha 
- [x] Apply fix to Flow plugins: https://github.com/dotansimha/graphql-code-generator/pull/1413 @kamilkisiela 
- [x] Refactor `typescript-mongo` @dotansimha  
- [x] Refactor `stencil-apollo` @ardatan
- [x] Remove Handlebars compiler package from all plugins, and add support for tree shaking. @ardatan 
- [x] Make sure to update packages in `gql-gen init`  @kamilkisiela 
- [x] Rename `XXX-documents` to `XXX-operations`
- [x] Move all flatten utils from `core` to `plugin-helpers` and clean `core` package. @kamilkisiela 
[...]

PR #1413, opened 2019-03-06
FIX: fixed bug with useFlowReadOnlyTypes on Enums

zanettin:
just fixing a bug which occurred after introducing `useFlowReadOnlyTypes` on handling Enums.

### current output
```
export const MyEnumValues = Object.freeze({
  +A: 'A',
  +B: 'B',
  +C: 'C'
});
```

### fixed output
```
export const MyEnumValues = Object.freeze({
  A: 'A',
  B: 'B',
  C: 'C'
});
```


dotansimha:
Nice, I'm merging, but I'll need to reimplement it here again: https://github.com/dotansimha/graphql-code-generator/pull/1353, because TypeScript plugins and Flow plugins have common logic now, and we refactored them.

zanettin:
thanks a lot @dotansimha 🎉

PR #1417, opened 2019-03-06
Test and update $ init + rename documents to operations

kamilkisiela:
- [x] add flow plugins
- [x] keep typescript + react related plugin selected by defaults when the project is react based


packages/graphql-codegen-cli/src/init/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  // Emit status to the terminal
+  log(`
+    Config file generated at ${chalk.bold(relativePath)}
+    
+      ${chalk.bold('$ npm install')}
+
+    To install the plugins.
+
+      ${chalk.bold(`$ npm run ${answers.script}`)}
```
kamilkisiela: need to refactor it to `bold` fn

dotansimha: It can also be a template literal tag 😉

packages/graphql-codegen-cli/src/init/questions.ts
```diff
@@ -0,0 +1,137 @@
...
+    },
+    {
+      name: 'Application built with React',
+      key: 'react',
+      value: withFlowOrTypescript([Tags.react, Tags.browser]),
+      checked: possibleTargets.React
+    },
+    {
+      name: 'Application built with Stencil',
+      key: 'react',
```
ardatan: I think this should be stencil

kamilkisiela:
```suggestion
      key: 'stencil',
```

like this?

kamilkisiela: @ardatan

kamilkisiela: @ardatan this suggestion thing is very nice!

ardatan: @kamilkisiela  I am on it

PR #1435, opened 2019-03-10
Fix behavior of gqlImport option

sandiiarov:
`gqlImport` option can be:
- `false` (GraphQL queries will be compiled to GraphQL AST without importing `gql` tag)
- `string` (Using custom `gql` tag modul)
- `undefined` (Using `import gql from 'graphql-tag'`)

### DIFF
--- a/docs/plugins/typescript-react-apollo.md
+++ b/docs/plugins/typescript-react-apollo.md
@@ -47,9 +47,10 @@ Or if you prefer:
 
 ## Configuration
 
-#### `gqlImport` (default value: `import gql from 'graphql-tag'`)
+#### `gqlImport` (default value: `graphql-tag`)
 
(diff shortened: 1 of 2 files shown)


kamilkisiela:
Hi, thanks for your contribution :)

We're preparing a new version of GraphQL Code Generator and the API changed a bit but your use case is now implemented:

```js
{
  gqlImport: 'gatsby#gql',
  noGraphQLTag: true       // turns string into DocumentNode
}
```

sandiiarov:
@kamilkisiela The lib has a bug now. I have described that bug here https://github.com/dotansimha/graphql-code-generator/pull/1347#issuecomment-468934446
Actually this PR inroduce a fix of that bug.

dotansimha:
@sandiiarov fixed in the recent refactor. thank you!

PR #1458, opened 2019-03-14
Move part of logic into a separate package

kamilkisiela:
I tried to move everything that allows to generate an output and is not related to CLI. It doesn't support multiple outputs but it could be used as a building block to other tools and because of that it's as tiny as possible.


kamilkisiela:
I also used `sideEffects: false` all `package.json` files except the one in the CLI (it might break stuff and we don't care that much about the size there).

Also here: https://github.com/ardatan/graphql-toolkit/pull/57

packages/graphql-codegen-cli/package.json
```diff
@@ -89,7 +90,8 @@
   },
   "main": "dist/commonjs/index.js",
   "module": "dist/esnext/index.js",
+  "typings": "dist/index.d.ts",
```
ardatan:
```suggestion
  "typings": "dist/esnext/index.d.ts",
```
I think we forgot that one.

packages/graphql-codegen-cli/package.json
```diff
@@ -89,7 +90,8 @@
...
   "typescript": {
-    "definition": "./dist/index.d.ts"
+    "definition": "dist/index.d.ts"
```
ardatan:
```suggestion
    "definition": "dist/esnext/index.d.ts"
```
This one, too

PR #1460, opened 2019-03-14
Generate different bundles instead of esm package

ardatan:
- Update graphql-toolkit
- Generate two different bundles; esnext and commonjs
- Remove `esm`, because it has problems with `jest`


packages/graphql-codegen-cli/tests/codegen.spec.ts
```diff
@@ -502,6 +502,8 @@ describe('Codegen Executor', () => {
         });
         throw new Error(SHOULD_NOT_THROW_STRING);
       } catch (e) {
+        // tslint:disable-next-line:no-console
+        console.log(e);
```
kamilkisiela: Should we keep it?

ardatan: No, I forgot that :)

Issue #1133, opened 2019-01-07
[TypeScript Resolvers] Apollo Server resolvers types missmatch


zelid:
I use typed resolvers generation and define my resolvers like:

```
import { QueryResolvers, MutationResolvers, SubscriptionResolvers, User } from './generated/graphql'

const query: QueryResolvers.Resolvers = {
  allUsers: (root, args, ctx) => {
    const result: User[] = []
    return result
  }
}

const mutation: MutationResolvers.Resolvers = {
  signup: (root, args, ctx) => {
    const res = ctx.gql.mutation.createUser({
      input: {
        user: {
          firstName: 'John Doe'
        }
      }
    })
    const result: User = { firstname: 'asdf', lastname: 'asdfasdf' }
    return result
  }
}

export default {
  mutation,
  query
}

```

When it comes to configure apollo-server like this:

```
import resolvers from './app/resolvers'

const server = new ApolloServer({
      typeDefs,
      resolvers,
      mocks: true,
      context: (c: any) => ({
        ...c,
        orm: connection,
        gql: postgraphileConnection
      })
    })
```

then apollo-server gives an error about types mismatch:

```
Type '{ mutation: Resolvers<GqlContext, {}>; query: Resolvers<GqlContext, {}>; }' is not assignable to type 'IResolvers<any, any>'.
  Property 'mutation' is incompatible with index signature.
    Type 'Resolvers<GqlContext, {}>' is not assignable to type '(() => any) | GraphQLScalarType | IEnumResolver | IResolverObject<any, any> | IResolverOptions<any, any>'.
      Type 'Resolvers<GqlContext, {}>' is not assignable to type 'IResolverObject<any, any>'.
        Index signature is missing in type 'Resolvers<GqlContext, {}>'.ts(2322)

types.d.ts(28, 5): The expected type comes from property 'resolvers' which is declared here on type 'Config & { cors?: boolean | CorsOptions; }'
```

What should be done in order to solve it?

ardatan:
Could you use title case for Query and Mutation resolvers instead of query and mutation?

zelid:
I've tried that as well, the error stays the same:

```
Type '{ Mutation: Resolvers<GqlContext, {}>; Query: Resolvers<GqlContext, {}>; }' is not assignable to type 'IResolvers<any, any>'.
  Property 'Mutation' is incompatible with index signature.
[...]

PR #1485, opened 2019-03-18
Fix index signature

kamilkisiela:
Related #1133

### DIFF
--- a/dev-test/githunt/flow.flow.js
+++ b/dev-test/githunt/flow.flow.js
@@ -2,23 +2,23 @@
 
 
 export type Scalars = {
-  ID: string,
+          ID: string,
   String: string,
   Boolean: boolean,
   Int: number,
   Float: number,
-};
+        };
 
 export type Comment = {
-  id: $ElementType<Scalars, 'Int'>,
+          id: $ElementType<Scalars, 'Int'>,
   postedBy: User,
   createdAt: $ElementType<Scalars, 'Float'>,
   content: $ElementType<Scalars, 'String'>,
   repoName: $ElementType<Scalars, 'String'>,
-};
+        };
 
 export type Entry = {
-  repository: Repository,
+          repository: Repository,
   postedBy: User,
   createdAt: $ElementType<Scalars, 'Float'>,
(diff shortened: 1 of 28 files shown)


packages/plugins/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -217,7 +217,7 @@ export class BaseResolversVisitor<
 
     const block = new DeclarationBlock(this._declarationBlockConfig)
       .export()
-      .asKind('interface')
+      .asKind('type')
```
dotansimha: Are you sure? I think `class` can't extend/implement a `type`.

dev-test/githunt/flow.flow.js
```diff
@@ -2,23 +2,23 @@
 
 
 export type Scalars = {
-  ID: string,
+          ID: string,
```
dotansimha: What happened here?

packages/plugins/visitor-plugin-common/src/base-resolvers-visitor.ts
kamilkisiela: It can, I guess when using `type A = { ... }` TS reads it as `type A = anonymous interface {...}`

https://stackblitz.com/edit/typescript-j8xye4?file=index.ts

packages/plugins/visitor-plugin-common/src/utils.ts
```diff
@@ -150,13 +152,17 @@ export class DeclarationBlock {
       }
 
       if (this._methodName) {
-        result += `${this._methodName}({${this._config.blockWrapper}
-${this._block}
-${this._config.blockWrapper}})`;
+        const block = `{${this._config.blockWrapper}
```
dotansimha: I think some has been reformatted here and now causes a lot of whitespaces.  @kamilkisiela

kamilkisiela: yep, I'm `yarn build`ing the fix

Issue #1498, opened 2019-03-19
[TypeScript-Resolvers] Same imported type for different GraphQL types


ardatan:
Before 1.0.0, it was working as expected.

codegen.yml
```yml
mappers:
   Admin: ./entities/UserEntity#UserEntity
   User: ./entities/UserEntity#
```

Result
```ts
import { UserEntity, UserEntity } from './entities/UserEntity'; 
```

And this causes `Duplicate identifier 'UserEntity'.` error.

dotansimha:
Fixed in 1.0.1 🚀

PR #1510, opened 2019-03-20
Fix/duplicate imports resolvers

ardatan:
Related to #1498


packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -258,6 +258,76 @@ describe('TypeScript Resolvers Plugin', () => {
...
+      {
+        mappers: {
+          MyType: './my-file#MyCustomOtherType',
+          MyOtherType: './my-file#MyCustomOtherType',
+        },
+      },
+      { outputFile: '' }
+    );
+
+    expect(result).toBeSimilarStringTo(`import { MyCustomOtherType } from './my-file';`);
```
kamilkisiela: Can we check if `MyCustomOtherType` is imported just once?

packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -258,6 +258,76 @@ describe('TypeScript Resolvers Plugin', () => {
...
+        id?: Resolver<Scalars['ID'], ParentType, Context>,
+      };
+    `);
+
+    expect(result).toBeSimilarStringTo(`
+      export type SubscriptionResolvers<Context = any, ParentType = Subscription> = {
+        somethingChanged?: SubscriptionResolver<Maybe<MyCustomOtherType>, ParentType, Context>,
+      };
+    `);
+    await validate(result);
```
kamilkisiela: Ohhh okey, we do that through ts compilation

packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
kamilkisiela: nvm :))

Issue #1497, opened 2019-03-19
[TypeScript-Resolvers] __resolveType doesn't have a correct signature


ardatan:
Schema
```graphql
interface User {
 // someFields
}
```

Generated
```ts
export type UserResolvers<Context = ModuleContext, ParentType = User> = {
  __resolveType: TypeResolveFn<SOME_IMPLEMENTED_TYPES>, // <= Lack of Parent generic
}
```

Expected
```ts
export type UserResolvers<Context = ModuleContext, ParentType = User> = {
  __resolveType: TypeResolveFn<SOME_IMPLEMENTED_TYPES, ParentType, Context>, // <= Lack of Parent generic
}
```

Usage
```ts
const User: UserResolvers = {
   __resolveType: (root) => ...
}
// root: {} here
```

dotansimha:
Fixed in 1.0.1 🚀

PR #1511, opened 2019-03-20
Fix __resolveType signature

ardatan:
Related #1497


kamilkisiela:
Does `__resolveType` gets a context in graphql-js?

dotansimha:
@kamilkisiela yes it does, it should be `(object, context)`

kamilkisiela:
okok

PR #1544, opened 2019-03-22
Add test case for typescript union

brabeji:
Hi @dotansimha !

I've created a failing test case for a possible bug when selecting fields of a union of types sharing common interface. This behavior is expected and supported by `apollo` so I think it's a valid bug.

Thank you for the library!


dotansimha:
Thank you @brabeji ! Great catch! 
I fixed this issue. Changed the logic of transforming the selections set. 
Now all fragments on each selection-set level are being collected, and then they are mapped for fragments that are implementing interfaces, and fragments that aren't, and then it created the intersection between them.

PR #1551, opened 2019-03-23
Removes typesPrefix from non-type definitions

### LINKED ISSUE #1550: Generated components, hooks and HOCs contain typesPrefix
**Describe the bug**
When using `typesPrefix: I` and when generating components, documents, hooks or HOCs, it also adds the type prefix to their names.

**To Reproduce**

1. Configure any project with queries, mutations and/or subscriptions to generate components, hooks or HOCs, and configure with `typesPrefix: I`

**Expected behavior**
I expected typesPrefix to prefix all type exports, but to leave other exports unchanged.

**Environment:**
- Codegen: 1.0.3

jeroenvisser101:
Definitions like Document, HOC, hooks, and Component shouldn't have
typesPrefix added to their generated names.

Related #1550


jeroenvisser101:
If this way of fixing is OK, I'm happy to also write some specs for it to prevent regression.

dotansimha:
Awesome! Thank you @jeroenvisser101 !

PR #1552, opened 2019-03-24
Update programatic usage in docs

dimitar-nikovski:
- convert schema to `DocumentNode` as required by `codegen`
- I also noticed that the `typescript` plugin also currently does not use the `outputFile` option to write the output to the file system.


dotansimha:
Awesome! Thank you @dimitar-nikovski !

Issue #1453, opened 2019-03-13
Subscription generated types are wrong


danfernand:
**Describe the bug**
When generating types from a subscription the types expect something inconsistent with what async iterator returns

Take the below example

Schema
```
const typeDefs = gql`
  type Subscription {
    postAdded: Post
  }

  type Query {
    posts: [Post]
  }

  type Mutation {
    addPost(author: String, comment: String): Post
  }

  type Post {
    author: String
    comment: String
  }
```

Resolver
```
const POST_ADDED = 'POST_ADDED';

const resolvers = {
  Subscription: {
    postAdded: {
      // Additional event labels can be passed to asyncIterator creation
      subscribe: () => pubsub.asyncIterator([POST_ADDED]),
    },
  },
  Query: {
    posts(root: any, args: any, context: any) {
      return postController.posts();
    },
  },
  Mutation: {
    addPost(root: any, args: any, context: any) {
      pubsub.publish(POST_ADDED, { postAdded: args });
      return postController.addPost(args);
    },
  },
};
```
Pubsub publish expects that there will be a wrapper, `postAdded` in the above example that is needed to publish

The generated types do not include this wrapper and make subscription resolvers type incorrect

The expectation is that this wrapper will be included in generated types

Current workaround

If you type the first generic to Subscription in your resolver things work

```
export namespace SubscriptionResolvers {
  export interface Resolvers<Context = ContextGen, TypeParent = {}> {
    postAdded?: PostAddedResolver<
      Subscription | null,
      TypeParent,
      Context
    >;
  }

  export type PostAddedResolver<
    R = Subscription | null,
    Parent = {},
    Context = ContextGen
  > = SubscriptionResolver<R, Parent, Context, PostAddedArgs>;
  export interface PostAddedArgs {}
}
```
 - OS: MacOs
 - Codegen: ^0.14.1
 - Node: v10.15.1

dotansimha:
Fixed in 1.0.4

PR #1556, opened 2019-03-25
Add tests for #1453

ardatan:
Related: #1453


ardatan:
@dotansimha Tests pass, but there is problem about publishing canary.

PR #1572, opened 2019-03-26
noSchemaStitching + warning in md config

kamilkisiela:
- [x] make sure `@warning bla bla bla` works

### DIFF
--- a/packages/plugins/typescript-resolvers/src/index.ts
+++ b/packages/plugins/typescript-resolvers/src/index.ts
@@ -58,14 +58,38 @@ export interface TypeScriptResolversPluginConfig extends RawResolversConfig {
    * ```
    */
   showUnusedMappers?: boolean;
+  /**
+   * @name noSchemaStitching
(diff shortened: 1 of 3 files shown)


packages/plugins/typescript-resolvers/src/index.ts
```diff
@@ -58,15 +58,39 @@ export interface TypeScriptResolversPluginConfig extends RawResolversConfig {
    * ```
    */
   showUnusedMappers?: boolean;
+  /**
+   * @name noSchemaStitching
+   * @type boolean
+   * @description Disables Schema Stitching support
+   * @default false
+   * @warning The default behavior will be reversed in the next major release. Support for Schema Stitching will be disabled by default.
```
kamilkisiela: @dotansimha I added `@warning` but we can rename it. It's just to provide more info in documentation (like deprecation and related).

It turns `@warning bla bla bla` into:

> bla bla bla

Issue #1591, opened 2019-03-28
Suggestion: Parameterized default mapper for typescript-resolvers


jonaskello:
**Is your feature request related to a problem? Please describe.**

Using the [default mapper](https://graphql-code-generator.com/docs/plugins/typescript-resolvers#defaultmapper-string) I cannot refer to the type that is being mapped.

**Describe the solution you'd like**

Today we can do this:

```yaml
plugins
  config:
    defaultMapper: any
```

From what I understand this feature can only replace all types by a single type (`any` in the example above).  I would like to be able to do this:

```yaml
plugins
  config:
    defaultMapper: Partial<{T}>
```

In this example `{T}` would get replaced with the type being mapped. For example if `User` is being mapped then it would be `Partial<User>`. This way I could make all types partial without losing type-safety. 

It could also be used to map types to an indirection. For example if I have this:

```ts
export type Parents = {
  User: UserSpecial;
  // Other parents
};
```

I could then do this:

```yaml
plugins
  config:
    defaultMapper: Parents["{T}"]
```

The benefit of this would be that I could refer to `Parents["User"]` and then easily change what this refers to (this would be another way to achieve what I suggested in #1589).

dotansimha:
Awesome! I'll try to add it :)

dotansimha:
WIP here: https://github.com/dotansimha/graphql-code-generator/pull/1593

dotansimha:
Implemented in 1.0.7 :)

Issue #1589, opened 2019-03-28
Suggestion: Always generate parent type for resolvers


jonaskello:
**Is your feature request related to a problem? Please describe.**

Naming of types that is returned by resolvers is inconsistent.

**Describe the solution you'd like**

In order to set the context for this suggestion, let's define the concept of a `FieldResolver` and a `TypeResolver`. A `FieldResolver` resolves a field on a type, and a `TypeResolver` is a map of `FieldResolver`s.

To give an example, this is generated today by the `typescript-resolvers` plugin:

```ts
export type UserResolvers<Context = any, ParentType = User> = {
  id: Resolver<Scalars["String"], ParentType, Context>;
  name: Resolver<Scalars["String"], ParentType, Context>;
};
```

In this example `UserResolvers` is a `TypeResolver` and `Resolver<...>` is a `FieldResolver`. (A perhaps clearer but also longer naming would be `UserTypeResolver` and `FieldResolver<...>`)

A `TypeResolver` accepts a single `ParentType` which is usually a semi-resolved type that the `TypeResolver` will fully resolve. I think this means that the return type of all `FieldResolver`s is not random, it must always the be exact same type that the `TypeResolver` accepts as `ParentType`. So if we have some types in the schema that have fields that resolves to a `User` type, they all need to return the exact `ParentType` that the `User` type's `TypeResolver` will accept.

The main challange with typings for type-resolvers is how to generate the `ParentType = User` part. Today it will by default assign the full schema type which means all `FieldResolver`s needs to return a fully resolved type which probably is not what we want (as noted in #1219). There is a nice [mapping feature](https://graphql-code-generator.com/docs/plugins/typescript-resolvers#mappers-overwrite-parents-and-resolved-values) that allows us to change this. For example we can map `User: UserParent`. This means that the `User` type's `TypeResolver` will have `ParentType = UserParent`.

Now to my suggestion. When writing a new `FieldResolver` that returns schema-type `User` I need to think about what typescript-type to return. Something like this:

```ts
export const queryResolver: GQLResolvers.QueryResolvers<Context, RootValue> = {
  async users(
    parent: RootValue,
    args: GQLResolvers.QueryUsersArgs,
    ctx: Context
  ): Promise<ReadonlyArray<SOMETHING>> {
      // Resolve and return "SOMETHING"
  },
};
```
[...]

PR #1593, opened 2019-03-28
Improve Resolvers plugins

dotansimha:
This PR will resolve the following issues related to resolvers packages:
- [x] https://github.com/dotansimha/graphql-code-generator/issues/1595
- [x] https://github.com/dotansimha/graphql-code-generator/issues/1591 
- [x] https://github.com/dotansimha/graphql-code-generator/issues/1589
- [x] Implement all in TypeScript
- [x] Implement all in Flow
- [x] Move `unusedMappers` config to the base package
- [x] Update documentation

### DIFF
--- a/packages/plugins/typescript-resolvers/tests/mapping.spec.ts
+++ b/packages/plugins/typescript-resolvers/tests/mapping.spec.ts
@@ -77,6 +77,7 @@ describe('ResolversTypes', () => {
       testSchema,
       [],
       {
+        noSchemaStitching: true,
         mappers: {
           ID: 'number',
           Chat: 'number',
@@ -121,6 +122,7 @@ describe('ResolversTypes', () => {
       schema,
       [],
       {
+        noSchemaStitching: true,
         defaultMapper: 'Partial<{T}>',
       },
       { outputFile: '' }
@@ -148,6 +150,7 @@ describe('ResolversTypes', () => {
       schema,
       [],
       {
+        noSchemaStitching: true,
         defaultMapper: './my-wrapper#CustomPartial<{T}>',
       },
       { outputFile: '' }
@@ -172,6 +175,16 @@ describe('ResolversTypes', () => {
   });
 
   it('Should map to a custom type on every level when {T} is used as default mapper', async () => {
+    const config = {
+      scalars: {
+        ID: 'number',
+      },
+      noSchemaStitching: true,
+      defaultMapper: 'Partial<{T}>',
+      mappers: {
+        User: 'number',
+      },
+    };
     const testSchema = buildSchema(/* GraphQL */ `
       type User {
         id: ID!
@@ -189,28 +202,18 @@ describe('ResolversTypes', () => {
         me: User
       }
     `);
-    const result = await plugin(
-      testSchema,
-      [],
-      {
-        defaultMapper: 'Partial<{T}>',
-        mappers: {
-          User: 'number',
-        },
-      },
-      { outputFile: '' }
-    );
+    const result = await plugin(testSchema, [], config, { outputFile: '' });
 
-    // expect(result).toBeSimilarStringTo(`
-    //   export type ResolversTypes = {
-    //     Query: Partial<Query>,
-    //     User: CustomUser,
-    //     ID: Scalars['ID'],
-    //     String: Scalars['String'],
-    //     Chat: Partial<Chat>,
-    //     Boolean: Scalars['Boolean'],
-    //   };
-    // `);
+    expect(result).toBeSimilarStringTo(`
+    export type ResolversTypes = {
+      Query: Omit<Partial<Query>, 'me'> & { me: Maybe<ResolversTypes['User']> },
+      User: number,
+      ID: Scalars['ID'],
+      String: Scalars['String'],
+      Chat: Omit<Partial<Chat>, 'owner' | 'members'> & { owner: ResolversTypes['User'], members: Maybe<Array<ResolversTypes['User']>> },
+      Boolean: Scalars['Boolean'],
+    };
+    `);
 
     const usage = `
       const resolvers: Resolvers = {
@@ -241,15 +244,7 @@ describe('ResolversTypes', () => {
       }
     `;
 
-    await validate(
-      [result, usage].join('\n\n'),
-      {
-        scalars: {
-          ID: 'number',
-        },
-      },
-      testSchema
-    );
+    await validate([result, usage].join('\n\n'), config, testSchema);
   });
 
   it('Should build ResolversTypes with defaultMapper set', async () => {
@@ -257,6 +252,7 @@ describe('ResolversTypes', () => {
       schema,
       [],
       {
+        noSchemaStitching: true,
         mappers: {
           Query: 'MyQueryType',
           MyType: 'MyTypeDb',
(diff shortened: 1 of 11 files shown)


jonaskello:
Nice work!! 
I have a question about how the replace using Omit works.
I think the idea is to make the types recursivly defined by ResolversTypes?

So with this schema:

```graphql
type MyType {
    foo: String!
    otherType: MyOtherType
  }
type MyOtherType {
    bar: String!
}
type Query {
    something: MyType!
}
```

We would by default get:

```ts
type ResolversTypes = {
    Query: Omit<Query, 'something'> & { something: ResolversTypes['MyType'] },
    MyType: Omit<MyType, 'otherType'> & { otherType: Maybe<ResolversTypes['MyOtherType']> },
    String: Scalars['String'],
    MyOtherType: MyOtherType,
}
```

This looks awesome since the return values of the resolvers now are recursively defined with it's own types instead of the original schema types.

My question is how we would make a Partial of `MyType` above without loosing the recursive definition?
I could add this mappers config:

```yaml
MyType: 'Partial<MyType>'
```

But I guess then I would get:

```ts
type ResolversTypes = {
    Query: Omit<Query, 'something'> & { something: ResolversTypes['MyType'] },
    MyType: Partial<MyType>,
    String: Scalars['String'],
    MyOtherType: MyOtherType,
}
```

So I would get a partial of the orignal `MyType` schema type where `MyType.otherType` is of type `MyOtherType` instead of `ResolversTypes['MyOtherType']`. What I really wanted was a partial of the recursively defined resolver type. Something like this:

```ts
type ResolversTypes = {
    Query: Omit<Query, 'something'> & { something: ResolversTypes['MyType'] },
    MyType: Partial<Omit<MyType, 'otherType'> & { otherType: Maybe<ResolversTypes['MyOtherType']> }>,
    String: Scalars['String'],
    MyOtherType: MyOtherType,
}
```

I'm not sure how but perhaps it would be possible to make the recursively defined resolver types available for additional mapping? Or maybe I am misunderstanding how it is supposed to work. Anyway, keep up the good work :-).

packages/plugins/typescript-resolvers/tests/mapping.spec.ts
```diff
@@ -0,0 +1,796 @@
...
+      {
+        defaultMapper: 'Partial<{T}>',
+        mappers: {
+          User: 'number',
+        },
+      },
+      { outputFile: '' }
+    );
+
+    // expect(result).toBeSimilarStringTo(`
```
kamilkisiela: I will remove that :(

packages/plugins/typescript-resolvers/tests/mapping.spec.ts
```diff
@@ -0,0 +1,796 @@
...
+      Node: CustomPartial<Node>,
+      ID: Scalars['ID'],
+      SomeNode: CustomPartial<SomeNode>,
+      MyUnion: CustomPartial<MyUnion>,
+      MyScalar: Scalars['MyScalar'],
+      Int: Scalars['Int'],
+    };`);
+  });
+
+  it('Should map to a custom type on every level when {T} is used as default mapper', async () => {
```
kamilkisiela: This one throws an error that `number` is not `User`, I think because setting a default mapper (`Partial<{T}>`) makes `ResolversTypes.Chat` to equal `Partial<Chat>` and doesn't not include `Omit` and stuff.

jonaskello:
Regarding my previous comment, perhaps something like this would be possible. Types specified in mappers is put in `MapperTypes` to not overwrite the default/original resolver type which has the Omit stuff. Types in `ResolversTypes` use `MapperTypes` if available, otherwise `ResolversTypes`. 

```ts
export type MapperTypes =  {
  readonly MyType: Partial<ResolversTypes["MyType"]>,
  readonly MyOtherType: ResolversTypes["MyOtherType"],
}

  readonly Query: Omit<Query, 'something'> & { readonly something: MapperTypes['MyType'] },
  readonly MyType: Omit<MyType, 'otherType'> & { readonly otherType: Maybe<MapperTypes['MyOtherType']> },
  readonly String: string,
  readonly MyOtherType: MyOtherType,
};
```

I don't really like it though since it becomes more complex and also it is not clear if types from `MapperTypes` or `ResolversTypes` should be used as return types for resolvers.

dotansimha:
@jonaskello We aim for replacing a type with another type in the level of its declaration, and not in the level of its usage, because we don't want to map fields, but types.
So if you map `MyType: 'Partial<MyType>'`, then `MyType` will get replaced by `Partial<MyType>`, but anywhere you use it, and not in a specific usage of it. 
Mapping fields inside types might be too complicated. 

Regarding `MapperTypes` - it might get too complicated and hard to understand. Our goal it so simplify the usage and map each GraphQL type to a single type. 

The priority `ResolversTypes`'s fields is: look for mappers, then look for default mapper, then look for scalar, otherwise, use the type generated by `typescript`.

packages/plugins/typescript-resolvers/tests/mapping.spec.ts
dotansimha: You are right. I moved the logic and applied it to any type, even when it's in use with `Partial<{T}>`. 
So now the logic of `Omit` is applied as well (basically, apply it always, expect when there is a specific mapper).

jonaskello:
@dotansimha Yes, I agree on the added complexity. Ideally we would have the possibility to use the Omitted types without adding complexity but I don't see a solution for that right now. Anyway, the changes in this PR is a big step forward for resolver typings without that possibility 👍

packages/plugins/typescript-resolvers/tests/mapping.spec.ts
```diff
@@ -0,0 +1,802 @@
...
+
+      type Query {
+        me: User
+      }
+    `);
+    const result = await plugin(testSchema, [], config, { outputFile: '' });
+
+    expect(result).toBeSimilarStringTo(`
+    export type ResolversTypes = {
+      Query: Omit<Partial<Query>, 'me'> & { me: Maybe<ResolversTypes['User']> },
```
kamilkisiela:
```suggestion
      Query: Partial<Omit<Query, 'me'> & { me: Maybe<ResolversTypes['User']>}>,
```

Shouldn't `Partial<T>` be on top of everything? Otherwise `me` is required.

packages/plugins/typescript-resolvers/tests/mapping.spec.ts
```diff
@@ -0,0 +1,802 @@
...
+      User: number,
+      ID: Scalars['ID'],
+      String: Scalars['String'],
+      Chat: Omit<Partial<Chat>, 'owner' | 'members'> & { owner: ResolversTypes['User'], members: Maybe<Array<ResolversTypes['User']>> },
```
kamilkisiela:
```suggestion
      Chat: Partial<Omit<Chat, 'owner' | 'members'> & { owner: ResolversTypes['User'], members: Maybe<Array<ResolversTypes['User']>>> },
```

packages/plugins/typescript-resolvers/tests/mapping.spec.ts
dotansimha: yeah you are right, i'll fix

kamilkisiela:
Looks good 👍

mshwery:
Looks like this [hasn't landed](https://github.com/dotansimha/graphql-code-generator/compare/v1.0.6...master) yet. Any idea on when the next release will be?

dotansimha:
Hi @mshwery 
We have auto-release for canary versions on each change, if you wish to try those new features now, you can use: `1.0.7-alpha-e006b14b.26`.

PR #1635, opened 2019-04-02
Should work with fragment on union type

### LINKED ISSUE #1624: Fragment on union type gives error schemaType.getInterfaces is not a function
**Describe the bug**
Fragment on union type gives error schemaType.getInterfaces is not a function

**To Reproduce**

There is a [reproduction repo here](https://github.com/jonaskello/graphql-codegen-issue1)

1. My GraphQL schema:

```graphql
type Query {
  fooBar: [FooBar!]!
}

union FooBar = Foo | Bar

type Foo {
  id: ID!
}

type Bar {
  id: ID!
}
```

2. My GraphQL operations:

```graphql
query {
  fooBar {
    ...FooBarFragment
  }
}

fragment FooBarFragment on FooBar {
  ... on Foo {
    id
  }
  ... on Bar {
    id
  }
}
```

```yml
schema:
  - "schema.graphql"
overwrite: true
generates:
  generated.ts:
    documents: "query.graphql"
    plugins:
      - "typescript"
      - "typescript-operations"
```

**Expected behavior**
To genereate without error.

**Environment:**

- OS: MacOS 10.13.6
- `@graphql-codegen/...`: 1.0.6
- NodeJS:

jonaskello:
Fixes #1624.

I know the fix is not ideal but the test-case might be helpful for anyone that has a better fix.

### DIFF
--- a/packages/plugins/typescript-operations/tests/ts-documents.spec.ts
+++ b/packages/plugins/typescript-operations/tests/ts-documents.spec.ts
@@ -1184,4 +1184,55 @@ describe('TypeScript Operations Plugin', () => {
       `);
     });
   });
+
+  describe('Issues', () => {
+    it('#1624 - Should work with fragment on union type', async () => {
+      const testSchema = buildSchema(`
+      type Query {
+        fooBar: [FooBar!]!
+      }
+      
+      union FooBar = Foo | Bar
+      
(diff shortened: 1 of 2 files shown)


dotansimha:
Thanks @jonaskello !

PR #1640, opened 2019-04-02
Gracefully handle an interface with no known implementors

mdlavin:
Before this change, the generated code would have included:

```
__resolveType: TypeResolveFn<, ParentType, Context>
```

which would fail to parse due to the missing type parameter


dotansimha:
Awesome! Thanks @mdlavin !

PR #1653, opened 2019-04-05
Set SubscriptionResolver type to full array type.

slagiewka:
The reason for that is that most Flow users will want to use the verbose version of array notation, according to https://github.com/gajus/eslint-plugin-flowtype/blob/master/.README/rules/array-style-simple-type.md

This change will also keep consistency with other plugins (i.e. `flow`) which generate `?Array<sometype>` (cf. https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/flow/src/flow-variables-to-object.ts#L25)


dotansimha:
Thanks @slagiewka !

PR #1671, opened 2019-04-08
do not apply omit to interface when mapper is configured

### LINKED ISSUE #1670: Generated ResolversTypes invalid when using mappers for interface types
When specifying mappers for implementations of an interface where one of the fields is referring to another mapped type, the generated resolver type consists of a union with fields of the graphql type, instead of those from the typescript interface.

I've created a (trivial) sample project where the invalid generated types are visible:

https://github.com/koenpunt/gql-codegen-interfaces

When you look at [resolver-types.ts#L108](https://github.com/koenpunt/gql-codegen-interfaces/blob/master/resolver-types.ts#L108), you'll see that the omit of the `pages` property is invalid (because that property doesn't exist on the typescript type). 

Also when you looks at [resolvers.ts#L43](https://github.com/koenpunt/gql-codegen-interfaces/blob/master/resolvers.ts#L43-45), you'll see that the `__resolveType` function doesn't accept the type that was specified in the mapper.

What a solution could be (and what I currently have applied by using patch-package), is to not apply the omit logic when the interface type has a mapper specified. 
But since interfaces are being considered a special case [here](https://github.com/dotansimha/graphql-code-generator/blob/9c856a4aeae9846328feaebe2786c55b2c12edf2/packages/plugins/visitor-plugin-common/src/base-resolvers-visitor.ts#L196), it might have implications I'm not aware of.

**Environment:**

- OS: 
- `@graphql-codegen/typescript-resolvers`: 1.0.7
- NodeJS: 10.14.1

koenpunt:
This PR makes sure that the omit logic is not applied when a mapper is specified for a graphql interface.

related #1670

P.s. is it possible to run the tests for a single package, instead of the whole suite? Preferably also without having to run `yarn build`.

### DIFF
--- a/packages/plugins/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -193,7 +193,7 @@ export class BaseResolversVisitor<TRawConfig extends RawResolversConfig = RawRes
         }
 
         const schemaType = allSchemaTypes[typeName];
-        if ((shouldApplyOmit && prev[typeName] !== 'any' && isObjectType(schemaType)) || isInterfaceType(schemaType)) {
+        if ((shouldApplyOmit && prev[typeName] !== 'any' && isObjectType(schemaType)) || (isInterfaceType(schemaType) && !this.config.mappers[typeName])) {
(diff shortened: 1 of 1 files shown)


dotansimha:
Looks great, thanks @koenpunt !

PR #1674, opened 2019-04-09
Introduce rootValueType in resolvers

kamilkisiela:
## Adds support for `rootValue` (`{}` by default)

There's a new option called `rootValueType`, points to a type representing the `rootValue`.

```graphql
type Query {
  hello: String!
  name: String!
}
```

```ts
{
  Query: {
    hello(rootValue) { // rootValue: { name: 'Arda' }
      return `Hello ${rootValue.name}!`; 
    },
    name: ({name}) => name,
  }
}
```

## Fixes `parent` object in root types (QMS)

Given the example above, the first argument (`rootValue`) was defined as `Query`:

```ts
{
  Query: {
    hello(rootValue) {
      // Type of rootValue was `ResolversTypes.Query` so doing
      rootValue.name(...)
      rootValue.hello(...)
      // was possible even though rootValue has no methods
    },
  }
}
```

## About signatures of subscription resolvers

Take a look at the diagram below:

subscription-resolvers

Because `subscribe()` is the one that passes value to `resolve()` and it's an `AsyncIterable` then what should we use here as a default type?


packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -713,19 +799,32 @@ describe('TypeScript Resolvers Plugin', () => {
         const resolvers: Resolvers = {
           Subscription: {
             postAdded: {
-              // Additional event labels can be passed to asyncIterator creation
               subscribe: () => pubsub.asyncIterator([POST_ADDED]),
```
kamilkisiela: The iterator resolves **a** value, we're not sure what it is.

packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -713,19 +799,32 @@ describe('TypeScript Resolvers Plugin', () => {
...
-            },
+              // Make sure we get and return a proper value
+              resolve: post => post
```
kamilkisiela: The result of `subscribe()` is passed here, we know what it should pass to nested resolvers so it's all good here.

packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -713,19 +799,32 @@ describe('TypeScript Resolvers Plugin', () => {
...
+            }
           },
           Mutation: {
-            addPost: (root, args) => {
-              pubsub.publish(POST_ADDED, { postAdded: args });
-              return args;
+            addPost: (root, { author, comment }) => {
```
kamilkisiela: Here the `root` argument is strongly typed - `{ version: number }`

packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -713,19 +799,32 @@ describe('TypeScript Resolvers Plugin', () => {
...
+              const post = {
+                author,
+                comment,
+              };
+
+              // RootValue should be accessible
+              console.log(root.version);
```
kamilkisiela: Because of that, doing `root.version` should be possible (no ts error)

packages/plugins/typescript-resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -680,6 +680,85 @@ describe('TypeScript Resolvers Plugin', () => {
...
+        query: MyQuery
+        mutation: MyMutation
+        subscription: MySubscription
+      }
+    `);
+    const content = await plugin(
+      testSchema,
+      [],
+      {
+        rootValueType: 'MyRoot',
```
kamilkisiela: That's the idea of `rootValueType`, it's the same for all root types and it's something different than `context` or even a `mapper` (it shouldn't be wrapped with `Partial` and it's the same signature for Query, Mutation and Subscription - using mapper will require to define the type for all three objects).

dotansimha:
Looks good @kamilkisiela . Just fix ci :)

Issue #1631, opened 2019-04-02
Suggestion for typescript-resolvers: Use generated resolver type in regular mappers


jonaskello:
**Is your feature request related to a problem? Please describe.**

Use generated resolver type in regular mappers.

**Describe the solution you'd like**

This suggestion builds on the features introduced in #1593.

The changes means that there are now special resolver return/parent types generated for all schema types, and these types are returned by default by resolvers. Also these resolver types are defined in each other by default, so one resolver type has fields of the other resolver types rather than the raw schema types.

However if we introduce a mapper it will break the chain of resolver types referencing each other as noted in [this comment](https://github.com/dotansimha/graphql-code-generator/pull/1593#issuecomment-477697479) and [this comment](https://github.com/dotansimha/graphql-code-generator/pull/1593#issuecomment-477751326).

To solve the above, perhaps we can introduce a parameter to the regular mappers the same way that the default mapper now have a parameter `{T}`. In the case or regular mappers the parameter would be the default generated type. 

Let's say by default we have this:

```graphql
type MyType {
    foo: String!
    otherType: MyOtherType
  }
type MyOtherType {
    bar: String!
}
type Query {
    something: MyType!
}
```

```ts
type ResolversTypes = {
    Query: Omit<Query, 'something'> & { something: ResolversTypes['MyType'] },
    MyType: Omit<MyType, 'otherType'> & { otherType: Maybe<ResolversTypes['MyOtherType']> },
    String: Scalars['String'],
    MyOtherType: MyOtherType,
}
```

Now we want to make `MyType` partial without breaking the chain of referencing resolver types. If mappers had a parameter which gave the default generated resolver type we could have this config:

```yaml
plugins
  config:
    mappers:
      MyType: Partial<{T}>
```

In this example the `{T}` would be `Omit<MyType, 'otherType'> & { otherType: Maybe<ResolversTypes['MyOtherType']> }` since that is the generated resolver-type. The output from this would be:

```ts
type ResolversTypes = {
    Query: Omit<Query, 'something'> & { something: ResolversTypes['MyType'] },
    MyType: Partial<Omit<MyType, 'otherType'> & { otherType: Maybe<ResolversTypes['MyOtherType']> }>,
    String: Scalars['String'],
    MyOtherType: MyOtherType,
}
```

kamilkisiela:
#1677

dotansimha:
Available in `1.1.0` 🚀

PR #1677, opened 2019-04-09
Allow to use {T} in mappers

kamilkisiela:
Related #1631

### DIFF
--- a/packages/plugins/typescript-resolvers/tests/mapping.spec.ts
+++ b/packages/plugins/typescript-resolvers/tests/mapping.spec.ts
@@ -132,16 +132,16 @@ describe('ResolversTypes', () => {
     export type ResolversTypes = {
       Query: Partial<Query>,
       MyType: Partial<MyType>,
-      String: Scalars['String'],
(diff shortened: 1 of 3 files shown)


packages/plugins/typescript-resolvers/tests/mapping.spec.ts
```diff
@@ -132,16 +132,16 @@ describe('ResolversTypes', () => {
     export type ResolversTypes = {
       Query: Partial<Query>,
       MyType: Partial<MyType>,
-      String: Scalars['String'],
+      String: Partial<Scalars['String']>,
```
kamilkisiela: One small change: if `defaultMapper` contains the `{T}` we apply it even on scalars. I think it makes sense because `Scalars['ID']` might be anything, a primitive but also an object.

kamilkisiela: It's not always `Partial<{T}>`, maybe someone wants to wrap it like this `{T} | Wrapped<{T}>`.

PR #1682, opened 2019-04-09
change generic parameter name of context type to prevent clash

### LINKED ISSUE #1681: Mapped Context type can't be named Context
**Describe the bug**

When specifying `contextType` in the config with a type called `Context`, the generated typescript file will contain multiple `Context` types, because `Context` is also the name of the generic parameter.

This should be easily solvable by changing the name of the generic parameter.

koenpunt:
when having a context interface name `Context`, it didn't work because the generic parameter was referring to itself. By naming it `ContextType` it's less common, and thus the change of having clash is smaller.

fixes #1681

### DIFF
--- a/dev-test/test-schema/resolvers-root.ts
+++ b/dev-test/test-schema/resolvers-root.ts
@@ -103,26 +103,26 @@ export type ResolversTypes = {
   Boolean: Scalars['Boolean'],
 };
 
-export type QueryRootResolvers<Context = any, ParentType = ResolversTypes['QueryRoot']> = {
-  allUsers?: Resolver<Array<Maybe<ResolversTypes['User']>>, ParentType, Context>,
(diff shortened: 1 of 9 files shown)


dotansimha:
Thanks @koenpunt . Can you please update the failing tests?

koenpunt:
I was trying to, but running the tests takes forever.. is there a way to speed that up? Also would be nice to not have to run `yarn build` every time before running the tests.

dotansimha:
You can run `yarn build` in the root to build all packages. Then you can use `yarn build` or `yarn test` inside a package directory. 
You can also speed up the tests by running `SKIP_VALIDATION=1 yarn test` (it will skip the typescript compilation tests)

kamilkisiela:
Maybe `T` prefix everywhere, including `TContext`?

koenpunt:
@kamilkisiela in this context (no pun intended) it's only an internal alias. Also `ContextType` matches with the already present `ParentType` used at the same locations.

koenpunt:
Done

Issue #1626, opened 2019-04-01
React-Apollo variables wrapped in Partial<>?


stevenmusumeche:
I'm experimenting with this library and auto-generating react-apollo components. When using the generated component, I do not get a type error when I don't provide a required variable to the query.  For example:

```ts
// original query
export const PRODUCT_DETAIL = gql`
  query ProductDetail($brandKey: String!, $productKey: String!) {
    getProductDetail(input: { brandKey: $brandKey, productKey: $productKey }) {
      id
      productNumber
      productKey
      erpProductKey
    }
  }
`
```

It generates a react-apollo component `ProductDetailComponent`.  When I use that component like this, with the missing brandKey and productKey, there are no type errors, even though those variables are required by the query. 

```ts
const myFunc = () => {
   return <ProductDetailComponent>{data => {
      // etc
    }}</ProductDetailComponent>
}
```

What is the reason for wrapping the required variables in a `Maybe`? I would expect a type error when I tried to use ProductDetailComponent without the required input variables.

dotansimha:
@stevenmusumeche, thank you for reporting it. Can you please share the generated component and type?

stevenmusumeche:
Here are the generated types:

```
export class ProductDetailComponent extends React.Component<
  Partial<
    ReactApollo.QueryProps<ProductDetailQuery, ProductDetailQueryVariables>
  >
> {
  render() {
    return (
      <ReactApollo.Query<ProductDetailQuery, ProductDetailQueryVariables>
        query={ProductDetailDocument}
        {...(this as any)['props'] as any}
      />
    )
  }
}

export type ProductDetailQueryVariables = {
  brandKey: Scalars['String']
  productKey: Scalars['String']
}
```

Usage like this works:

```
<ProductDetailComponent
          variables={{
            brandKey: 'some-string'
            productKey: 'some-other-string'
          }}
        >
// etc
</ProductDetailComponent>
```

I also get a type error (correctly) if I leave off one of the required variables

```
<ProductDetailComponent
          variables={{
            brandKey: 'some-string' // type error, missing productKey
          }}
        >
// etc
</ProductDetailComponent>
```

However, if I leave off the `variables` property completely, I get no type errors:

```
<ProductDetailComponent>
// etc
</ProductDetailComponent>
```

dotansimha:
Fixed in `1.1.0` 🚀

PR #1685, opened 2019-04-10
Enhancements for plugins

ardatan:
- [x] Change the way we import from external dependencies in generated files instead of using ‘* as LibraryName’.
- [x] Update stencil-apollo
- [ ] Solve the issue #1626 
- [ ] Fix tests


dev-test/githunt/types.reactApollo.tsx
```diff
@@ -224,7 +224,7 @@ export type VoteMutation = ({ __typename?: 'Mutation' } & { vote: Maybe<({ __typ
 
 import gql from 'graphql-tag';
 import * as React from 'react';
-import * as ReactApollo from 'react-apollo';
```
kamilkisiela: @ardatan it's treated as a namespace, so it won't increase bundle size, totally safe.

kamilkisiela: https://github.com/kamilkisiela/tree-shake-test/tree/apollo

Here's a playground

PR #1687, opened 2019-04-10
Make variables prop required if any of argument is non-null

ardatan:
Related to #1626


dotansimha:
@ardatan can you please add `export` to the `Omit` type?

Issue #1633, opened 2019-04-02
Suggestion: Allow __typename in types returned by resolvers


jonaskello:
**Is your feature request related to a problem? Please describe.**

When working with resolvers that returns an union type there is a need to resolve the concrete type. Graphql supports [several ways](https://stackoverflow.com/a/53082574/2761797) to do this. One way is the `__resolveType` function. However there is also a built-in default implementation of the `__resolveType` function  which will check the `__typename` property of the returned value. This is convinient and allows us to write resolvers like this without having to implement `__resolveType` ourselves:

```ts
  async resolverThatReturnsUnionType(parent, args ctx) {
    return {
      __typename: "MyConcreteType",
      myprop: 1
    } as any; // any required becuase generated types does not include __typename
  },
```

However the generated types for resolvers does not allow `__typename` to be included. 

**Describe the solution you'd like**
I propose that `__typename` is included in an optional fashsion for these types, same as for operation types.

**Describe alternatives you've considered**
I could add a lot of mappers and add the `__typename` field, but since this functionality is built-in to `graphql` I think it should also be built-in for the type generator tool.

jonaskello:
Ideally, the optional `__typename` field could be added to the raw schema types generated by the `typescript` plugin. That would solve the above and also my more unusual use-case mentioned [here](https://github.com/dotansimha/graphql-code-generator/issues/1350#issuecomment-467345175).

This change would be non-breaking if the `__typename` field is generated as optional. I guess the `typescript` plugin could have a setting for this if people really don't want it generated.

jonaskello:
These are the use-cases for `__typename` that I currently know of:

1. You can ask for `__typename` in an operation sent to the server.
2. You can use `__typename` in resolvers for resolving types of union and interfaces instead of `__resolveType`.
3. You can use `__typename` to construct an unique ID for a normalized cache (I think apollo allows you to construct your own ID like this).

Given that the above use-cases spans across `typescript-operations` and `typescript-resovlers` I think it would make sense to include `__typename` in the base types generated by the `typescript` plugin.

I could work on a PR for this if it would be accepted.

leebenson:
👍 for this. We have a ton of code that branches based on a union's/interface's returned `__typename`, and being able to select that as a first-class field gives this parity with the GraphQL spec.

Currently, we're working around this by having a concrete field called `type` that returns the same thing, but that's redundant since it's being auto-generated by our GraphQL server.

[...]

Issue #649, opened 2018-09-27
[TypeScript] Add __typename to typings


eljenso:
I use GraphQL interfaces.
When querying for a field that might be any type implementing a specific interface I need to know which implementation it is, for this I currently use `__typename`.

Would be great if you would add the `__typename` field for all typings (should probably be an optional field).

dotansimha:
@eljenso can you please provide an example for where it's missing? 
We added `__typename` for all client-side generated interfaces...

eljenso:
Here is part of my schema: https://gist.github.com/eljenso/b2b43f30cdc3e9aa765f10df99de67a3
Running `apollo schema:download --endpoint 'http://localhost:4000'` yields this `schema.json`: https://gist.github.com/eljenso/6a8a53ce5b36edb19fb3d7a6fd3913cb
And after running `npx gql-gen --schema schema.json --template graphql-codegen-typescript-template --out ./src/typings/`, the resulting `types.ts` is this: https://gist.github.com/eljenso/5139abfbf4bed47c705120f5de46bedc

As you can see, interfaces `IElement` and `Section` do not have a `__typename` field.

dotansimha:
@eljenso Can you explain you usage of `__typename` in the generated typings for the server side? They are generated automatically by `graphql` and you shouldn't deal with them as far as I know.

The `__typename` is an internal field, and usually it's usage is for client-side to know what is the declared type of each structure. 

Btw, you don't have to download the schema first, you can just use `--schema http://localhost:4000` ;)

eljenso:
> The __typename is an internal field, and usually it's usage is for client-side to know what is the declared type of each structure.

This is exactly what I want to do but the typings generated by `graphql-codegen-typescript-template` do not have `__typename`.
I have some GraphQL types which do only differ in name (i.e. fields are all the same), so I need `__typename` client-side in order to differentiate between those TS interfaces.

jonaskello:
I would also like `__typename` to be optionally added to the full schema types. My use-case is that I have my own client-side caching solution and the ability to modify the cache (for optimistic updates). So when I modify the cache on the client-side I use the full schema types. This is done because the optimistic update is allowed to add any field within the full type, but no fields that are not in the full type. So for this reason I'm using the full types on the client-side. Not for queries but strictly for cache manipulation.

Elijen:
[...]

PR #1701, opened 2019-04-11
Add __typename to base TS and flow types

jefflewis:
As seen in these issues, `__typename` being included in the base type will help with comparing types returned by unions.

* https://github.com/dotansimha/graphql-code-generator/issues/1633
* https://github.com/dotansimha/graphql-code-generator/issues/649

I hooked into the `skipTypenames` configuration flag. If `undefined` or `false`, `__typename` will be generated as an optional type in base types.


dotansimha:
Hi @jefflewis , thank you for your contribution! 
I'll review it very soon.

kamilkisiela:
And maybe `__typename` should be only defined in interfaces?

dotansimha:
@kamilkisiela why only for interfaces? I think we can add it for types as well.
My only concern is that now we should use `Pick` with `__typename` instead of manually adding it. But maybe we can do it after merging this PR.

jefflewis:
@kamilkisiela Adding `__typename` to the generated type is just as important here. Many of my generated GQL types are included in unions. Not having `typename` for these is exactly what I'm trying to fix.

@dotansimha I was unsure of a good way to refactor this to use a `Pick`, or to make it non-optional (note how the read-only flow type is not being used here) as I was also unsure of how to get that addressed.

It seems that both could be addressed in a future commit. Anyone needing typenames would be helped by this. and adding `skipTypename: false` would get the current behavior. It seems changing the default would be okay, but would similar "break" people's configs since the default is already true for operation types.

kamilkisiela:
@jefflewis ok, seems reasonable. 

About the default, I think since `graphql-js` doesn't add `__typename` property, that's what the developer does, then we shouldn't show it to people, it might be misleading.

```ts
{
  users: (parent, args, context, info) {
    // parent.__typename is accessible but undefined
  }
}
```

jefflewis:
@kamalgill  I'm not sure I understand what you mean here... 

If `__typename` is output in the generated type, it will be accessible and defined. If the config is set to not generate, it will not be accessible or defined.

Changing the default to not generate `__typename` seems fine, but there's a lot more code to refactor to accomodate that change, since `skipTypename` is used in the `typescript-operations` plugin. Either way is fine with me, since all I need is to get the typename into the generated types! 😄

jefflewis:
I had to rebase after all of the packages moved into folders...

cristianosantos95:
`skipTypename` is already on the docs (https://graphql-code-generator.com/docs/plugins/flow-operations) but the flow definitions aren't being generated with `__typename`. Am I missing something or it's not working because this PR is still open?

tokenvolt:
@cristianosantos95 this is for GraphQL operations, i.e. query, mutation, subscription and fragment, but not for the GraphQL types.

jefflewis:
Is there anything I need to change about this before it gets merged and released? I'm more than willing to update the PR with any necessary changes.

cristianosantos95:
@tokenvolt you mean this pull request or the one that's on the docs? Either way, none of my generated definitions have `__typename` (GraphQL operations and types). Is there a way to add it?

tokenvolt:
@cristianosantos95 on the docs. This PR adds `__typename` to GraphQL objects as well.

vasilich6107:
Hi! Thanks for library and PR.
I’m awaiting of this PR to be merged)

dotansimha:
Hi @jefflewis , Sorry for the delay, I hope I'll find the time to merge it in the next days.

jefflewis:
No worries @dotansimha! Ping me if you need me to make any changes.

Jonatthu:
@jefflewis Is this helping or related to this issue?
https://github.com/dotansimha/graphql-code-generator/issues/643#issuecomment-491026828

andreyvital:
@dotansimha sorry to push on this, but do you have an expectation of when to merge this PR? I'm not getting `__typename` the generated definitions for TS. Thanks!

BeeeQueue:
We are also waiting eagerly for this PR to be merged and released, we need it to migrate from Apollo's type generation

dotansimha:
Hi all, sorry for the delay. I'm rebasing this PR manually now, and then I'll merge it to `master`. It still requires some changes because we need now to `Pick<>` the `__typename` as well, so it might take a bit more time (a day or so maybe) to have a stable release with it.

dotansimha:
I think, at the moment, I'll merge it as-is to allow the base plugins to add `__typename`, and `x-operations` plugins will add `__typename` by themself (like we do today).
The reason for that is might be a bit more complicated to do `Pick<Type, '__typename'>` when `__typename?` is optional, and then to remove the optional sign from it (in case the query contains `__typename` explicitly).
Everything should work as-is, and maybe in a future release we'll use `Pick` to get the `__typename` as well (it's just an optimization, so it's not required at the moment).

packages/plugins/typescript/typescript/tests/typescript.spec.ts
```diff
@@ -287,6 +291,7 @@ describe('TypeScript', () => {
 
       expect(result).toBeSimilarStringTo(`
         export type MyType = {
+          __typename?: "MyType",
```
lookfirst: this should be readonly __typename

dotansimha: It could, we have a configuration for immutable types, be default they are not.

Issue #1678, opened 2019-04-09
[TypeScript Compatibility] Should work with strict mode 


dotansimha:
Without strict mode:
![image](https://user-images.githubusercontent.com/3680083/55800235-e18d9200-5adb-11e9-9176-1fc677e4b122.png)

With strict mode:
![image](https://user-images.githubusercontent.com/3680083/55800248-ece0bd80-5adb-11e9-969b-2701e876c641.png)

I think we need to add support for it, by wrapping the proxy type with `NonNullable`:

```
export type Users = NonNullable<NonNullable<GetAccountUsersQuery['account']>['users']>[0];
```

josemarluedke:
We are trying to upgrade to v1 to solve some other issue, so we started using the compatibility layer to help on the migrations and we hit this as well. We definably want to keep strict mode enabled. Any help on this would be highly appreciated.

dotansimha:
@josemarluedke I'm working on fixing that. 
Did you had other issues with the rest of the plugins?

dotansimha:
Fixed in https://github.com/dotansimha/graphql-code-generator/commit/57e0304d100f4a35460fb0a3b7395e036529cbd2

jrr:
I think I'm hitting another variant of this, accessing a field on a record in an array:
```gql
# schema
type Mutation {
  employeeCreateWidgets(data: EmployeeCreateWidgetArgs!): [Widget]
}
```

```gql
# mutation
mutation EmployeeCreateWidgets($data: EmployeeCreateWidgetArgs!) {
  employeeCreateWidgets(data: $data) {
    employee {
      id
      name
      }
    }
  }
```

```ts
// generated code
export namespace EmployeeCreateWidgets {
  export type Variables = EmployeeCreateWidgetsMutationVariables;
  export type Mutation = EmployeeCreateWidgetsMutation;
  export type EmployeeCreateWidgets = (NonNullable<
    EmployeeCreateWidgetsMutation["employeeCreateWidgets"]
  >)[0];
  export type Employee = (NonNullable<
    EmployeeCreateWidgetsMutation["employeeCreateWidgets"]
  >)[0]["employee"];
}
```

That final array access (_"employee"_) is squiggled, with the following error:

> Property 'employee' does not exist on type 'Maybe<{ __typename?: "Widget" | undefined; } & { employee: { __typename?: "Employee" | undefined; } & Pick<Employee, "id" | "name">; }>'.ts(2339)

[...]

PR #1719, opened 2019-04-16
Handle another TS strict mode case

jrr:
This is meant to address https://github.com/dotansimha/graphql-code-generator/issues/1678#issuecomment-482420680 .


dotansimha:
That's perfect! Thanks @jrr !

Issue #1725, opened 2019-04-17
Flow resolvers signature shouldn't use optional


dotansimha:
Related:
https://github.com/dotansimha/graphql-code-generator/issues/489

dotansimha:
Fixed in `1.1.0` 🚀

PR #1726, opened 2019-04-17
fix(flow-resolvers): non optional arguments in resolvers

loicplaire:
Fixes https://github.com/dotansimha/graphql-code-generator/issues/1725.

## The changes

- As discussed, the resolvers arguments shouldn't be optional. 
- There was also no test coverage for these type definitions so I added a snapshot test which covers them as well as the existing "string comparison" tests (removed since they were made obsolete by the snapshot test).


dotansimha:
That's awesome!!! Thank you @loicplaire . 
Will merge right after ci passes.

PR #1738, opened 2019-04-18
Added transformUnderscore configuration to namingConvention

roman-may:
Added the field transformUnderscore to the namingConvention object to make it possible to override the default behavior to preserve Underscores. Without this it is not possible to fully customize the naming convention for the generated types, only the individual parts between underscores.

I was not sure if this is the best place for the configuration. Let me know if I should change it.


dotansimha:
Thanks @roman-may :)

Issue #1723, opened 2019-04-17
Generate MutationFn types when not using HOCs


bopfer:
When `withHOC` is true, `MutationFn` types will be generated for each mutation.  But, with HOCs off they are not generated. 

**Describe the solution you'd like**

It would be nice if they would be generated when not using HOCs as well.  Or have an option to generate them.  I am using the generated Components, but often pass the mutation function around to callbacks.  Currently, my team is manually creating all the MutationFn for each mutation.

I could turn on HOCs, but we are not using them and I would rather not add to our bundle size by having them in there.

dotansimha:
Fixed in 1.1.1 🎉

JClackett:
I think this breaks the react-apollo-hooks plugin, as it generates MutationFn types using "react-apollo" as the namespace i.e. ReactApollo.MutationFn, however react-apollo doesn't necessarily have to be installed to use react-apollo-hooks so I'm getting the error ...

```
Exported type alias 'MyAmazingMutationFn' has or is using private name 'ReactApollo'.
```

My plugin config is this:

```yaml
    config:
      withHooks: true
      withComponent: false
      withHOC: false
```

Ideally it would use the ReactApolloHooks namespace if hooks is true and the others false
```
import * as ReactApolloHooks from "react-apollo-hooks"
```

PR #1739, opened 2019-04-18
Add withMutationFn

ardatan:
Related to #1723


SimenB:
Defaulting this to `true` is a breaking change if you don't use `react-apollo` (i.e. only use hooks).

image

dotansimha:
@ardatan ^

Issue #1672, opened 2019-04-09
Set HOC config.alias to improve debuggability


slikts:
[config.alias] in react-apollo docs.

Currently the generated HOCs don't set any options, but it would be useful if `config.alias` was set because it improves the representation of the HOC in React DevTools.

When the alias is left as default, the wrapped components are named `Apollo(WrappedComponent)`, while setting the `alias` could make them look like `withHocName(WrappedComponent)`.

[config.alias]: https://www.apollographql.com/docs/react/api/react-apollo#graphql-config-alias

dotansimha:
Fixed in 1.1.1 🎉

PR #1740, opened 2019-04-18
Set HOC alias

ardatan:
Related to #1672

### DIFF
--- a/packages/plugins/typescript/react-apollo/src/visitor.ts
+++ b/packages/plugins/typescript/react-apollo/src/visitor.ts
@@ -68,7 +68,10 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
   ${operationResultType},
   ${operationVariablesTypes},
   ${propsTypeName}<TChildProps>>) {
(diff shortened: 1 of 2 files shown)


packages/plugins/typescript/react-apollo/src/visitor.ts
```diff
@@ -68,7 +68,10 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
   ${operationResultType},
   ${operationVariablesTypes},
   ${propsTypeName}<TChildProps>>) {
-    return ReactApollo.with${titleCase(node.operation)}<TProps, ${operationResultType}, ${operationVariablesTypes}, ${propsTypeName}<TChildProps>>(${documentVariableName}, operationOptions);
+    return ReactApollo.with${titleCase(node.operation)}<TProps, ${operationResultType}, ${operationVariablesTypes}, ${propsTypeName}<TChildProps>>(${documentVariableName}, {
+      alias: 'with${operationName}',
```
dotansimha: Can you please check if `operationName` is set before using it? Because it might be empty

ardatan: Do you ask for only here or in other parts as well?
because we don't check for it in anywhere in this plugin.

dotansimha: Hmm so let's leave it as is. maybe later we can add it on a higher level.

PR #1743, opened 2019-04-18
Update import statement in getting-started docs

betterin30days:
The docs referenced a method from the incorrect package. `generate` is available in `@graphql-codegen/cli` and `codegen` is available in `@graphql-codegen/core`. From the usage in the code that followed, it seems `generate` is the intended method.


dotansimha:
Thank you @betterin30days !

PR #1746, opened 2019-04-19
docs(getting-started): import generate from cli not core

skvale:
https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/generate-and-save.ts#L11

This exported method must have been moved from core to cli without updating the docs?


dotansimha:
Thanks!

Issue #1747, opened 2019-04-19
Using {T} in mapper type imported from file produces wrong import


robak86:
**Describe the bug**

The problem is related to custom wrapper type imported from a file using ```./someFile$SomeType<{T}> ``` syntax. This feature already works for global types #1631, but for types coming from file generates wrong imports. 

**To Reproduce**

1. My custom mapper type

```typescript
export type UserExtension<T> = Partial<T> & {
    someDatabaseRelationId: number
}
```
2. My `codegen.yml` config file:

```yml
plugins:
      - typescript
      - typescript-resolvers
config:
    mappers:
       User: ./userExtension#UserExtension<{T}>
```

3. Generated types

```typescript 
import { UserExtension<{T}> } from './userExtension; // wrong import

export type ResolversTypes = {
  Concept: UserExtension<User> //correct type
}
```

**Expected behavior**

Correct types are generated

```typescript 
import { UserExtension } from './userExtension; // wrong import

export type ResolversTypes = {
  Concept: UserExtension<User> //correct type
}
```

**Environment:**

- OS: macOS 10.14.3
- `@graphql-codegen/1.1.0`: 
- NodeJS: v11.12.0

robak86:
Fixed at #1748

dotansimha:
Fixed in 1.1.1 🎉

PR #1748, opened 2019-04-19
Remove interpolation fragment from import symbol for concrete type ma…

robak86:
Fix for #1747


dotansimha:
Awesome. Thank you @robak86 !

PR #1782, opened 2019-04-27
Docs fix and import React Apollo when withMutationFn is configured

Whazor:
Small fix in docs. Also update to import ReactApollo when `withMutationFn` is enabled but ReactApollo features are disabled.


Whazor:
also, parameter does not work....

Whazor:
Ok. That does work. I have a different problem because `_buildMutationFn` uses `ReactApollo` but when`this.config.withComponent || this.config.withHOC` is false the module does not get imported.

dotansimha:
Thank you @Whazor ❤️

PR #1814, opened 2019-05-06
Urql support

### LINKED ISSUE #1575: Add new template or config option to `apollo-react` to support `urql`
Please consider adding a special template or config for `apollo-react` generator to support
https://github.com/FormidableLabs/urql/
It's quite popular and will be nice to have.

mxstbr:
I mainly just copied the react-apollo plugin and rewrote the generated code to use the urql API, which differs from the `react-apollo` and `react-apollo-hooks` a bit.

- [x] Generate hooks
- [x] Generate components
- [x] Fix options
- [x] Test whether generated code actually works

Closes #1575

### DIFF
--- a/packages/plugins/typescript/urql/src/visitor.ts
+++ b/packages/plugins/typescript/urql/src/visitor.ts
@@ -6,19 +6,17 @@ import { toPascalCase } from '@graphql-codegen/plugin-helpers';
 import { titleCase } from 'change-case';
 
 export interface UrqlPluginConfig extends ClientSideBasePluginConfig {
-  withHOC: boolean;
   withComponent: boolean;
   withHooks: boolean;
-  withMutationFn: boolean;
+  urqlImportFrom: string;
 }
 
 export class UrqlVisitor extends ClientSideBaseVisitor<UrqlRawPluginConfig, UrqlPluginConfig> {
   constructor(fragments: FragmentDefinitionNode[], rawConfig: UrqlRawPluginConfig) {
     super(fragments, rawConfig, {
-      withHOC: getConfigValue(rawConfig.withHOC, true),
       withComponent: getConfigValue(rawConfig.withComponent, true),
       withHooks: getConfigValue(rawConfig.withHooks, false),
-      withMutationFn: getConfigValue(rawConfig.withMutationFn, true),
+      urqlImportFrom: getConfigValue(rawConfig.urqlImportFrom, null),
     } as any);
 
     autoBind(this);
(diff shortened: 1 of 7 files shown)


mxstbr:
Testing this in our project, this fails because `Urql.UseQuery|MutationArgs` is not exported from `urql`. /cc @kitten

packages/plugins/typescript/urql/src/visitor.ts
```diff
@@ -0,0 +1,75 @@
...
+  urqlImportFrom: string;
+}
+
+export class UrqlVisitor extends ClientSideBaseVisitor<UrqlRawPluginConfig, UrqlPluginConfig> {
+  constructor(fragments: FragmentDefinitionNode[], rawConfig: UrqlRawPluginConfig) {
+    super(fragments, rawConfig, {
+      withComponent: getConfigValue(rawConfig.withComponent, true),
+      withHooks: getConfigValue(rawConfig.withHooks, false),
+      urqlImportFrom: getConfigValue(rawConfig.urqlImportFrom, null),
+    } as any);
```
dotansimha: I think you should be able to remove `as any` now.

mxstbr:
Blocked by https://github.com/FormidableLabs/urql/pull/235, otherwise this is almost ready pending some cleanup

packages/plugins/typescript/urql/src/visitor.ts
mxstbr: Done!

dotansimha:
@mxstbr that's looks great. 
Note that the base client-side also has configuration for `noGraphQlTag` that precompile the GraphQL operations into a `DocumentNode` during generation, so I think we need to make sure it's compatible with `urql`.

mxstbr:
Note that components support currently fails with this error:

```
TypeScript error: Type '{ variables: QueryVariablesName; requestPolicy?: "cache-first" | "cache-only" | "network-only" | "cache-and-network" | undefined; query: any; }' is missing the following properties from type 'QueryHandlerProps': client, children, key  TS2739
```

I am not sure how to resolve that. It's caused by `Urql.Query` expecting more props than we provide to the custom components.

mxstbr:
Fixed component generation in the latest commit, pending the publication of a new version of urql with this patch applied https://github.com/FormidableLabs/urql/pull/236 this works great! :100:

@dotansimha this should be ready to ship?!?!

dotansimha:
Thanks @mxstbr ! Looks great. 
(btw, you can test and use it now, before the next stable release, because we are publishing an alpha version per each commit)

Issue #1806, opened 2019-05-02
Security Issue NODE_TLS_REJECT_UNAUTHORIZED is set to 0


TSiege:
Found a major security concern. [This package sets my project's `NODE_TLS_REJECT_UNAUTHORIZED` to 0 for the entire process](https://github.com/dotansimha/graphql-code-generator/blob/de7ad41f4f93dbf47e4da4b4919ac35acbd791a6/packages/graphql-codegen-cli/src/generate-and-save.ts#L9) thereby disabling SSL. I found this when searching after seeing this security warning from node.
```
Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
```

I've traced it back to this commit... https://github.com/dotansimha/graphql-code-generator/pull/147/commits/8b0bc1ea8408badb845696e35df85c6a84a4563a

simPod:
Can it be backported to v0.18 pls?

dotansimha:
@simPod on it now, sorry for the delay.

dotansimha:
@simPod 0.18.2 is available now.

simPod:
@dotansimha thanks, contacted NPM about fix for https://www.npmjs.com/advisories/834 (there's no release with fix for v1 AFAIK)

dotansimha:
Fixed in `1.2.0`.

simPod:
@dotansimha Hi, NPM cannot find the release https://www.npmjs.com/package/graphql-code-generator

![image](https://user-images.githubusercontent.com/327717/58027634-5f9c7a00-7b19-11e9-9931-bb2d3ae75007.png)

dotansimha:
@simPod we moved to organization libraries after 1.0, `graphql-code-generator` is now `@graphql-codegen/cli` and plugins are also moved (for example `graphql-codegen-typescript` is now `@graphql-codegen/typescript`.

simPod:
@dotansimha right, I forgot that! thanks

PR #1820, opened 2019-05-06
Remove use of NODE_TLS_REJECT_UNAUTHORIZED env var.

michaelnyu:
This PR addresses issue: https://github.com/dotansimha/graphql-code-generator/issues/1806.

Following the recommendations from, this PR removes the use of the `NODE_TLS_REJECT_UNAUTHORIZED` environment variable: https://nodejs.org/api/cli.html#cli_node_tls_reject_unauthorized_value

> If value equals '0', certificate validation is disabled for TLS connections. This makes TLS, and HTTPS by extension, insecure. *The use of this environment variable is strongly discouraged.*

### DIFF
--- a/packages/graphql-codegen-cli/src/generate-and-save.ts
+++ b/packages/graphql-codegen-cli/src/generate-and-save.ts
@@ -6,8 +6,6 @@ import { sync as mkdirpSync } from 'mkdirp';
 import { dirname } from 'path';
 import { debugLog } from './utils/debugging';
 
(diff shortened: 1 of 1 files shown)


dotansimha:
That's awesome! Thanks @michaelnyu !

silentsakky:
Can we get a release containing this fix?

dotansimha:
@silentsakky a new version will be available later today :)

Issue #1828, opened 2019-05-07
Webpack loader for .graphql files


mxstbr:
**Is your feature request related to a problem? Please describe.**

Generating files that are checked into source control is suboptimal when working on a team where these files constantly change—especially when there's only a single file (ref #1292).

**Describe the solution you'd like**

It would be awesome if there was a webpack loader that transformed `.graphql` files into JavaScript modules that export client code. For example:

```GraphQL
# getViewer.graphql
query getViewer {
  viewer {
    id
  }
}
```

Once the webpack loader is added to the configuration, you could import from that `.graphql` file the client code, for example when using the urql or react-apollo plugins:

```JS
import { useGetViewer, GetViewerComponent } from '../getViewer.graphql';
// Now you can use useGetViewer and GetViewerComponent!
```

Note how there's no intermediary files generated that are checked into source control, everything happens inside of webpack and is opaque to the user.

This would also require a change to the generated TypeScript declarations to make sure TypeScript understands what is exported from all those `.graphql` files.

dotansimha:
I think by combining some of the types generated by `typescript-operations` and the TypeScript modules generated by `typescript-graphql-files-modules` could work.
We just need to note that some parts of the `typescript-operations` are not declarative (for example, `enum` should be generated as `type` and not as `enum`, and there is a flag for that), because `d.ts` files can only contain TypeScript declarations and not code.

Then, we can create a very simple Webpack plugin that will build the HOC/Hooks/Components in runtime and export it (and then just match it to the declared one in the `d.ts` file).

slikts:
Having a webpack loader for the runtime parts is a nice idea that's marred slightly by the fact that the types would still need to be generated to the filesystem for typechecking to work, but there's no real way around that. A webpack loader would also lend itself naturally to solving issues like #1570 (incremental builds) and #1797 (selective building).

slikts:
Here's [an example](https://github.com/olegstepura/typed-css-modules-loader/blob/master/index.js) of a webpack loader that emits typings for CSS modules.

mxstbr:
Funnily enough we already use that exact webpack loader in our project. If the `@graphql-codegen/webpack-loader` worked that exact way I would be _very_ happy!

CarloPalinckx:
[...]

PR #1835, opened 2019-05-08
Add first boilerplate setup for webpack loader

mxstbr:
Reference #1828, this solely does the first setup for the package and nothing else.

@dotansimha how are we going to run the raw file contents through the graphql-codegen plugins?

### DIFF
--- a/packages/graphql-codegen-webpack-loader/.gitignore
+++ b/packages/graphql-codegen-webpack-loader/.gitignore
@@ -1,0 +1,6 @@
+node_modules
+npm-debug.log
+dist
+temp
+yarn-error.log
+tests/coverage/
--- a/packages/graphql-codegen-webpack-loader/.npmignore
+++ b/packages/graphql-codegen-webpack-loader/.npmignore
@@ -1,0 +1,6 @@
+src
+node_modules
+tests
+!dist
+jest.config.js
+tsconfig.json
--- a/packages/graphql-codegen-webpack-loader/README.md
+++ b/packages/graphql-codegen-webpack-loader/README.md
@@ -1,0 +1,9 @@
+# GraphQL Code Generator
+
+<p align="center">
+    <img src="https://github.com/dotansimha/graphql-code-generator/blob/master/logo.png?raw=true" />
+</p>
+
+Live demo and full documentation: [graphql-code-generator.com](https://graphql-code-generator.com)
+
+Project repository: [graphql-code-generator](https://github.com/dotansimha/graphql-code-generator)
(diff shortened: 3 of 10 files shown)


dotansimha:
That's awesome! Thank you @mxstbr ! 

A few thoughts:
- I think maybe we can create a separate package for that, that compiles `.graphql` files into a ready-to-use HOC/Component/Hooks, and allow it to be pure JavaScript.
- Then, we can create a codegen plugin that supports this package by generating typings for it (using `declare module`). This way the package could be used as pure JS without the codegen, and the codegen can support it with typings (we can do it for both Flow an TS).

mxstbr:
That sounds like a great plan to me! I'm not well versed enough in the codebase to tackle that, so I'll leave it up to you :pray:

eddeee888:
Hello! Just wondering what's the state of this? Would love to help to push this forward

> Then, we can create a codegen plugin that supports this package by generating typings for it (using declare module). This way the package could be used as pure JS without the codegen, and the codegen can support it with typings

Is this similar to `typescript-react-apollo` plugin? If I understand this correctly, we create declaration files `d.ts` instead of `.tsx` files?

mxstbr:
I have not enough expertise in the graphql-codegen codebase to pick this up further, I would love for somebody else to pick this up and implement the functionality!

dotansimha:
So now we are able to load document from the original source file, thanks to @eddeee888 . I think it's one step forward in order to get easy integration with Webpack.

CarloPalinckx:
If we could skip outputting implementation and limit the loader to just output declaration files matching the .graphql modules we could solve a lot of confusion that has led to the "near-operation-file" preset. Because all a user would be doing is import the .graphql file 🤩:

In a `globals.d.ts` file:

```
declare module '*getProducts.graphql' {
// generated types
}
```

in a `Products` component: 
```
import useGetProductsQuery from 'getProducts.graphql';

...
const { data, loading, error } = useGetProductsQuery();
...
```

A tradeoff to this aproach is that we would probably needs some in memory caching to keep it quick.

eddeee888:
Hello! I have made a small repo to see if this is doable ( also to play with webpack loader since I have never done it before! )

https://github.com/eddeee888/experiment-apollo-hooks-loader

And here's [the loader](https://github.com/eddeee888/experiment-apollo-hooks-loader/blob/master/loaders/src/apollo-hooks-loader.ts)

If this is in the right direction, I can try to bring it over and iterate on it?

dotansimha:
WOW @eddeee888 ! 
This is awesome!!! I'll try it locally later this week.

GiancarlosIO:
:eyes:

dotansimha:
Let's continue the discussion here: https://github.com/dotansimha/graphql-code-generator/issues/3494 
Thank you :)

Issue #1632, opened 2019-04-02
Types for subscription operations seem to have an extra wrapper


jonaskello:
**Describe the bug**

Types for subscription operations seem to have an extra wrapper. Maybe this is an effect of  #1453 where the wrapper seems to have been removed for the resolver, but it is not removed for the operation types?

**To Reproduce**

1. My GraphQL schema:

```graphql
type Subscription {
  postAdded: Post
}

type Query {
  posts: [Post]
}

type Post {
  author: String
  comment: String
}
```

2. My GraphQL operations:

```graphql
subscription MyPostAdded {
  postAdded {
    author
  }
}
```

3. My `codegen.yml` config file:

```yml
schema:
  - "schema.graphql"
overwrite: true
generates:
  generated.ts:
    documents: "query.graphql"
    plugins:
      - "typescript"
      - "typescript-operations"
```

**Expected behavior**

This is generated for the resolvers (which I think is correct according to #1453), there is no wrapper for `postAdded`:

```ts
export type SubscriptionResolvers<Context = any, ParentType = Subscription> = {
  postAdded?: SubscriptionResolver<Maybe<Post>, ParentType, Context>;
};
```

This is generated for the operations, here we still have a wrapper for `postAdded`:

```ts
export type MyPostAddedSubscription = { __typename?: "Subscription" } & {
  postAdded: Maybe<{ __typename?: "Post" } & Pick<Post, "author">>;
};
````

I would expect it to instead generate this for the operation (no wrapper, same as the resolver):

```ts
export type MyPostAddedSubscription =
  postAdded: Maybe<{ __typename?: "Post" } & Pick<Post, "author">>;
````

**Environment:**

- OS: MacOS
- `@graphql-codegen/...`:  1.0.6
- NodeJS: 10

**Additional context**

I think the confusion is that a subscription can only have a single root-field by the specification. This means no wrapper is needed?

**Comment by jonaskello** [comment]
Actually I don't think this error report is accurate :-).

I am getting some strange behaviour though and I'm not sure the typings are 100%  correct.

I have somthing like this resolver:

```ts
export const subscriptionResolver: SubscriptionResolvers<Context, RootValue> = {
  postAdded: {
    subscribe: (parent, args, context) => {
      return pubsub.postAddedAsyncIterator();
    },
    resolve: (payload, args, context) => {
      return payload as any;
    },
```

[...]

PR #1842, opened 2019-05-10
Fix SubscriptionSubscribeFn interface

const86:
The subscribe callback of subscription 'foo' should normally
return an object like { foo: value } where value is of type
of the subscription. Add that wrapping layer to the interface.


const86:
This should address bug #1632.

dotansimha:
Hi @const86 ! Thank you for your contribution! 
Does your PR is a breaking change? because I noticed that you added the new generics in between and not as last.

/cc @kamilkisiela can you please take a look?

const86:
I'm changing existing types, so I guess it's a breaking change. Unless you consider previous code so broken that nobody could used it anyway.

dotansimha:
@const86 my point is, that you changed it from `export interface ISubscriptionResolverObject<Result, Parent, Context, Args>` to `export interface ISubscriptionResolverObject<Result, Key: string, Parent, Context, Args>` , which is a breaking change for the API, because if someone uses it as `ISubscriptionResolverObject<A, B, C>` it will be invalid now. 
The non-breaking way of doing that is `export interface ISubscriptionResolverObject<Result, Parent, Context, Args, Key: string>`. But it all the usages of those interfaces are internal, then it's fine (but I'm not so sure about that)

const86:
I see. Tell me if I should move new argument to the end of the list to make things easier.
I believe, it will be a breaking change in that interface anyway, unless that new last argument is optional with  default resulting in the old behavior. Overall behavior is incompatible with current version. If somebody has subscription resolvers matching generated interface, they will see errors with new version.

dotansimha:
I see, thanks @const86 ! 
@ardatan @kamilkisiela can you please review this?

DanielMSchmidt:
Awesome work @const86, I was about to dig into how to do a PR like this 👍

const86:
@kamilkisiela, I've update the PR. Now subscription resolver is a union of two interfaces. One is like the old one, with even relaxed result type of `subscribe`. That doesn't have to be `TResult`, only the `resolve` has to.
Another is like my first option: `subscribe` yields `{ [name]: TResult }` and `resolve` is optional. If you still want to provide `resolve`, it must again return a `TResult`.
I guess, now we can even call it backward compatible.

const86:
@kamilkisiela, @dotansimha, anything else I should do?

dotansimha:
@kamilkisiela what do you think? 

If someone wants to test it, I pushed the changes to the source repo so it will be available as alpha release: `1.2.1-alpha-0a69c2be.18`

Yuyz0112:
@dotansimha I have tested the `1.2.1-alpha-0a69c2be.18` release and it works in my situation.

Is there any chance we can merge this into the master branch?

dotansimha:
@kamilkisiela can you please take another look at this issue?

Issue #1834, opened 2019-05-08
Make suffix of generated React components customizable


jlkiri:
**Is your feature request related to a problem? Please describe.**

Since components generated by `typescript-react-apollo` all have `Component` suffix in their name, it is not easy to distinguish between components that do the query work and those that don't in my code. It would be nice to still be able to distinguish the query components from others by using the `Query` suffix, e.g. not `MyItemsComponent` but `MyItemsQuery`. However, there is no way to specify this in the plugin config.

**Describe the solution you'd like**

Make it possible to specify `customSuffix` in plugin config inside `codegen.yml`. E.g. `customSuffix: 'Query'`.

**Describe alternatives you've considered**

**Additional context**

dotansimha:
Available in `1.2.0`.

PR #1843, opened 2019-05-10
Make suffix of generated React Apollo components customizable

jlkiri:
This PR addresses the following issue: #1834 

The default suffix is `Component`. It should be possible to specify the desired suffix (string) via `customSuffix` property in plugin config in `codegen.yml`.

I only made changes to the `typescript-react-apollo` plugin.


dotansimha:
Hi @jlkiri ! Thank you for your contribution! 
I think `componentSuffix` could be a better name for this, what do you think?

jlkiri:
@dotansimha This indeed sounds like a better name. I'll change it.

dotansimha:
Thanks @jlkiri !

PR #1875, opened 2019-05-15
Allow specifying customModulePathPrefix with typescript-graphql-files-modules plugin

Charliekenney23:
This change would allow specifying a `customModulePathPrefix` when using the `typescript-graphql-files-modules` plugin to generate typings for GraphQL files.

Consider the following project structure:

```
/src
 /api
   /user-service
     queries.graphql
     mutations.graphql
     schema.graphql
     fragments.graphql
   /message-service
     queries.graphql
     mutations.graphql
     schema.graphql
     fragments.graphql
  index.ts
codegen.yml
```

Currently all files generated using the plugin have the following format. The generic module definition glob makes type discovery with intellisense difficult.

```typescript
module "*/FILE_NAME.graphql" { ... }
```

With the changes in place, using the following config would allow for a more specific module definition path.

```yaml
# ...
generates: ./src/api/user-service/queries.d.ts
  schema: ./src/api/user-service/schema.graphql
  documents: ./src/api/user-service/*.graphql
  plugins:
    - typescript-graphql-files-modules
  config:
    customModulePathPrefix: api/user-service/
```

```typescript
module "*/api/user-service/queries.graphql" { ... }
```

### DIFF
--- a/packages/plugins/typescript/graphql-files-modules/src/index.ts
+++ b/packages/plugins/typescript/graphql-files-modules/src/index.ts
@@ -2,7 +2,30 @@ import { basename } from 'path';
 import { Types, PluginFunction, PluginValidateFn } from '@graphql-codegen/plugin-helpers';
 import { GraphQLSchema, OperationDefinitionNode } from 'graphql';
 
-export const plugin: PluginFunction = async (schema: GraphQLSchema, documents: Types.DocumentFile[]): Promise<string> => {
+export interface TypeScriptFilesModulesPluginConfig {
+  /**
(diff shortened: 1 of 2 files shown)


dotansimha:
Hi @Charliekenney23 ! Thank you for this contribution! 
I think we can use `modulePathPrefix`, which is a better naming, what do you think?

Charliekenney23:
I suppose the word 'custom' seems a bit redundant in a config block, `modulePathPrefix` seems fine.

dotansimha:
Perfect! Than you @Charliekenney23 !

PR #1879, opened 2019-05-16
Add option for non-optional typename in operations

piotr-m-jurek:


piotr-m-jurek:
The flow needs corrections, will work on that :)

piotr-m-jurek:
The flow needs corrections, but please check if this even makes sense

packages/plugins/typescript/operations/src/index.ts
```diff
@@ -42,6 +42,20 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
...
+   * @description Automatically adds `__typename` field to the generated types, even when they are not specified
+   * in the selection set, and makes it non-optional
+   *
+   * @example
+   * ```yml
+   * config:
+   *   forceTypename: true
+   * ```
+   */
+  forceTypename?: boolean;
```
dotansimha: I think that now the `BaseVisitor` is the one that manages the configuration for `__typename`, this should be defined there (and this config interface inherits from it, so it will be included here as well).

dotansimha: See this: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts#L87-L100

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -7,6 +7,7 @@ import { TypeScriptDocumentsPluginConfig } from './index';
 export interface TypeScriptDocumentsParsedConfig extends ParsedDocumentsConfig {
   avoidOptionals: boolean;
   immutableTypes: boolean;
+  forceTypename: boolean;
```
dotansimha: Can be moved to `BaseVisitor`.

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -16,11 +17,12 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<TypeScriptD
       {
         avoidOptionals: config.avoidOptionals || false,
         immutableTypes: config.immutableTypes || false,
+        forceTypename: config.forceTypename || false,
```
dotansimha: Can be moved to `BaseVisitor`.

dotansimha:
That's perfect! Thank you @piotr-m-jurek !

PR #1893, opened 2019-05-19
Fix relative imports generation in `near-operation-file` preset

### LINKED ISSUE #1892: Wrong import paths with `near-operation-file` preset (on Windows)
In the recent canary version, with new `near-operation-file` preset, import paths in "near" files are malformed. 

My guess is it that has something to do with Windows file separator, but I wasn't able to replicate it consistently as I got two different behaviours. Either the relative path contains backslashes: 
```js
import * as Types from '..\..\..\__generated__/types';`
```
or there aren't any slashes at all:
```js
import * as Types from "......__generated__/types";
```

Two different projects, the same machine. 

**`codegen.yml` config file:**

```yml
overwrite: true
schema: "schema.graphql"
documents: 'src/**/*.{graphql,tsx}'
generates:
  src/__generated__/types.ts:
    plugins:
      - "typescript"
  src/:
    preset: near-operation-file
    presetConfig:
      extension: .generated.ts
      baseTypesPath: __generated__/types.ts
    plugins:
      - typescript-operations
```

**Environment:**

- OS: Windows 10
- `@graphql-codegen/...@1.1.4-alpha-62ee81ed.50`: 
- NodeJS: v10.15.3

vhenzl:
Fixes #1892 (backslashes being used in generated imports on Windows) and also problem with a nested file import where the import would be missing slash after "current directory" dot (`.nested/types` instead of `./nested/types`).

I've used a regex to replace backslashes with slashes. Let me know if you want me to use a library like https://www.npmjs.com/package/normalize-path or https://www.npmjs.com/package/slash instead.

### DIFF
--- a/packages/presets/near-operation-file/src/utils.ts
+++ b/packages/presets/near-operation-file/src/utils.ts
@@ -1,17 +1,17 @@
-import { parse, dirname, relative } from 'path';
+import { parse, dirname, relative, join } from 'path';
 import { DocumentNode, visit, FragmentSpreadNode, FragmentDefinitionNode } from 'graphql';
 import { FragmentNameToFile } from './index';
 
 export function appendExtensionToFilePath(baseFilePath: string, extension: string) {
   const parsedPath = parse(baseFilePath);
 
-  return parsedPath.dir + '/' + parsedPath.name + extension;
+  return join(parsedPath.dir, parsedPath.name + extension).replace(/\\/g, '/');
 }
 
 export function clearExtension(path: string): string {
   const parsedPath = parse(path);
 
-  return parsedPath.dir + '/' + parsedPath.name;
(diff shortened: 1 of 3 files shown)


packages/presets/near-operation-file/src/utils.ts
```diff
@@ -39,12 +39,18 @@ export function extractExternalFragmentsInUse(documentNode: DocumentNode | Fragm
 
 export function fixLocalFile(path: string): string {
   if (!path.startsWith('..') && !path.startsWith('&&')) {
```
vhenzl: When `path` could start with `&&`? IMO this condition doesn't make sense.

dotansimha: Can be removed, added it for tests

dotansimha:
Awesome! Thank you @vhenzl !

Issue #1896, opened 2019-05-20
1.2.0: add plugin prepends all root configs; non-string values cause error


jrr:
**Describe the bug**
1.2.0 does this:

```
$> gql-gen -o
  ✔ Parse configuration
  ❯ Generate outputs
    ❯ Generate modules/client/graphql/types.gen.tsx
      ✔ Load GraphQL schemas
      ✔ Load GraphQL documents
      ✖ Generate
        → a.startsWith is not a function
    ✔ Generate modules/client/graphql/client-types.gen.ts
    ✔ Generate modules/graphql-api/server-types.gen.ts


 Found 1 error

  ✖ modules/client/graphql/types.gen.tsx
    TypeError: a.startsWith is not a function


Something went wrong
```

I suspect it may have broken with this change https://github.com/dotansimha/graphql-code-generator/pull/1869, as it's the only instance of `a.startsWith` I could find in the repo.

I haven't gotten to the bottom of this yet, but here's a mostly-minimized repro config:

```yml
generates:
  modules/client/graphql/types.gen.tsx:
    config:
      strict: true
    schema:
      - "modules/graphql-api/schema.graphql"
      - "modules/client/graphql/schema.graphql"
    documents:
      - "modules/client/graphql/*/*.graphql"
    plugins:
      - add:
          - // hello
      - typescript
```

It looks like perhaps there's some interaction of `strict: true` and the _add_ plugin; remove either one and the error goes away.

jrr:
Ah-ha! The _add_ plugin is picking up my root-level configs.

Given the above codegen.yml, and this code from the add plugin:

https://github.com/dotansimha/graphql-code-generator/blob/9a70bf6dcc8b7278308e1759dc5b58127288ad6a/packages/plugins/other/add/src/index.ts#L6-L13

A little bit of logging:

```
$> gql-gen -o
plugin { add: [ '// hello' ] }
config: { '0': '// hello', strict: true }
asArray: [ '// hello', true ]
prepending // hello
prepending true
plugin { typescript: {} }
prepending export type Maybe<T> = T | null;
a is string |// hello|
a is boolean |true|
  ✔ Parse configuration
  ❯ Generate outputs
    ❯ Generate modules/client/graphql/types.gen.tsx
      ✔ Load GraphQL schemas
      ✔ Load GraphQL documents
      ✖ Generate
        → a.startsWith is not a function


 Found 1 error

  ✖ modules/client/graphql/types.gen.tsx
[...]

PR #1921, opened 2019-05-27
Update add plugin with output section selection

AnotherHermit:
Add functionality to add plugin so that it is possible to select which output section the text should end up in.

This supports such use cases such as wrapping the output in a namespace, etc.

### DIFF
--- a/docs/plugins/add.md
+++ b/docs/plugins/add.md
@@ -10,10 +10,19 @@ You can use this plugin to add custom code, imports, comments and more to your o
 ## Installation
 
     $ yarn add -D @graphql-codegen/add
-    
-    
+
 > Notice that the order of plugins in the yaml config matters.
 > If you need to add something to the top of the file, for example `/* eslint-disable */`, makes sure to place the `add` plugin before the other plugins, so their output won't be on top of the `add` addition.
+
+## Input parameters
+
+To control in which section of output the text will be added, the input can be specified
+as an object. The object takes two parameters:
+
+- `placement`: Optional specification of output section of added text
+  - Input: `'prepend' | 'content' | 'append'`
+  - Default: `'prepend'`
+- `content`: Content to add to file, can be text or an array/list of strings
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


dotansimha:
That's perfect, @AnotherHermit ! And you also fixed https://github.com/dotansimha/graphql-code-generator/issues/1896 ;)

benquarmby:
Just upgraded to 1.3.0 of this plugin. The configuration shape changed without backwards compatibility. This was technically a breaking change as well as a new feature. Ideally it would have bumped to 2.0.0.

dotansimha:
@benquarmby can you please explain why it's breaking? You can still specify a string, and it will work as before.

benquarmby:
Old configuration:
```
overwrite: true
schema:
  - "../../graphql/graphql-*/index.gql"
documents: null
generates:
  ./src/generated.ts:
    plugins:
      - add:
          - "/* eslint-disable */"
```
New configuration:
```
overwrite: true
schema:
  - "../../graphql/graphql-*/index.gql"
documents: null
generates:
  ./src/generated.ts:
    plugins:
      - add:
          content:
            - "/* eslint-disable */"
```
Trying to use the old config on 1.3.0 throws an error and requires code changes to fix.

AnotherHermit:
Which version did you use before? And what does the error say?

benquarmby:
```
-        "@graphql-codegen/add": "1.1.0",
-        "@graphql-codegen/cli": "1.1.0",
-        "@graphql-codegen/typescript": "1.1.0",
+        "@graphql-codegen/add": "1.3.0",
+        "@graphql-codegen/cli": "1.3.0",
+        "@graphql-codegen/typescript": "1.3.0",
```

When using the old config with 1.3.0 we see an `Add plugin, missing content` message
image

AnotherHermit:
If I am not mistaken I think there was kind of a bug (#1896) introduced in 1.2.0 that made it difficult to support the kind of input you had in your config. The problem being that if you have any top level config (not related to the add plugin) in that file the parser would throw an error. I tried to make the fix as backwards compatible as possible, but did not handle all cases since the migration path is pretty simple.

benquarmby:
Got it. Yeah, seems to be related. We've moved forward by being more explicit with the config (adding `content` property).

PR #1923, opened 2019-05-28
Interface Option

### LINKED ISSUE #1852: TypeScript: Option to generate interfaces
While it's true that TypeScript type literals are *almost* identical to interfaces (as mentioned in #1810), one of the key exceptions is that interfaces are [open to declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html), where types are not.

This is very important for generated code, where declaration merging is the only avenue to apply manual tweaks to type information.

Please add an option to generate interfaces rather than types.

benquarmby:
~Currently work in progress. Happy to get feedback on the approach.~

Ready for review.

Related #1852

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -160,8 +160,6 @@ export class BaseTypesVisitor<TRawConfig extends RawTypesConfig = RawTypesConfig
   }
 
   InterfaceTypeDefinition(node: InterfaceTypeDefinitionNode, key: number | string, parent: any): string {
-    const optionalTypename = this.config.nonOptionalTypename ? '__typename' : '__typename?';
-    const allFields = [...(this.config.addTypename ? [indent(`${optionalTypename}: '${node.name}',`)] : []), ...node.fields];
     const argumentsBlock = this.buildArgumentsBlock(parent[key] as InterfaceTypeDefinitionNode);
 
     let declarationBlock = new DeclarationBlock(this._declarationBlockConfig)
@@ -170,7 +168,7 @@ export class BaseTypesVisitor<TRawConfig extends RawTypesConfig = RawTypesConfig
       .withName(this.convertName(node))
       .withComment((node.description as any) as string);
 
-    const interfaceDefinition = declarationBlock.withBlock(allFields.join('\n')).string;
+    const interfaceDefinition = declarationBlock.withBlock(node.fields.join('\n')).string;
 
     return [interfaceDefinition, argumentsBlock].filter(f => f).join('\n\n');
   }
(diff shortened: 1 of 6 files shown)


packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -160,17 +198,15 @@ export class BaseTypesVisitor<TRawConfig extends RawTypesConfig = RawTypesConfig
   }
 
   InterfaceTypeDefinition(node: InterfaceTypeDefinitionNode, key: number | string, parent: any): string {
-    const optionalTypename = this.config.nonOptionalTypename ? '__typename' : '__typename?';
```
benquarmby: GraphQL interfaces do not have a `__typename`, only concrete types do. Other than being off-spec, it caused problems for TypeScript interfaces.

I realize this is a breaking change and the tests will need to be updated.

packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -143,11 +166,26 @@ export class BaseTypesVisitor<TRawConfig extends RawTypesConfig = RawTypesConfig
...
+
+    const buildInterfaces = () => {
+      if (!originalNode.interfaces || !node.interfaces.length) {
+        return '';
+      }
+
+      const interfaces = originalNode.interfaces.map(i => this.convertName(i));
+
+      if (type === 'interface') {
+        return ' extends ' + interfaces.join(', ') + (allFields.length ? ' ' : ' {}');
```
benquarmby: Required a stand alone function to differentiate between interface extension and type unions.

packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -81,7 +104,7 @@ export class BaseTypesVisitor<TRawConfig extends RawTypesConfig = RawTypesConfig
 
     return new DeclarationBlock(this._declarationBlockConfig)
       .export()
-      .asKind('type')
+      .asKind(this._parsedConfig.declarationKind.scalar)
```
benquarmby: This is the gist of the PR. Use config for the output type rather than a static kind.

packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
dotansimha: You are right, thank you for pointing that out.

dotansimha:
Thanks @benquarmby ! That's awesome! 
Let me know how it goes, and if you need my help with that :)

benquarmby:
Still to do:
* Cover interface behavior with tests
* Add documentation for new configuration

benquarmby:
@dotansimha: I believe this is ready for review. Please let me know what you think.

benquarmby:
@dotansimha: Bump. Please take a look and let me know if this PR needs any extra work to get it merged. My team has just come up with a new use case for interface declaration merging, so it would be great to keep this moving.

dotansimha:
Hi @benquarmby , sorry for the delay (I was in a vacation 😎).
This PR looks great, and I think we can merge it :) 
Thank you!

benquarmby:
Fantastic news, thanks @dotansimha. Look forward to the next release!

What is the typical turn around on publishing new versions?

dotansimha:
@benquarmby We usually try to collect a few features and bug fixes, test it on our projects, and then release (basically, very soon).

Issue #1897, opened 2019-05-20
[TS React Apollo] Variables should be optional for Query components


cecchi:
**Describe the bug**

The generated types for React Apollo `<Query>` components explicitly override the "variables" prop and makes it required:

```ts
export type NavHierarchyComponentProps = Omit<
  Omit<ReactApollo.QueryProps<NavHierarchyQuery, NavHierarchyQueryVariables>, 'query'>,
  'variables'
> & { variables: NavHierarchyQueryVariables }
```

It looks as though this was done intentionally in https://github.com/dotansimha/graphql-code-generator/pull/1687, and then undone for mutations in https://github.com/dotansimha/graphql-code-generator/issues/1741, so I'm probably missing some context as to why that was desirable. 

Our use-case is that we use the `skip` prop to conditionally execute the query only once the variables are available. Something like this:

```jsx
<NavHierarchyComponent 
  variables={someContext ? { foo: someContext.bar } : undefined} 
  skip={someContext === undefined} 
/>
```

This fails with the current typings since it expects `variables` to be explicitly non-optional.

**Expected behavior**

Either use the built-in React Apollo types for `variables`, or add some kind of clever typing that allows the `variables` to be optional if the `skip` prop is provided. I believe simply using the typings from `react-apollo` would solve my use-case, but I'm sure there is a reason it was overridden in the first place.

**Edit:** It looks as though this was done since the `react-apollo` typings were wrapping the variables in `Partial`, meaning the types were not as restrictive as they could be (https://github.com/dotansimha/graphql-code-generator/issues/1626). 

In my opinion it's best to lean on the Apollo typings as much as possible, so I would suggest removing the override (`Omit` + `& { variables ... }`) altogether. If they wrap the types in Partial (incorrectly?), that should be dealt with on their end. Anyway, as far as I can tell, this is no longer the case. These are the typings from the latest version of `react-apollo`:

```tsx
export interface QueryOpts<TGraphQLVariables = OperationVariables> {
    variables?: TGraphQLVariables;
}
```

...Which looks correct to me. Variables can be omitted altogether, but if they are present they must exactly match the typings.

ardatan:
If at least one of the variables is required, `variables` property would be required; because this means that query is unable to work without variables. So, codegen generates the component of that query with a required `variables` property.

cecchi:
[...]

PR #1926, opened 2019-05-28
Make variables optional if skip is provided

ardatan:
Related to https://github.com/dotansimha/graphql-code-generator/issues/1897

### DIFF
--- a/dev-test/githunt/types.reactApollo.hooks.tsx
+++ b/dev-test/githunt/types.reactApollo.hooks.tsx
@@ -295,22 +295,22 @@ export const OnCommentAddedDocument = gql`
   }
 }
     `;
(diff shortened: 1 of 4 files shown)


cecchi:
Thanks! One thing: I think this disallows `skip` being `undefined` with `variables` being provided? We should treat `undefined` and `null` as equivalent to `false`.

ardatan:
Good catch! :) Fixed

benjie:
This also disables settings `skip` true when variables are supplied; could we instead have `skip?: boolean` when variables are supplied?

boom:
This change currently breaks passing a variable to the `skip` property. See @benjie comment that 
```typescript
& ({ variables: FeedQueryVariables; skip?: false; } | { skip: true; });
``` 

should not reference boolean values but rather refer to the type `boolean`.

boom:
Opened https://github.com/dotansimha/graphql-code-generator/pull/2251 to resolve.

PR #1958, opened 2019-06-04
Update documentation for `skipTypename` and `nonOptionalTypename`

dfreeman:
This adds documentation for `nonOptionalTypename` (introduced in https://github.com/dotansimha/graphql-code-generator/pull/1879) and clarifies the behavior of `skipTypename` a bit.

I _think_ I put this in the right markdown document, given that `skipTypename` is also here, but happy to relocate if there's a better place 🙂

### DIFF
--- a/docs/generated-config/base-documents-visitor.md
+++ b/docs/generated-config/base-documents-visitor.md
@@ -1,10 +1,21 @@
 ### skipTypename (`boolean`, default value: `false`)
 
-Automatically adds `__typename` field to the generated types, even when they are not specified in the selection set.
+When `false`, automatically adds an optional `__typename` field to the generated types, even when not specified in the selection set.
 
 #### Usage Example
 
 ```yml
 config:
   skipTypename: true
+```
+
+### nonOptionalTypename (`boolean`, default value: `false`)
+
+When `true`, makes the `__typename` field non-optional on generated types, even when not specified in the selection set.
+
+#### Usage Example
+
+```yml
+config:
+  nonOptionalTypename: true
 ```


dotansimha:
Hi @dfreeman ! 

Thank you for this PR! We have a documentation mechanism that does that automatically from the code.
The specific relevant file is here: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/typescript/src/index.ts#L9 

So basically you need to update it there, in a comment, above the property, and then it will be automatically generated during the docs website deployment.

dfreeman:
Thanks for the pointer, @dotansimha! I think I'm a bit confused, because the source of the [docs for `skipTypename`](https://graphql-code-generator.com/docs/plugins/typescript-operations#skiptypename-boolean-default-value-false) appears to be [this doc comment](https://github.com/dotansimha/graphql-code-generator/blob/195484651f12e46bc9266532b2b6170beb1afcaf/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts#L88-L100), but `nonOptionalTypename` appears [right below it](https://github.com/dotansimha/graphql-code-generator/blob/195484651f12e46bc9266532b2b6170beb1afcaf/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts#L102-L115) in that same file and is already documented, but is nowhere to be found on the documentation website.

Is it possible the docs haven't been regenerated or republished since that was added?

dotansimha:
@dfreeman yeah we had to rollback because of an error with the website app, so probably that's what causes it (related: https://github.com/dotansimha/graphql-code-generator/pull/1964). You can update it there if you think something is missing. Thank you!

dfreeman:
Alright, I'll go ahead and close this out for now then. Thanks!

Issue #1863, opened 2019-05-14
Support @apollo-react/hooks


slikts:
The official hooks support for React Apollo [has been released](https://twitter.com/peggyrayzis/status/1128280687470100480?s=20) as a beta version, and it'd be nice to have support for it in gql-gen. The API is close to react-apollo-hooks, which is already supported, so it shouldn't take major changes (probably fewer changes than in #1814, although I've only looked at it in passing).

ardatan:
It's supposed to work with `hooksImportFrom: @apollo-react/hooks` configuration. We can keep this issue until it is confirmed.

SimenB:
The types are wrong (typed as `QueryResult<any, any>`: https://unpkg.com/@apollo/react-hooks@0.1.0-beta.2/lib/useQuery.d.ts). Seems like the shared code sandbox (https://codesandbox.io/s/github/peggyrayzis/redux-to-graphql/tree/master/client) uses `const res: QueryResult<ApolloTypes.GetSubreddit, ApolloTypes.GetSubredditVariables> = useQuery()` instead. Not sure if that's on purpose or not, but it means that the generated code is not correct

ardatan:
I think it is better to wait the stable release. @dotansimha What do you think?

SimenB:
Might be nice to release an alpha or something supporting it (it's just moving the generics a bit and adding an import), just so people can test it? Completely understand if you don't wanna put work into it though, it seems silly (probably an oversight) to not allow passing in the generics at the callsite

dotansimha:
I think it worth to add the option to use it, maybe by just doing: `hooksImportFrom: @apollo-react/hooks` we can change the output to match their signature.

SimenB:
I asked: https://github.com/apollographql/react-apollo/pull/2892#issuecomment-492949070

ardatan:
@SimenB As I see in the current state of that PR in apollo,
```tsx
export function useVoteMutation(baseOptions?: ReactApolloHooks.MutationHookOptions<VoteMutation, VoteMutationVariables>) {
  return ReactApolloHooks.useMutation<VoteMutation, VoteMutationVariables>(VoteDocument, baseOptions);
};
```
This generated code fits their signature since we don't specify any return type etc. We only pass generics and `Options` interface as a parameter that also fits Apollo's `Options` interface.

SimenB:
The types are closer now. However, an actual difference is in `QueryHookOptions`.

[...]

PR #1962, opened 2019-06-05
Fix QueryHookOptions for @apollo/react-hooks

edorivai:
Implements the solution as suggested by @dotansimha in https://github.com/dotansimha/graphql-code-generator/issues/1863#issuecomment-496586940

It checks whether `hooksImportFrom` is set to `@apollo/react-hooks`, if so, it outputs the signature that `@apollo/react-hooks` expects, otherwise it defaults to the signature that is compatible with `react-apollo-hooks`.

One quick thing to mention, I tested this change rather quickly by manually editing the visitor in my node_modules, let me know if somethings not right!


dotansimha:
Perfect! Thank you @edorivai !

Issue #1959, opened 2019-06-04
Homepage 'try it out' section is broken


moodysalem:
**Describe the bug**
The home page try it out section is broken for me

**To Reproduce**

1. Visit https://graphql-code-generator.com/
2. Click try it out, which navigates to the next section
3. Nothing there to try out
4. Errors in the console

```
11.a655f578.chunk.js:1 Invariant Violation: Minified React error #321; visit https://reactjs.org/docs/error-decoder.html?invariant=321 for the full message or use the non-minified dev environment for full errors and additional helpful warnings. 
    at https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:397878
    at g (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:397981)
    at U (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:401138)
    at Object.useContext (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:402064)
    at a (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:579192)
    at https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:15432
    at https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:18267
    at qi (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:137334)
    at ko (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:143321)
    at Io (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:151904)
pa @ 11.a655f578.chunk.js:1
11.a655f578.chunk.js:1 Uncaught Invariant Violation: Minified React error #321; visit https://reactjs.org/docs/error-decoder.html?invariant=321 for the full message or use the non-minified dev environment for full errors and additional helpful warnings. 
    at https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:397878
    at g (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:397981)
    at U (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:401138)
    at Object.useContext (https://graphql-code-generator.com/live-demo/static/js/11.a655f578.chunk.js:1:402064)
[...]

PR #1964, opened 2019-06-05
Update the live demo to include the missing stencil apollo dependency

moodysalem:
I think this should fix https://github.com/dotansimha/graphql-code-generator/issues/1959

Just based on the fact that when I tried to run this package locally, it wouldn't run without this dependency


dotansimha:
Awesome! Thank you @moodysalem ! I'll merge it and try to deploy the website

dotansimha:
@moodysalem it seems like it's still failing, see this temporary deploy: https://5cf8ba9a09fd2e00091e2058--graphql-code-generator.netlify.com/ 

Any idea why? :(

Issue #1934, opened 2019-05-29
Wrong Typescript type for Query with a variable with default value


jflewis:
**Describe the bug**
 
When generating Typescript types for a query with default variables, the generated query variables interface specifies the field as `Maybe<Scalars["Boolean"]>`. However, I believe the correct implementation should be `field?: Scalars["Boolean']` so the field isn't required when defining the variables in a query. Having the field be mandatory for a optional field doesn't seem quite right.

**To Reproduce**

1. My GraphQL schema:

```graphql
# Put your schema here
type Query {
    kit(id: ID!): Kit!
  }
```

2. My GraphQL operations:

```graphql
# Put your operations here
query kit($withRawData: Boolean = false, $id: ID!, $languageId: ID!) {
    kit(id: $id) {
      id
      kitId
      nickname
      status
      trackingNumber
      deliveryService
      personalizedReports {
        id
        reportType
      }
      data(languageId: $languageId)
      rawData @include(if: $withRawData) {
        id
        rawDataType
        url
      }
    }
  }
```

3. My `codegen.yml` config file:

```yml
# Put your YML here
pluckConfig:
  modules:
    - name: "apollo-boost"
      identifier: "gql"
schema: http://localhost:3001
documents: graphql/**/*.{ts,tsx}
generates:
  generated/apolloComponents.tsx:
    plugins:
      - typescript
      - typescript-operations
      - typescript-react-apollo
```

**Expected behavior**

**Environment:**

- OS: mac os
- `@graphql-codegen/...`:  "^1.0.6"
- NodeJS: 8.10

**Additional context**

example.) compiler fails with missing property of `withRawData`
```
<KitComponent
              variables={{
                id: kit.id,
                languageId: user.language.id || DEFAULT_LANGUAGE_ID
              }}
```

Aleksion:
Seeing the same thing

n1ru4l:
@Aleksion @jflewis I tackled this issue in https://github.com/dotansimha/graphql-code-generator/pull/2018 

You could help me out by verifying it is behaving correct in a real-world application.

dotansimha:
Fixed by @n1ru4l ❤️ 
If someone wants to test it before we release a stable version, you can use `1.2.2-alpha-d6b07f60.28`.

dotansimha:
Fixed in 1.3.0 🎉

PR #2018, opened 2019-06-18
fix(typescript-operations): wrong type for variable with default value ( #1934)

n1ru4l:
Related #1934

`typescript-operations` generates false types for operation variables width default values.

Example:

```graphql
query foo($var: String = "test") {
  foo(var: $var)
}
```

**Generated Types:**

```typescript
export type FooQueryVariables = {
  var: Maybe<Scalars['String']>
};
```

**Expected Types:**

```diff
export type FooQueryVariables = {
-  var: Maybe<Scalars['String']>
+  var?: Maybe<Scalars['String']>
};
```

More details can be found in https://github.com/dotansimha/graphql-code-generator/issues/1934

### DIFF
--- a/packages/plugins/typescript/operations/src/ts-operation-variables-to-object.ts
+++ b/packages/plugins/typescript/operations/src/ts-operation-variables-to-object.ts
@@ -1,13 +1,6 @@
 import { TypeScriptOperationVariablesToObject as TSOperationVariablesToObject } from '@graphql-codegen/typescript';
 
 export class TypeScriptOperationVariablesToObject extends TSOperationVariablesToObject {
-  protected formatFieldString(fieldName: string, isNonNullType: boolean, hasDefaultValue: boolean): string {
-    if (hasDefaultValue) {
-      return `${fieldName}?`;
-    }
-
-    return super.formatFieldString(fieldName, isNonNullType, hasDefaultValue);
-  }
   protected formatTypeString(fieldType: string, isNonNullType: boolean, hasDefaultValue: boolean): string {
     return fieldType;
   }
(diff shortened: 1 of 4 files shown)


dotansimha:
Hi @n1ru4l ! Thank you for this PR.
I think we can even drop `TypeScriptOperationVariablesToObject` and use `TSOperationVariablesToObject` directly.

The base implementation of `TSOperationVariablesToObject#formatTypeString` does:

```ts
  protected formatFieldString(fieldName: string, isNonNullType: boolean, hasDefaultValue: boolean): string {
    if (hasDefaultValue || isNonNullType || this._avoidOptionals) {
      return fieldName;
    } else {
      return `${fieldName}?`;
    }
  }
```

And then, the `TypeScriptOperationVariablesToObject#formatTypeString` overrides it with: 

```ts
  protected formatTypeString(fieldType: string, isNonNullType: boolean, hasDefaultValue: boolean): string {
    return fieldType;
  }
```

Maybe the subclass is not even necessary there? and we can use the same one? I can't really remember, I think we should check that.

n1ru4l:
@dotansimha I changed the implementation, I still have some concerns about the `avoidOptionals` option.

What should be the correct output (given `avoidOptionals` is set to true):

```graphql
type MyType {
  foo(
    a: String = "default",
    b: String! = "default",
    c: String,
    d: String!
  ): String
}
```

Output 1

```ts
export type MyTypeFooArgs = {
  a: Maybe<Scalars['String']> | undefined,
  b: Scalars['String'] | undefined,
  c: Maybe<Scalars['String']>,
  d: Scalars['String']
};
```

Output 2

```ts
export type MyTypeFooArgs = {
  a?: Maybe<Scalars['String']>,
  b?: Scalars['String'],
  c: Maybe<Scalars['String']>,
  d: Scalars['String']
};

dotansimha:
@n1ru4l , I think we can keep those as optionals, because we can be sure that GraphQL engine will make sure to fill in the default value. 
The reason we chose to use `avoid` keyword is to keep us the option to generate optionals in some cases. 

I think option 2 is better :) 

Thank you for your work on this PR!

n1ru4l:
@dotansimha That makes sense to me 😊

I added a test and altered the implementation! Let me know what you think about it!

dotansimha:
Looks great! Thank you @n1ru4l !

n1ru4l:
@dotansimha Thank you for guiding me  🎉

PR #2035, opened 2019-06-20
Add support to use URLs as default value in the yml file

naglik:


dotansimha:
That's perfect! Thank you!

PR #2063, opened 2019-06-26
Incorrect example in write-your-plugin

DylanRJohnston:
In the description just above you state that documents is an array of type `{ filePath: string, content: DocumentNode }` but then in the subsequent example `definitions` is accessed directly instead of first via `content`.


DylanRJohnston:
Also worth noting that on `OperationDefinition` name can be undefined and on `SchemaDefinition` and `SchemaExtension` name is not present at all meaning the example will crash at runtime.

dotansimha:
Thank you!
The codegen core warn when there is an unnamed operation (and also for duplicates).

PR #2071, opened 2019-06-27
Use the proper output filename when invoking prettier

mvestergaard:
When using `near-operation-file` where the configuration doesn't point to a file, prettier was invoked incorrectly

For example, with this:
```yml
 ./packages/app/src:
    preset: near-operation-file
```

The filename passed to prettier was `./packages/app/src` and not the actual generated file name


dotansimha:
Thank you @mvestergaard !

PR #2072, opened 2019-06-27
Fix watch mode when using presets

### LINKED ISSUE #2069: Watch mode exits after first run when using near-operation-file
**Describe the bug**
 
When using the `near-operation-file` preset watch mode exits right after the first iteration.

**To Reproduce**
Run `gql-gen` with `near-operation-file` configured

1. My GraphQL schema:

```graphql
# Schema is huge. If you can't reproduce, I'll try to find the minimal reproduction setup
```

2. My GraphQL operations:

```graphql
# Same as above
```

```yml
schema: ./packages/app/src/graphql/schema/schema.graphql
overwrite: true
documents:
  - ./packages/app/src/**/*.fragment.ts
  - ./packages/app/src/**/*.query.ts
  - ./packages/app/src/**/*.mutation.ts
generates:
  ./packages/app/src/graphql/schema/index.ts:
    plugins:
      - add: '/* eslint-disable */'
      - typescript
    config:
      scalars:
        Date: Date
        DateTime: Date
        DateTimeOffset: Date
        Guid: string
        Decimal: number
        Seconds: number
        Milliseconds: number
  ./packages/app/src/graphql/schema/introspection.ts:
    plugins:
      - add: '/* eslint-disable */'
      - fragment-matcher
  ./packages/app/src:
    preset: near-operation-file
    presetConfig:
      extension: .generated.tsx
      baseTypesPath: graphql/schema
    plugins:
      - add: '/* eslint-disable */'
      - typescript-operations
      - typescript-react-apollo
    config:
      withHOC: false
      withComponent: false
      withHooks: true
```

**Expected behavior**

Watch mode should keep running

**Environment:**

- OS: Windows 10
- `@graphql-codegen/1.3.1`: 
- NodeJS: 12.3.1

**Additional context**

image

## Other issues with `near-operation-file`
- Output is quite spammy. As also seen above. Each generated file produces a `√ Parse configuration` line
- While running, a weird warning appears (i think it comes from prettier, see screenshot below)
- Execution is quite slow compared to generating a single file. At least 300% slower.
- When referencing fragments, alot of unused imports are added to the generated files (i think this has been reported already)

image

mvestergaard:
This is a fairly naive attempt at fixing watch mode while using a preset.

The problem stems from the key of the configuration using the preset pointing to a folder, and in the end being ignored by chokidar.

The easy fix would be to just leave out any `generates` configuration that uses a prefix, but I've tried to make it do the right thing, by combining the key (assuming it's a path) with the configured extension (if one exists)

For config
```yml
some/folder:
  preset: near-operation-file
  presetConfig:
      extension: .generated.tsx
```

You should end up with an ignore of `some/folder/**/*.generated.tsx`

Related #2069

### DIFF
--- a/packages/graphql-codegen-cli/src/utils/watcher.ts
+++ b/packages/graphql-codegen-cli/src/utils/watcher.ts
@@ -6,6 +6,7 @@ import * as isGlob from 'is-glob';
 import * as logSymbols from 'log-symbols';
 import { debugLog } from './debugging';
 import { getLogger } from './logger';
+import path from 'path';
 
 function log(msg: string) {
   // double spaces to inline the message with Listr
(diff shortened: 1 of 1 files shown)


dotansimha:
Great catch, thank you so much, @mvestergaard !

PR #2073, opened 2019-06-27
Improve performance by not building schema AST over and over

mvestergaard:
Each plugin being executed was generating an AST of the schema over and over again, which was especially painful when using `near-operations-file`.

With this change the AST is built early, and only rebuilt in situations where it isn't passed to the plugin for whatever reason, or a plugin changes the schema.

In my project it decreased execution time from 10s to 4s (250%)

Before
image

After
image

There's probably still more that could be done, but this takes care of the worst bits

Addresses the performance issues mentioned on #2069

### DIFF
--- a/packages/graphql-codegen-cli/src/codegen.ts
+++ b/packages/graphql-codegen-cli/src/codegen.ts
@@ -16,7 +16,7 @@ export async function executeCodegen(config: Types.Config): Promise<Types.FileOu
   function wrapTask(task: () => void | Promise<void>, source?: string) {
     return async () => {
       try {
-        await task();
+        await Promise.resolve(task());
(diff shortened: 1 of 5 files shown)


mvestergaard:
Another obvious speed improvement, is to make the prettier formatting optional. In the flamegraph above, after these changes, prettier makes up 40-50%

Here's a flamegraph with prettier removed
image

It brought execution time in my scenario down to 2.5 seconds (another 160%, or 400% increase total)

dotansimha:
WOW, that's super interesting. Thank you so much for your efforts! 
And you are right, the `near-operation-file` (and presets in general) is still in experiments. Thank you so much for helping us improving this.

PR #2074, opened 2019-06-27
Add config option for disabling output formatting

mvestergaard:
See https://github.com/dotansimha/graphql-code-generator/pull/2073#issuecomment-506531049


dotansimha:
Thank you, @mvestergaard ! 
Actually it was doing this as optional, so only when prettier was available it runs it. But this is not good enough because we couldn't disable it. 
This is a very important PR! 

Can you please rebase it, and then I'll merge?

mvestergaard:
Rebase done

PR #2076, opened 2019-06-28
Fix: incorrect comment position

ken0x0a:
property should be after comment to recognize correctly.

NO CHANGE FOR FUNCTIONALITY


dotansimha:
Thank you @ken0x0a !

PR #2085, opened 2019-07-01
Set comment descriptions to false by default

jdreg95:
By setting the comment descriptions to false, comments in the `.graphql` files will also be included in introspection after code generation.

Currently (with the default `true`), comments are not included in introspection.


dotansimha:
Thank you @jdreg95 !

PR #2090, opened 2019-07-01
Strip all extraneous type import characters

blakeembrey:
Closes the same issue as https://github.com/dotansimha/graphql-code-generator/pull/1166 that's re-appeared in the refactor.


dotansimha:
Thank you @blakeembrey !

Issue #2083, opened 2019-07-01
New plugin that generates Typescript source file from GraphQL files


dex4er:
Hi,

I wrote some additional plugin https://github.com/dex4er/js-graphql-codegen-typescript-document-nodes

It generates Typescript files from GraphQL schema, ie.:

```ts
import {DocumentNode} from 'graphql';
import gql from 'graphql-tag';

export const viewerQuery: DocumentNode = gql`
  query Viewer {
    viewer {
      login
      name
    }
  }
`;
```

It would be great if this plugin will be a part of your framework. Just feel free to include it and change it if you find it useful.

dotansimha:
Hi @dex4er ! That's awesome! 

Do you want to create a PR for it?

dex4er:
@dotansimha ok, I never worked with lerna but I'll try

PR #2098, opened 2019-07-03
Typescript document nodes

dex4er:
Hi. This is PR for #2083 

This is a new plugin that generates Typescript files from GraphQL schema which use graphql-tag module so basically it embeds GraphQL files in TypeScript files.

Ah, BTW:

`tslint` complained about quotes and this check should be disabled there because formatting is already checked and fixed by `prettier`.


tslint.json
```diff
@@ -29,9 +29,9 @@
     "object-literal-sort-keys": false,
     "no-conditional-assignment": true,
     "one-line": [true, "check-open-brace", "check-whitespace"],
-    "quotemark": [true, "single"],
```
ardatan: Why do we need to change this?

dex4er: https://prettier.io/docs/en/comparison.html

and 

https://alexjover.com/blog/use-prettier-with-tslint-and-be-happy/

Please consider to use `tslint-config-prettier` because tslint often collides with prettier and then this config relaxes tslint's settings in favour of prettier.

I don't mind if you will change the settings so you will use workflow that maches your needs.

dex4er: I think it was a line that made a conflict between prettier and tslint:

https://github.com/dex4er/graphql-code-generator/blob/aab6161c668b2d18e41730934e29a0d5e13a1417/packages/plugins/typescript/document-nodes/src/index.ts#L113

ardatan: You can use template literals there instead of `"`, and I also think this prettier and tslint stuff is not directly related to this PR. So we should discuss this on another issue in my opinion.

dex4er: Agree

packages/plugins/typescript/document-nodes/lib/typescript-document-nodes.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
```
dotansimha: Can we remove this one?

packages/plugins/typescript/document-nodes/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+import { GraphQLSchema, OperationDefinitionNode } from 'graphql';
+import { print } from 'graphql/language/printer';
+
+export type NamingConvention = 'lowerCamelCase' | 'UpperCamelCase' | 'UPPER_CASE';
+
+function upperFirst(str: string): string {
+  return str[0].toUpperCase() + str.slice(1);
+}
+
+function camelize(str: string): string {
```
dotansimha: You can use `change-case` library for those transformation, it's already part of the codegen's dependencies.

packages/plugins/typescript/document-nodes/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  nameSuffix?: string;
+}
+
+export const plugin: PluginFunction<TypeScriptDocumentNodesPluginConfig> = (schema: GraphQLSchema, documents: Types.DocumentFile[], config: TypeScriptDocumentNodesPluginConfig): string => {
+  const { namingConvention, namePrefix = '', nameSuffix = '' } = config;
+  const content = documents
+    .filter(documentFile => documentFile.filePath.length > 0)
+    .map(documentFile =>
+      documentFile.content.definitions
+        .filter((d: OperationDefinitionNode) => d.name && d.name.value)
```
dotansimha: Maybe add `console.warn` here to notify about the ignored operations?

dotansimha:
It looks good. I think this plugin could be very useful for doing TypeScript compilation (because we can't include `.graphql` in the ts output directory).
Maybe we can also add a feature like `noGraphQlTag`? this way we can provide a ready-to-use and compiled `DocumentNode` (instead of doing that during runtime with `graphql-tag`). 
We had the feature in the TypeScript operations plugin, it's done like that: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts#L112-L122

packages/plugins/typescript/document-nodes/lib/typescript-document-nodes.js
dex4er: Oh... I think it was some artifact from lerna.

packages/plugins/typescript/document-nodes/src/index.ts
dex4er: +1

packages/plugins/typescript/document-nodes/src/index.ts
dex4er: I'm not convinced. This pattern is copy-pasted from `graphql-files-modules`. Should I change it in all existing plugins?

dex4er:
> It looks good, thanks @dex4er!. I think this plugin could be very useful for doing TypeScript compilation (because we can't include `.graphql` in the ts output directory).
> Maybe we can also add a feature like `noGraphQlTag`? this way we can provide a ready-to-use and compiled `DocumentNode` (instead of doing that during runtime with `graphql-tag`).
> We had the feature in the TypeScript operations plugin, it's done like that: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts#L112-L122

I think it might be a good idea but I would prefer this PR accepted first. It would be easier to add new features later if this plugin exists already in the project.

packages/plugins/typescript/document-nodes/src/index.ts
dex4er: I'm going to change this parameter to be compatible with namingConvention from visitor-plugin (`module#method`) so it will be more flexible

packages/plugins/typescript/document-nodes/src/index.ts
dex4er: @dotansimha I double checked this code. It works like additional sanitisation and filtering it is very easy but making some warning on missing value might be more difficult and probably unnecessary.

dotansimha:
Perfect. Thank you @dex4er ! Amazing work.

PR #2111, opened 2019-07-05
Add failing test for duplicate names

mxstbr:
This adds a failing test for issue #1954 with a minimal reproduction of duplicate type names being generated. I tried fixing this, but am not sure where in the codebase the fix should go.

@dotansimha would you mind tackling this?


dotansimha:
Thanks @mxstbr , I added `addUnderscoreToArgsType` that resolves this. I couldn't see any other use-cases that might cause this issue.

Issue #2110, opened 2019-07-05
Make suffix of generated query response types customizable


jlkiri:
**Is your feature request related to a problem? Please describe.**

`componentSuffix` option of `typescript-react-apollo` plugin allows us to generate React Apollo components with arbitrary suffixes. The default is `XXXComponent` but many would like to generate something like `XXXQuery`. However, this format is already reserved for generated query response types, e.g.

```javascript
export type PlayerQuery = ({ __typename?: 'Query' } &
{ player: ({ __typename?: 'PlayerModel' } & Pick<PlayerModel, 'id' | 'username' | 'character'>) });
```

**Describe the solution you'd like**

Although type names and component names should not clash, it would be preferrable to be able to also customize names of generated query responses via some similar option in config. I think `querySuffix` is appropriate.

**Describe alternatives you've considered**

Is this better done somehow by extending the execution flow?

**Additional context**

jlkiri:
I have a working solution but not sure if it is the best way. I can send a PR.

dotansimha:
@jlkiri sure, can you please share it here or submit a PR? I :)

dotansimha:
Available in 1.4.0 🎉

PR #2116, opened 2019-07-06
Make suffix of generated query response types customizable

jlkiri:
This PR addresses the following issue: #2110 

I made changes to `base-documents-visitor` that allow users to specify (via config) a custom suffix of generated operation result type names (like `PlayerQuery` or `PlayerMutation` types). Adding a suffix `Result` to e.g. `PlayerQuery` type would generate `PlayerQueryResult`.

To add a suffix (which type is `string`), you have to use `operationResultSuffix` in `config.yml`.


dotansimha:
Hi @jlkiri ! 
Thank you.
I think you should allow other base plugins to know about this new flag as well. 
The `typescript-react-apollo` (based on `base-client-side-visitor`) is using the generated type name, so it might cause a mismatch if it won't append this suffix as well.

jlkiri:
As far as I could understand, `base-client-side-visitor` is the only one I had to additionally change. I also added an annotation.

dotansimha:
Looks great, thanks @jlkiri !

PR #2125, opened 2019-07-08
Import path module as variable instead of default in watcher

therealtbs:
The usage later on gets transpiled by TypeScript into:
```javascript
ignored.push(path_1.default.join(entry.filename, '**', '*' + extension));
```
Note the `default`, which is incorrect for a commonjs module and causes a crash.

This PR fixes that


dotansimha:
@therealtbs , actually, @mvestergaard just fixed it a minute after you, merged his PR first. 

Thank you!!

therealtbs:
No worries! :relieved: Just glad it's fixed

PR #2126, opened 2019-07-08
Fix import causing issues when compiled to commonjs

mvestergaard:
The use of `import path from 'path'` was incorrectly compiled to commonjs as `path_1.default.join`


dotansimha:
Thank you @mvestergaard !

PR #2137, opened 2019-07-09
Fixes issue where base types file is not referenced

mvestergaard:
This is a suggestion to fix the issue mentioned in https://github.com/dotansimha/graphql-code-generator/issues/2069#issuecomment-509346480

When a query only contains a fragment spread:
```graphql
query {
  user {
    ...UserFragment
  }
}
```
as opposed to
```graphql
query {
  user {
    id
    ...UserFragment
  }
}
```

The import of the main types file, i.e. `import * as Types from '../types'` was not included.


dotansimha:
Thank you @mvestergaard ! 
I think @mxstbr had the same issue recently.

PR #2139, opened 2019-07-09
Guard against duplicate fragment names in near-operations-file

mvestergaard:
Right now if you have multiple fragments with the same name, it will not fail, and you may end up with the wrong fragment being used.

Ideally I would prefer that it figured out using the correct fragment based on the imports in the source file, however that could be pretty complicated.

For now, this will at least prevent you from having duplicate fragment names.


dotansimha:
That's great! Thank you @mvestergaard ! 
Can you please rebase? And then I'll merge this.

mvestergaard:
Done

Issue #2160, opened 2019-07-12
Wrong component prop type names when using typescript-react-apollo


jlkiri:
**Describe the bug**
As a result of #2116, using `operationResultSuffix` leads to generation of wrong React Apollo component prop type names. They are not updated, but operation result type name is generated correctly.

**To Reproduce**
1. Specify `operationResultSuffix` in `config.yml` (e.g. `"Response"`), when using `typescript-react-apollo` plugin.
2. Generate output.
3. Check the generated file and make sure that `TData` in generated `ReactApollo.DataProps<TData, TGraphQLVariables>` has a wrong type name. E.g. `MyOperationQuery` instead of `MyOperationQueryResponse`.

**Expected behavior**
Component prop type names are updated too when `operationResultSuffix` is used.

**Environment:**
Any

**Additional context**
This is my mistake and I will fix in a PR.

mohsen89z:
Same! Also, Query argument's name is wrong!

jlkiri:
I did some tests and it looks like only generated React Apollo HOC props are affected by the change. The only current workaround is to set `withHOC` to `false`.  
This is a pretty easy fix, I'll send a PR.

dotansimha:
Fixed in 1.5.0 🎉

PR #2166, opened 2019-07-13
Fix HOC data props with custom type names

jlkiri:
This PR addresses the following issue: #2160 

When custom operation result type name is specified by `operationResultSuffix` config option, React Apollo HOC props are generated with a wrong operation type name. This PR fixes the way HOC props are generated.

Expected example (with `operationResultSuffix` set to `Response`): 
```javascript
export type BooksProps<TChildProps = {}> = Partial<
  ReactApollo.DataProps<BooksQueryResponse, BooksQueryVariables>
> &
  TChildProps;
```


dotansimha:
Perfect, thank you @jlkiri !

PR #2171, opened 2019-07-15
Add variables type parameter to Urql.useMutation hook

cometkim:
First of all, thank you for making a great tool.

Here is simple patch for add variables type to useMutation hooks so to be specified type instead of `object`

```ts
export declare const useMutation: <T = any, V = object>(query: string | DocumentNode) => [UseMutationState<T>, (variables?: V | undefined) => Promise<OperationResult<T>>];
```


dotansimha:
Perfect, thank you @cometkim !

Issue #1668, opened 2019-04-08
Possible to produce classes vs types ? 


appsolutegeek:
Hi,

I am using the typescript plugin. All types are used by using type keyword. In fact my tslint complains about 

```
Use an interface instead of a type literal. (interface-over-type-literal)tslint(1)
```

Although what I really need is for the generator to create classes. I use mappers and passing in the type isn't valid.

ie.

```
  this.mapperService.mapper
      .createMap<Responses.IUserAudit, GraphQL.UserAudit>(
        MapperHelper.signatureFromStrings("Entity.IUserAudit", "ViewModel.UserAudit"),
        GraphQL.UserAudit
      )
      .forMember("who", opt => opt.mapFrom(source => source.who))
      .forMember("when", opt => opt.mapFrom(source => source.when))
      )
```

GraphQL.UserAudit isn't valid as it is this

```
export type UserAudit = {
  what?: Maybe<UserAuditWhat>;
  relatedId?: Maybe<Scalars["String"]>;
  when?: Maybe<Scalars["String"]>;
  who?: Maybe<Scalars["String"]>;
};
```

and the createMap takes

```
export declare type TGenericClass<T = any> = {
    new (...args: any[]): T;
};
```

so it needs a concrete class.

Is there some kind of configuration to switch between interfaces and classes and types ?

Thanks in advance

appsolutegeek:
I suppose if classes can't be produced, it would be nice to be able to produce interfaces.

borremosch:
+1, I am using graphql-code-generator in conjunction with type-graphql. I have a service which has both a GraphQL server and a GraphQL client. I need to get some of the types that are received on the client side of this service (graphql-code-generated) and expose them again on the server side of the service. But for this I require the types to be classes as well, otherwise I cannot use them in a type-graphql-specific decorator.

kamilkisiela:
@borremosch @appsolutegeek I see no reason to move from types to interfaces. How is one different form another if it comes to your example?

borremosch:
@kamilkisiela in my case, it would only help to have classes. Interfaces would not add anything over types. But actually it will require more than just classes (i.e. the proper decorators) to be able to use graphql-code-generator types in conjunction with a type-graphql server, so it would not make a lot of sense to include that type of functionality in the core graphql-codegen typescript plugin.

kamilkisiela:
Sounds like a plugin, if you want I could help you implement it :)

dotansimha:
[...]

PR #2173, opened 2019-07-15
Add type-graphql support

borremosch:
Following my issue in #1668, I have extended the @graphql-code-generator/typescript plugin with an extra flag `outputTypeGraphQL` that will output type-graphql classes instead of interfaces/types, and register any emitted enums with type-graphql. This enables exposing the types generated by graphql-code-generator in a downstream GraphQL server.

I have tested it on my own code and it works, outside of a nasty issue where the reflection metadata code generated by the TypeScript compiler will try to reference a class that is defined later in the generated file, crashing the application using the generated types. I have found similar issues concerning this functionality [here](https://github.com/19majkel94/type-graphql/issues/130), [here](https://github.com/rbuckton/reflect-metadata/issues/12) and [here](https://github.com/Microsoft/TypeScript/issues/4521). I have contemplated a couple of solutions to this issue:

- Output graphql-code-generator types into separate files. This cannot be accomplished without quite a bit of refactoring in graphql-code-generator. It will work by fixing the commonjs limitation described in [this type-graphql issue](https://github.com/19majkel94/type-graphql/issues/130).
- Use a Webpack/Babel postprocessing plug-in to move all emitted decorators below the required class definition. This is a manual process done by the end user.
- A ridiculously simple solution is to define a type `type FixDecorator<T> = T;`, and then defining all fields using `field: FixDecorator<ReferencedClass>;` rather than `field: ReferencedClass`. This makes the TypeScript compiler output `Object` as the type in the emitted metadata, fixing the missing reference.

Any thoughts on how to solve this issue are greatly appreciated. I will also be opening an issue on the type-graphql GitHub.

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -182,8 +182,7 @@ export class BaseTypesVisitor<TRawConfig extends RawTypesConfig = RawTypesConfig
       .withContent(possibleTypes).string;
   }
 
-  ObjectTypeDefinition(node: ObjectTypeDefinitionNode, key: number | string, parent: any): string {
-    const originalNode = parent[key] as ObjectTypeDefinitionNode;
+  getObjectTypeDeclarationBlock(node: ObjectTypeDefinitionNode, originalNode: ObjectTypeDefinitionNode): DeclarationBlock {
     const optionalTypename = this.config.nonOptionalTypename ? '__typename' : '__typename?';
     const allFields = [...(this.config.addTypename ? [indent(`${optionalTypename}: '${node.name}',`)] : []), ...node.fields];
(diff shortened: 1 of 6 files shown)


dotansimha:
Hi @borremosch ! 
Thank you for this PR, this is amazing work! 

From a brief view - is there a reason not to create a new plugin for type-graphql and use the `typescript` package? 
This might be an issue, because `typescript` plugin outputs the "base types" for `typescript-operations` and `typescript-resolvers`, and changing the declaration to `class` might break it for all of them. 
You can still use and depend on `typescript` visitor and extend it if you wish, but I think it belong to a new plugin (`@graphql-codegen/type-graphql`?)

borremosch:
@dotansimha I had not even considered extending the typescript plugin as a possibility. I have created a new pull request in #2177. This one can be closed

Issue #2112, opened 2019-07-05
@apollo/react-common support


OneCyrus:
**Is your feature request related to a problem? Please describe.**
please add support for @apollo/react-common (react-apollo 3). currently the generated types don't work with anymore.

**Describe the solution you'd like**
generated types which work with the new @apollo/react-common package. HOC, Components and Hooks.

**Describe alternatives you've considered**
there's no alternative.

**Additional context**
@apollo/react-common changes some types so there are breaking changes. probably it's just a different mapping to the new types.

dotansimha:
Hi @OneCyrus ! 
I'm not sure, but I think this PR solves it: https://github.com/dotansimha/graphql-code-generator/pull/2077 
You can try it with `1.3.1-alpha-d928e0d7.72` because it's not released yet.

OneCyrus:
ah ok. that looks like a start. from overlooking it seems to be incomplete. will give this a try and report my findings.

dotansimha:
@OneCyrus Thanks, waiting for your update on the missing features :)

OneCyrus:
just tried the alpha with the following config:

```
generates: {
    "./src/shared/api/api.tsx": {
      config: {
        resolvers: false,
        noNamespaces: true,
        reactApolloVersion: 3,
        reactApolloImportFrom: "@apollo/react-hooks",
        withHooks: true,
        withHOC : true,
        withComponent: true
      },
      plugins: [
        "typescript",
        "typescript-operations",
        "typescript-react-apollo",
        "typescript-compatibility"
      ]
    }
  }
}; 
```

and here are the thins which i needed to do:

### 1. replace imports

replace
```
import * as ReactApollo from "@apollo/react-hooks";
import * as ReactApolloHooks from "react-apollo-hooks";
```
with
```
import { Query, QueryComponentOptions, Mutation, MutationComponentOptions } from "@apollo/react-components";
import { withQuery, withMutation, OperationOption, DataProps } from "@apollo/react-hoc";
import { MutationFunction } from "@apollo/react-common";
import { useMutation, MutationHookOptions, QueryHookOptions, useQuery } from "@apollo/react-hooks";
```

### 2. replace types/strings

i replaced the following strings with new values

| old value|new value|
|-----------|-------------|
|ReactApollo. | "" (nothing)|
|ReactApolloHooks. | "" (nothing)|
|MutateProps| MutationComponentOptions|
|MutationProps | MutationComponentOptions|
[...]

PR #2182, opened 2019-07-17
Support new React Apollo API

ardatan:
Related #2112
I think we should wait for stable release.
`@apollo/*` libraries can be used with the version;
` - @graphql-codegen/typescript-react-apollo => 1.4.1-alpha-3b1e744b.8+3b1e744b`


yakovlevyuri:
`@apollo/*` libraries have been bumped to v.3 🎊

SimenB:
Did this drop support for `react-apollo-hooks` entirely?

ardatan:
You are right, it is a breaking change and we are sorry about that.
With a bit more context, the maintainer of `react-apollo-hooks` said that his package would be deprecated in favor of the stable release of `@apollo/react-hooks`, so we’ve moved to the new package after the stable release of it.
The name and the signature of generated hooks should be almost same in both packages, so the only thing changed is the package name.
But if you want to go with the old package, you can use older version for react-apollo plugin of codegen.
I understand it is still a breaking change but we believe it is a relatively small and manageable one

Issue #1885, opened 2019-05-17
[typescript-operations] it should be possible to set avoidOptionals only for input types


helios1138:
- currently, it's only possible to set `avoidOptionals` globally, for all types
- what would be preferred, is to have a way to set this option to `true` for output types (since empty values are returned as `null`, not `undefined` in the responses), but set it to `false` for input types (since variables treat `undefined` as null and it's a normal practice to omit empty values when providing variables, rather than providing `null`s

dotansimha:
Hi @yoavain-sundaysky ! Thank you for reporting this issue.
You are right, at the moment the optionals configuration can only be turned on/off and is not configurable.
We can give it a try in the future and add it (and we always welcome PRs).

slikts:
An example use case for `avoidOptionals` is to model mock data in strictly the same way the server would return it, i.e., with `null` for requested nullable fields. Meanwhile, nullable fields in variable and input types should still be optional, because being able to elide the fields is also how it works in GraphQL, and having to always explicitly provide a `null` or `[]` is verbose and poor DX.

A workaround to currently avoid using `avoidOptionals` is a helper that removes optionality:
```ts
type DeepNonNullable<T> = {
  [P in keyof T]-?: T[P] extends {} ? DeepNonNullable<T[P]> : T[P];
}
```

dotansimha:
Implemented in https://github.com/dotansimha/graphql-code-generator/pull/2191 . Thanks @ekfn

dotansimha:
Available in 1.5.0 🎉

thefliik:
@dotansimha it doesn't look like this issue should be closed yet. In my testing, #2191 fixes this issue for the `typescript` plugin, but not for the `typescript-operations` plugin.

It appears that `typescript-operations` still only accepts a boolean value for `avoidOptionals`. Attempting to pass an object value is treated the same as passing `avoidOptionals: true`.

dotansimha:
@thefliik you are right, I added a validation for the type of `avoidOptionals` in `typescript-operations`:
```
        avoidOptionals: typeof config.avoidOptionals === 'boolean' ? getConfigValue(config.avoidOptionals, false) : false,

```

Now it will ignore the object value if it set globally.

dotansimha:
Fixed in 1.6.0

PR #2191, opened 2019-07-18
Avoid input optionals

ekfn:
Relevant to  https://github.com/dotansimha/graphql-code-generator/issues/1885

### DIFF
--- a/packages/plugins/typescript/typescript/src/avoid-optionals.ts
+++ b/packages/plugins/typescript/typescript/src/avoid-optionals.ts
@@ -1,0 +1,20 @@
+import { AvoidOptionalsConfig } from './types';
+
+export const DEFAULT_AVIOD_OPTIONALS: AvoidOptionalsConfig = {
+  object: false,
+  inputValue: false,
+};
+
+export function normalizeAvoidOptionals(avoidOptionals?: boolean | AvoidOptionalsConfig): AvoidOptionalsConfig {
+  if (typeof avoidOptionals === 'boolean') {
+    return {
+      object: avoidOptionals,
+      inputValue: avoidOptionals,
+    };
+  }
+
+  return {
+    ...DEFAULT_AVIOD_OPTIONALS,
+    ...avoidOptionals,
+  };
+}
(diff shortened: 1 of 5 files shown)


dotansimha:
Thank you @ekfn ! 

thefliik:
That moment when you run into a problem only to find someone else has already created a PR to solve it 😃  🎉 

This being said, is it appropriate for this API to be different from the [`declarationKind` config](https://graphql-code-generator.com/docs/plugins/typescript#declarationkind-declarationkindconfig) API ?

Specifically,  `declarationKind` has a single config argument. You can either provide a single string value, or you can provide a config object which allows different configs depending on the GQL schema element type. Rather than letting `avoidOptionals` accept a single boolean or a config object, I see this PR adds a separate config argument, `avoidInputOptionals`. 

On the one hand, `avoidOptionals` only applies to two GQL schema types: `FieldDefinition` and `InputValueDefinition` while `declarationKind` can apply to `ScalarsDefinition`, `InputObjectTypeDefinition`, `ObjectTypeDefinition`, etc. Letting the `avoidOptionals` config argument accept an object may make people think they can configure other definition types rather than just Field and InputValue.

On the other hand, implementing this change by keeping a single `avoidOptionals` argument, that accepts either a config object or a boolean, will create a more consistent API and is potentially more "future proofed", as other keys could be added to the config object in the future. Admittedly, I can't think of anything else that might be added in the future.

All this is to say that I, personally, would vote for something like this to be the API:

```yml
config:
  avoidOptionals:
    object: true
    inputValue: false

# OR

config:
  avoidOptionals: true
```

It's not a big deal though.

dotansimha:
Thank you @ekfn ! 
What do you think about @thefliik 's suggestion? I think it's a bit more nice API.

ekfn:
Yes, the API is a bit nicer and has better backward compatibility, although I am not familiar with all parts of this repo, I will try to implement it.

ekfn:
@dotansimha I've updated pull request

dotansimha:
That's perfect! Thank you @ekfn ! how was the experience of working on this? the code base was easy to understand?

ekfn:
@dotansimha the code base is good, TS was very helpful, but the tests were a bit confusing (I think they do not cover all test cases).
Glad that I was able to help. :)

PR #2224, opened 2019-07-24
Fix links to @graphql-codegen/typescript

glenwinters:
This fixes a couple links to the typescript plugin page that currently result in a 404.


dotansimha:
Thank you @glenwinters !

PR #2227, opened 2019-07-24
process.env.DEBUG

Qeneke:
I don't want to receive codegen logs, but when I set the value DEBUG to undefined it affects the whole project


dotansimha:
Thanks @Qeneke !

PR #2233, opened 2019-07-26
Fix external enums to be re-exported from the generated file

### LINKED ISSUE #2206: External enums not re-exported from generated `types.ts` break near operations
**Describe the bug**
If external enums (either explicit mapping `Role:./enums#MyRole` or the new single file option #2038) used in combination with `near-operation-file` preset and `typescript-operations`, and an enum is used as a field argument, the enum is incorrectly imported in the "near" file. 
Or in other words, external enums aren't re-exported from generated `types.ts`.

```ts
// src/UsersByRole.generated.ts
import * as Types from "./types";

export type UsersByRoleQueryVariables = {
  role: Types.Role; // 💥 has no exported member 'Role'
};

export type UsersByRoleQuery = { __typename?: "Query" } & {
  usersByRole: Array<
    { __typename?: "User" } & Pick<Types.User, "id" | "username">
  >;
};

```

**To Reproduce**

GraphQL schema:

```graphql
# schema.graphql
schema {
  query: Query
}

type Query {
  usersByRole(role: Role!): [User!]!
}

enum Role {
  USER,
  ADMIN,
}

type User {
  id: ID!
  username: String!
  role: Role!
}
```

GraphQL operations:

```graphql
# src/UsersByRole.graphql
query UsersByRole($role: Role!) {
  usersByRole(role: $role) {
    id
    username
  }
}
```

Config file:

```yml
# codegen.yml
schema: schema.graphql
documents: src/**/*.graphql
config:
  enumValues:
    Role: ./enums#MyRole
generates:
  src/types.ts:
    - typescript
  src/:
    preset: near-operation-file
    presetConfig:
      baseTypesPath: types.ts
    plugins:
      - typescript-operations
```

File with external enums:

```ts
// src/enums.ts
export enum MyRole {
    User = "USER",
    Admin = "ADMIN",
}
```

Code sandbox example: https://codesandbox.io/s/graphql-codegen-issue-template-g5m7g

**Expected behaviour**
Generated `UsersByRole.generated.ts` correctly imports `Role` enum. Either directly from `enums.ts` or, probably better, from `types.ts` where is re-exported from `enums.ts`. 

**Environment:**

- OS: Win 10
- `@graphql-codegen/...`: canary, 1.4.0, 1.3.1,…
- NodeJS: 10

vhenzl:
Related #2206.

It works but I'm not sure if the approach is correct. Maybe there should be `getEnumsExports` method in `BaseTypesVisitor` (to complement `getEnumsImports`)?

### DIFF
--- a/packages/plugins/typescript/typescript/src/visitor.ts
+++ b/packages/plugins/typescript/typescript/src/visitor.ts
@@ -78,7 +78,7 @@ export class TsVisitor<TRawConfig extends TypeScriptPluginConfig = TypeScriptPlu
 
     // In case of mapped external enum string
     if (this.config.enumValues[enumName] && this.config.enumValues[enumName].sourceFile) {
-      return null;
+      return `export { ${this.config.enumValues[enumName].typeIdentifier} };\n`;
     }
 
     if (this.config.enumsAsTypes) {
--- a/packages/plugins/typescript/typescript/tests/typescript.spec.ts
+++ b/packages/plugins/typescript/typescript/tests/typescript.spec.ts
@@ -1573,6 +1573,26 @@ describe('TypeScript', () => {
 
       validateTs(result);
     });
+
+    it('Should re-export external enums', async () => {
+      const schema = buildSchema(`enum MyEnum { A, B, C } enum MyEnum2 { X, Y, Z }`);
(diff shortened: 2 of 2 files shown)


dotansimha:
It looks good. Thank you @vhenzl !

Issue #2240, opened 2019-07-28
Typescript Operations: Generated types for selection set that have unions/interfaces are incorrect


n1ru4l:
**Describe the bug**

The generated types for a field that returns a union/interface type is unnecessarily nested and incorrect.

**Reproduction on Codesandbox**:

https://codesandbox.io/s/graphql-codegen-issue-template-h3kpf?fontsize=14

- See `types.tsx`@ Line 70 for incorrect types
- See `expected-types.tsx` @ Line 70 for correct types (how it should be)
- See `example-client-usage.tsx` for example usage, switch import for generated file (L4 - 7).

Diff: 
```diff
diff --git a/types.tsx b/expected-types.tsx
index fdf99b1..4e39bf1 100644
--- a/types.tsx
+++ b/expected-types.tsx
@@ -69,13 +69,16 @@ export type UserRegisterMutationVariables = {
 
 export type UserRegisterMutation = { __typename?: "Mutation" } & {
   userRegister:
-    | (({
-        __typename?: "UserRegisterInvalidInputError" | "CountryBlockedError";
-      } & Pick<Error, "message">) &
-        ({ __typename?: "UserRegisterInvalidInputError" } & Pick<
-          UserRegisterInvalidInputError,
-          "loginErrorMessage" | "emailErrorMessage" | "passwordErrorMessage"
-        >))
+    | ({
+        __typename?: "CountryBlockedError";
+      } & Pick<Error, "message">)
+    | ({ __typename?: "UserRegisterInvalidInputError" } & Pick<
+        UserRegisterInvalidInputError,
+        | "message"
+        | "loginErrorMessage"
+        | "emailErrorMessage"
+        | "passwordErrorMessage"
+      >)
     | ({ __typename?: "UserRegisterResultSuccess" } & {
         user: { __typename?: "User" } & Pick<User, "id" | "login">;
       });
```

**Reproduction manual (prefer codesandbox)**:

1. My GraphQL schema:

```graphql
type User {
  id: ID!
  login: String!
}

type UserRegisterResultSuccess {
  user: User!
}

interface Error {
  message: String!
}

type UserRegisterInvalidInputError implements Error {
  message: String!
  loginErrorMessage: String
  emailErrorMessage: String
  passwordErrorMessage: String
}

type CountryBlockedError implements Error {
  message: String!
}

input UserRegisterInput {
  login: String!
  email: String!
  password: String!
}

union UserRegisterResult =
[...]

Issue #2200, opened 2019-07-19
Typescript Operations don't handle unions properly


tempname11:
**Describe the bug**
The typescript-operations plugin generates incorrect types for unions.

There are two problems, as far as I can see:

1) Not all union cases are generated (the server may actually return every one of them, regardless of what cases the query fragments specify).
2) If fragments operate on interfaces, the types are joined by `&` in the overall type, instead of just the cases which implement this interface.

**To Reproduce**
Minimal repro is at https://github.com/tempname11/gcg-example
See the output at https://gist.github.com/tempname11/c1f9a8ca97530bae078cfd55654f3068

example1 is correct:
```
  example1:
    | ({ __typename: "Foo" } & Pick<Foo, "foo">)
    | ({ __typename: "Bar" } & Pick<Bar, "bar">);
```
However, both example2 and example 3 are incorrect:
```
  example2: { __typename: "Foo" } & Pick<Foo, "foo">;
  example3: ({ __typename: "Foo" | "Bar" } & Pick<Common, "common">) &
({ __typename: "Foo" } & Pick<Foo, "foo">);
>;
```

**Expected behavior**
```
  example2:
    | ({ __typename: "Foo" } & Pick<Foo, "foo">)
    | ({ __typename: "Bar" })
  example3:
    | ({ __typename: "Foo" } & Pick<Foo, "foo"> & Pick<Common, "common">))
    | ({ __typename: "Bar" } & Pick<Common, "common">))
>;
```

**Environment:**

- OS: macOS 10.14.5
- `@graphql-codegen/...`: 1.4.0
- NodeJS: 10.16.0

n1ru4l:
I created a duplicate with some more context and a codesandbox example: https://github.com/dotansimha/graphql-code-generator/issues/2240

n1ru4l:
I just created a pull request with failing tests: #2248

dotansimha:
Thank you! Will take a look very soon.

dotansimha:
Closing this for now, let's continue discussing it here: https://github.com/dotansimha/graphql-code-generator/issues/2240 (it has more context and examples)

PR #2248, opened 2019-07-29
fix(typescript/flow-operations): generate correct union/interface type definitions

n1ru4l:
- Changes the tests for union/interfaces according to #2240 and #2200
- Fixes code generation

**TODO:**

- [x] fix `preResolveTypes` option
- [x] clean up code (kind of^^)

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -15,8 +15,12 @@ import {
   TypeMetaFieldDef,
   isScalarType,
   GraphQLInterfaceType,
+  SelectionNode,
+  isListType,
+  isNonNullType,
+  GraphQLOutputType,
 } from 'graphql';
-import { getBaseType, quoteIfNeeded, isRootType } from './utils';
+import { getBaseType, quoteIfNeeded, isRootType, getBaseTypeNode } from './utils';
 import { ScalarsMap, ConvertNameFn, LoadedFragment } from './types';
 import { GraphQLObjectType, GraphQLNonNull, GraphQLList } from 'graphql';
 import { BaseVisitorConvertOptions } from './base-visitor';
@@ -33,6 +37,37 @@ function isMetadataFieldName(name: string) {
 const metadataFieldMap: Record<string, GraphQLField<any, any>> = {
   __schema: SchemaMetaFieldDef,
   __type: TypeMetaFieldDef,
+};
+
+const getFieldNodeNameValue = (node: FieldNode): string => {
+  return (node.alias || node.name).value;
+};
+
+const mergeSelectionSets = (selectionSet1: SelectionSetNode, selectionSet2: SelectionSetNode) => {
+  const newSelections = [...selectionSet1.selections];
+  for (const selection2 of selectionSet2.selections) {
+    if (selection2.kind === 'FragmentSpread') {
+      newSelections.push(selection2);
+      continue;
+    }
+    if (selection2.kind !== 'Field') {
+      throw new TypeError('Invalid state.');
+    }
+
+    const match = newSelections.find(selection1 => selection1.kind === 'Field' && getFieldNodeNameValue(selection1) === getFieldNodeNameValue(selection2));
+
+    if (match) {
+      // recursively merge all selection sets
+      if (match.kind === 'Field' && match.selectionSet && selection2.selectionSet) {
+        mergeSelectionSets(match.selectionSet, selection2.selectionSet);
+      }
+      continue;
+    }
+    newSelections.push(selection2);
+  }
+
+  // replace existing selections
+  selectionSet1.selections = newSelections;
 };
 
 export class SelectionSetToObject {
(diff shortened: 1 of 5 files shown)


n1ru4l:
the tests should now cover all important cases

n1ru4l:
So it seems in order to fix this we have to change the process on how the code is generated:

Currently, this is what I have in mind:

**Step 1**: Operation Documents must be normalized

We should normalize the operation AST by removing duplicated fields and inlining `inline fragments` (on non-union/interface fields).

This can be done by using `relay-compiler` with the `FlattenTransform` and the `SkipRedundantNodesTransform` transform. Of course, we could also build our own (e.g. I already built a flatten operation transform that works without the relay-compiler)

**Step 2**: Traverse the normalized document AST

After we have normalized the document AST we can traverse it with confidence that no duplicated types are generated


@dotansimha @tempname11  How do you think about this? I am not sure whether I am missing something...

n1ru4l:
Only flow tests are failing now, they need to be updated. AFAI t understand the types were also incorrect.

n1ru4l:
@dotansimha I need to update the Flow Tests and then I will try cleaning up the code.

dotansimha:
Thanks @n1ru4l ! 
What do you think left in order to complete this PR?

n1ru4l:
Clean up the code and add tests for queries that use both  inline and non inline fragments.

n1ru4l:
Ok, now I have to figure out what to do in case a fragment is specified on an enum type. Any suggestions welcome 👀

For now, it seems like the following is correct:


```graphql
type Error {
  message: String!
}

type User {
  id: ID!
  login: String!
}

union UserResult = User | Error

type Query {
  user: UserResult!
}
```

```graphql
fragment UserResultFragment on UserResult {
  ... on User {
    login
  }
  ... on Error {
    message
  }
}

query TheQuery {
  user {
    ...UserResultFragment
    ... on User {
      id
    }
  }
}
```


```ts
export type User = {
  id: string
  login: string
}

export type Error = {
  message: string
}

export type UserResultFragmentFragment = (
  (
    { __typename?: "User"}
    & Pick<User, 'login'>
  ) | (
    { __typename?: "Error"}
    & Pick<Error, 'message'>
  )
)

export type TheQueryQuery = (
  { __typename?: 'Query' }
  & {
   user: (
     (
       { __typename?: "User" }
       & Pick<User, 'id'>
     ) | (
       { __typename?: "Error" }
     )
   ) & UserResultFragmentFragment
  }
);

function handle(input: TheQueryQuery) {
  if (input.user.__typename === "User") {
    input.user.login
  } else if (input.user.__typename === "Error") {
    input.user.message
  }
}
```

https://www.typescriptlang.org/play/#code/KYDwDg9gTgLgBDAnmYcCqBnYU4F44DeAUHHAJYAmAXHBjFGQHYDmJcANhM0zXQy0QC+RIqEiwEyVAFEoUaHkJsAtsAwYAhs2C96TVsNHho8JCnRYoAJTUBXdjABiULasZOXzN-HwAKNv6kpARwAPqhZsCMGqoA-DQARJjYCcJBcABkcAAKZADGANYAPMlQADRwAOSc3IyVAHxsAJRwAD5wgUEh4ZHRcYmy8lCpbKRZuYVFg9AVlarqWsANzURNImImkuYAKgAWwACKttiIRyeKnd0RUn3A8VVnUIiVcGmZSkG2ljSdQb-pV16MTuiVKCVeo3S43yxVKs0oy3ScBa7X+XTC1xQt3uCWmwwhSLWQRaWVKNgw9g8riiVK8NLYwiaAG4RAAzWyMPIwMgQRhwXYaRgUdjAXxMMC2GA0PaHY5PR6IFrEUhkVkdcWSgB0X2wmp6N2BeFw+CSlgSSshGpg2ssmpqTAZcGA7Cw5DVYsYEutOqgesxUUNxpNePNHyCVptuvmmm0DKEQA

n1ru4l:
I now only need to fix the `preResolveTypes` options and clean up the code.

n1ru4l:
@dotansimha ready for review 🎉

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -35,6 +39,37 @@ const metadataFieldMap: Record<string, GraphQLField<any, any>> = {
...
+      if (match.kind === 'Field' && match.selectionSet && selection2.selectionSet) {
+        mergeSelectionSets(match.selectionSet, selection2.selectionSet);
+      }
+      continue;
+    }
+    newSelections.push(selection2);
+  }
+
+  // replace existing selections
+  selectionSet1.selections = newSelections;
```
dotansimha: Maybe instead of mutating `selectionSet1`, we should create a new object and return it?
I think it might effect other pieces of code that use `selectionSet1.selections`.

dotansimha:
I pushed the recent changes into a branch in this repo (instead of the fork). So now it's available as alpha if someone wants to try it: `1.5.1-alpha-3f341e3e.23`.

PR #2242, opened 2019-07-29
Revert "Make variables optional if skip is provided (#1926)"

palindrom615:
Please see #2145 issue. 

This reverts commit 6508ba45ff32608f353aaba0dc50c149a3ff1ad0.

PR #2251, opened 2019-07-30
Allow for skip to be a boolean type

boom:
- Hardcoded true or false does not allow for variables of type boolean to be passed into the skip property. This resolves the type error for boolean variables used.

Resolves bug introduced from https://github.com/dotansimha/graphql-code-generator/pull/1926

### DIFF
--- a/dev-test/githunt/types.reactApollo.customSuffix.tsx
+++ b/dev-test/githunt/types.reactApollo.customSuffix.tsx
@@ -340,7 +340,7 @@ export const CommentDocument = gql`
   }
 }
     ${CommentsPageCommentFragmentDoc}`;
-export type CommentComponentProps = Omit<ReactApollo.QueryProps<CommentQueryMyOperation, CommentQueryVariables>, 'query'> & ({ variables: CommentQueryVariables; skip?: false; } | { skip: true; });
+export type CommentComponentProps = Omit<ReactApollo.QueryProps<CommentQueryMyOperation, CommentQueryVariables>, 'query'> & ({ variables: CommentQueryVariables; skip?: boolean; } | { skip: boolean; });
 
     export const CommentComponent = (props: CommentComponentProps) => (
       <ReactApollo.Query<CommentQueryMyOperation, CommentQueryVariables> query={CommentDocument} {...props} />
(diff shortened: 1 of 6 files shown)


dev-test/githunt/types.reactApollo.customSuffix.tsx
```diff
@@ -340,7 +340,7 @@ export const CommentDocument = gql`
   }
 }
     ${CommentsPageCommentFragmentDoc}`;
-export type CommentComponentProps = Omit<ReactApollo.QueryProps<CommentQueryMyOperation, CommentQueryVariables>, 'query'> & ({ variables: CommentQueryVariables; skip?: false; } | { skip: true; });
+export type CommentComponentProps = Omit<ReactApollo.QueryProps<CommentQueryMyOperation, CommentQueryVariables>, 'query'> & ({ variables: CommentQueryVariables; skip?: false; } | { skip: boolean; });
```
benjie: Can we change the false to boolean also?

boom: Good call. Updated.

ardatan:
Looks like same
#2242 https://github.com/dotansimha/graphql-code-generator/pull/2242

boom:
> Looks like same
> #2242 #2242

Believe the revert is not necessary but either will do from my perspective. Was trying to honor the initial intent of https://github.com/dotansimha/graphql-code-generator/pull/1926 in this PR.

n1ru4l:
I do not think this is necessary?

This code works fine:
```tsx
const { data } = useMeQuery({ skip: auth.isLoggedIn === false });
```

On Componentsthe follwoing works:

```tsx
<SpotifyTrackSearchComponent
  {...(inputValue
    ? { variables: { data: { term: inputValue } } }
    : { skip: true })}
/>;
```

benjie:
It makes for simpler code;

```
const [selectedUserId, set] = useState(null);
...
<MyQuery variables={{userId: selectedUserId}} skip={!selectedUserId}>
  {(data) => ...}
</MyQuery>
```

(Also this is the pattern used in the Apollo docs.)

n1ru4l:
But wouldn't this be a type error as the variables would be incorrect (expected string got null)?

benjie:
Okay; assume it’s a value that is valid (e.g. zero) but you know won’t work.

benjie:
(Or that there’s an external trigger for it, some other state separate from variables.)

benjie:
Here’s a concrete use case: searching to see if a username has already been taken. The value of `username` will always be a string, starting with `""`. Your skip may be `!usernameIsValid || username.length < 3` since there’s no point checking until they’ve typed at least three characters.

n1ru4l:
The use case seems legit to me!

But wouldn't this change allow to omit setting the variables even if `skip` is `false`?
```jsx
const [selectedUserId, set] = useState(null);
...
<MyQuery skip={!selectedUserId}>
  {(data) => ...}
</MyQuery>
```

Issue #2210, opened 2019-07-22
[client-side-base-visitor]   Allow importing documents from an external source - making it possible to use webpack loaders


eddeee888:
**Is your feature request related to a problem? Please describe.**

Allow importing document nodes from external such as`.graphql`, `.ts` files. This makes it more flexible to import documents. If this is done, we can utilise the [webpack loader](https://www.apollographql.com/docs/react/recipes/webpack/)!

**Describe the solution you'd like**

Given the following document:

```graphql
mutation Login($input: LoginInput!) {
  login(input: $input) {
    id
    email
  }
}
```

We should be able to generate the following:

```tsx
// Login.generated.tsx
import * as Types from "../__generated__/types";
import * as operations from "./Login.graphql"; // <-- This is the line we are generating

// Other imports / types we already generate

export const LoginComponent = (props: LoginComponentProps) => {
  <ReactApollo.Mutation<LoginMutation, LoginMutationVariables> 
    mutation={operations.Login} {...props} />
}
```

Most of these types are already generated by the existing plugins/presets (Which is awesome!). 

The only thing blocking this is [here](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts#L253). This means we always parse the document string generated by stringifying the parsed document or use the `graphql-tag` to generate the document at run time.

My proposal is that we add a config option (naming could be changed): `documentSource` which could be `graphQLTag`, `documentNode` or `external`:

- `graphQLTag`: is the default option i.e. importing `graphql-tag` and parsing document at runtime using `gql`
- `documentNode`: is the equivalent of `noGraphQLTag`
- `external`: import documents from a file. This could be a `.graphql` file for the case of webpack. Documents could also come from a `.ts` or `.tsx` file. e.g. users may want to export parsed document manually. When we use setting, we simply don't include the `documentString` in the generated template.

**Usage for `external`**

`external` mode could have the following API:

- `.graphql#near-operation-file`: This means we get the documents from the graphql file next to the file. For this mode, we will have to pass the documents from params into the plugin [here](https://github.com/dotansimha/graphql-code-generator/blob/f074de1b64cb489dc02ba199f73d9b2e066ac4a1/packages/plugins/typescript/react-apollo/src/index.ts#L182)

- `path/to/file/ABC.ts`, `path/to/file/ABC.graphql`: This means we get all documents from a certain file.

**Problems**

[...]

PR #2261, opened 2019-07-31
Add option to import document nodes from an external file

eddeee888:
## Overview

This adds 2 config options to `client-side-base-visitor`:

- `documentMode`: This replaces the existing `noGraphQLTag`. This has 3 possible options: `documentNode` which works the same way as `noGraphQLTag = true`, `graphQLTag` which works the same way as `noGraphQLTag = false` and `external` which is new

- `importDocumentNodeExternallyFrom`: To be used in together with `documentNode = "external"`

Related issue: https://github.com/dotansimha/graphql-code-generator/issues/2210

## `external` mode

This can be used if we want to generate the document nodes manually and import them in. If this mode is used, we can use `importDocumentNodeExternallyFrom` as the file path to said file.

## Support for webpack loader

`importDocumentNodeExternallyFrom` also has a special option: `near-operation-file`. This option is intended to be used with `near-operation-file` preset. When they are used together, the operations will be imported from the document file. If the document file was a `.graphql`, we can use the webpack loader as documented [here](https://www.apollographql.com/docs/react/recipes/webpack/)

## Changes:

- `@graphql-codegen/visitor-plugin-common`: keep `noGraphQLTag` for backward compatibility. This doesn't get into the parsed config
- `@graphql-codegen/typescript-react-apollo`: pass `documents` into the plugin as this is needed by the base plugin to generate the import line

TODO:

- [x] Handle filename properly for `near-operation-file`. We only need the file name instead of the whole path to file
- [x] Review config option names
- [x] Check if we need to make `near-operation-file` option a bit more special? Maybe we can prefix it with something to mark that it is not just another string? e.g. `#near-operation-file`, `!near-operation-file` ?
- [x] Check other plugins to make sure there's no regression
- [x] Check if we need to make the hardcoded `Operations` customisable?
- [x] Check that webpack loader works with the `near-operation-file` option i.e. if there's only one operation in a file, do we have to `import Operation from '...'` instead of `import * as Operation from '...'`

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -6,6 +6,7 @@ import gqlTag from 'graphql-tag';
 import { toPascalCase, Types } from '@graphql-codegen/plugin-helpers';
 import { getConfigValue } from './utils';
 import { LoadedFragment } from './types';
+import { basename } from 'path';
 export interface RawClientSideBasePluginConfig extends RawConfig {
   gqlImport?: string;
   noExport?: boolean;
@@ -218,7 +219,7 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
         break;
       case 'external':
         if (this.config.importDocumentNodeExternallyFrom === 'near-operation-file' && this._documents.length === 1) {
-          imports.push(`import * as Operations from './${this._documents[0].filePath}';`);
+          imports.push(`import * as Operations from './${basename(this._documents[0].filePath)}';`);
         } else {
           imports.push(`import * as Operations from '${this.config.importDocumentNodeExternallyFrom}';`);
         }
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -206,7 +206,6 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
   }
 
   public getImports(): string[] {
-    const gqlImport = this._parseImport(this.config.gqlImport || 'graphql-tag');
     let imports = [];
 
     switch (this.config.documentMode) {
@@ -214,13 +213,13 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
         imports.push(`import { DocumentNode } from 'graphql';`);
         break;
       case 'graphQLTag':
+        const gqlImport = this._parseImport(this.config.gqlImport || 'graphql-tag');
         imports.push(`import ${gqlImport.propName ? `{ ${gqlImport.propName === 'gql' ? 'gql' : `${gqlImport.propName} as gql`} }` : 'gql'} from '${gqlImport.moduleName}';`);
         break;
       case 'external':
         if (this.config.importDocumentNodeExternallyFrom === 'near-operation-file' && this._documents.length === 1) {
           imports.push(`import * as Operations from './${this._documents[0].filePath}';`);
-        }
-        else {
+        } else {
           imports.push(`import * as Operations from '${this.config.importDocumentNodeExternallyFrom}';`);
         }
         break;
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -106,7 +106,12 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
       gqlImport: rawConfig.gqlImport || null,
       noExport: !!rawConfig.noExport,
       operationResultSuffix: getConfigValue(rawConfig.operationResultSuffix, ''),
-      documentMode: typeof rawConfig.noGraphQLTag === 'boolean' && rawConfig.noGraphQLTag === false ? 'documentMode' : getConfigValue(rawConfig.documentMode, DocumentMode.graphQLTag),
+      documentMode: ((rawConfig: RawClientSideBasePluginConfig) => {
+        if (typeof rawConfig.noGraphQLTag === 'boolean') {
+          return rawConfig.noGraphQLTag ? DocumentMode.documentNode : DocumentMode.graphQLTag;
+        }
+        return getConfigValue(rawConfig.documentMode, DocumentMode.graphQLTag);
+      })(rawConfig),
       importDocumentNodeExternallyFrom: getConfigValue(rawConfig.importDocumentNodeExternallyFrom, ''),
       ...additionalConfig,
     } as any);
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -45,13 +45,37 @@ export interface ClientSideBasePluginConfig extends ParsedConfig {
   noExport: boolean;
 
   /**
-   * TODO: eddeee - write stuff
+   * @name documentMode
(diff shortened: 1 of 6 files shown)


eddeee888:
Hello! Would love to have this reviewed please!

Some questions:

1. Are `documentMode` and `importDocumentNodeExternallyFrom` clear as settings names?
2. `importDocumentNodeExternallyFrom` just has `near-operation-file` as a string, should we make it more special? Should we have just have another setting called `importDocumentNodeExternallyFromNearOperationFile` as a `boolean`? Or should we just mark it with some special character?
3. Operations are imported into `Operations` variable. This is not customisable at the moment. I don't think we need to give it an option until someone needs to costimise it?
4. Is the way we are importing it can be used for webpack loader? I haven't used it enough to be 100% sure
5. This is not going to work for `typescript-angular-apollo` or `typescript-uqrl` as I'm not sure how they work. Can this go first and those can be implemented later?

dotansimha:
Thank you @eddeee888 ! This is a nice idea. 
Config values looks good. I think we should keep `noGraphQLTag` for backward compatibility (and if it's `true`, we can set it to `documentMode`).
It's fine to leave the near-operation-file config as-is, but make sure to document in the `@example` section.

I think the webpack graphql-loader exposes a default export when there is only one file, and a named export when there are multiple, using the `operationName`. So if the document has no operation name, we need to import it as default export, otherwise we can import it with a named import (even if there is only one).

eddeee888:
> I think we should keep noGraphQLTag for backward compatibility (and if it's true, we can set it to documentMode).

Good point! I have put this back

> I think the webpack graphql-loader exposes a default export when there is only one file, and a named export when there are multiple, using the operationName. So if the document has no operation name, we need to import it as default export, otherwise we can import it with a named import (even if there is only one).

I see! 

I tried creating an unnamed operation and it seems like we are not even creating an operation component. Codesandbox [here](https://codesandbox.io/s/gcg-14-experiments-c8l15), look for `Unnamed.operation.graphql`

In that case, I think we don't have to import the document since there's no component to use the documents? 🤔 

If that's the case, I think this PR can be merged - unless there's something else you want changed :) ?

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -63,21 +71,61 @@ export interface ClientSideBasePluginConfig extends ParsedConfig {
...
 
 export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginConfig = RawClientSideBasePluginConfig, TPluginConfig extends ClientSideBasePluginConfig = ClientSideBasePluginConfig> extends BaseVisitor<TRawConfig, TPluginConfig> {
   protected _collectedOperations: OperationDefinitionNode[] = [];
+  protected _documents: Types.DocumentFile[] = [];
 
-  constructor(protected _fragments: LoadedFragment[], rawConfig: TRawConfig, additionalConfig: Partial<TPluginConfig>) {
+  constructor(protected _fragments: LoadedFragment[], rawConfig: TRawConfig, additionalConfig: Partial<TPluginConfig>, documents?: Types.DocumentFile[]) {
     super(rawConfig, {
       dedupeOperationSuffix: getConfigValue(rawConfig.dedupeOperationSuffix, false),
       noGraphQLTag: getConfigValue(rawConfig.noGraphQLTag, false),
```
dotansimha: I think you can remove `noGraphQLTag` from the parsed config, and just do? 
```
documentMode: typeof rawConfig.noGraphQLTag === 'boolean' && typeof rawConfig.noGraphQLTag === false ? 'documentMode' : getConfigValue(rawConfig.documentMode, 'graphQLTag'),
importDocumentNodeExternallyFrom: getConfigValue(rawConfig.importDocumentNodeExternallyFrom, ''),
```

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -109,7 +157,7 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
 
   protected _includeFragments(fragments: string[]): string {
     if (fragments && fragments.length > 0) {
-      if (this.config.noGraphQLTag) {
+      if (this._getDocumentMode() === 'documentNode') {
```
dotansimha: Maybe add all available options of `documentMode` as TypeScript enum?

dotansimha:
Thanks @eddeee888 
I added minor notes, and I think we can merge it once it's fixed and CI passes.

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
eddeee888: Interesting! I've always wondered why we had raw and parsed config. I have fixed this in https://github.com/dotansimha/graphql-code-generator/pull/2261/commits/42ffe06eb4adaa14050a1a5590ecfa2437b5cdbf

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
eddeee888: I was using union type - I have changed this to enum in https://github.com/dotansimha/graphql-code-generator/pull/2261/commits/85ed7654b6d49c85053c09642ecbf7ebb3eae3f0

eddeee888:
@dotansimha Build is green finally :)

However, it looks like this merge may remove the `dedupeOperationSuffix` implementation...? e.g. https://github.com/dotansimha/graphql-code-generator/pull/2261/files#diff-8416dbe3f0939d02fe01bf885bcf3a8cL59-L62

I wonder if it's because I've forced push my fork a few times...? I haven't forked many codebases before so I'm not 100% sure what effects it would have on the master codebase in this case 😅

dotansimha:
not your force push, but the rebase caused it. Can you please fix that? and then I'll merge :) 
Thanks!

eddeee888:
Ah of course! Also, I'm supposed to rebase to your master, not my master. Monday brainfarts are bad!

Anyways it's ready to go please have a 👀 . Thanks for helping me get this over the line!

dotansimha:
Perfect! Thank you @eddeee888 !

qiwang97:
@eddeee888 
i use this config

```yml
  src/services/generated/schema.ts:
    plugins:
      - typescript
    config:
      skipTypename: true
  src/services/generated/operations.ts:
    preset: import-types
    presetConfig:
      typesPath: ./schema
    plugins:
      - typescript-operations
    config:
      skipTypename: true
  src/services/generated/request.ts:
    plugins:
      - typescript-graphql-request
    config:
      documentMode: external
      importDocumentNodeExternallyFrom: ./operations
```

But `importDocumentNodeExternallyFrom` doesnot work. The `Operations` prefix is lost like this:

```
import { GraphQLClient } from 'graphql-request'
import { print } from 'graphql'
import * as Operations from './operations'

export type SdkFunctionWrapper = <T>(action: () => Promise<T>) => Promise<T>

const defaultWrapper: SdkFunctionWrapper = sdkFunction => sdkFunction()
export function getSdk(client: GraphQLClient, withWrapper: SdkFunctionWrapper = defaultWrapper) {
  return {
    Upload(variables: UploadMutationVariables): Promise<UploadMutation> {
      return withWrapper(() => client.request<UploadMutation>(print(UploadDocument), variables))
    },
  }
}
```
`UploadMutationVariables` should be `Operations.UploadMutationVariables`.
Anything to do to solve it?

eddeee888:
Hello @qiwang97! The `importDocumentNodeExternallyFrom` setting is part of the common visitor plugin. The consuming plugin ( in this case `typescript-graphql-request` ) needs to handle the variable usage by itself I think.

For example, this is how the `react-apollo` plugin does it:  https://github.com/dotansimha/graphql-code-generator/pull/2261/files#diff-ce84bc42a6b0c5a7f856c37781eb4e65R131

Issue #2302, opened 2019-08-07
Add support for useLazyQuery


immortalx:
@apollo/react-hooks added useLazyQuery
https://www.apollographql.com/docs/react/api/react-hooks/#uselazyquery

It's possible to support it?

Adding a configuration like:
withLazyQuery: true

And generating a useQueryLazyComponent for each query, i wonder if it's the best way to do it.

n1ru4l:
Maybe it would be a better solution to allow controlling what kind of code you want to generate for which query/mutation? E.g. by using directives on the operation definition?

I may not want to generate a lazy query component for each query in my project? Or is this something that should be handled by the bundler aka tree shaking?

immortalx:
A similar scenario already happens if you use hooks and hoc/component, in the end it/should be handled by the bundler.

ardatan:
That would be nice! I'll add this soon.

ardatan:
You can try it in the following canary version;
```
 1.5.1-alpha-e4e10196.6+e4e10196
```

immortalx:
Wow, that was fast. Seems to be working fine for me!
Thank you @ardatan

dotansimha:
Added in 1.6.0.

DarkLite1:
It would be great if useLazyQuery could also be added to the Vue plugin #5258

ejkkan:
Hey!
So is lazy queries supported?
I've been trying to figure out how and browsing docs but without success yet.
What an I missing?
Does anyone have a code sample?

n1ru4l:
@ejkkan Ir is not supported in the apollo react hooks plugin. However, you can either use the `client-preset` or `typed-document-node` plugin for generating typed document nodes that you can then use with `useLazyQuery` from apollo client.

PR #2309, opened 2019-08-07
Generate lazy query hooks on react-apollo plugin

ardatan:
Related #2302


packages/plugins/typescript/react-apollo/tests/react-apollo.spec.ts
```diff
@@ -997,6 +997,39 @@ export function useListenToCommentsSubscription(baseOptions?: ApolloReactHooks.S
...
+        schema,
+        docs,
+        { withHooks: true, withComponent: false, withHOC: false },
+        {
+          outputFile: 'graphql.tsx',
+        }
+      )) as Types.ComplexPluginOutput;
+
+      expect(content.content).toBeSimilarStringTo(`
+  export function useFeedLazyQuery(baseOptions?: ApolloReactHooks.LazyQueryHookOptions<FeedQuery, FeedQueryVariables>) {
```
koenpunt: Wouldn't a name like `useLazyFeedQuery` be more correct?

dotansimha: We decided to keep it as-is for now, I think both forms are correct

PR #2331, opened 2019-08-10
[typescript-react-apollo] Add support for importing document node from external file for hooks and HOCs

eddeee888:
This PR adds support for importing document node from an external file for hooks i.e. `useQuery`, `useLazyQuery`, `useMutation`, `useSubscription`. 

The current implementation only works with `withComponent` config. As the result, hook document nodes are not used correctly. This can be seen here: https://codesandbox.io/s/gcg-15-experiments-c8l15 (Look for `src/GetProject.operation.generated.tsx`)

### DIFF
--- a/packages/plugins/typescript/react-apollo/src/visitor.ts
+++ b/packages/plugins/typescript/react-apollo/src/visitor.ts
@@ -68,6 +68,10 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
     return OMIT_TYPE;
   }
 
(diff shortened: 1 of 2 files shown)


dotansimha:
@eddeee888 thank you!!!
we can apply it to all available options? (hooks/components/hocs) ?

eddeee888:
You are right @dotansimha ! I never used `withHOC` so I forgot we could do that 😬 Thanks for reminding me!

I have added that and also updated the tests to show explicitly which config (i.e. `withHOC`, `withHooks` or `withComponent`) we are trying to test in each

dotansimha:
Thank you @eddeee888 ! Looks great.

Issue #2165, opened 2019-07-13
[ts-react-apollo] Types should able to import from other file when it is the only plugin used


veeramarni:
**Is your feature request related to a problem? Please describe.**

If we need to generate an independent file with only `typescript-react-apollo` plugin, it needs types generated from other plugins to be imported externally. 

**Describe the solution you'd like**

For example, if we have an option like `typesImport: @external-file` in `typescript-react-apollo` plugin

```codegen.yml
generates:
  packages/sample-client-state/src/interfaces/generated-models.ts:
    documents: "packages/sample-client-state/src/**/*.gql"
    schema: 
      - ./packages/sample-client-state/src/**/*.graphql
    config:
      enumValues: ./enum
      constEnums: true
      skipTypename: true
      noNamespaces: true
      contextType: ./context#MyContext
      typesPrefix: I
      withMutationFn: false
      withHOC: false
      withComponent: false
    plugins:
      - add: |
          /* tslint:disable */
      - typescript
      - typescript-operations
      - typescript-resolvers
      - typescript-react-apollo
  packages/sample-client-react/src/generated/models.tsx:
    documents: "packages/sample-client-state/src/**/*.gql"
    schema: 
      - ./packages/sample-client-state/src/**/*.graphql
    config:
        typesPrefix: I
        typesImport: @external-file
        withMutationFn: true
        withHOC: true
        withComponent: true
        noGraphQLTag: true
    plugins:
      - typescript-react-apollo
```

generated `.ts` file
```
.....
export type ISelectFileMutationVariables = {
  content?: Maybe<ISelectEditor_Input>;
};

export type ISelectFileMutation = Pick<IMutation, "selectFile">;
....
```

generated `.tsx` file
```
import { ISelectFileMutation, ISelectFileMutationVariables } from '@external-file'
....
export type ISelectFileMutationFn = ReactApollo.MutationFn<
  ISelectFileMutation,
  ISelectFileMutationVariables
>;
export type SelectFileComponentProps = Omit<
  ReactApollo.MutationProps<ISelectFileMutation, ISelectFileMutationVariables>,
  "mutation"
>;

export const SelectFileComponent = (props: SelectFileComponentProps) => (
  <ReactApollo.Mutation<ISelectFileMutation, ISelectFileMutationVariables>
    mutation={SelectFileDocument}
    {...props}
  />
);

export type ISelectFileProps<TChildProps = {}> = Partial<
[...]

PR #2344, opened 2019-08-12
Prefix apollo react

veeramarni:
Inorder the preset to work for `typescript-react-apollo`. However, this may not be the best way and it would be better the core `visitor` class can handle `preset` configuration so that all `plugins` uniformly get the same `prefix`. Relates to #2165


veeramarni:
@dotansimha Wondering if you got chance to review this PR and other one #2344

packages/plugins/typescript/react-apollo/src/visitor.ts
```diff
@@ -80,8 +83,8 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
   }
 
   private _buildHocProps(operationName: string, operationType: string): string {
-    const typeVariableName = this.convertName(operationName + toPascalCase(operationType) + this._parsedConfig.operationResultSuffix);
-    const variablesVarName = this.convertName(operationName + toPascalCase(operationType) + 'Variables');
+    const typeVariableName = this._prefix + this.convertName(operationName + toPascalCase(operationType) + this._parsedConfig.operationResultSuffix);
```
dotansimha: `convertName` accepts a second argument of options, which contains `prefix` field. I think it will be a better way to add prefix.

packages/plugins/typescript/react-apollo/src/visitor.ts
```diff
@@ -218,6 +221,8 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
   }
 
   protected buildOperation(node: OperationDefinitionNode, documentVariableName: string, operationType: string, operationResultType: string, operationVariablesTypes: string): string {
+    operationResultType = `${this._prefix}${operationResultType}`;
```
dotansimha: `buildOperation` should get the built `operationResultType`, and not build it.

dotansimha:
@veeramarni please see my comments. Thanks!

veeramarni:
@dotansimha Will update the PR shortly.

packages/plugins/typescript/react-apollo/src/visitor.ts
veeramarni: I have tried passing some like 

```
    const typeVariableName = this.convertName(operationName + toPascalCase(operationType) + this._parsedConfig.operationResultSuffix, { prefix: this._prefix });
    const variablesVarName = this.convertName(operationName + toPascalCase(operationType) + 'Variables', { prefix: this._prefix });
```
but the result was

`ISchemaTypestestStringMutation` instead of `SchemaTypes.TestStringMutation`

packages/plugins/typescript/react-apollo/src/visitor.ts
veeramarni: @dotansimha should we add the `prefix` here so it will be converted prior to sending as arguments.
https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts#L340

veeramarni:
@dotansimha Can you please respond to my comments?

dotansimha:
Hi @veeramarni ! Thank you for the answer. Can you please just rebase and resolve the conflicts?

PR #2360, opened 2019-08-13
[typescript-react-apollo] Fix problem with unnecessary fragment imports for external mode

eddeee888:
Normally, fragments are turned into document nodes so we can include them in the operations document nodes they are used in. In the case of `documentMode = 'external'`, the operation document nodes are created externally and imported into the generated files. Therefore, there's no need to print fragment variables in the generated file anymore

This problem can be seen in this sandbox: https://codesandbox.io/s/gcg-15-experiments-c8l15. There are fragments in `webpack/GetProject_Status.fragment.graphql` and `webpack/CreateProject.operation.graphql`, and these will be converted into variables with suffix `FragmentDoc` in generated files (Note that we are not even importing the `graphql-tag` module).

This PR ensures that no fragments are generated when `external` mode is used. 

NOTE: I'm having problem running `validateTypeScript` in the new tests. I've tried looking at the implementation of `tsDocumentsPlugin` and still not able to figure out what it's supposed to do. Would love some pointer please :) 

TODO:

- [ ] Check why `validateTypeScript` does not work

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -250,7 +250,7 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
   }
 
   public get fragments(): string {
-    if (this._fragments.length === 0) {
+    if (this._fragments.length === 0 || this.config.documentMode === DocumentMode.external) {
       return '';
     }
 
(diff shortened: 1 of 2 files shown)


dotansimha:
Hi @eddeee888 ! 
This PR looks good, can you please explain the issue with `validateTypeScript`? Do you get an error?

eddeee888:
Hi @dotansimha , thanks for explaining how it works! Sooo... It was because I was using the wrong type for the fragments in the tests! 🤦‍♂

PR #2375, opened 2019-08-15
fix: properly dedupe react hooks names

anderson-amorim:
Right now react hook's names are not being deduped even if the option `dedupeOperationSuffix` is enabled. This PR fixes it.


ardatan:
Thank you @anderson-amorim !

PR #2376, opened 2019-08-15
docs: hint how to use generated resolvers directly

stoically:
Thanks for the awesome work on graphql-codegen!

Added just a short hint in the docs on how to use generated resolvers directly since I had to do some substantial digging to find that bit of information. If you don't find that useful or it's already in the docs and I just missed it then feel free to close the PR.


dotansimha:
Thanks @stoically !

ardatan:
There was a typo :)
> Generated resolvers can be passed directly into [graph-tools]

stoically:
@dotansimha Thanks for merging!

@ardatan oups, introducing typos in only one sentence, that's something. Thanks for pointing it out, opened #2377

PR #2392, opened 2019-08-17
[typescript-react-apollo] Fix external mode to avoid importing operations in files that don't have operations

eddeee888:
Currently, in external mode, we are importing nodes even for files without operations i.e. files with just fragments. This behaviour can be seen here: https://codesandbox.io/s/gcg-16-experiments-c8l15 in `src/webpack/GetProject_Id.fragment.generated.tsx`

This PR adds a check to make sure we don't import operations unnecessarily


dotansimha:
Thanks @eddeee888 !

PR #2408, opened 2019-08-20
[fix] hook functions don't need a ; at the end

lookfirst:
This change hopefully breaks tests. ;-)


dotansimha:
Thank @lookfirst , merged and fixed tests.

PR #2418, opened 2019-08-21
 feat: allow using file paths for `scalars:` config like you can for enumValues

ForbesLindesay:
This allows you to do:

```yaml
config:
  scalars:
    DateTime: Date
    JSON: { [key: string]: any }
    CustomType: ../scalars#CustomType
```

or even:

```yaml
config:
  scalars: ../scalars
```


dotansimha:
That's great! Thank you @ForbesLindesay ! 
Those changes effects some parts of the internal core, and effects almost all plugins. Can you please add some unit tests to make sure those new forms of scalars are generated correctly? 

Thanks!

ForbesLindesay:
It took me a pretty long time to make this change relative to what I expected. I don't think I'm going to find time to add tests any time soon.

dotansimha:
All good @ForbesLindesay , I'll add some tests later today. Thank you!

shiatsumat:
When I use the `scalars: [module_path]` notation (on ver. 1.7.0), the generator reports `Cannot use string scalars mapping when building without a schema`;
since I have actually specified the schema and `enumValues: [module_path]` works fine, probably it is because the custom scalar enumerator for `scalars: [module_path]` notation is not working right.

dotansimha:
@shiatsumat can you please share your config file or a reproduction of the issue?

shiatsumat:
Thank you for reply. My `.codegen.yml` is as follows.

```yml
schema: ./graphql/schema.graphql
documents: ./graphql/usage/*.graphql
generates:
  ./graphql/_gen_/resolve.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      preResolveTypes: true
  ./graphql/_gen_/api.ts:
    plugins:
      - typescript
      - typescript-operations
      - typescript-react-apollo
    config:
      preResolveTypes: true
      withComponent: false
      withHOC: false
      withHooks: true
config:
  scalars: ../scalars
  enumValues: ../enums
```

The `graphql` folder consists of the following files.
```
graphql
├── _gen_
│   ├── api.ts
│   └── resolve.ts
├── scalars.ts
├── schema.graphql
├── usage
│   └── usage.graphql
└── values.ts
```

shiatsumat:
@dotansimha Any ideas...?

ForbesLindesay:
This is because the relevant code doesn't have access to the schema, and this PR was getting huge already. You'll need to add a PR to ensure that the schema is actually passed into the relevant function where that exception is thrown.

shiatsumat:
@ForbesLindesay You are absolutely right. I myself cannot find the way to fix the code now though...

dotansimha:
I'll take a look, thanks @ForbesLindesay @shiatsumat

dotansimha:
@shiatsumat @ForbesLindesay 
I think I managed to fix this issue, by making sure to always pass the schema to `buildScalars` method. (https://github.com/dotansimha/graphql-code-generator/pull/2600)
Can you please try the alpha version? `1.7.1-alpha-8ed4b450.74`

shiatsumat:
@dotansimha Wow great, it works!!

dchambers:
Does anyone know what the contents of `scalars.ts` needs to look like for this to work?

dchambers:
In case anyone else comes this way looking for the same answer, it turned out that `scalars.ts` needed to look like this:

```ts
export type PositiveInt = number
export type CountryCode = string
export type PostalCode = string
// etc...
```

GraphQL Code Generator doesn't actually process the file itself however; it merely refers to it in the generated code it creates, with type import statements like this:

```ts
import { PositiveInt } from '../scalar-types'
import { CountryCode } from '../scalar-types'
import { PostalCode } from '../scalar-types'
// etc...
```

In my case, types of all of my scalars were defined using the `extensions.codegenScalarType` pattern that `graphql-scalars` adheres to, so I wrote the following generation script (`generate-graphql-scalars.ts`) so I could use that config rather than redundantly redefine everything:

```ts
import { writeFileSync } from 'fs'
import { scalarResolvers } from './src/scalar-resolvers' // NOTE: these are my scalar resolvers as an object, ready to spread into my resolver config within Apollo server

const getKeys = Object.keys as <T extends object>(obj: T) => Array<keyof T>

const generateTypes = getKeys(scalarResolvers)
    .map((name) => `export type ${name} = ${scalarResolvers[name].extensions.codegenScalarType}`)
    .join('\n')

writeFileSync('./src/scalar-types.gen.ts', generateTypes)
```

The `prepack` for my library now has the following steps:

1. Generate the scalar types using `generate-graphql-scalars.ts`.
2. Generate the schema types using `codegen`.
3. Generate the transpiled code using `tsc`.

Since `generate-graphql-scalars.ts` imports TS from my `src` directory that can not yet be built (it depends on the generated types) I use `babel-node` to run that TypeScript, as follows:

```
babel-node --presets @babel/preset-typescript --plugins @babel/plugin-transform-modules-commonjs --extensions '.ts' ./generate-graphql-scalars.ts
```

Hope that helps someone!

Urigo:
@dchambers do you think we could add something to the docs to make it easier for people?
or even a change in the API?

dchambers:
> @dchambers do you think we could add something to the docs to make it easier for people? or even a change in the API?

Hi @Urigo 👋 . Yes, absolutely. The current documentation for `scalars` is as follows:

image

whereas I would ideally hope to see something more like this:

# scalars

type: `ScalarsMap`

Extends or overrides the built-in scalars and custom GraphQL scalars to a custom type.

## Usage Examples

### Inline Definitions

```yml
config:
  scalars:
    DateTime: Date
    JSON: { [key: string]: any }
    CustomType: ../scalars#CustomType
```

where `scalars.ts` contains:

```ts
export type CustomType = string
```

### External Definitions

```yml
config:
  scalars: ../scalars.ts
```

where `scalars.ts` contains:

```ts
export type DateTime = Date
export type JSON = { [key: string]: any }
export type CustomType = string
```

dchambers:
In regards to the potential for API changes, I believe the ideal outcome would be support for pointing directly at `graphql-scalar` conformant scalar resolvers, as an alternative means by which scalar types can be retrieved -- particularly since `graphql-scalar` is now also maintained by The Guild.

For example, the 'External Definitions' section in the example above could be sub-divided into two parts:

### External Type Definitions

```yml
config:
  scalars: ../scalars.ts
```

where `scalars.ts` contains:

```ts
export type DateTime = Date
export type JSON = { [key: string]: any }
export type CustomType = string
```

### External Scalar Resolver Definitions

```yml
config:
  scalars: ../scalar-resolvers.ts
```

where `scalar-resolvers.ts` contains:

```ts
import { RegularExpression } from 'graphql-scalars'
export { DateTime, JSON } from 'graphql-scalars'

export const CustomType = new RegularExpression('CustomType', /^ABC$/);
```

Urigo:
@dchambers that is super helpful, thank you!
Do you want to maybe try to help us work on these things?
We can support and guide you as you make your contributions with anything you might need!

PR #2430, opened 2019-08-23
Make GraphQL operation alias unique from HOC name

uhLeeshUh:
- The use of the same name for an HOC operation and its alias (i.e. `withCreateAppointment`) causes a shadowed variable problem when trying to call the alias function from props and requires an additional alias to be assigned under the `name` property of an operation like below:

```
export default withCreateAppointment({
  name: 'createAppointment',
})(MyComponent);
```
- This PR changes HOC aliases to the camel-cased version of their operation name to make them usable as intended


dotansimha:
That's correct. Thank you for this PR, @uhLeeshUh !

PR #2441, opened 2019-08-25
fix: --watch flag can watch schema children.

tonyfromundefined:
- Fixed an issue with watch flags not working properly in child schemas when using `import-graphql` syntax inside a schema.


dotansimha:
Hi @tonyfromundefined 
Thank you for this PR. Unit tests for this change are missing, and they are important because it might cause issues (because `graphql-import` might throw an error)

dotansimha:
Closing for now. Thanks @tonyfromundefined. If you wish to have this PR added, please add some tests to it and response to the latest comment.

If someone needs to workaround this, you can set an array of paths to `watch:` flag and it will watch additional files.

PR #2442, opened 2019-08-25
Fix importDocumentNodeExternallyFrom on CRA TypeScript

cometkim:
:wave: 

I have trouble when using graphql-codegen with CRA.

I tried to use `importDocumentNodeExternallyFrom: near-operation-file` with in-source-code documents becuase I don't wanna separate document file(`.graphql`) from component code.

in my config:
```yaml
documents:
  - src/**/*.{ts,tsx}
  - '!src/**/*.generated.{ts,tsx}'

generates
  ./src/generated/types.ts:
    plugins:
      - typescript
      - typescript-resolvers

  ./src/:
    preset: near-operation-file
    presetConfig:
      extension: .generated.tsx
      baseTypesPath: generated/types.ts
    plugins:
      - typescript-operations
      - typescript-react-apollo
    config:
      gqlImport: graphql.macro#gql
      documentMode: external
      reactApolloVersion: 3
      withComponent: false
      withHOC: false
      withHooks: true
```

But I got an error: `An import path cannot end with a '.tsx' extension`

image

I couldn't find a way to suppress/ignore the specific line without eject CRA or this change. CRA is force my tsconfig.json compilerOptions.moduleResolution to be `node`.

So...

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -285,7 +285,7 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
         if (this.config.importDocumentNodeExternallyFrom === 'near-operation-file' && this._documents.length === 1) {
           imports.push(`import * as Operations from './${basename(this._documents[0].filePath)}';`);
         } else {
-          imports.push(`import * as Operations from '${this.config.importDocumentNodeExternallyFrom}';`);
+          imports.push(`import * as Operations from '${this.config.importDocumentNodeExternallyFrom.replace(/\.tsx?$/, '')}';`);
         }
         break;
       default:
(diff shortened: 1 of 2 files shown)


dotansimha:
Thank you @cometkim!
Does the extension should be removed anyway? why is it adding `.tsx` there?

cometkim:
Hi @dotansimha 

> why is it adding .tsx there?

Because of my source code (`App.tsx`) is.

I don't wanna add `.graphql` files separately, or copy queries into the generated files witch is already defined once in my source code. 

the `documentMode: external` option is what exactly I want.

> Does the extension should be removed anyway?

Yes, only for TS modules(`.ts`, `.tsx`). Some TS projects, including CRA, use `compilerOptions.moduleResolution: node` that doesn't allow extensions when importing node modules.

dotansimha:
I see, thanks @cometkim ! 
I think `.js`/`.jsx` should also be removed from paths. 
I merged your PR manually because I added support for more extensions. 

Thank you!

PR #2448, opened 2019-08-27
feat(type-graphql): allow overriding decorators

NickBolles:
I'm looking to generate interface type objects with type-graphql so that I can implement them while simultaneously extending another type

I just threw this together, it might not be the best way to achieve this though.


dotansimha:
Thanks @NickBolles !

PR #2483, opened 2019-09-02
Use The Guild's contact api

kamilkisiela:


dotansimha:
@kamilkisiela can you please also remove the lambda from the website and it's dependencies?

kamilkisiela:
@dotansimha no lambda

PR #2487, opened 2019-09-03
update doc to use correct url for graphql.js types

pmbanugo:
This points the URL to DocumentNode and GraphQLSchema API to the correct URL. They moved the types to graphql-js repo.

I checked and didn't see if there's any guideline for prefixing commits or PR titles so I hope this works :D


dotansimha:
Perfect! Thanks!

Issue #2489, opened 2019-09-03
Typescript generated with errors


PedroLamas:
**Describe the bug**
With the latest versions, I am getting some errors in my generated typescript files.

Notice the final `|  |  |  |  |  |  |  |` on the following generated code:

```javascript
export type PostMdxFragment = (
  Pick<Mdx, 'id' | 'body' | 'excerpt'>
  & { fields: Maybe<Pick<MdxFields, 'slug'>>, frontmatter: Maybe<(
    Pick<MdxFrontmatter, 'categories' | 'date' | 'dateFormatted' | 'last_modified_at' | 'layout' | 'tags' | 'title'>
    & { image: Maybe<(
      Pick<File, 'publicURL'>
      & { childImageSharp: Maybe<{ fluid: Maybe<Pick<ImageSharpFluid, 'presentationWidth'>
          & GatsbyImageSharpFluid_WithWebpFragment
        > }> }
    )> }
  )>, file: Maybe<Pick<File, 'base'> |  |  |  |  |  |  |  | > }
);
```

**To Reproduce**

1. clone/download this repo: https://github.com/PedroLamas/pedrolamas.com
2. run `yarn` to install dependencies
3. run `yarn develop` to start
4. on a second terminal, run ` yarn codegen`
5. check the generated "src\utils\graphql.generated.ts" file, and look for the bit in the example above.

1. My GraphQL schema:

Mostly inferred from my typescript code, but after running `yarn develop` from the steps above, you can access it via http://localhost:8000/__graphql

2. My GraphQL operations:

(Inferred from typescript)

3. My `codegen.yml` config file:

```yml
overwrite: true
schema: 'http://localhost:8000/___graphql'
documents:
  - ./src/**/*.{ts,tsx}
  - ./node_modules/gatsby-*/**/*.js
generates:
  src/utils/graphql.generated.ts:
    config:
      skipTypename: true
      enumsAsTypes: true
    plugins:
      - 'typescript'
      - 'typescript-operations'
  #     - "typescript-react-apollo"
  # ./graphql.schema.json:
  #   plugins:
  #     - "introspection"
```

**Expected behavior**

Generate typescript code without errors!

**Environment:**

- OS: Windows 10 version 1903 (build 18362.10015)
- `@graphql-codegen/...`: 1.6.1
- NodeJS: 10.15.3

n1ru4l:
This should be fixed by https://github.com/dotansimha/graphql-code-generator/pull/2490

[...]

PR #2490, opened 2019-09-03
fix(visitor-plugin-common): also include empty selection set types/generate valid type output

n1ru4l:
This fixes the issue occurring in #2489


n1ru4l:
There is an issue with flow types that must be fixed: https://github.com/dotansimha/graphql-code-generator/pull/2490/commits/d49dbbab7c05eb2723ca8c40ad75509f538ab6f2

I am on vacation for the next two weeks so someone else could pick this up and finish it. Typescript part is fine.

dotansimha:
Thanks @n1ru4l ! We'll take a look soon :)

dotansimha:
Thanks @n1ru4l , I'm doing some more fixes for the selection set transformation, also changed this current fix to output more clear code (without empty objects). 
I'll push it to master when it's done. 

Thanks!!

PR #2508, opened 2019-09-06
Federation in all plugins

kamilkisiela:
ardatan/graphql-toolkit#233


dotansimha:
@kamilkisiela maybe we can add something like? 

```
schema:
   - my-schema.graphql
   - federation
```

And it will add all required schema types?

kamilkisiela:
@dotansimha by `- federation` you mean a string that equals `federation` or a `federation.graphql`?

I think it should be inside the config because otherwise how do you know plugins should generate something federation-specific?

I'm not sure what's bad in putting federation in common lib. There's a shared logic between multiple plugins so it is common.

dotansimha:
@kamilkisiela ci :)

Issue #2495, opened 2019-09-04
TypeScript: Fragment + Union gives duplicated (wrong) __typename


whtsky:
**To Reproduce**
https://codesandbox.io/s/graphql-codegen-issue-template-ttfo0

1. My GraphQL schema:

```graphql
  scalar IPV4
  scalar IPV6
  type IPV4Route {
    address: IPV4
    gateway: IPV4
  }
  type IPV6Route {
    address: IPV6
    gateway: IPV6
  }
  union RouteUnion = IPV4Route | IPV6Route
  type Query {
    routes: [RouteUnion!]!
  }
```

2. My GraphQL operations:

```graphql

  fragment NetRoute on RouteUnion {
    __typename
    ... on IPV4Route {
      ipv4Address: address
      ipv4Gateway: gateway
    }
    ... on IPV6Route {
      ipv6Address: address
      ipv6Gateway: gateway
    }
  }
  query QQ {
    routes {
      ...NetRoute
    }
  }
```

3. Output:

```typescript
type NetRouteFragment = (
  { __typename: 'IPV4Route' }
  & { ipv4Address: Ipv4Route['address'], ipv4Gateway: Ipv4Route['gateway'] }
) | (
  { __typename: 'IPV6Route' }
  & { ipv6Address: Ipv6Route['address'], ipv6Gateway: Ipv6Route['gateway'] }
);

type QqQuery = (
  { __typename?: 'Query' }
  & { routes: Array<({ __typename?: 'IPV4Route' } | { __typename?: 'IPV6Route' })
    & NetRouteFragment
  > }
);
```

**Expected output**

```typescript
type NetRouteFragment = (
  { __typename: 'IPV4Route' }
  & { ipv4Address: Ipv4Route['address'], ipv4Gateway: Ipv4Route['gateway'] }
) | (
  { __typename: 'IPV6Route' }
  & { ipv6Address: Ipv6Route['address'], ipv6Gateway: Ipv6Route['gateway'] }
);

type QqQuery = (
  { __typename?: 'Query' }
  & { routes: Array<NetRouteFragment> }
);
```

**Environment:**

- `@graphql-codegen/typescript-operations`: 1.7.0

**Additional context**

dotansimha:
Hi @whtsky ! 

Can you please explain why is that an issue? 

[...]

PR #2540, opened 2019-09-12
[TypeScript Operations] try to fix issues with incorrect __typename and fragments

dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/2495


n1ru4l:
Yeah I think this is probably the route we must go :/ Unless there is an option to deep merge typescript types

dotansimha:
@n1ru4l 
I tried to use Relay's transformations (similar to what you did in your plugin), and it seems like we are getting a flatten selection sets without fragments at all, so we can build a complete set of fields per each type, under each level of the selection set. 
The latest code seems to work fine, I'm getting:

```ts
    export type UserQueryQuery = (
      { __typename?: 'Query' }
      & { user: (
        { __typename?: 'User' }
        & Pick<User, 'id' | 'login'>
      ) | (
        { __typename?: 'Error2' }
        & Pick<Error2, 'message'>
      ) | (
        { __typename?: 'Error3' }
        & Pick<Error3, 'message'>
        & { info: Maybe<(
          { __typename?: 'AdditionalInfo' }
          & Pick<AdditionalInfo, 'message2' | 'message'>
        )> }
      ) }
    );
```

From this selection set: https://github.com/dotansimha/graphql-code-generator/blob/aad87d83efb8b52bde286978aea765944abf67f4/packages/plugins/typescript/operations/tests/ts-documents.spec.ts#L2761 

Which is great because everything it merged and the typescript type represents exactly the return value of the query. 

The only issue is, that it will make it a bit more difficult to use, because the type we generate for the fragment won't be in use by the query type.

@n1ru4l @ardatan @kamilkisiela what do you think?

dotansimha:
I tried to create sub-types, it makes sub-types per each fragment spread according to the possible types, and then it uses the subtypes in the query that uses it. Still need to do some more fixes. 

The output it a bit more messy, but it's backward compatible so maybe we can go with that.

n1ru4l:
Using relay compiler will definitely break stuff for some users, e.g. they will have to add declarations for all custom fragments to their schema files and people that might use persisted queries will have more optimized queries in their production code that are not registered on the server.

I am not totally against this approach (as I am already using relay-compiler and therefore do not experience these issues 😂). However I think that that should definitely be a breaking change that should be well documented and communicated towards the users of GraphQL Code Generator.

Regarding the approach I suggested:

https://www.typescriptlang.org/play/#code/C4TwDgpgBAglC8UDeBYAUFTUCGAuKAzsAE4CWAdgOYA06WUARvqhvVnoSRTXW5k5zJVemAL7pxadKEhQAQgmQic+IkJ6t+zZfQ5rutTWwDGqrsKOTJ08NADCiuADJ56dMYD25IlABmHj0UAClMoOwBKBAA+JSNjADoGeOxlBKTjVMTEiSA

Seems like that should work

n1ru4l:
Generating types for the fragments always makes sense as you might want to use a fragment type for props on a component. For now let’s implement the sub-type approach first.

Also, should we export the subtypes or only the Fragment type?

dotansimha:
@n1ru4l I think I'll keep the local as default, but allow to export them.

dotansimha:
@n1ru4l I fixes all of the issues :) And fixes all tests, and it grouped and generates fragment spreads correctly. 
Also updates some of the tests with usages and compile it with `tsc`. 

TODO:
- [ ] Change the default export of fragment sub-types to `false`
- [ ] Fix this: https://github.com/dotansimha/graphql-code-generator/issues/2566 
- [ ] Add `relay-compiler` with `optimizeSelectionSet: true` config.

n1ru4l:
Instead of using relay-compiler we could also add fragments as inline fragments and the existing algorithm should generate the flat types

dotansimha:
@n1ru4l Interesting, your are right, this could work. But are there other transformations in Relay compiler that might be relevant?

n1ru4l:
Also correct, Relay will transform the operations, the approach I mentioned will only affect types!

n1ru4l:
Wouldn't it make more sense to provide the relay-compiler optimizations via a plugin vs an option? The option could imply that it "simply works", however, in reality, you will have some limitations in how you use GraphQL e.g. every operation MUST have a name, you MUST add directive definitions to your schema for all non-standard directive types (e.q. `@client`, `@connection`). There might be even more!

@dotansimha If you want we can move my plugin over to this repository in order to have it all colocated.

dotansimha:
@n1ru4l The idea behind add an option for Relay optimization is to allow developers to write their operations as they used to, but optimize it while generating in order to create better TypeScript types (and not to optimize the operation before running it).  

And yeah, I think you can move it to this repo, it will also make sure that we won't break it :D

n1ru4l:
> But are there other transformations in Relay compiler that might be relevant?

I do not think there is anything further provided that would benefit the generated types.

I would say let's try the convert fragments to inline fragments (for type generation) when `optimizeSelectionSet` is set to `true`.

Maybe we should even change the wording of this? I think it is a bit misleading.

`optimizeSelectionSet` makes me assume that it also alters the selection sets of the operations.

Might something like `flattenGeneratedTypes` be more fitting?

Will try to create a pull request for adding relay-optimizer-plugin this week!

dotansimha:
I added `flattenGeneratedTypes`. it's using Relay compiler for now (it also includes only the operations, without the fragments, because they are not needed after flattening). 

Also verified that everything works with compatibility, so we are good to go with this PR. 

:)

n1ru4l:
Try to add a test with `@client` and `@connection` directive on an operation.

dotansimha:
@n1ru4l Is there a reason for this not to work with those directives? 
I tried and it seems fine, it's being treated just like any other field.

n1ru4l:
It will fail in case you do not add to your schema

```
directive @client on FIELD
directive @connection on FIELD
```

to your schema, which you probably shouldn't do as they are client-only. Providing those via the server schema does not make sense.

dotansimha:
@n1ru4l But anyway, if someone uses client-side schema, they needs to provide those directives to the codegen (otherwise it will fail before running this plugin).

Issue #2556, opened 2019-09-15
Parent Resolver Type Uses Return Resolver Type


blakeembrey:
**Describe the bug**

When using a union type in GraphQL, the output types for `ResolversParentTypes` will use `ResolversTypes` which results in a `Promise` wrapped value. However, in GraphQL, you can't receive a `Promise<T>` in the parent position. Looks to be caused by this line: https://github.com/dotansimha/graphql-code-generator/blob/f5250087ef756f02803259abeecf318cd4180e9c/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts#L344. This code is used to generate both parent resolvers types and resolvers types, but doesn't differentiate within a union.

**To Reproduce**

1. My GraphQL schema:

```graphql
union Union = A | B
```

2. My GraphQL operations:

```graphql
# Put your operations here
```

3. My `codegen.yml` config file:

```yml
schema: "schema.graphql"
generates:
  src/__generated__/graphql.ts:
    plugins:
      - typescript
      - typescript-resolvers
```

**Expected behavior**

It should use the parent resolvers type for a union within parent resolvers.

**Additional context**

dbeauchamp:
Is there an ETA on this being released?

dotansimha:
Fixed in 1.8.0 🚀

PR #2557, opened 2019-09-15
Fix `getTypeToUse` within parent resolvers type

blakeembrey:
Related https://github.com/dotansimha/graphql-code-generator/issues/2556.


dotansimha:
Thanks @blakeembrey . You are right, it should use `ResolversParentTypes`. 
I rebased your PR and fixed the failing tests :)

PR #2572, opened 2019-09-16
typescript/react-apollo: Generate docblocks for hooks

### LINKED ISSUE #2558: JSDoc
It would be cool to generate [JSDoc](https://devdocs.io/jsdoc/) annotations from graphql descriptions. A lot of modern code editors get inline documentation when writing a function. (ex. [VSCode](https://code.visualstudio.com/docs/languages/javascript#_jsdoc-support))

CarloPalinckx:
This PR:
- Adds docblock generation to the react-apollo plugin. 
- Adds jest-docblock (from their official repo) dependency to parse docblocks in tests.
- A test to cover the new functionality.

Few things I would like to receive some feedback on:
- Should we also be looking into docblocks for HOC and Component API's?
- What exactly would we want to put in the docblocks, since the generated code is only used by typescript/flow users, I wonder how much benefit annotations like typedef have.

resolves #2558


dotansimha:
That's awesome! Thank you @CarloPalinckx !

levrik:
Why is this feature not toggleable? We're checking in our generated files and this generated a huge amount of noise. Also we're not seeing any value in that for us.

n1ru4l:
@levrik Pretty sure a pulll request for this will be accepted! It should be pretty easy too!

CarloPalinckx:
@levrik @n1ru4l I'll have a look 👍, default behaviour would be to add the comments? I understand that you could see it as noise but I think IDE integrations outweigh clutter in generated files.

CarloPalinckx:
@levrik a configuration option was added to disable the docblocks https://github.com/dotansimha/graphql-code-generator/pull/2717

levrik:
@CarloPalinckx awesome, thanks.

PR #2591, opened 2019-09-19
Prepare for federation on top of plugins

kamilkisiela:
@dotansimha should we care about plugin's config? Otherwise I could remove those lines https://github.com/dotansimha/graphql-code-generator/pull/2591/files#diff-001d4eb7df07b3b30562eaba70d2f29aR182-R194

### DIFF
--- a/dev-test/codegen.yml
+++ b/dev-test/codegen.yml
@@ -20,14 +20,14 @@ generates:
       - add: // tslint:disable
       - typescript
       - typescript-resolvers
-  # ./dev-test/test-schema/resolvers-federation.ts:
-  #   schema: ./dev-test/test-schema/schema-federation.graphql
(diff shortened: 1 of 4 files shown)


dotansimha:
@kamilkisiela can you please explain your question? I didn't understand

kamilkisiela:
```yaml
config: ... # parent config
generates: 
  ./output.ts:
    config: ... # config
    plugins:
      - typescript-resolvers: 
        .. # plugin config
```

Is "plugin config" a thing? If so, should we accept `federation: true` in there?

This pr checks if federation is enabled/disabled in `config`, then in `parent config`, prioritizing `config` (if federation is enabled in `parent config` and disabled in `config` then we turn off federation for this particular output).

PR #2625, opened 2019-09-24
move @n1ru4l/graphql-codegen-relay-optimizer-plugin over to graphql-codegen monorepo

n1ru4l:
move @n1ru4l/graphql-codegen-relay-optimizer-plugin https://github.com/n1ru4l/graphql-codegen-relay-plugins/tree/master/packages/graphql-codegen-relay-optimizer-plugin over to graphql-codegen monorepo

____

There is also an example app in my monorepo: https://github.com/n1ru4l/graphql-codegen-relay-plugins/tree/master/packages/todo-app-example

Should we move that one over here as well? It kind of served as a test to ensure there are no conflicts with the other plugi^ns.


dotansimha:
@n1ru4l Thank you!
Maybe the test repo could be a unit test in this repo? It will be easier to test it this way.

n1ru4l:
@dotansimha Any examples for a unit test that runs the codegen cli with multiple plugins?

dotansimha:
@n1ru4l You need the cli itself or just running the plugins is ok? 
Because if you needs the plugins, you can import them directly in your tests (from the `src` or as a node library, and then run it and merge the outputs together.

n1ru4l:
@dotansimha I can add these tests later in case anything breaks. Should I also update the docs?

dotansimha:
Yes please :)
Thanks @n1ru4l !

n1ru4l:
@dotansimha Done 👍

PR #2689, opened 2019-10-05
Custom output default value based on the selected plugins

JureSotosek:
Hey hey, another one from me 👋🏼

**The problem I had:**
When running `yarn graphql-codegen init`, you are prompted with a question: "Where to write the output:", with the default value thats always "src/generated/graphql.ts". A problem occurs if you have for example chosen `flow` plugin that requires its extension to be `.js` or the `react-apollo` plugin that requires its extension to be `.jsx`. 

**My solution:**
The default value should be inferred from the plugins you have selected. There is also an inferred hierarchy of the extensions.

For example: If you have multiple plugins selected and one of them requires `.tsx`, no matter what other plugins requires, `.tsx` will be the suggested default.

If you have multiple plugins selected and one of them requires `.ts` and no other requires `.tsx`, no matter what other plugins require, `.ts` will be the suggested default.

`.js` will be a suggested default only if no other plugins require `.ts` or `.tsx`.

**Example**
Before, when `react-apollo` plugin was selected, you were prompted with a default `.ts` output file:

Screenshot 2019-10-05 at 18 02 54

Now you are prompted with a default `.tsx` output file:
 
Screenshot 2019-10-05 at 17 50 01

Let me know what you think ✌🏼

### DIFF
--- a/packages/graphql-codegen-cli/src/init/plugins.ts
+++ b/packages/graphql-codegen-cli/src/init/plugins.ts
@@ -9,6 +9,7 @@ export const plugins: Array<PluginOption> = [
     pathInRepo: 'typescript/typescript',
     available: hasTag(Tags.typescript),
     shouldBeSelected: tags => oneOf(tags, Tags.angular, Tags.stencil) || allOf(tags, Tags.typescript, Tags.react) || noneOf(tags, Tags.flow),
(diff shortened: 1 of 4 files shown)


dotansimha:
That's perfect! Thank you @JureSotosek !

PR #2691, opened 2019-10-05
Update extensions in the docs

JureSotosek:
Another one, kind of connected to #2689, I feel like extensions of output files in the docs should by corresponding to the plugins used.

Especially in the case of `apollo-react` where using a `.ts` output file throws an error.

Also, in the case of `flow` plugins, if you are using `flow` then you are probably not using `typescript` and the extension should be just `.js`.

Another small one I fixed, the last `yml` code snippet in the `react-apollo` plugin docs was wrongly formated, here is how it looked:
Screenshot 2019-10-06 at 00 31 21

### DIFF
--- a/docs/generated-config/flow-operations.md
+++ b/docs/generated-config/flow-operations.md
@@ -8,7 +8,7 @@ Generates Flow types as Exact types.
 
 ```yml
 generates:
-path/to/file.ts:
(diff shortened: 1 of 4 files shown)


dotansimha:
great catch, thank you @JureSotosek 😍

PR #2725, opened 2019-10-10
Fix Bug when using react-apollo and transformUnderscore

damianfrizzi:
We encountered a bug when using hooks with typescript-react-apollo and the option transformUnderscore enabled.

Configuration file:
```
overwrite: true
schema: '...'
documents: ['...']
generates:
  something/:
    preset: near-operation-file
    presetConfig:
      extension: .generated.tsx
      baseTypesPath: types.ts
    plugins:
      - typescript-operations
      - typescript-react-apollo
    config:
      withHooks: true
      withComponent: false
      withHOC: false
      skipTypename: true
      namingConvention:
        typeNames: change-case#pascalCase
        enumValues: keep
        transformUnderscore: true
      documentMode: external
```

This generates the correct types (`GET_SOMEHTING` leads to `GetSomethingQuery`) and the correct hooks (`use_GET_SOMEHTING` leads to `useGetSomethingQuery`) but the generics are wrong (`Get_SomethingQuery` rather than `GetSomethingQuery`) and therefore can't be found.

With the fix provided in this PR the generics will be correctly generated as well.

### DIFF
--- a/packages/plugins/typescript/react-apollo/src/visitor.ts
+++ b/packages/plugins/typescript/react-apollo/src/visitor.ts
@@ -285,8 +285,16 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
   }
 
   protected buildOperation(node: OperationDefinitionNode, documentVariableName: string, operationType: string, operationResultType: string, operationVariablesTypes: string): string {
-    operationResultType = `${this._prefix}${operationResultType}`;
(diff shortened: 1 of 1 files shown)


dotansimha:
Thank you @damianfrizzi ! 
I moved this fix to: https://github.com/dotansimha/graphql-code-generator/pull/2743 because I applied some more fixed related to `transformUnderscore`.

PR #2732, opened 2019-10-11
Added start for webpack/babel compatible declarations

CarloPalinckx:
@dotansimha @mxstbr I've started working on making a plugin able to output "declarations only". In order to be compatible with something like a babel-macro or webpack-loader.

A few things I would like some feedback on before continuing:

I think there are two approaches we can take. Either we make "declarations only" a configuration option per plugin. Or we can adjust the graphql-module plugin to somehow include the modified output of specific compatible plugins.

I could also use some pointers on how to get the document file path inside a plugin nicely, that way we can declare the module with the path accordingly.

Would love to hear your thoughts 🙌


dotansimha:
Nice! Thank you @CarloPalinckx ! 

@kamilkisiela can you please take a look? you are the last one that tried to integrate Webpack with the Codegen ;)

Issue #2588, opened 2019-09-19
ID Scalar is incorrectly typed


n1ru4l:
**Describe the bug**

When you return a number as an `ID` type in a resolver, it causes a typescript error. However, `graphql.js` does also accept returning numbers which are then automatically cast to strings.

See https://graphql.github.io/graphql-spec/draft/#sec-ID

TL;DR:

ID as operation argument/return type from resolver -> string or number
ID as operation selection set value -> string

ardatan:
@n1ru4l I think it is correct right now because if you have a different type than string in your entity(parent) type, you can use mappers to define it like below;
```ts
export interface UserEntity {
    id: number;
   name: string;
}
```
```gql
type User {
  id: ID
  name: String
}
```
```ts
export const Query: QueryResolvers = {
   user(_, { id }) => ({ id: 2, name: 'TestUser' }),
}
```

n1ru4l:
> So in that case, it should be fine, because if we change ID to string | number, it wouldn't be consistent with client which is always string.

Wouldn't it be possible to distinguish between resolver return `ID` and client-consumed `ID`?

ardatan:
@n1ru4l We can do that but we use those generated types in client as well. They shouldn't be used as parent types in my opinion because the data coming from another source is generally different than the one fetched by client. I think we should keep that and you should use `mappers` as I said.
@dotansimha What do you think?

dotansimha:
I think we can generate different types for client and server, that shouldn't be a problem. 
I think the codegen assumes that `ID` is `string` in GraphQL, but the actual real value is a number or a string. So I think the should maybe change it in the codegen, even if it's a breaking change for some developers (because it caused by an actual bug in the codegen).

dotansimha:
The actual default value of the `ID` scalar should be `string | number`, to match the implementation of the scalar in `graphql.js`: https://github.com/graphql/graphql-js/blob/730f5cc6e0c7b7994cb11fe1a9785700fa4dc5db/src/type/scalars.js#L256 

@n1ru4l @ardatan @kamilkisiela what do you think? Do you think it will cause any breaking changes?

n1ru4l:
https://github.com/graphql/graphql-js/blob/730f5cc6e0c7b7994cb11fe1a9785700fa4dc5db/src/type/scalars.js#L239-L247

ID as Input -> String or Number
ID as Output -> String

dotansimha:
[...]

PR #2757, opened 2019-10-15
Added support for custom scalars

Choem:
Hello, I did come across a problem that when I provided a custom scalar, it would try and find the type in the graphql-type package. This type did of course not exist and it would define it as any in the code generated graphql file.

So I thought it would be nice to have custom scalar support in the plugin :)

```
    generate({
      overwrite: true,
      schema: `http://${minikubeIP.trim()}/micro-service/graphql`,
      documents: 'src/queries/micro-service/**/*.graphql',
      generates: {
        'src/generated/micro-service-graphql.ts': {
          plugins: [{ [TYPESCRIPT_TYPE_GRAPHQL_PLUGIN]: { scalars: { DateTime: "Date" } } }, TYPESCRIPT_OPERATIONS_PLUGIN]
        },
        'src/generated/micro-service-graphql-modules.d.ts': {
          plugins: [TYPESCRIPT_GRAPHQL_FILES_MODULES_PLUGIN]
        },
        'src/generated/micro-service.graphql': {
          plugins: [TYPESCRIPT_SCHEMA_AST_PLUGIN]
        }
      }
    })
```

Would generate something like this.

Added to the scalars with correct type;
```
export type Scalars = {
  ID: string,
  String: string,
  Boolean: boolean,
  Int: number,
  Float: number,
  DateTime: Date,
};
```

If it would find the type as a native js type it would produce a field like this.
```
@TypeGraphQL.ObjectType()
export class Class {
  __typename?: 'Class';

  @TypeGraphQL.Field(type => Date)
  field!: Scalars['DateTime'];
};
```

An important note to Apollo client users! 
I'm using Apollo client myself and there [isn't support to automatically interpret custom scalars](https://www.apollographql.com/docs/graphql-tools/scalars/#custom-scalars). So sadly you have to map the values to the correct type yourself. There also doesn't seem to be any plans for integrating it because of this [thread](https://github.com/apollographql/apollo-client/issues/585).

For example in a resolver, we have an Apollo client that does a call and get objects back. The field contains a date type but still needs to be converted so one way of doing this:

```
  const objects = (await microServiceClient.query<
      QueryType,
      QueryVariables
    >({
      query: QueryDoc,
      variables: {
        variable
      }  
    })).data.objects;

    return objects.map(object => ({
      ...object,
      field: new Date(object.field)
    }));
```

### DIFF
--- a/packages/plugins/typescript/type-graphql/src/visitor.ts
+++ b/packages/plugins/typescript/type-graphql/src/visitor.ts
@@ -172,7 +172,20 @@ export class TypeGraphQLVisitor<TRawConfig extends TypeGraphQLPluginConfig = Typ
     const isArray = !!nonNullableType.match(ARRAY_REGEX);
     const singularType = nonNullableType.replace(ARRAY_REGEX, '$1');
     const isScalar = !!singularType.match(SCALAR_REGEX);
-    const type = singularType.replace(SCALAR_REGEX, (match, type) => (global[type] ? type : `TypeGraphQL.${type}`));
+    const type = singularType.replace(SCALAR_REGEX, (match, type) => {
+      if (TYPE_GRAPHQL_SCALARS.includes(type)) {
+        // This is a TypeGraphQL type
+        return `TypeGraphQL.${type}`;
+      } else if (global[type]) {
+        // This is a JS native type
+        return type;
+      } else if (this.scalars[type]) {
+        // This is a type specified in the configuration
+        return this.scalars[type];
+      } else {
+        throw new Error(`Unknown scalar type ${type}`);
+      }
+    });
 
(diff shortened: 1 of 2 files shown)


Choem:
I'll check it this afternoon, thanks for your fast reply.

dotansimha:
Nevermind actually, @n1ru4l is right here: https://github.com/dotansimha/graphql-code-generator/issues/2588#issuecomment-542184382 

We can get this merged now :) 

Thanks @Choem !

ChrisLahaye:
We require BigInt in the scalar map but this doesn't work since it is also a global / native type which are given preference. Wouldn't it make more sense to give explicit configuration, the scalar map, preference?

Issue #2676, opened 2019-10-03
federation bug: Cannot find name 'ReferenceResolver'.


ForbesLindesay:
**Describe the bug**
 

When federation is enabled, you now get the error: `Cannot find name 'ReferenceResolver'.`

**To Reproduce**

[![Edit graphql-codegen-issue-template](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/graphql-codegen-issue-template-wdfcr?fontsize=14)

1. run yarn generate
2. run yarn typecheck

My GraphQL schema:

```graphql
type Query {
    user(id: ID!): User!
}

type User @key(fields: "id") {
    id: ID!
    username: String!
    email: String!
}
```

My `codegen.yml` config file:

```yml
schema: schema.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      federation: true
```

**Expected behavior**

It should output the type, https://github.com/dotansimha/graphql-code-generator/blob/587599f47336afff05e42f2cd11c74c69bd5ab40/packages/plugins/typescript/resolvers/src/index.ts#L90-L96 depends on a property that doesn't get set until after the rest of the code is generated.

dotansimha:
@ForbesLindesay Can you please try to move the federation config upper?

```
schema: schema.graphql
config:
  federation: true
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-resolvers
```

@kamilkisiela

zhenxzheng:
@dotansimha, I see the same bug with v1.8. moving config federation to upper didn't help

dotansimha:
@kamilkisiela can you please take a look? ❤️

ruanyl:
I have the same issue, running codegen v1.8.1
even after moving the config upper, it doesn't help

ruanyl:
@dotansimha was it `GraphQLReferenceResolver` from
https://github.com/apollographql/apollo-server/blob/d5015f4ea00cadb2a74b09956344e6f65c084629/packages/apollo-federation/src/types.ts#L57

ruanyl:
The workaround that I have now is to declare it in `global.d.ts`
```
import {GraphQLResolveInfo} from 'graphql';

declare global {
  type ReferenceResolver<TReference = object, TContext = any, ResolverInfo = GraphQLResolveInfo> = (
    reference: TReference,
    context: TContext,
    info: ResolverInfo
  ) => any;
}
```

[...]

PR #2760, opened 2019-10-15
Fix for missing `ReferenceResolver` import

kamilkisiela:


dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/2676#issuecomment-542201357

Issue #2772, opened 2019-10-16
flow-operations generates semi exact types


vasilich6107:
**Describe the bug**
The generated types are somehow not fully exact

**To Reproduce**
Generate types with flow operaions

1. My GraphQL schema:

```graphql
type Query {
    getReportDefaults: ReportDefaults!
}

type ReportDefaults {
    arrival: Destination
    dateFrom: String
    dateTo: String
    departure: Destination
    type: ReportType
    voyageNumber: String
}

type Destination {
    id: ID!
    name: String!
    type:String!
}
```

2. My GraphQL operations:

```graphql
query getReportDefaults {
        getReportDefaults {
            dateFrom
            dateTo
            voyageNumber
            type
            departure {
                id
                name
                type
            }
            arrival {
                id
                name
                type
            }
        }
    }
```

3. My `codegen.yml` config file:

```yml
overwrite: true
generates:
  src/types/graphql.js:
    schema: {
      "[URL]": {
        headers: {
          Authorization: 'Bearer ...'
        }
      }
    }
    documents:
      - "./src/apollo/queries/reports.js"
    plugins:
      - "flow"
      - "flow-operations"
    config:
      scalars:
        DateTime: string
      skipTypename: false
      nonOptionalTypename: true
      useFlowExactObjects: true
```

**Expected behavior**
```
export type GetReportDefaultsQuery = {|
    ...{| __typename: 'Query' |},
    ...{|
        getReportDefaults: {|
            ...{| __typename: 'ReportDefaults' |},
            ...ReportDefaultsFragment
        |}
    |}
|};

export type ReportDefaultsFragment = {|
    ...{| __typename: 'ReportDefaults' |},
    ...$Pick<ReportDefaults, {| dateFrom: *, dateTo: *, voyageNumber: *, type: * |}>,
    ...{|
        departure: ?{|
            ...{ __typename: 'Destination' },
            ...DestinationFragment
        |},
        arrival: ?{|
            ...{| __typename: 'Destination' |},
            ...DestinationFragment
        |}
    |}
|};
```

**Actual behavior(a lot of pipes are missing)** 
```
export type GetReportDefaultsQuery = {
    ...{ __typename: 'Query' },
    ...{|
        getReportDefaults: {
            ...{ __typename: 'ReportDefaults' },
            ...ReportDefaultsFragment
        }
    |}
};

export type ReportDefaultsFragment = {
    ...{ __typename: 'ReportDefaults' },
[...]

PR #2795, opened 2019-10-21
[Flow] Make interface implementing types exact when object spreading.

slagiewka:
The issue described in #2772 is also visible with general Flow schema types generation. 

Due to that Flow checks will fail because inexact type is incompatible with exact overall type.

If these changes are OK, I can also dig into fixing it with `flow-operations`.

As side note: @dotansimha what do you think about the option to make interface types NOT exact (with option or not)? The issue here is that if we define a resolver return type with an exact type, we cannot return a type that is a superset to that interface. Making it inexact will help to return anything compliant with that actual interface.

### DIFF
--- a/packages/plugins/flow/flow/src/visitor.ts
+++ b/packages/plugins/flow/flow/src/visitor.ts
@@ -100,7 +100,17 @@ export class FlowVisitor extends BaseTypesVisitor<FlowPluginConfig, FlowPluginPa
       return allFields.join('\n');
     }
 
-    return `...{\n${allFields.map(s => indent(s)).join('\n')}\n  }`;
+    return (
+      new DeclarationBlock(this._declarationBlockConfig)
+        .withContent('...')
+        .withBlock(allFields.join('\n'))
+        .ignoreBlockSemicolon(true)
+        .string.slice(0, -1) // Removes trailing newline from stringified block.
+        .split('\n')
+        // First line gets indented independently when joined in appendInterfacesAndFieldsToBlock.
+        .map((s, i) => (i !== 0 ? indent(s) : s))
+        .join('\n')
+    );
   }
 
   EnumTypeDefinition(node: EnumTypeDefinitionNode): string {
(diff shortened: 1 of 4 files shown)


packages/plugins/flow/flow/src/visitor.ts
```diff
@@ -100,7 +100,17 @@ export class FlowVisitor extends BaseTypesVisitor<FlowPluginConfig, FlowPluginPa
       return allFields.join('\n');
     }
 
-    return `...{\n${allFields.map(s => indent(s)).join('\n')}\n  }`;
+    return (
+      new DeclarationBlock(this._declarationBlockConfig)
+        .withContent('...')
+        .withBlock(allFields.join('\n'))
+        .ignoreBlockSemicolon(true)
+        .string.slice(0, -1) // Removes trailing newline from stringified block.
```
slagiewka: It would be great to have the trailing newline configurable as well, the same way block semicolon was added by me now.

packages/plugins/flow/flow/src/visitor.ts
```diff
@@ -100,7 +100,17 @@ export class FlowVisitor extends BaseTypesVisitor<FlowPluginConfig, FlowPluginPa
...
+        .split('\n')
+        // First line gets indented independently when joined in appendInterfacesAndFieldsToBlock.
+        .map((s, i) => (i !== 0 ? indent(s) : s))
```
slagiewka: If you have better ideas how to solve this - I'm open to suggestions.

slagiewka:
@dotansimha any thoughts? If you would like me to change something in this PR just let me know and we will proceed 😉

whollacsek:
Is there an ETA on this?

dotansimha:
@slagiewka @whollacsek 
Sorry for the delay. I implemented it by adding `|` where needed. Used you tests and snapshot for comparison. 

https://github.com/dotansimha/graphql-code-generator/pull/3007

Thank you @slagiewka !

Issue #2746, opened 2019-10-14
Typescript-Resolver: custom `resolveInfo` type


ab-pm:
Hi, we are using the Postgraphile library which not only uses a [custom context](https://graphql-code-generator.com/docs/plugins/typescript-resolvers#custom-context-type) but also [has an extended `resolveInfo` argument](https://www.graphile.org/postgraphile/make-extend-schema-plugin/#the-selectgraphqlresultfromtable-helper). They do [provide their own `AugmentedGraphQLFieldResolver` type](https://github.com/graphile/graphile-engine/blob/1d589633698b4bdfb2923226a330b6345505dd77/packages/graphile-utils/src/makeExtendSchemaPlugin.ts#L44-L56) but since it returns `any` it's not really compatible with the proper resolver types created by `graphl-codegen`.

Feature request: Could you please add another configuration option so that one could choose the type of the `ResolveFn`'s `info` argument?
```ts
export type ResolverFn<TResult, TParent, TContext, TArgs> = (
  parent: TParent,
  args: TArgs,
  context: TContext,
  info: GraphQLResolveInfo
//      ^^^^^^^^^^^^^^^^^^ customise this
) => Promise<TResult> | TResult;
```

dotansimha:
Implemented in 1.8.2.

PR #2803, opened 2019-10-22
Support custom ResolveInfo

ardatan:
Related #2746

### DIFF
--- a/packages/plugins/typescript/resolvers/src/index.ts
+++ b/packages/plugins/typescript/resolvers/src/index.ts
@@ -195,7 +195,6 @@ export type DirectiveResolverFn<TResult = {}, TParent = {}, TContext = {}, TArgs
 ) => TResult | Promise<TResult>;
 `;
 
-
   const resolversTypeMapping = visitor.buildResolversTypes();
   const resolversParentTypeMapping = visitor.buildResolversParentTypes();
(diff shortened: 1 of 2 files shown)


packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -186,8 +208,25 @@ export type DirectiveResolverFn<TResult = {}, TParent = {}, TContext = {}, TArgs
...
   }
 
+  const prepend: string[] = [];
+
+  if (imports.length) {
+    prepend.push(`import { ${imports.join(', ')} } from 'graphql';`);
+  }
+
+  if (config.customResolveInfo) {
+    const [importName, moduleName] = config.customResolveInfo.split('#');
```
dotansimha: I think it's better to use here `parseMapper` method, it will handle the `#` but also other use cases.

packages/plugins/typescript/resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -631,6 +631,15 @@ describe('TypeScript Resolvers Plugin', () => {
     await validate(result, {}, testSchema);
   });
 
+  it('Should generate the correct imports when customResolveInfo defined', async () => {
```
dotansimha: Where it's defined? :X

ardatan:
@dotansimha I updated tests

Issue #2416, opened 2019-08-21
Unused imports are generated when using fragment


mizdra:
**Describe the bug**
`tsc --noUnusedLocals` warn generated files because the files include unused imports when using fragment.

**To Reproduce**

- Sample repository: https://github.com/mizdra/graphql-code-generator-sample
```
$ git clone https://github.com/mizdra/graphql-code-generator-sample.git
$ cd graphql-code-generator-sample
$ yarn install
$ yarn run build:graphql
$ yarn run check:typecheck # tsc --noUnusedLocals
```

1. My GraphQL schema:

```graphql
type Query {
  articles: [Article!]!
}
type Article {
  id: ID!
  title: String!
  body: String!
}
```

2. My GraphQL operations:

```tsx
// ArticleListItem.tsx
export function ArticleListItem() { ... }
ArticleListItem.Fragment = gql`
  fragment SimpleArticle on Article {
    title
    body
  }
`
```

```tsx
// ArticleList.tsx
import { ArticleListItem } from './ArticleListItem'

const GET_ARTICLES = gql`
  query GetArticles {
    articles {
      ...SimpleArticle
    }
  }
  ${ArticleListItem.Fragment}
`
```

3. My `codegen.yml` config file:

```yml
overwrite: true
schema: schema.graphql
generates:
  src/types.generated.ts:
    - typescript
  src/:
    documents: 'src/**/*.{ts,tsx}'
    preset: near-operation-file
    presetConfig:
      extension: .generated.tsx
      baseTypesPath: types.generated.ts
    plugins:
      - typescript-operations
```

**Expected behavior**
Unused imports should be removed from generated files.

**Environment:**

- OS: macOS Mojave 10.14.6
- `@graphql-codegen/cli`: 1.6.1
- `@graphql-codegen/near-operation-file-preset`: 1.6.1
- `@graphql-codegen/typescript`: 1.6.1
- `@graphql-codegen/typescript-operations`: 1.6.1
- `typescript`: 3.5.3
- NodeJS: 12.4.0

**Additional context**

dotansimha:
Fixed and available as alpha `1.6.2-alpha-aeb3f880.46`, @mizdra can you please try it?

dotansimha:
Fixed in 1.7.0

mizdra:
Using 1.7.0, my problem was resolved. Thank you!

mizdra:
[...]

PR #2805, opened 2019-10-22
[Plugin] `oclif` CLI Generator

kalzoo:
First, thanks @dotansimha, @kamilkisiela, and @ardatan for your great work on this package! I didn't know what I was missing until I used the `react-apollo` plugin for the first time, and it's saved me many, many hours of work since then.

The plugin in this PR, `typescript-oclif`, generates command files for the [oclif](https://www.npmjs.com/package/oclif) package as described in the included docs. It includes simple snapshot unit tests and a [link to a working example repo](https://github.com/kalzoo/graphql-codegen-oclif-example). Of course, that example needs an `yarn|npm link` to this package in the meantime.

While it does work as-is, there are still a few features we may want to add:

* I'd like to give the user a way to neatly format the results returned by `graphql-request` - am considering an optional `formatPath` config for the plugin, which would be a path to a module with a default export that formats/post-processes all request responses
* On the topic of paths, the `near-operation-file` preset has access to the [base output directory](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/presets/near-operation-file/src/index.ts#L183), but the plugins it invokes don't appear to get that information. So, the config has this weirdness where for the preset it's a file path that's resolved by depth [here](https://github.com/kalzoo/graphql-codegen-oclif-example/blob/master/codegen.yml#L10) vs the plugin [here](https://github.com/kalzoo/graphql-codegen-oclif-example/blob/master/codegen.yml#L13), where it's not. This makes the generated plugin output fragile (unless the user has absolute import paths configured), since it assumes a uniform depth for all generated files (which might not be the case). Ie, ` import client from ../../client` may be a valid import in `src/commands/something/file.ts` but not `src/commands/other.ts`.
* I'd like to support file uploads, and so will look into supporting `apollo-client` rather than/alongside `graphql-request`.
* I'd like to be able to determine which input arguments are of `Enum` type, and while that information is present in the [schema](https://github.com/dotansimha/graphql-code-generator/blob/master/dev-test/githunt/schema.json#L115) itself, it doesn't appear to be available to the plugin anywhere, even within the `schema` constructor parameter. This may easily be a misunderstanding of graphql on my part.
* I'd like to be able to support descriptions for arguments somehow, but I'm not sure where that information could live in a client document, since there can't be directives in the arguments section (otherwise: `query GetUser($name: name @oclif(description: "This is a name")) { ... }`), and since directives only seem to support scalar-typed values (otherwise: `@oclif(arguments: { name: { description: 'This is a name' } } )`)

### Less Important

* I'd like to be able to remove/override the extra imports that aren't added by this plugin itself (ie `import * as Types ...` and `graphql-tag`). [Related?](https://github.com/dotansimha/graphql-code-generator/issues/2416)
* I'd like to be able to log warnings in code generation, but console statements seem to be swallowed by the CLI tool
* Would prefer to use the more complex SWAPI rather than GitHunt, but importing it for tests failed for reasons that aren't immediately clear to me:

```
Invalid or incomplete introspection result. Ensure that you are passing "data" property of introspection response and no "errors" was returned alongside: { data: { __schema: [Object] } }

      12 |   const schemaText = require('../../../../../dev-test/star-wars/schema.json');
    > 13 |   const schema = buildClientSchema(schemaText);
```

Looking forward to hearing what you think!


dotansimha:
This is very nice idea, @kalzoo ! I'll review it later this week :)

kalzoo:
@dotansimha any thoughts? 

In the meantime, I've thought of a change: rather than including `graphql-request` in every command module, each one should call a method imported from the base path (by default `execute`, but overridable in config), that receives the query document and all variables - that way implementers can define the query preparation, execution, and post-processing/display however they want.

This would address the `formatPath` and 'file uploads' questions above, but the jury is still out on the other ones.

kalzoo:
@dotansimha , two asks:
1. Could you provide insight into any/all of the questions from the first post?
2. Would it be better if I vended this plugin separately from the `graphql-code-generator` package? In the docs you mentioned it's better to keep it here for maintainability, but it looks like it might not have a chance of being merged.

dotansimha:
This is awesome! Thank you so much @kalzoo !

PR #2808, opened 2019-10-22
Interpolate multiple env variables

simPod:


dotansimha:
Great! Thank you @simPod

PR #2809, opened 2019-10-22
Use GraphQL Config

kamilkisiela:


dotansimha:
@kamilkisiela @Urigo @ardatan Merging this into `next`, and we'll include this with 1.9 (once new `graphql-config` will be available as alpha ;) )

PR #2836, opened 2019-10-26
Fix broken links from graphql-js

alivenotions:
There were some broken links, pointing to the correct location now


ardatan:
Thank you @alivenotions !

PR #2837, opened 2019-10-26
add filename to target file

alivenotions:
In the [Using in runtime](https://graphql-code-generator.com/docs/getting-started/#using-in-runtime) section, the `generates` file points to a directory and hence throws:
`Error: EISDIR: illegal operation on a directory`

It gets fixed by providing the file name.


ardatan:
Thank you @alivenotions !

PR #2850, opened 2019-10-30
make fragment-matcher compatible with Apollo Client 3

kel-sakal-biyik:
Apollo announced Client version 3. It is in beta right now. The configuration for the fragment matcher has been changed. 

The new configuration can be found [here](https://www.apollographql.com/docs/react/v3.0-beta/data/fragments/#defining-possibletypes-manually)

This pull request aims to make fragment-matcher plugin compatible with the new Apollo Client v3.


dotansimha:
Looks good. I think we should have something like `apolloClientVersion: 2 | 3` in order to allow developers to choose (without limit it to the codegen version).

Urigo:
Thank you @kel-sakal-biyik !!

dylanwulf:
This is awesome, thank you for making it happen so quickly!

PR #2855, opened 2019-10-31
Updated docs for union directive.

Nivek92:


dotansimha:
You are right :) Thanks @Nivek92 !

PR #2883, opened 2019-11-05
fix broken links in using-visitor.md

kroeder:
All visitor examples in https://graphql-code-generator.com/docs/custom-codegen/using-visitor point to 404. I assume the directory structure has changed after this documentation was written

This PR fixes all example links


dotansimha:
Awesome! Thank you @kroeder !

Issue #2415, opened 2019-08-21
Generic Apollo Client Plugin (Not Frontend-Specific)


travigd:
**Is your feature request related to a problem? Please describe.**

I'm trying to use Apollo from a Node server. I use it on the frontend with a React app and I love it together with the generated types but it's a little more problematic to use in a frontend-framework-agnostic way because all of the Apollo plugins are specific to React/Angular/etc.

**Describe the solution you'd like**

The React Apollo plugin is capable of generating `useGetFooQuery()` functions that are strongly typed. It'd be nice if there was a plugin to generate similar functions for the base Apollo client. Something like...

```ts
export function executeGetFooQuery(
  client: ApolloClient,
  // TVariables should be the specific type associated with GetFooQuery (in this case, just {})
  // Omit query to avoid an end-user trying to override the query
  options: Omit<QueryOptions<TVariables>, "query">
) {
  // TResult should be the specific type associated with GetFooQuery
  return client.query<TResult>({...options, query: GetFooDocument});
}
```

The `options` args should probably be optional if there are no required variables and non-optional otherwise. :^)

**Describe alternatives you've considered**

Currently, I can just do something like
```ts
import {GetFooQuery} from "./generated.ts";
client.query<GetFooQuery>({ query: gql`query GetFoo { ... }` });
```
but it requires me manually specifying the type parameter.

**Additional context**

If I should open this issue somewhere else, let me know! Thanks for all the hard work.

dotansimha:
Sounds like a valid use-case :)
We'll maybe add this in the future. Feel free to send a PR if you wish :)

schmod:
I'd also be interested to see something like this for [graphql-request](https://www.npmjs.com/package/graphql-request), and may take a stab at a PR.

schmod:
thanks @dotansimha ;-)

P4sca1:
As `graphql-request` has already been implemented, are there any news on the `apollo-client` implementation? @dotansimha

dotansimha:
@P4sca1 Sorry, couldn't find the time to add it. If someone has the time, it should be pretty simple (most of is is copy/paste of the graphql-request plugin).

akozhemiakin:
Maybe we should abstract over concrete client implementation? After all it is all about passing document and variables and returning some typed result wrapped in promise. I prepared a draft: #2978

dotansimha:
[...]

PR #2978, opened 2019-11-17
Introduce typescript-generic-sdk plugin

akozhemiakin:
Possible solution for #2415 which abstracts over concrete graphql client implementation

### DIFF
--- a/packages/plugins/typescript/generic-sdk/.gitignore
+++ b/packages/plugins/typescript/generic-sdk/.gitignore
@@ -1,0 +1,5 @@
+node_modules
+npm-debug.log
+dist
+temp
+yarn-error.log
--- a/packages/plugins/typescript/generic-sdk/.npmignore
+++ b/packages/plugins/typescript/generic-sdk/.npmignore
@@ -1,0 +1,5 @@
+src
+node_modules
+tests
+!dist
+example
--- a/packages/plugins/typescript/generic-sdk/package.json
+++ b/packages/plugins/typescript/generic-sdk/package.json
@@ -1,0 +1,44 @@
+{
+  "name": "@graphql-codegen/typescript-generic-sdk",
+  "version": "1.8.3",
+  "description": "GraphQL Code Generator plugin for generating a ready-to-use client-agnostic SDK based on GraphQL operations",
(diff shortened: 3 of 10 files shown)


akozhemiakin:
It would be awesome to have higher kinded types feature in typescript to abstract over wrapper type. For example apollo return result in some wrapper and with this generic implementation we forced to unwrap it in requester.

dotansimha:
Very nice! I think the effort of having to create a wrapper for client is fine. 
I'll review it soon :) 

Thanks @akozhemiakin !

akozhemiakin:
Thank you for the response and for the great project. I just pushed another commit which allows to pass options to the requester.
Also I really want to find a way to generalize output type. Maybe I'll give a try to fp-ts HKT or something like that. In Scala it would be painless, but Typescript, unfortunately doesn't have so much power it its type system yet.

akozhemiakin:
Btw, here is the Apollo requester implementation created to be used with this plugin:
https://gist.github.com/akozhemiakin/731b0c1e99eb89b01f80f08f9146b6b6

dotansimha:
Sorry for the delay, @akozhemiakin . 
We are currently working on a library that could use that feature, so I'm reviewing and testing it now. I assume we can merge this very soon.

dotansimha:
Merged, I also pushed a fix to align the version number in `package.json`.
Alpha will be available very soon.

dotansimha:
Alpha version now available: `1.12.2-alpha-085fe906.18`.

DrChrispoper:
I can only find documentation for `typescript-graphql-request`. Does this mean there is no `typescript-generic-sdk` or as https://github.com/dotansimha/graphql-code-generator/issues/3458 sugguests there is just no documentation for it. I am looking to use it with Apollo Client.

akozhemiakin:
@DrChrispoper First of all, take my excuses for not providing docs along with the original request, I'll try to schedule it.

Meanwhile, have you seen this gist?
https://gist.github.com/akozhemiakin/731b0c1e99eb89b01f80f08f9146b6b6

It is how we use `typescript-generic-sdk` with Apollo in our project.

DrChrispoper:
Thanks for reaching out so fast, yeah I've seen it thanks! The docs would be great mainly to know that it exists hehe!

PR #2993, opened 2019-11-21
Fix closing of multiline comment blocks

yamaha252:
Now it's impossible to fix `**/` multiline comment ending to be them linted correctly


dotansimha:
Thanks @yamaha252 !

Issue #2973, opened 2019-11-15
Add option to `typescript-graphql-files-modules` to generate complete paths


pietdevries94:
Currently the modules generated for the files have a wildcard as path. Example: `declare module '*/Project.gql' {`

I would like an option to generate the files with the absolute path (`src/graphql/queries/Project.gql`). This way auto-imports will work correctly in VS Code

dotansimha:
I implemented it in this PR: https://github.com/dotansimha/graphql-code-generator/pull/3010

With these changed, you should be able to set now `relativeToCwd: true` to generate a complete path to CWD, and you can customize the prefix (default for `*/`).

@pietdevries94

dotansimha:
Available in 1.9.0

tyankatsu0105:
@dotansimha 
How can I set the option for typescript-files-modules?

vhiairrassary:
@tyankatsu0105 For me it works using something like:
```yaml
generates:
  src/types/graphql.d.ts:
    plugins:
      - 'typescript-graphql-files-modules'

    config:
      relativeToCwd: true
```

bennypowers:
Would be awesome if these would generate files in-place, so that I'd get a `.d.ts` file for each document

PR #3010, opened 2019-11-25
Fixed for typescript-files-modules plugin

dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/2973


tyankatsu0105:
@dotansimha 
How can I set the option for typescript-files-modules?

PR #3014, opened 2019-11-26
Add assumeValidSDL config option

jdolle:
To allow more flexibility, I've exposed `graphql`'s assumeValidSDL option. (ref: https://github.com/graphql/graphql-js/blob/master/src/utilities/buildASTSchema.d.ts#L45)

I am intentionally using multiple directives and have a custom schema compiler. There are PRs to allow multiple directives (https://github.com/sangria-graphql/sangria/issues/409 https://github.com/graphql/graphql-js/pull/1541), but adding this option allows for more possibilities and more flexibility for peoples' schemas.

### DIFF
--- a/docs/getting-started/codegen-config.md
+++ b/docs/getting-started/codegen-config.md
@@ -65,6 +65,8 @@ Here are the supported options that you can define in the config file (see [sour
 
   - **`pluckConfig.globalIdentifier`** - Overrides the name of the default GraphQL name identifier.
 
+- **`assumeValidSDL`** - A flag to assume the SDL is valid
+
 ## Environment Variables
 
 You can use environment variables in your `codegen.yml` file::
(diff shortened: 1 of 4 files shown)


jdolle:
Not sure why the build failed -- I ran test, build, build-website, and lint and they all passed.

dotansimha:
@jdolle It's on our CI, just ignore it :) 

Thanks for the PR, I'll review it soon

dotansimha:
Thanks @jdolle 

I'm not sure that adding it in the root is the right choice, because we might want to allow this only for specific inputs. 
What do you think?

@ardatan @kamilkisiela

jdolle:
Perhaps it would be better placed under the schema, similarly to the custom loader? I can make that change if so.

ardatan:
@dotansimha This is something we should handle in graphql-toolkit's schema loaders. So we can move this to graphql-toolkit.

dotansimha:
Sounds good, @jdolle , it can be under schema config. 
@ardatan you are right, but we can merge it here and then later move it.

ardatan:
https://github.com/dotansimha/graphql-code-generator/pull/3147
This PR covers that functionality as an option to `schema` field like below;
```yml
schema:
  - your-schema-ptr:
      assumeValidSDL: true
```

jdolle:
Closing since #3147 covers this. Thanks

PR #3092, opened 2019-12-08
Try to fix alpha releases 

dotansimha:


ardatan:
It is fixed in other PR :)

Issue #3091, opened 2019-12-07
Make near-operation-file language agnostic


micimize:
**Problem**
I want to use `near-operation-file` for dart type generation, but `baseTypesPath` generates a js-specific import.

**Solution**
It could be made language agnostic by offering an `importTemplateString` config that could look something like
```js
`import "${baseTypesPath}" as ${importTypesNamespace}`
// or for simply overriding with a literal
`import 'package:my_package/base_types.dart';` // dart absolute import
```

dotansimha:
You are right, we should do that :) 
Which plugin are you trying to use with that?

micimize:
@dotansimha I'm reworking [graphql-to-dart](https://github.com/micimize/graphql-to-dart) to work via `near-operation-file`

micimize:
We could use a simple template interpolation helper:
```javascript
function interpolateTemplate(template, params) {
  const names = Object.keys(params);
  const vals = Object.values(params);
  return new Function(...names, `return \`${template}\`;`)(...vals);
}
interpolateTemplate('${foo} bar', { foo: 'biz' })
```

micimize:
Hacking around this with `add` for now
```yaml
      - add: |-
          // GENERATED CODE - DO NOT MODIFY BY HAND

          import 'package:savvy_app/graphql/schema.dart';
          /*
      - add: 
          placement: 'content'
          content: '*/'
```

micimize:
Actually that doesn't work for fragments, which will need other configurations also. Am going to fork and see what I can do

dotansimha:
This was merged already, so we can close this issue. 
Thank you @micimize !

PR #3109, opened 2019-12-12
Language agnostic near-operation-file preset helpers

micimize:
I've refactored `near-operation-file` to extract the fragment resolution logic into `fragment-resolver.ts` and the main document processing logic into `resolve-document-imports.ts`, which is then re-exported as `resolveDocumentImports` for use by other preset builders, thus satisfying #3091.

This cuts the main plugin logic down to the following, aside from prep and post processing:
```typescript
function resolveImportPath(baseOutputDir: string, relativeOutputPath: string, sourcePath: string) {
  const shouldAbsolute = !sourcePath.startsWith('~');
  if (shouldAbsolute) {
    const absGeneratedFilePath = resolve(baseDir, relativeOutputPath);
    const absImportFilePath = resolve(baseDir, sourcePath);
    return resolveRelativeImport(absGeneratedFilePath, absImportFilePath);
  } else {
    return sourcePath.replace(`~`, '');
  }
}

const sources = resolveDocumentImports(options, schemaObject, {
  generateFilePath(filePath: string) {
    const newFilePath = defineFilepathSubfolder(filePath, folder);
    return appendExtensionToFilePath(newFilePath, extension);
  },
  fragmentSuffix: 'Fragment',
  generateImportStatement({ relativeOutputPath, importSource, baseOutputDir }) {
    const importPath = resolveImportPath(baseOutputDir, relativeOutputPath, importSource.path);
    const importNames = importSource.names && importSource.names.length ? `{ ${importSource.names} }` : '*';
    const importAlias = importSource.namespace ? ` as ${importSource.namespace}` : '';
    return `import ${importNames}${importAlias} from '${importPath}';${importAlias ? '\n' : ''}`;
  },

  schemaTypesSource: {
    path: shouldAbsolute ? join(options.baseOutputDir, baseTypesPath) : baseTypesPath,
    namespace: importTypesNamespace,
  },
});
```

<details>
<summary>Old Description</summary>

This is a first pass at #3091, but I'm not sure it's the right way to go. Really, it might be better to export a function that takes the arguments I added to the config here, to make it easier for plugin publishers to implement their own preset. That approach:
* would actually be better for my use case
* would allow for the usage of function arguments instead of the templates

I still like the idea of exposing the template strings in the end user config, as even with the "preset wrapper" approach end users might want to customize the `import` statements. However, the fact that we're already interpolating environment variables makes the API awkward (didn't this was the case at first)
</details>

### DIFF
--- a/packages/presets/near-operation-file/package.json
+++ b/packages/presets/near-operation-file/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@graphql-codegen/near-operation-file-preset",
-  "version": "1.9.2",
+  "version": "1.9.2-alpha1",
   "description": "GraphQL Code Generator preset for generating operation code near the operation file",
   "repository": "git@github.com:dotansimha/graphql-code-generator.git",
   "license": "MIT",
--- a/packages/presets/near-operation-file/package.json
+++ b/packages/presets/near-operation-file/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@graphql-codegen/near-operation-file-preset",
-  "version": "1.9.3",
+  "version": "1.9.1",
   "description": "GraphQL Code Generator preset for generating operation code near the operation file",
   "repository": "git@github.com:dotansimha/graphql-code-generator.git",
   "license": "MIT",
--- a/packages/presets/near-operation-file/package.json
+++ b/packages/presets/near-operation-file/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@graphql-codegen/near-operation-file-preset",
-  "version": "1.9.1",
+  "version": "1.9.2",
   "description": "GraphQL Code Generator preset for generating operation code near the operation file",
   "repository": "git@github.com:dotansimha/graphql-code-generator.git",
   "license": "MIT",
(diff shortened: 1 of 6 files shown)


micimize:
@dotansimha could you take a look at this when you get the chance? I'm already using it in the latest https://github.com/micimize/graphql-to-dart beta

packages/presets/near-operation-file/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@graphql-codegen/near-operation-file-preset",
-  "version": "1.9.1",
+  "version": "1.9.3",
```
dotansimha: @micimize can you please reset this one?

packages/presets/near-operation-file/src/index.ts
```diff
@@ -18,7 +21,7 @@ export type NearOperationFileConfig = {
    * src/:
    *  preset: near-operation-file
    *  presetConfig:
-   *    baseTypesPath: types.ts
+   *    schemaTypesPath: types.ts
```
dotansimha: Do we really need to change that? because it might be a breaking change

dotansimha:
Looks great, thanks @micimize . can you please revert the version bump? Then I'll merge this.

micimize:
@dotansimha reverted the version bump 👍

PR #3156, opened 2019-12-21
[kotlin] remove constructor method from generated code

omuomugin:
Thanks for this great tool.
In data class in Kotlin, usually constructors are not needed. 
Therefore in this PR I removed constructors from generated code.

I guess nothing will affect and it will be more Kotlin like code.

If I was missing something please tell me.

Refs
https://kotlinlang.org/docs/reference/data-classes.html

### DIFF
--- a/packages/plugins/java/kotlin/src/visitor.ts
+++ b/packages/plugins/java/kotlin/src/visitor.ts
@@ -66,7 +66,7 @@ export class KotlinResolversVisitor extends BaseVisitor<KotlinResolversPluginRaw
 
     return `${comment}enum class ${enumName}(val label: String) {
 ${enumValues}
-        
+
   companion object {
     @JvmStatic
     fun valueOfLabel(label: String): ${enumName}? {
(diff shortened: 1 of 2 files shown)


dotansimha:
Thanks @omuomugin !

msmsmsmsms:
Maybe I am missing something here, but I do not think that the constructor was redundant.
The constructor was accepting a Map<String, Any>, which allowed us (using graphql-java) to just throw env.arguments in there. 

Using @graphql-codegen/kotlin@1.8.1 our code looked sth like this:
```
// generated
data class QueryWikisArgs(
    val page: Int = 0,
    val size: Int = 20
) {
  constructor(args: Map<String, Any>) : this(
      args["page"] as Int? ?: 0,
      args["size"] as Int? ?: 20
  )
}

// our resolver/fetcher
val wikis: DataFetcher<List<WikiDto>> = DataFetcher { env ->
        val queryArgs = QueryWikisArgs(env.arguments)    //   <-- does not work with data class without a constructor
       ...
}
```

Our old code does not work without the constructor.

Using @graphql-codegen/kotlin@1.17.1 our code would look like:
```
// generated
data class QueryWikisArgs(
    val page: Int = 0,
    val size: Int = 20
)

// our resolver/fetcher
val wikis: DataFetcher<List<WikiDto>> = DataFetcher { env ->
        val page = env.arguments["page"] as Int? ?: 0,
        val size = env.arguments["size"] as Int? ?: 20
        val queryArgs = QueryWikisArgs(page, size)
       ...
}
```

That kinda defeats of the code generator, because changes in our schema still require manual work and type casts in our code afterwards.
Maybe I am just not aware of some Kotlin language feature related to data classes?

PR #3157, opened 2019-12-21
Imperative query draft

KATT:
I needed a way in my app to imperatively call `client.query` and get the promise, was kinda surprised this wasn't built-in, but maybe I'm missing something?

Here's a work in progress that outputs something like this:

```typescript
export function queryFeed<T>(
  client: ApolloClient.ApolloClient<T>,
  options: Omit<
    ApolloClient.QueryOptions<FeedQueryVariables>,
    'query'
  >,
) {
  return client.query<FeedQuery>({
    query: FeedDocument,
    ...options,
  });
}
```

My use case is that I have a list of items on a page and want to prefetch the data on each item's "detail page". Might be another solution out there too.

### DIFF
--- a/dev-test/githunt/types.reactApollo.imperative.tsx
+++ b/dev-test/githunt/types.reactApollo.imperative.tsx
@@ -1,0 +1,410 @@
+// tslint:disable
+import gql from 'graphql-tag';
+import * as ApolloReactCommon from '@apollo/react-common';
+import * as ApolloClient from 'apollo-client';
+export type Maybe<T> = T | null;
+
+/** All built-in and custom scalars, mapped to their actual values */
+export type Scalars = {
+  ID: string;
+  String: string;
+  Boolean: boolean;
+  Int: number;
+  Float: number;
+};
+
+/** A comment about an entry, submitted by a user */
+export type Comment = {
+  __typename?: 'Comment';
+  /** The SQL ID of this entry */
+  id: Scalars['Int'];
+  /** The GitHub user who posted the comment */
+  postedBy: User;
+  /** A timestamp of when the comment was posted */
+  createdAt: Scalars['Float'];
+  /** The text of the comment */
+  content: Scalars['String'];
+  /** The repository which this comment is about */
+  repoName: Scalars['String'];
+};
+
+/** Information about a GitHub repository submitted to GitHunt */
+export type Entry = {
+  __typename?: 'Entry';
+  /** Information about the repository from GitHub */
+  repository: Repository;
+  /** The GitHub user who submitted this entry */
+  postedBy: User;
+  /** A timestamp of when the entry was submitted */
+  createdAt: Scalars['Float'];
+  /** The score of this repository, upvotes - downvotes */
+  score: Scalars['Int'];
+  /** The hot score of this repository */
+  hotScore: Scalars['Float'];
+  /** Comments posted about this repository */
+  comments: Array<Maybe<Comment>>;
+  /** The number of comments posted about this repository */
+  commentCount: Scalars['Int'];
+  /** The SQL ID of this entry */
+  id: Scalars['Int'];
+  /** XXX to be changed */
+  vote: Vote;
+};
+
+/** Information about a GitHub repository submitted to GitHunt */
+export type EntryCommentsArgs = {
+  limit?: Maybe<Scalars['Int']>;
+  offset?: Maybe<Scalars['Int']>;
+};
+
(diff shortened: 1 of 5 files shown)


dev-test/githunt/types.reactApollo.imperative.tsx
```diff
@@ -0,0 +1,411 @@
...
+  ${CommentsPageCommentFragmentDoc}
+`;
+export type CommentQueryResult = ApolloReactCommon.QueryResult<CommentQuery, CommentQueryVariables>;
+
+export function queryComment<T>(client: ApolloClient.ApolloClient<T>, options: Omit<ApolloClient.QueryOptions<CommentQueryVariables>, 'query'>) {
+  return client.query<CommentQuery>({
+    query: CommentDocument,
+    ...options,
+  });
+}
```
KATT: 👍

KATT:
I also wrote this monster to pass in a generated useLazyQuery and a set of variables to sequentially  load

```typescript
import * as ApolloReactHooks from '@apollo/react-hooks';
import { useState, useRef, useEffect } from 'react';
import { isEqual } from 'lodash';

export type LazyQueryFn<TQuery, TData, TVars> = (
  opts?: ApolloReactHooks.LazyQueryHookOptions<TQuery, TVars>,
) => ApolloReactHooks.QueryTuple<TData, TVars>;

export function usePreloadLazyQuery<TQuery, TData, TVars>(
  useLazyQueryFn: LazyQueryFn<TQuery, TData, TVars>,
  variablesList: TVars[],
) {
  const [loadOne, { data, loading }] = useLazyQueryFn({
    fetchPolicy: 'cache-first',
  });
  const [queue, setQueue] = useState(variablesList);
  const prevVariablesList = useRef(variablesList);
  const abortController = useRef(new window.AbortController());

  useEffect(() => {
    if (!isEqual(variablesList, prevVariablesList.current)) {
      // reset queue
      setQueue(variablesList);
    }
    prevVariablesList.current = variablesList;
  }, [variablesList]);

  // when component unmounts
  useEffect(() => {
    return () => {
      // TODO FIXME: below triggers forever loading with `useApolloNetworkStatus`
      // abortController?.current.abort();
    };
  }, [loadOne]);

  useEffect(() => {
    if (loading && !data) {
      // prev one isn't
      // - loaded, or
      // - propagated through cache
      return;
    }
    const variables = queue.shift();

    if (!variables) {
      // done
      return;
    }
    loadOne({
      variables,
      context: {
        fetchOptions: {
          signal: abortController.current.signal,
        },
      },
    });
  }, [data, loadOne, loading, queue]);
}
````

File: dev-test/githunt/types.reactApollo.imperative.tsx (hunk 1 above)
Comment (ardatan) [reply dev-test/githunt/types.reactApollo.imperative.tsx]: This can be another plugin because it is not react-specific, is it?

**Comment by ardatan** [comment]
@KATT I think this is not a React-specific code generation, so making this a seperate plugin would be better idea in my opinion.

**Comment by CarloPalinckx** [comment]
@KATT 
> I needed a way in my app to imperatively call client.query and get the promise
What is your actual use-case for this? It seems to me that you can just use the generate lazy queries for this?

If you want don't want to use the client from apollo's context then you can also just pass in the client as an option to the lazy query. You can also load them sequentially if you want with the `onCompleted` callback.

```tsx
const [loadGreeting] = useLazyQuery(GET_GREETING);
const [loadName] = useLazyQuery(GET_NAME, {
  onCompleted(response) {
      loadGreeting({
        variables: {
            name: response.name,    
        },
      }),
  });
});

loadName();
```
It seems that these changes are more a client-library issue rather than something you would fix with a code-generator. Would be happy to dig out this issue further tho if I missed something 🙌

KATT:
@CarloPalinckx oh yeah it could be cleaned up a bit by wrapping it in a promise. Not working on the project I needed this anymore though :)

rintaun:
I've recently started using next.js server-side rendering, and have ended up using `client.query` directly on the server side, e.g.

```typescript
export const getServerSideProps: GetServerSideProps<Props> = async ({
  params: { id },
}) => {
  const apolloClient = initializeApollo()

  await apolloClient.query({
    query: GetStoryDocument,
    variables: { id },
  })

  return {
    props: {
      initialApolloState: apolloClient.cache.extract(),
      id,
    },
  }
}
```

The problem with that approach is the loss of types by using `GetStoryDocument` directly (in fact, there is a possible bug lurking in the snippet above 😱). The approach suggested in this PR exactly fits that use case, and is what I came here to propose.

PR #3167, opened 2019-12-24
Fix apollo angular external mode

wassim-k:
There are currently 2 issues with generating apollo-angular operations with documentMode = external

1) It throws an exception in `ClientSideBaseVisitor` when it attempts to access `documents.length` because `ApolloAngularVisitor` is missing the `document` parameter when calling the parent constructor.
2) The document variable name is missing `Operations.` namespace which is defined by external mode.

This PR solves both issues.
For the second issue, it borrows from react's plugin implementation of the method `_getDocumentNodeVariable`

Finally, thanks for all the great work.


dotansimha:
Awesome. thanks @wassim-k :)

PR #3171, opened 2019-12-27
Add importDocumentNodeFragments config option

wassim-k:
I think `documentNode` mode has some great potential because unlike `external` and `graphQLTag` modes, it doesn't require client side dependency on `graphql-tag` nor does it require a custom webpack loader. And the latter is a big plus for angular projects due to build customization limitations.

Unfortunately, I found it unusable in practice due to 2 limitations:
1) Nested fragments are missing from generated code.
For example, if an operation has a fragment spread, and that fragment has a fragment spread of its own, that nested fragment definition is not included in the generated AST, causing a "No fragment named ..." error.
I'm not entirely sure if this limitation is by design or if it's an actual bug.
2) External fragments are embedded in the generated AST instead of being referenced.
This becomes even more of an issue once the first limitation is fixed, because 5 levels of fragment definition nesting will result in a large of amount of generated code, and if that fragment is referenced in multiple queries and mutations, then that's the same code duplicated in multiple files.

This draft PR is an effort to kick off a discussion about the best way to solve those limitations and attempts to do so by introducing a new flag `importDocumentNodeFragments`.
Any thoughts on whether this indeed should be a flag or replace the default behaviour of `documentNode` since this should only affect the generated code? (is that a breaking change?)

This is what the generated code looks like with this fix (minus formatting):
```typescript
import { SomeFragmentDoc } from './some.fragment';
import { OtherFragmentDoc } from './other.fragment';

export const SomeQueryDocument: DocumentNode = {
    "kind":"Document",
    "definitions": [
        {
        "kind":"OperationDefinition",
        "operation":"query"
        "name": ...
        ...
        },
        ...SomeFragmentDoc.definitions,
        ...OtherFragmentDoc.definitions
    ]
};
```

Looking forward to hearing your toughts & feedback.


packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -204,21 +223,25 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
...
       const gqlObj = gqlTag(doc);
-
       if (gqlObj && gqlObj['loc']) {
         delete gqlObj.loc;
       }
-
+      if (this.config.importDocumentNodeFragments && fragments.length > 0) {
+        const fragmentsSpreads = fragments.filter((name, i, all) => all.indexOf(name) === i).map(name => `...${name}.definitions`);
+        const definitions = [...gqlObj.definitions.map(JSON.stringify), ...fragmentsSpreads].join();
+        return `{"kind":"${Kind.DOCUMENT}","definitions":[${definitions}]}`;
```
dotansimha: It's a JS object, so no need for `"`

dotansimha:
That's a great idea! I think maybe this could be a new mode? then we can reuse the same existing configuration, and just extend it. so it could be `documentNode` for current behaviour or `documentNodeImportFragments` for the new one. 
What do you think?

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
wassim-k: I mainly used this format to keep it consistent with the rest of the stringified document since it's using `JSON.stringify`, and also to keep it consistent with the default `documentNode` mode which outputs the document with `"`. thoughts?

wassim-k:
I think it's a good idea, but since the current `documentNode` behaviour is incomplete, I would also like to a fix it for completeness sake so that it includes all required fragments, not only the ones directly referenced in the current operation.  
In order to achieve that, we can replace:
```typescript
return `${fragments
    .filter((name, i, all) => all.indexOf(name) === i)
    .map(name => {
        const found = this._fragments.find(f => `${f.name}FragmentDoc` === name);
        if (found) {
            return print(found.node);
        }
        return null;
    })
    .filter(a => a)
    .join('\n')}`;
```
with:
```typescript
return this._fragments.map(fragment => print(fragment.node)).join('');
```

Let me know if you're happy with those changes?

wassim-k:
I've updated the PR per suggestions.

wassim-k:
@dotansimha, can you please let me know what's the next step in progressing this PR?

jeffreyffs:
I just ran into this:
```Nested fragments are missing from generated code.
For example, if an operation has a fragment spread, and that fragment has a fragment spread of its own, that nested fragment definition is not included in the generated AST, causing a "No fragment named ..." error.```

and would love to see this fixed :)

dotansimha:
Thanks @wassim-k ! 
I'm merging this, and then I'll add some unit tests to make sure it works as we need.

PR #3172, opened 2019-12-27
[kotlin] Map ID to String instead of Any

omuomugin:
As I looked into code generation logic for TypeScript it seems that `ID` is mapped to `string`.
However in Kotlin it was mapped to `Any` so I changed to `String`.

Thanks.


dotansimha:
Thanks @omuomugin !

Issue #3102, opened 2019-12-10
ts1005 error with declarationKind: class


andreialecu:
```
config:
  declarationKind:
    input: 'class'
```
Results in classes being generated with commas instead of semicolons at end of lines, resulting in TypeScript compilation errors: `TS1005: ';' expected`

As a concrete example:

```
export class SomeInput {
  code: Scalars['String'], // should be ;
};
```

dotansimha:
It's actually more than that, we need to have `declare module` in some way, and have special care of `scalars` type. I'll take a look soon :) 

Thanks @andreialecu

andreialecu:
In my case, simply replacing all commas with semicolons makes everything work properly, but I'm not using any custom scalar types, just the built in ones.

dotansimha:
@andreialecu can you please share the complete (fixed) output? so I'll be able to test that directly in TS Playground. 
Thanks

andreialecu:
@dotansimha I'm not sure what you mean, the only problem with the output with this config

```
overwrite: true
schema: "http://localhost:1338/graphql"
documents: "./source/**/*.graphql"
config:
  declarationKind:
    input: 'class'
generates:
  source/generated/graphql.ts:
    plugins:
      - "typescript"
      - "typescript-operations"
      - "typescript-apollo-angular"
  ./graphql.schema.json:
    plugins:
      - "introspection"
```
Is the comma at end of class declarations instead of a semicolon. Everything else is fine, and I have a pretty big graphql schema. The generated `graphql.ts` file is over 1000 lines long. Everything is fine except for those lines.

andreialecu:
Here's an example from the generated output.

![image](https://user-images.githubusercontent.com/697707/70925310-b3316080-2033-11ea-8368-7a46b60de7ad.png)

dotansimha:
I see, but I suspect `Scalars` might cause issues with other TypeScript configurations, that's why I asked for the complete file, I want to verify that.

dinana:
I'm facing the same issue and here's the complete output with scalars class
![image](https://user-images.githubusercontent.com/1944157/71579481-66568b00-2b05-11ea-8ef9-c8d4a9ff96a3.png)

dotansimha:
@dinana @andreialecu I think I managed to fix that issue in: https://github.com/dotansimha/graphql-code-generator/pull/3183 

Can you please try `1.9.2-alpha-f125efa1.44`?

dotansimha:
[...]

PR #3183, opened 2019-12-31
Fix declarationKind=class syntax errors

dotansimha:
related: https://github.com/dotansimha/graphql-code-generator/issues/3102


738:
@dotansimha As well as `class`, i think `interface`, `type`, `abstract class` should be changed into semicolon, not comma. (except `enum`)

can I contribute?

dotansimha:
@738 this could really help :) PRs are always welcome!

Issue #3137, opened 2019-12-17
Enums with numeric values defined in enumValues are not generated correctly


BigsonLvrocha:
**Describe the bug**
 

There are two, enum given numeric values in enumValues config options are generated as string and value `0` will not be converted at all

**To Reproduce**

give numeric values to map enum fields in enumValues config option

1. My GraphQL schema:

```graphql
enum FooEnum {
  FOO
  BAR
}
```

2. My GraphQL operations:

no problems with operation, just with generated type, graphql will convert correctly

3. My `codegen.yml` config file:

```yaml
overwrite: true
schema: './data/schema.graphql'
documents: null
generates:
  packages/api/src/generated/schema.ts:
    plugins:
      - add: '/* istanbul ignore file */'
      - typescript
      - typescript-resolvers
    config:
      namingConvention:
        enumValues: keep
      enumValues:
        FooEnum:
          FOO: 1
          BAR: 0
```

**Expected behavior**

The generated code should be
```typescript
export enum FooEnum {
  FOO = 1
  BAR = 0
}
```
but it is 
```typescript
export enum FooEnum {
  FOO = '1'
  BAR = 'BAR'
}
```

https://codesandbox.io/s/graphql-codegen-issue-template-1lmxr

**Environment:**

- OS: Linux
- `@graphql-codegen/...`: 
```json
{
    "@graphql-codegen/add": "^1.8.3",
    "@graphql-codegen/cli": "^1.9.1",
    "@graphql-codegen/introspection": "^1.9.1",
    "@graphql-codegen/typescript": "^1.9.1",
    "@graphql-codegen/typescript-resolvers": "^1.9.1",
}
```
- NodeJS: 12.13.0

**Additional context**

graphql will convert correctly because I have the enum resolver defined explicitly as

```typescript
export const resolvers = {
  FooEnum: {
    FOO: 1,
    BAR: 0,
  },
};
```

dotansimha:
@BigsonLvrocha I think I managed to fix that in: https://github.com/dotansimha/graphql-code-generator/pull/3187 

Can you please try `1.9.2-alpha-4be25b2b.82`?

dotansimha:
Available in `v1.10.0`.

BigsonLvrocha:
I was on a trip, got the computer now, it's fixed, thanks :)

PR #3187, opened 2019-12-31
Fix for enumsValues and explicit numeric value

dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/3137

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/utils.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/utils.ts
@@ -48,7 +48,7 @@ export function block(array) {
 }
 
 export function wrapWithSingleQuotes(value: string | number | NameNode): string {
-  if (typeof value === 'number' || (typeof value === 'string' && !isNaN(parseInt(value)))) {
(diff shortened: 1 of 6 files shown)


packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -47,8 +47,12 @@ export function block(array) {
   return array && array.length !== 0 ? '{\n' + array.join('\n') + '\n}' : '';
 }
 
-export function wrapWithSingleQuotes(str: string | NameNode): string {
-  return `'${str}'`;
+export function wrapWithSingleQuotes(value: string | number | NameNode): string {
+  if (typeof value === 'number' || (typeof value === 'string' && !isNaN(parseInt(value)))) {
```
ardatan: This can be dangerous because `parseInt('2a')` will return 2 as in `parseInt('2')`

ardatan: The final code can be;
```ts
if (parseFloat(value).toString() === value.toString()) {
   return value; // Number
}
```

PR #3223, opened 2020-01-06
Plugin: JSDoc operations

CarloPalinckx:
This PR adds support for generating JSDoc types based on your graphql schema.


packages/plugins/other/jsdoc-operations/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@graphql-codegen/jsdoc",
+  "version": "1.8.3",
```
dotansimha: Can you please bump it to the latest? `1.10.0`

packages/plugins/other/jsdoc-operations/src/index.ts
```diff
@@ -0,0 +1,118 @@
+import { PluginFunction } from '@graphql-codegen/plugin-helpers';
+import { printSchema, parse, visit } from 'graphql';
+
+type NamedType = {
```
dotansimha: I think maybe you can use the visitor to return directly the output string, it might simplify the implementation because you'll be able to skip the need to those proxy types.

packages/plugins/other/jsdoc-operations/src/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+type TypeDef = {
+  name: string;
+  operation: string;
+  properties?: Array<Property>;
+  types?: Array<NamedType>;
+};
+
+const transformScalar = (scalar: string) => {
+  const scalarMap = {
```
dotansimha: You can use `@graphql-codegen/visitor-plugin-common` and import it (`DEFAULT_SCALARS`) from there

packages/plugins/other/jsdoc-operations/src/index.ts
CarloPalinckx: Was actually thinking about reusing `createTypedef` at some later point for other plugins as well, since the creation of a JSDoc typedef is consistent across those uses. I'd prefer a bit more extensive interfacing over having to duplicate the printing behavior :). What do you think?

packages/plugins/other/jsdoc-operations/package.json
CarloPalinckx: I'll leave this comment unresolved so I can bump it once if this PR get's closer to a merge 🙂

dotansimha:
@CarloPalinckx I think this PR is almost ready, right? 
Can you please align the versions, rebase and make sure CI works? Then we can merge this :) 

Thank you!!

CarloPalinckx:
@dotansimha this only generates jsdoc blocks for types in the schema and not for the operations, do you think this is mergeable regardless? I doubted that a bit because it diverts a lot from other plugins (like typescript-operations).

CarloPalinckx:
@dotansimha also, this PR doesn't apply the naming conventions that the users configure. Any pointer for that would be welcome 🙌

CarloPalinckx:
Just hit a few bumps in this PR when generating operation types, JSDoc does not have something akin to typescript's `Pick`. Making this PR suddenly a lot harder. The only solution I see for generating operation types is:

Schema:
```gql
type Bar {
  bar: Int!
}

type Query {
  foo: Int!
  bar: Bar!
}
```

With operation:
```gql
query GetFoo {
  foo
  bar {
    bar
  }
}
```

Would result into a bunch of intermediate types to be generated:
```js
/** 
 * @typedef {Object} GetFooQueryBar
 * @property {number} bar
 */

/**
 * @typedef {Object} GetFooQuery
 * @property {number} foo
 * @property {GetFooQueryBar} bar
 */
```
Wich will only increase exponentionally the deeper you query goes.

This isn't undoable but I wonder if anyone sees a better solution for this 🙂

CarloPalinckx:
@dotansimha sorry to let this PR stale for so long. I am allocating some time to get this fixed this week. Since generating operation types for JSdoc (last comment) I suggest we omit it from this PR for the initial release, that way users will at least have JSdoc types based on their GQL schema, which is already a nice feature on it's own imo. What do you think?

dotansimha:
@CarloPalinckx This sounds good, maybe we can have it for schema, and in the future extend that to operations. 
Thank you!

CarloPalinckx:
> This sounds good, maybe we can have it for schema, and in the future extend that to operations.

@dotansimha it should be ready for a merge if the review doesn't bring up issues. I'll rebase it again and will allign the version 👍

dotansimha:
This is awesome. Thank you and sorry for the late reply :)

PR #3288, opened 2020-01-14
Parse & stringify gql when documentMode=string

tmcw:
documentMode=string is perfect for graphql-request, but it simply
inlines any referenced fragments. This means that fragments that
are included more than once in a single query for example, by being
included by another fragment in that query, as well as at the top level -
are thus duplicated in the query, and then fail when send to the server.

This PR parses and stringifies queries ahead of time such that
no duplicates occur.

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -175,7 +175,7 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
 
   protected _includeFragments(fragments: string[]): string {
     if (fragments && fragments.length > 0) {
-      if (this.config.documentMode === DocumentMode.documentNode) {
+      if (this.config.documentMode === DocumentMode.documentNode || this.config.documentMode === DocumentMode.string) {
         return `${fragments
           .filter((name, i, all) => all.indexOf(name) === i)
           .map(name => {
@@ -222,7 +222,7 @@ export class ClientSideBaseVisitor<TRawConfig extends RawClientSideBasePluginCon
 
       return JSON.stringify(gqlObj);
     } else if (this.config.documentMode === DocumentMode.string) {
-      return '`' + doc + '`';
+      return '`' + print(gqlTag(doc)) + '`';
     }
 
     return 'gql`' + doc + '`';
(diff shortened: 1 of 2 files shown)


ardatan:
@tmcw Thank you for your PR. Even if this works, this shouldn't be the way to do that because this is only a plugin that prints some output related to schemas and documents which are loaded already by other packages/libraries etc. Loading, fixing and modifying the schema shouldn't be done by plugin itself in my opinion :)

tmcw:
Okay, it's just confusing in that sense why graphql-code-generator would provide a string output at all, because people will always need to parse & regenerate those strings. The goal here is that if you have a string output mode, the queries it generates should be runnable. Right now they aren't - because they include duplicates, and the one thing that this PR fixes is that it removes those duplicates. Otherwise everyone needs to ship a graphql parser & stringifier, which - afaict, is really not necessary in the simplest cases.

ardatan:
@tmcw If the document string needs to be parsed and stringified, this is the problem of the loader that loads the document. So this should be done somewhere else (loader maybe codegen's core package) because `graphql-request` is not the only place we have string output for documents.
Also `print(parse(` looks like a workaround for a broken document, instead there might be a function or something that deduplicates stuff in that document.
@dotansimha What do you think?

tmcw:
The documents themselves aren't broken - this is addressing well-formed documents that end up including the same fragment twice, like:

- A doc
  - B frag
  - C frag
    -B frag

Because the C frag includes B frag, B is included twice. The documents on their own are valid, but they can't be naively combined.

ardatan:
@tmcw Let me change my comment then
> If the document string needs to be parsed and stringified, this is the problem of the loader that loads the document. So this should be done somewhere else (loader maybe codegen's core package) because graphql-request is not the only place we have string output for documents.
Also print(parse( looks like a workaround for a `needs-to-be-changed` document, instead there might be a function or something that deduplicates stuff in that document.

ardatan:
@dotansimha What do you think?

dotansimha:
@ardatan , I agree with you. Some plugins works with strings, and it’s not codegen’s responsibility to manage that.
But - what do you think should be done in order to improve that?

I think we need something like `removeDuplicateFragments(…)` when used with `documentMode=string`. We do need to remove duplicates, otherwise it’s not executable, and the duplication is created by the codegen - so we need to fix that

How do you suggest to do this otherwise?

dotansimha:
Tracking here: https://github.com/prisma-labs/graphql-request/issues/176
When this will be supported, it will make it easier for this library.

Issue #3112, opened 2019-12-13
[typescript-resolvers] Allow resolvers to return lazy parameters or promises


cranberyxl:
**Is your feature request related to a problem? Please describe.**
I use a relay connection style in my schema. I have a single method that builds the connection fields, but returns them as functions like:

```
{
  pageInfo: () => PageInfo,
  count: () => number,
  edges: () => Edge[]
}
```
Unfortunately, the Resolver type expects them to be json or a promise returning json.

**Describe the solution you'd like**
I'd like to find a way that all of the fields can resolve as json, promise, or function without having a mapper for every connection. 

**Describe alternatives you've considered**
I looked at trying to return each of the connection fields as their own resolver, but I end up writing a lot more code and have to make a mapper for every connection. 

**Additional context**
Let me know if I'm overlooking a config, or way to writing this. 

Here's what I want my resolver to look like:
```
const eventsResolver: QueryResolvers['events'] = async (_, { input }) =>
  buildConnection(EventsQuery, input);
```

My Interfaces:
```
interface Node {
  id: ID!
}

interface Edge {
  cursor: String!
  node: Node!
}

type PageInfo {
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
}

interface Connection {
  pageInfo: PageInfo!
  count: Int!
  edges: [Edge!]!
}
```

dotansimha:
@cranberyxl You can use `defaultMapper` for that: https://graphql-code-generator.com/docs/plugins/typescript-resolvers#defaultmapper-string

ardatan:
@cranberyxl You mean providing custom `return` and `parent` types or providing factory functions of that type.
For example you have `User` like below in your schema;
```gql
type User {
 id: ID
 name: String
 books: [Book]
}
```
and the resolvers are likebelow;
```ts
{
 Query: {
   user: (_, args) => {
      return {
        id: 1,
        name: 'Arda',
        books: () => getBooks() // So books are given as function to be called on demand.
      }
    }
 }
}
```

Is that what you asked or something else?

cranberyxl:
@ardatan That's correct, because you can return any JSON from any resolver that satisfies it's children. Which may be a function or a promise.

dotansimha:
@ardatan do you think it's something we can implement in the generated output?

rintaun:
I think the underlying problem here is that, anywhere that can return a value could instead return:
* a Promise for that value
* a function that returns the value
[...]

PR #3298, opened 2020-01-15
fix: resolvers can be values rather than functions

rintaun:
Allows `Resolver` to be a value rather than just a function by adding `TResult` and `Promise<TResult>` as possible types.

Resolvers are generally functions, but they are also allowed to be values.

Related: #3112

### DIFF
--- a/packages/plugins/typescript/resolvers/src/index.ts
+++ b/packages/plugins/typescript/resolvers/src/index.ts
@@ -96,7 +96,7 @@ export type StitchingResolver<TResult, TParent, TContext, TArgs> = {
   const resolverFnUsage = `ResolverFn<TResult, TParent, TContext, TArgs>`;
   const stitchingResolverUsage = `StitchingResolver<TResult, TParent, TContext, TArgs>`;
 
-  const resolverDefParts = [resolverType, `  | TResult`, `  | Promise<TResult>`];
+  const resolverDefParts = [resolverType, `TResult`, `Promise<TResult>`, resolverFnUsage];
 
   if (visitor.hasFederation()) {
     defsToInclude.push(`export type ReferenceResolver<TResult, TReference, TContext> = (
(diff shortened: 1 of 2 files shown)


dotansimha:

packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -105,15 +107,23 @@ export type StitchingResolver<TResult, TParent, TContext, TArgs> = {
   }
 
   if (noSchemaStitching) {
-    // Resolver = ResolverFn;
-    defsToInclude.push(`${resolverType} ${resolverFnUsage};`);
+    // Resolver =
+    // | TResult
+    // | Promise<TResult>
+    // | ResolverFn;
+    resolverDefParts.push(`  | ${resolverFnUsage};`);
```
dotansimha: Why adding ` | ` to every string? You can add it without it, and then use `.join(' | ')`

rintaun: That's a good point. Now that there are always multiple, it shouldn't be necessary anymore. I'll update the PR

dotansimha: Thanks!
@kamilkisiela is the resolvers expert, he'll review it soon :)

kamilkisiela:
I'm not quite sure GraphQL allows to use primitive values or Promises as resolvers. I agree they are correct when being returned via function and passed on to child types.

```graphql
type Story {
  id: ID!
  isManageable: Boolean!
}

type Query {
  stories: [Story!]
  story(id: ID!): Story
}
```

```typescript
interface Story {
  id: string;
  title: string;
  isManageable: (() => boolean);
}
const stories: Array<Story>;

const resolvers = {
  Query: {
    story(_, { id }) {
      return stories.find(u => u.id === id);
    },
    stories,
  }
} 
```


The `Query.story` resolver will work, `isManageable` will be resolved correctly but I don't thing `Query.stories` will return something non-null value. Correct me if I'm wrong.

The `Resolver` type produced by Codegen with your addition will allow to put strings, numbers, objects and arrays or even all of them wrapped with a Promise. 

What we expect is a resolver that returns an object with fields where each can be a primitive value, a function or a promise.

rintaun:
I stand corrected. I don't know why I was under the impression that non-functional resolvers worked... Sorry for wasting your time 😬

PR #3299, opened 2020-01-15
feat: allow field definitions to be wrapped

rintaun:
Adds the optional (disabled by default) `FieldWrapper` type for wrapping field definitions on objects and interfaces (but _not_ input types).

In addition to returning a value of the actual type for a field, objects can also return Promises to that type or functions that do the same. It is currently not possible to achieve that by reconfiguring the code generator.

This adds the ability to wrap field definitions in a `FieldWrapper`, which defaults to `FieldWrapper<T> = T`, thus not changing behavior in any way. To get the above-described behavior, a user would need to set `fieldWrapperValue` to `T | Promise<T> | (() => T | Promise<T>)` or something similar in addition to enabling field wrapping by setting `wrapFieldDefinitions` to `true`.

That may be overkill, but I wanted to avoid altering the current existing behavior unexpectedly.

Closes: #3112

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -787,6 +787,10 @@ export type IDirectiveResolvers${contextType} = ${name}<ContextType>;`
     return `${resolversType}['${name}']`;
   }
 
+  protected getParentTypeForSignature(node: FieldDefinitionNode) {
+    return 'ParentType';
+  }
+
   protected transformParentGenericType(parentType: string): string {
     return `ParentType extends ${parentType} = ${parentType}`;
   }
@@ -829,7 +833,11 @@ export type IDirectiveResolvers${contextType} = ${name}<ContextType>;`
         }
       }
 
-      const parentTypeSignature = this._federation.transformParentType({ fieldNode: original, parentType, parentTypeSignature: 'ParentType' });
+      const parentTypeSignature = this._federation.transformParentType({
+        fieldNode: original,
+        parentType,
+        parentTypeSignature: this.getParentTypeForSignature(node),
+      });
       const mappedTypeKey = isSubscriptionType ? `${mappedType}, "${node.name}"` : mappedType;
 
       let signature: {
--- a/packages/plugins/typescript/resolvers/src/config.ts
+++ b/packages/plugins/typescript/resolvers/src/config.ts
@@ -39,6 +39,25 @@ export interface TypeScriptResolversPluginConfig extends RawResolversConfig {
    */
   noSchemaStitching?: boolean;
   /**
+   * @name wrapFieldDefinitions
+   * @type boolean
+   * @description Set to `true` in order to wrap field definitions with `FieldWrapper`.
+   * This is useful to allow return types such as Promises and functions. Needed for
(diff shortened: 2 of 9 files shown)


rintaun:
I just found a slight incompatibility with `ReferenceResolver` from the federation support in `typescript-resolvers`.

Since `ParentType` uses `ResolversParentTypes['SomeType']`, the value of fields in the `reference` object are typed as whatever `FieldWrapper` returns -- that is, however, incorrect, since those behave more like input types instead, i.e. they will always be resolved values rather than Promises or functions.

I have a fix that basically unwraps the types. I'll be pushing a new commit with that in the next day or two.

rintaun:
So I don't lose it, and as a sanity check, this is what I'm thinking. Given the following schema

```graphql
type Story @key(fields: "id") {
  id: ID!
  title: String!
}
```

The generated resolvers type would look something like this...

```typescript
export type FieldWrapper<T> = T | Promise<T> | (() => T | Promise<T>)
export type UnwrappedObject<T> = {
  [P in keyof T]: T[P] extends FieldWrapper<infer R> ? R : T[P]
}

export interface StoryResolvers<
  ContextType = IContext,
  ParentType extends ResolversParentTypes['Story'] = ResolversParentTypes['Story']
> {
  __resolveReference?: ReferenceResolver<
    Maybe<ResolversTypes['Story']>,
    { __typename: 'Story' } & Pick<UnwrappedObject<ParentType>, 'id'>,
    ContextType
  >
  id?: Resolver<ResolversTypes['ID'], ParentType, ContextType>
  title: String!
}
```

The key being that instead of `Pick<ParentType, 'id'>` as you would currently expect to see, you instead get `Pick<UnwrappedObject<ParentType>, 'id'>`, which pulls the wrapper off of each field.

rintaun:
This commit should address the issue I mentioned above.

dotansimha:
@kamilkisiela can you please review this?

kamilkisiela:
I believe you can just add `() => (T | Promise<T>)` [here](https://github.com/dotansimha/graphql-code-generator/blob/63fe83d65c52dd554753bfc1fd0ababc43c14353/packages/plugins/typescript/resolvers/src/index.ts#L69). We already handle return types of resolvers.

rintaun:
@kamilkisiela That is a similar problem, but this isn't about return types of resolvers; it is about return types of `type` objects, e.g. given

```graphql
type Story {
  id: ID!
  isManageable: Boolean!
}
```

it will current generate

```typescript
export interface Story {
  __typename?: 'Story'
  id: Scalars['ID']
  isManageable: Scalars['Boolean']
}
```

However, that `Scalars['ID']` doesn't can also be a promise or function, just like the return value from a resolver. Right now there is not a way to handle objects that have mixed values and functions for fields, even though they are valid, e.g. the following type:

```typescript
interface StoryModel {
  id: string
  isManageable(): boolean
}
```

See #3112 for some more concrete details on what this fixes

kamilkisiela:
Oh I get it, so now it LGTM

dotansimha:
Great! 
@rintaun can you please just rebase this PR?

rintaun:
Yep, I'll do that this evening. 👍

Thanks 🙂

rintaun:
@dotansimha my apologies for the delay. I've pushed the rebase 😄

Issue #3326, opened 2020-01-20
Incorrect "Not all operations have an unique name" on 1.11.2


benjie:
**Describe the bug**

1.11.2 is incorrectly stating "Not all operations have an unique name" when I attempt to run codegen; 1.8.3 was working fine.

**To Reproduce**

Codegen works fine on the Graphile Starter `codegen-fine` branch: https://github.com/graphile/starter/tree/codegen-fine

However, when we upgrade graphql-codegen from 1.8.3 to 1.11.2 we get the aforementioned error; see: https://github.com/graphile/starter/tree/codegen-bad which is one commit ahead of `codegen-fine`.

I've dug into this further; see "Additional context"

Running `graphql-codegen` results in:

```
 ✔ Parse configuration
  ❯ Generate outputs
    ❯ Generate index.tsx
      ✔ Load GraphQL schemas
      ✔ Load GraphQL documents
      ✖ Generate
        → Not all operations have an unique name: AddEmail


 Found 1 error

  ✖ index.tsx
    Not all operations have an unique name
  
    * AddEmail found in:
  
          - /home/benjie/Dev/starter/@app/client/src/graphql/AddEmail.graphql
          - /home/benjie/Dev/starter/@app/client/src/graphql/DeleteEmail.graphql
          - /home/benjie/Dev/starter/@app/client/src/graphql/EmailsForm_User.graphql
          - /home/benjie/Dev/starter/@app/client/src/graphql/SettingsEmails.graphql
          - /home/benjie/Dev/starter/@app/client/src/graphql/SettingsProfile.graphql
    Error: Not all operations have an unique name: AddEmail
        at validateDuplicateDocuments (/home/benjie/Dev/starter/node_modules/@graphql-codegen/core/index.cjs.js:200:1
5)
        at Object.codegen (/home/benjie/Dev/starter/node_modules/@graphql-codegen/core/index.cjs.js:59:9)
        at process (/home/benjie/Dev/starter/node_modules/@graphql-codegen/cli/bin.js:783:67)
        at Array.map (<anonymous>)
        at /home/benjie/Dev/starter/node_modules/@graphql-codegen/cli/bin.js:790:63
        at processTicksAndRejections (internal/process/task_queues.js:93:5)
        at async Task.task (/home/benjie/Dev/starter/node_modules/@graphql-codegen/cli/bin.js:582:17)
    Error: Not all operations have an unique name: AddEmail
[...]

PR #3328, opened 2020-01-20
Better validation for duplicate operations

ardatan:
Related #3326


lorado:
Hi @ardatan 
I just wanted to test your changes. With `1.11.3-alpha-841a8693.43` it works now. I didn't dive in the code, so I don't know if there is still something to do, but my issue was solved.  Waiting for merging and next release. Thanks!

lorado:
@ardatan sorry, I was a little too fast with my comment... Generation of types works without error, but it seems that some of my queries are simply ignored. i.e. I have two queries: `GetQualifications` and `GetAllQualifications`. In the latest version of the Package `GetQualifications` gets ignored. 

I currently rolled back to `1.9.1` again, as there it works without problems. Let me know, if i can somehow help you catch the error or test new fix in my project

ardatan:
@lorado I think I need a reproduction. Codesandbox would be the best :)

Issue #3322, opened 2020-01-19
typescript-resolvers uses wrong type for all-optional args


sgoll:
**Describe the bug**

`typescript-resolvers` can be used to generate type signatures for field resolvers.

When all args of a field are optional, the type of each arg should be `T | null | undefined` as missing args are passed to the resolver as `undefined` when resolving the query.

However, `typescript-resolvers` only generates the signature `T | null` which may lead to subtle bugs, e.g. when `args.foo !== null` is used to check if an arg is set.

If at least one of the args to a field is required, the correct signature is generated.

**To Reproduce**

Use https://graphql-code-generator.com/#live-demo to see the bug in action.

1. My GraphQL schema:

```graphql
schema {
  query: Query
}

type Query {
  allOptional(optional: Int): Boolean
  someRequired(optional: Int, required: Int!): Boolean
}
```

2. My `codegen.yml` config file:

```yml
generates:
  server-types.ts:
    plugins:
      - typescript
      - typescript-resolvers
```

3. The generated output (notice the missing `RequireFields` wrapper in `allOptional`):

```typescript
export type QueryResolvers<ContextType = any, ParentType extends ResolversParentTypes['Query'] = ResolversParentTypes['Query']> = {
  allOptional?: Resolver<Maybe<ResolversTypes['Boolean']>, ParentType, ContextType, QueryAllOptionalArgs>,
  someRequired?: Resolver<Maybe<ResolversTypes['Boolean']>, ParentType, ContextType, RequireFields<QuerySomeRequiredArgs, 'required'>>,
};
```

**Expected behavior**

The output should generate a type for `allOptional`'s field resolver that ensures that all args may be `undefined`.

`RequireFields` does this implicitly by marking any fields not listed in its second generic argument as optional (the first part in the following definition, before `&`):

```typescript
export type RequireFields<T, K extends keyof T> = { [X in Exclude<keyof T, K>]?: T[X] } & { [P in K]-?: NonNullable<T[P]> };
```

That is why the `optional` arg in the field `someRequired` in the example above is correctly typed as `T | null | undefined`.

It seems that `RequireFields<QueryAllOptionalArgs, never>` could achieve this, i.e. instead of dropping `RequireFields` when no args are required use `never` for the list of required arg names/keys.

**Environment:**

- OS: any
- `@graphql-codegen/typescript-resolvers`: 1.11.0
- NodeJS: any

dotansimha:
Fixed in `v1.13.0`

PR #3344, opened 2020-01-22
fix: include undefined as type for all-optional argument lists

sgoll:
The motivation behind this change is lined out in more detail in dotansimha/graphql-code-generator#3322. The basic problem is that `RequireFields` includes provisions to make non-required arguments optional (via `?:` mapping) but this wrapper type only gets applied when there is at least one required argument in the list.

However, argument lists with all-optional arguments must have the optional mapping applied as well. If not, they end up including only `null` but not `undefined` in their types which can lead to subtle runtime errors.

Related #3322


dotansimha:
That's awesome :) Sorry for the late response.
Thanks @sgoll !

PR #3346, opened 2020-01-22
Use graphql-toolkit signature for documents and clean relay optimizer codes

ardatan:
Less iteration and it would increase the performance for large projects.


dfreeman:
This was a breaking change to the plugin API that went out in a minor release, wasn't it? Was this messaged anywhere?

dotansimha:
I think you are right, @dfreeman :( It got missed from the changelog :(

PR #3384, opened 2020-01-26
Fix edge cases from multiple fragment spreads in `typescript-compatibility` plugin

jamesreggio:
We encountered some codegen issues in the `typescript-compatibility` plugin pertaining to multiple fragment spreads.

The primary problem can be illustrated through this simple example:

```graphql
fragment FullUser on User {
  id
  ...UserInfo
  ...UserFriends
}
```

This fragment will generate the following namespace:

```typescript
export namespace FullUser {
  export type Fragment = UserFriendsFragment; // This is wrong. It omits the fields from UserInfoFragment
  // ...
}
```

Effectively, when you have more than one named fragment spread, the type of the last fragment will become the type of the containing fragment.

Fixing this issue took me down a bit of a rabbit hole, because I had to work to intuit the purpose of the `Kind.FRAGMENT_SPREAD` case in the code generator. Ultimately, I discovered that that case was not specifically necessary, and that the `Kind.INLINE_FRAGMENT` case suffered from the same problem.

I added a number of regression tests to illustrate the various cases. There's one edge case that was very difficult to solve, which I've explained in the diff below.

### DIFF
--- a/packages/plugins/typescript/compatibility/src/selection-set-to-types.ts
+++ b/packages/plugins/typescript/compatibility/src/selection-set-to-types.ts
@@ -83,13 +83,6 @@ export function selectionSetToTypes(
           break;
         }
 
-        case Kind.FRAGMENT_SPREAD: {
-          const fragmentName = baseVisitor.convertName(selection.name.value, { suffix: 'Fragment' });
-          result[typeToUse] = { export: 'type', name: fragmentName };
-
-          break;
-        }
-
         case Kind.INLINE_FRAGMENT: {
           const typeCondition = selection.typeCondition.name.value;
           const fragmentName = baseVisitor.convertName(typeCondition, { suffix: 'InlineFragment' });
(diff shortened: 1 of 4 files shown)


packages/plugins/typescript/compatibility/src/selection-set-to-types.ts
```diff
@@ -77,13 +70,31 @@ export function selectionSetToTypes(
...
+            const subSelections = selection.selectionSet.selections
+              .map(subSelection => {
+                switch (subSelection.kind) {
+                  case Kind.FIELD:
+                    return `'${subSelection.name.value}'`;
+                  case Kind.FRAGMENT_SPREAD:
+                    return `keyof ${baseVisitor.convertName(subSelection.name.value, { suffix: 'Fragment' })}`;
+                  case Kind.INLINE_FRAGMENT:
+                    encounteredNestedInlineFragment = true;
+                    return null;
```
jamesreggio: Nested inline fragments cannot be effectively handled because their name cannot be resolved until the next call of `handleTypeNameDuplicates`, which requires recursion into `selectionSetToTypes`. This is an obscure use case, so we just throw.

I think a warning would be better, but I haven't seen infrastructure for emitting warnings. If so, please let me know and we can change this.

jamesreggio:
@dotansimha can you help me understand why the tests are failing? It looks like the magic `/* GraphQL */` comment isn't being interpreted correctly, and as a result, all of my new tests are failing in the AST-construction phase.

jamesreggio:
Hi @dotansimha, any chance you can take a look at this?

dotansimha:
@jamesreggio Sorry for the delay! We did a lot of improvements to build/test flow, so it should work now.
I rebased your branch, fixes the snapshots as well.

Let's see if it passes now, I think we can merge it :) 

Thank you so much!

jamesreggio:
Thank you so much, @dotansimha!

Issue #3412, opened 2020-01-29
Failed to install using yarn on node 8 (mkdirp 1.0.3)


polRk:
Failed to install using yarn on node 8
```
"@graphql-codegen/cli@^1.11.2":
  version "1.12.1"
  resolved "https://registry.yarnpkg.com/@graphql-codegen/cli/-/cli-1.12.1.tgz#7d7535f0dc6f12285ec987dd7b734de5e5395f67"
  integrity sha512-ObbuUaBC48i8glFXcBNo9oBMfv9HZ4JQtP52UX6Ppv/KUOISQWCYvC1YmtWIkuVYRz65Ds2kHpNS9YknVf7eEQ==
  dependencies:
    mkdirp "1.0.3"
```

My firebase functions using node 8, and i can't install dependencies

```
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
error mkdirp@1.0.3: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.16.1"
```

polRk:
I don't want to use --ignore-engines

polRk:
```
⚠  functions[webhooks-deleteBot(us-central1)]: Deployment error.
Build failed: {"error": {"canonicalCode": "INVALID_ARGUMENT", "errorMessage": "`yarn_install` had stderr output:\nerror mkdirp@1.0.3: The engine \"node\" is incompatible with this module. Expected version \">=10\".\nerror Found incompatible module\n\nerror: `yarn_install` returned code: 1", "errorType": "InternalError", "errorId": "EBA42403"}}
⚠  functions[bots-viber(us-central1)]: Deployment error.
Build failed: {"error": {"canonicalCode": "INVALID_ARGUMENT", "errorMessage": "`yarn_install` had stderr output:\nerror mkdirp@1.0.3: The engine \"node\" is incompatible with this module. Expected version \">=10\".\nerror Found incompatible module\n\nerror: `yarn_install` returned code: 1", "errorType": "InternalError", "errorId": "EBA42403"}}
⚠  functions[auth-userCreate(us-central1)]: Deployment error.
[...]

PR #3413, opened 2020-01-29
Use mkdirp "^0.5.1" instead of 1.0.3

polRk:
fix: #3412


ardatan:
See my comment
https://github.com/dotansimha/graphql-code-generator/issues/3412#issuecomment-579718583

Issue #3259, opened 2020-01-09
Fragment Doc get lost after upgrading above the 1.9.1


ivansky:
**Describe the bug**
On 1.9.1 there is an import from Graphql file to use `SomeFragmentDoc` string.
After upgrading above 1.9.1 it doesn't work properly, I didn't tested alpha version, there are too much of them.
I don't know maybe it's intended behaviour and I need to use some kinda import feature, but I didn't need before.

**To Reproduce**
There is a repo
https://github.com/ivansky/graphql-codegen-test-repo/tree/babc5efd6e757bcfb1fe3dcfeac2851f00f6866b

And commit that describes the issue
https://github.com/ivansky/graphql-codegen-test-repo/commit/babc5efd6e757bcfb1fe3dcfeac2851f00f6866b

1. My GraphQL schema:

```graphql
schema {
    query: Query
}

type Query {
    check(
        product: ProductCheckInput!
    ): ProductCheckResult!
}

interface Service {
    id: ID!
    productId: String!
}

type AService implements Service {
    id: ID!
    productId: String!
    dooms: [[ADoom!]!]!
}

type ALoad {
    code: String!
}

type ADoom {
    loads: [ALoad!]!
    code: String!
    description: String
    searchCount: Int
}

type Product {
    rules: [String!]!
    id: ID!
    notes: [String!]!
    services: [Service!]!
}

type ProductCheckResult {
    product: Product
    success: Boolean!
}

input ProductCheckInput {
    searchId: String!
    searchProductId: String!
}

scalar Date

scalar DateTime

scalar Time

type Auth {
    accessToken: String!
}

type Client {
    auth: Auth
}

extend type Query {
    client: Client!
}

```

2. My GraphQL operations:

**AServiceFragment.graphql**
```graphql
fragment AServiceFragment on AService {
    id
    productId
}
```

**CheckProduct.graphql**
```graphql
query CheckProduct($product: ProductCheckInput!) {
    check(product: $product) {
        success
        product {
            notes
            rules
            services {
                ...AServiceFragment
                ...on AService {
                    dooms {
                        code
                        loads {
                            code
                        }
                    }
                }
            }
        }
    }
}
```

3. My `codegen.yml` config file:

```yml
overwrite: true
config:
  namingConvention:
    typeNames: change-case#pascalCase
    transformUnderscore: true
schema:
  - ./schema.api.graphql
  - ./schema.client.graphql
generates:
  types/types.ts:
    documents:
[...]

PR #3436, opened 2020-02-03
Fix missing FragmentDoc import

ardatan:
Related https://github.com/dotansimha/graphql-code-generator/issues/3259


mvestergaard:
This is broken with the following config on 1.12.2, works fine on 1.12.1
```
documentMode: external
```

Issue #2932, opened 2019-11-11
`mappers` option results in duplicate identifiers in the generated TS file


helios1138:
having some types remapped using the `mappers` option leaves both the new (imported) type and the old (generated) type in the generated file.

apparently it wasn't an issue before, but now the new Typescript does not like that and I get
```
TS2440: Import declaration conflicts with local declaration
```

ardatan:
@helios1138 Could you share your `codegen.yml` file?

helios1138:
@ardatan 
```yml
overwrite: true
schema: src/graphql/**/*.graphql
documents: null
generates:
  src/types.ts:
    config:
      namingConvention:
        enumValues: keep
      scalars:
        Date: Date
        ID: ID<Node['__typename']>
        Upload: Promise<FileUpload>
      mappers:
        Node: ./graphql/resolvers#Node
        User: ./services/Users#User
        Thread: ./services/Threads#Thread
        # more types here...
      contextType: ./graphql/Context#Context
    plugins:
      - add:
          content:
            - import { ID } from './lib/idHelpers'
            - import { Connection } from 'core/src/Mongodb/pagination/getConnection'
            - import { FileUpload } from 'core/src/graphqlUpload'
      - typescript
      - typescript-resolvers
```
and so I get e.g. both imported types from `mappers`
```ts
import { Thread } from './services/Threads'
```
and the default generated types
```ts
export type Thread = Node & {
  __typename?: 'Thread'
  id: Scalars['ID']
  title: Scalars['String']
  // more fields here...
}
```

helios1138:
@ardatan I assume it's because `typescript` plugin does not know about `typescript-resolvers`' `mappers` option, so it doesn't know to ignore them. Is there a way to manually "skip" some types from being generated?

janhartmann:
I am hitting this issue as well, when upgrading to TypeScript 3.7.2 

The `as X` would be helpful or something like `import * as Something from "./something"`.

ardatan:
@helios1138 @janhartmann You can try to import those like below;
```yaml
mappers:
        Node: import('./graphql/resolvers').Node
        User: import('./services/Users').User
        Thread: import('./services/Threads').Thread
```

janhartmann:
That works! Generated code in my case is:

```MachineCluster: ResolverTypeWrapper<import('../api/location/LocationApi.swagger.generated').MachineCluster>,```

helios1138:
[...]

PR #3451, opened 2020-02-06
Add mapperTypeSuffix option

koenpunt:
this option allows you to add a suffix to all imported types to prevent name clashes with graphql types.

Followup of my comment here: https://github.com/dotansimha/graphql-code-generator/issues/2932#issuecomment-582862732

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/tests/parse-mapper.spec.ts
+++ b/packages/plugins/other/visitor-plugin-common/tests/parse-mapper.spec.ts
@@ -68,4 +68,75 @@ describe('parseMapper', () => {
       source: 'file',
     });
   });
+
+  describe('suffix', () => {
+    it('Should return the correct values for a simple named mapper', () => {
+      const result = parseMapper('MyType', null, 'Model');
+
+      expect(result).toEqual({
+        isExternal: false,
+        type: 'MyType',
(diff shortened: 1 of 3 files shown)


dotansimha:
Thank you @koenpunt ! This is a nice idea. Can you please add some basic unit tests for this? Thanks :)

packages/plugins/other/visitor-plugin-common/tests/parse-mapper.spec.ts
```diff
@@ -68,4 +68,75 @@ describe('parseMapper', () => {
...
+      const result = parseMapper('file#default', 'MyGqlType', 'Model');
+
+      expect(result).toEqual({
+        default: true,
+        isExternal: true,
+        import: 'MyGqlType',
+        type: 'MyGqlType',
+        source: 'file',
+      });
+    });
```
koenpunt: I'm not completely sure if this is the desired outcome; the `MyGqlType` is a type that's derived from the schema, right?

koenpunt:
Unit tests are added for the `parseMapper` function. Do I need to add a test for the option itself somewhere too?

dotansimha:
Looks good. Thanks @koenpunt !

PR #3479, opened 2020-02-11
Add withRefetchFn option

### LINKED ISSUE #3478: Generate functions to use for refetchQueries

The [`refetchQueries`](https://www.apollographql.com/docs/react/data/mutations/#options) seems super useful for us in contrast with update functions, since we would mostly be duplicating server logic. However, we are not fond of the fact that the variables aren't type checked to match the query:

```ts
{
  refetchQueries: [
    { query: GetUserDocument, variables: { name: 'user1' } }, // should type check
    { query: GetUserDocument, variables: { foo: 1 } }, // should not type check
  ],
}
```

It would be great if there could be an option that generates something of the form
```ts
export function refetchGetUserQuery(variables: GetUserQueryVariables) {
  return { query: GetUserDocument, variables: variables }
}
```

**Additional context**

brandon-leapyear:
Related: #3478

### DIFF
--- a/dev-test/codegen.yml
+++ b/dev-test/codegen.yml
@@ -299,6 +299,14 @@ generates:
     plugins:
       - typescript
       - typescript-operations
+  ./dev-test/star-wars/types.refetchFn.tsx:
+    schema: ./dev-test/star-wars/schema.json
+    documents: ./dev-test/star-wars/**/*.graphql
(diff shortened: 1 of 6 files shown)


dotansimha:
That's perfect :) Thank you @brandon-leapyear !

PR #3488, opened 2020-02-12
Add "-D" installation flag

xamgore:
Getting Started contains `-D` flag, same should be done by the plugins guides.

image


dotansimha:
Awesome! Thanks @xamgore !

PR #3489, opened 2020-02-12
feat: in watch mode remove stale files from the previous build

nodkz:
### In watch mode remove stale files from the previous build

Code is quite simple:
- ONLY in `watch` mode:
  - it remembers files from build to `previouslyGeneratedFilenames` array
  - for next builds it tries to find files in `previouslyGeneratedFilenames` which are not generated in the current build - such files are `stale`
  - if `overwrite: true` then `stale` files will be deleted.

I tested locally and it works perfectly. 

But I don't find a proper way how to write test cases for Jest. I didn't find any test for watch mode.


nodkz:
It will be great if you bump a new version. Thanks!

dotansimha:
Merged :) I'll update soon when the alpha will be published. 
We plan to have a new stable release soon ;)

PR #3505, opened 2020-02-14
Fix declarationKind=interface, type syntax errors

738:
related: #3102  #3183 

If declarationKind is `class`, `interface`, `type` in `typescript`, and `class`, `interface` in `flow`, the punctuation should be semicolon(;), not comma(,).


dotansimha:
This is awesome!!! Thank you so much @738 !

Issue #3237, opened 2020-01-06
Keys with collection values will be stringified as YAML due to JS Object restrictions


sergsalo:
Recently started to see this warning during codegen:

 (node:17948) YAMLWarning: Keys with collection values will be stringified as YAML due to JS Object restrictions. Use mapAsMap: true to avoid this. at line 6, column 11:

    JSON: { [key: string]: any }

codegen.yml:

```
overwrite: true
schema: "src/schema/**/*.graphql"
documents: null
scalars:
    DateTime: Date
    JSON: { [key: string]: any }
config:
    contextType: ../context#Context
generates:
  src/generated/graphql.ts:
    plugins:
      - "typescript"
      - "typescript-resolvers"
```

not sure where/how to use "mapAsMap" to fix the issue

ardatan:
Could you try to change that line to `JSON: "{ [key: string]: any }"`?

sergsalo:
yeah, that seems to work just fine.

sergsalo:
thanks for quick workaround.

ardatan:
You're welcome! The first one was not valid for YAML syntax. But the second one is valid because you want to pass a string that will be used in the generated code which is generated as string as well.

PR #3522, opened 2020-02-19
docs: fix config example in base-visitor.md (scalars - JSON)

738:
related: #3237 

```yaml
config:
  scalars:
    DateTime: Date
    JSON: { [key: string]: any }
```

This config file makes TS type definition file like this.

```typescript
export type Scalars = {
  ID: string;
  String: string;
  Boolean: boolean;
  Int: number;
  Float: number;
  JSON: {"[ key: string ]":"any"};
};
```

So it should be changed into this code.

```yaml
config:
  scalars:
    DateTime: Date
    JSON: "{ [key: string]: any }"
```

```typescript
export type Scalars = {
  ID: string;
  String: string;
  Boolean: boolean;
  Int: number;
  Float: number;
  JSON: { [key: string]: any };
};
```

### DIFF
--- a/docs/generated-config/base-visitor.md
+++ b/docs/generated-config/base-visitor.md
@@ -10,7 +10,7 @@ Extends or overrides the built-in scalars and custom GraphQL scalars to a custom
 config:
   scalars:
     DateTime: Date
-    JSON: { [key: string]: any }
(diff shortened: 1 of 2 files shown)


dotansimha:
Thank you @738 !

PR #3524, opened 2020-02-19
Use fragment prefix/suffix in client side base visitor

mjbcopland:
These config options are [declared in ClientSideBasePluginConfig](https://github.com/dotansimha/graphql-code-generator/blob/v1.12.2/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts#L72-L73) but not being used. Looks like a typo to use the document prefix/suffix.


dotansimha:
Perfect! Thank you @mjbcopland !

PR #3553, opened 2020-02-25
[typescript-resolvers] `customResolverFn` option

eturino:
Allows the ability to fully customise the `ResolverFn` type.

You can provide your custom ResolverFn instead of the default one.

Most of the time, this is not needed, and `customResolveInfo` is enough to modify the `info`.

One reason for this can be to make it compatible with other libraries in which the resolver functions are different than the rest, and for example the `info` type depends on the `TParent`.

One example of this kind of libraries is Postgraphile, in which `info`'s type is:

```typescript
  info: GraphQLResolveInfo & {
    graphile: GraphileHelpers<TParent>;
  }
```

### DIFF
--- a/dev-test/test-schema/resolvers-federation.ts
+++ b/dev-test/test-schema/resolvers-federation.ts
@@ -31,8 +31,6 @@ export type User = {
 
 export type ResolverTypeWrapper<T> = Promise<T> | T;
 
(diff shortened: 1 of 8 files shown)


eturino:
Related #2746

eturino:
@dotansimha is this acceptable?

dotansimha:
Looks good. Thank you @eturino !

PR #3565, opened 2020-02-28
[Plugin][typescript-apollo-angular]: Add configurable suffixes for operations

spaghettiismymiddlename:
This PR allows users to specify a naming scheme for operations other than the currently hard-coded suffix of `GQL` that is applied to everything and may not be suitable for every team/application.

We are currently using the changes in this PR* as a custom version of this plugin but we would love to get official support for this, and hope that this can offer some value to other users.

*_bar making `GQL` the default suffix, which was added to the PR so there are no breaking changes_


dotansimha:
This is perfect! Thank you so much @spaghettiismymiddlename !

PR #3569, opened 2020-02-28
[Typescript] enum as const: add support const assertions from TypeScript 3.4

DragorWW:
TypeScript 3.4 introduces a new construct for literal values called const assertions. Its syntax is a type assertion with const in place of the type name (e.g. `{} as const`)

Example:
```
// Type '{ readonly text: "hello" }'
let z = { text: "hello" } as const;
```

Read more about `const assertions`  in:
- [Typescript release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html)
- [Pull Request](https://github.com/Microsoft/TypeScript/pull/29510)

This PR add parameter `enumAsConst` for generate enum like object literals

```
enum MyEnum {
  A_B_C
  X_Y_Z
  _TEST
  My_Value
}
```

output
```
export const MyEnum = {
  ABC: 'A_B_C',
  XYZ: 'X_Y_Z',
  Test: '_TEST',
  MyValue: 'My_Value'
} as const;
export type MyEnum = typeof MyEnum[keyof typeof MyEnum];
```

enum as const pattern allow usage string literal [Playground Link](https://www.typescriptlang.org/play/?ssl=27&ssc=1&pln=27&pc=9#code/KYOwrgtgBAsgngUXNA3gKClAggfQEI4DCUAvFAOS4GHkA0GUAGjgJo4BapFzb7dDOACoIAyoK7khowf0zwcANQCGAGzDAJ85WuDk0AXzRpgADwAOAewBOAFygBjCyADOd+EkiEnrrukxY8QgAuCioiWSYWdhDyHg4IwWBXGKkxCPhtdRitVXU9fSglZwdvGwBuY3NrOxs4Mw13ZC8XOzJa+osAM1hEJtKAbQBrYDguqHbgMcbPUoBdCqNOsBB7GwBLJ3Gkmw8IAAolEOmIAEooFChDJZX1zZtt3ebXA6PemZazi8MjNHvXXb2xwAdGFCCcKn8dsg9pR8OFwUZIY9SoC3hAQXCwRCHn0WqjkS0QYEEUjcc9YdRyOCgA);

PS:
I think `constEnums`, `enumsAsTypes` and `enumAsConst` need replace to `enum: 'enum' | 'const enum', 'object const'`

### DIFF
--- a/docs/generated-config/typescript.md
+++ b/docs/generated-config/typescript.md
@@ -60,6 +60,21 @@ path/to/file.ts:
    enumsAsTypes: true
 ```
 
+### enumsAsConst (`boolean`, default value: `false`)
+
+Generates enum as TypeScript `const assertions` instead of `enum`. This can even be used to enable enum-like patterns in plain JavaScript code if you choose not to use TypeScript’s enum construct.
+
+#### Usage Example
+
+```yml
+generates:
+path/to/file.ts:
+ plugins:
+   - typescript
+ config:
+   enumsAsConst: true
+```
+
 ### immutableTypes (`boolean`, default value: `false`)
 
 Generates immutable types by adding `readonly` to properties and uses `ReadonlyArray`.
--- a/packages/plugins/typescript/typescript/src/config.ts
+++ b/packages/plugins/typescript/typescript/src/config.ts
@@ -69,6 +69,23 @@ export interface TypeScriptPluginConfig extends RawTypesConfig {
    */
   enumsAsTypes?: boolean;
   /**
+   * @name enumsAsConst
+   * @type boolean
+   * @description Generates enum as TypeScript `const assertions` instead of `enum`. This can even be used to enable enum-like patterns in plain JavaScript code if you choose not to use TypeScript’s enum construct.
+   * @default false
+   *
+   * @example
+   * ```yml
+   * generates:
+   * path/to/file.ts:
+   *  plugins:
+   *    - typescript
+   *  config:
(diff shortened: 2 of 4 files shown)


dotansimha:
Hi @DragorWW ! 
Thank you for this PR. We currently have `constEnums`, how is that different? it seems like the output will be the same?

DragorWW:
@dotansimha thanks for the answer

`const enums` works in the same way as regular `enum` in typescript, except that the compiled code does not have it, and the values `inline`, [more details here](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums)

I apologize for not describing why a new enum generation mode is needed.

In the modern frontend, typescript is used exclusively as a type checking system.
That's why all the things that are not in pure JS: enum, private class method, namespace try not to use.


In TypeScript 3.4 const assertions were added - and this allowed to abandon `enum` in favor of plain js object, without loss of typing


The main difference between (`enum` / `const enum`) and typed object (`as const`) is that in the code we can pass enum values of type as a string, here is an example

```js
enum MyEnum {
  A_B_C = 'A_B_C',
}

export const MyEnumConst = {
  ABC: 'A_B_C',
} as const;
export type MyEnumConst = typeof MyEnumConst[keyof typeof MyEnumConst];


const a: MyEnum = 'A_B_C'; // ERROR
const b: MyEnumConst = 'A_B_C'; 
```

dotansimha:
Got it! 
Thank you @DragorWW :)

Issue #3584, opened 2020-03-05
Enum doesn't work in input type of java code


ducthienbui97:
**Describe the bug**
 
For the schema below
```graphql
# Put your schema here
input UserData {
    userName: String!
    givenName: String
    familyName: String
    password: String
    role: UserRole
}

enum UserRole {
    ADMIN,
    USER
}
```
The java generated code is:
```java
  public enum UserRole {
    Admin("ADMIN"),
    User("USER");
    
    public final String label;
     
    UserRole(String label) {
      this.label = label;
    }
    
    private static final Map<String, UserRole> BY_LABEL = new HashMap<>();
      
    static {
        for (UserRole e : values()) {
            BY_LABEL.put(e.label, e);
        }
    }
    
    public static UserRole valueOfLabel(String label) {
      return BY_LABEL.get(label);
    }
  }

```
``` java
public UserDataInput(Map<String, Object> args) {
      if (args != null) {
        this._userName = (String) args.get("userName");
        this._givenName = (String) args.get("givenName");
        this._familyName = (String) args.get("familyName");
        this._password = (String) args.get("password");
        this._role = (UserRole) args.get("role");
      }
    }
```
If Graphql-Java parse schema file, `args.get("role")` will be a string either [`"USER"` or `"ADMIN"`](https://github.com/graphql-java/graphql-java/blob/8a5fe9101504d8c22bbc426bd2a5f22279816510/src/test/groovy/graphql/schema/GraphQLEnumTypeTest.groovy#L161) therefore can't be casted to UserRole type.
Solution is to check if `args.get("role") instanceof UserRole` then use the current function
else use `UserRole.valueOfLabel(args.get("role"))`

I can make a MR you think this change is OK.

**To Reproduce**

1. My GraphQL schema:

```graphql
# Put your schema here
input UserData {
    userName: String!
    givenName: String
    familyName: String
    password: String
    role: UserRole
}

enum UserRole {
    ADMIN,
    USER
}
```

2. My GraphQL operations:

```graphql
# Put your operations here
```

3. My `codegen.yml` config file:

```yml
# Put your YML here
generates:
  ../app/graphql/generated/Types.java:
    plugins:
      - java
    config:
      package:
        graphql.generated
      listType:
        java.util.List
      namingConvetion:
        typeNames: keep
```

**Expected behavior**

**Environment:**

- OS: Ubuntu
- `@graphql-codegen/java`: 1.12.2
- NodeJS: 12.16.1

[...]

PR #3585, opened 2020-03-05
fix(java): handle enum input when args is not enum

ducthienbui97:
Related #3584 
if a argument in input is enum, and if the argument is not of that  enum type. Use enum.valueOfLabel to try to convert argument to that enum.


dotansimha:
Great! Thank you @ducthienbui97 !

PR #3586, opened 2020-03-05
add support for offix useOfflineMustation

auryn31:
Add Support to create offix offline mutations


dotansimha:
Thank you @auryn31 ! 
Do you think it must be as a new plugin? maybe it could be integrated into the existing `react-apollo` plugin ?

auryn31:
Hy @dotansimha,
I thought about extending the current `react-apollo`, but I didn't want to bring the offix dependency into this plugin.
But if you prefere the other way, we could bring it in there.

dotansimha:
I see. Can you please just verify that this plugin can work together with the existing `react-apollo` plugin in the same output? Because I think we can keep it separated, but this could be an extension to the other types generated by the existing plugin.

auryn31:
Thanks for the response. It works wit the Apollo plugin. I use it together with this one :)

Issue #3604, opened 2020-03-09
Generated java code throws null pointer exception if nullable array of input argument is null 


ducthienbui97:
**Describe the bug**
 
For schema below
```graphql
type Mutation {
  update(users: [UserData]): [User]
}

type User {
  id: ID!
}

input UserData {
  id: String!
}
```
Java output will be
```java
public class Types {
  public static class MutationUpdateArgs {
    private Iterable<UserDataInput> _users;
  
    public MutationUpdateArgs(Map<String, Object> args) {
      if (args != null) {
        this._users = ((List<Map<String, Object>>) args.get("users")).stream().map(UserDataInput::new).collect(Collectors.toList());
      }
    }
  
    public Iterable<UserDataInput> getUsers() { return this._users; }
  }
```
if ``args.get("users")`` is null then``((List<Map<String, Object>>) args.get("users")).stream().map(UserDataInput::new).collect(Collectors.toList())`` will throw null pointer exception even if the array is nullable.
It should check null before use ``args.get("users")``. Desired output should be something like this
```java
if (args != null) {
  if (args.get("users") != null) {
    this._users = ((List<Map<String, Object>>) args.get("users")).stream().map(UserDataInput::new).collect(Collectors.toList());
  }
}
```
**To Reproduce**

1. My GraphQL schema:

```graphql
type Mutation {
  update(users: [UserData]): [User]
}

type User {
  id: ID!
}

input UserData {
  id: String!
}
```

2. My GraphQL operations:

```graphql
# Put your operations here
```

3. My `codegen.yml` config file:

```yml
# Put your YML here
overwrite: true
schema: "schema.graphqls"
  /graphql/generated/Types.java:
    plugins:
      - java
    config:
      package:
        graphql.generated
      listType:
        java.util.List
      namingConvetion:
        typeNames: keep
```

**Expected behavior**

Generated code should check null of the nullable array input arguments before call stream of it.

**Environment:**

- OS: Ubuntu
- `@graphql-codegen/java`:  1.12.2 
- NodeJS: 12.16.1

**Additional context**

PR #3605, opened 2020-03-09
Check null args for nullabe input type array

ducthienbui97:
If `args.get("input") == null` then `(List<Map<String, Object>>) args.get("input).stream()` will throws NullPointerException since it's trying to access null.

Checking null before accessing to make sure there's no NullPointerException thrown
Related #3604


dotansimha:
Perfect! Thank you @ducthienbui97 :)

PR #3608, opened 2020-03-09
Generate TS deprecations from GQL deprecated directive

### LINKED ISSUE #3141: Generate @deprecated annotations in typescript doc comments
I've got a GraphQL schema that has some `@deprecated` types and/or fields. While the descriptions already translate into tsdoc comments with the `typescript` plugin, the deprecations are not mentioned anywhere in the generated type declarations and the typescript compiler won't warn me when using those properties. I would expect a [`@deprecated` tag](https://api-extractor.com/pages/tsdoc/tag_deprecated/) to get created in the doc comment of the respective typescript declaration.

adelsz:
Adds TS deprecation annotations based on GQL deprecation directives.
Implements #3141


packages/plugins/typescript/typescript/src/visitor.ts
```diff
@@ -88,8 +88,14 @@ export class TsVisitor<TRawConfig extends TypeScriptPluginConfig = TypeScriptPlu
     const typeString = this.config.wrapFieldDefinitions ? `FieldWrapper<${node.type}>` : ((node.type as any) as string);
     const originalFieldNode = parent[key] as FieldDefinitionNode;
     const addOptionalSign = !this.config.avoidOptionals.object && originalFieldNode.type.kind !== Kind.NON_NULL_TYPE;
-    const comment = transformComment((node.description as any) as string, 1);
     const { type } = this.config.declarationKind;
+    let commentText: string = node.description as any;
```
dotansimha: @adelsz I think this could be moved to the Base visitor? this way it will be applied to `flow` plugins as well, and not only `typescript`.

dotansimha:
This looks great, thanks @adelsz! 
I think the implementation could be moved as-is to the base types visitor, this way it will work for flow plugins as well. (tests can remain in the same place, under `typescript` plugin)

adelsz:
@dotansimha Moved `getFieldComment` to base visitor.

Issue #3574, opened 2020-03-01
extend typescript-graphql-request to allow retries of requests


andycmaj:
i use the typescript-graphql-request template extensively, and one thing i think it's missing is the ability to control error handling/retries for the network calls made by the SDK.
I'd like to be able to retry transient network errors with retries.

There are a few ways i can think of to implement this.

## build configuraable retries into template

add retry policies into the template and allow configuration/overriding

* `withRetryPolicy` helper
```
import polly, { AsyncRetryable } from 'polly-js';

const defaultPolicy = polly()
  .handle(err => {
    return true;
  })
  .waitAndRetry(2);

const withRetryPolicy = <T>(
  action: () => Promise<T>,
// Potentially pass more config into here or pass an override policy
) => {
  return defaultPolicy.executeForPromise(async info => {
    return await action();
  });
};

export default withRetryPolicy;
```

* codegen `visitor`
```
...
    if (this.config.retryPolicy) {
      // TODO: default config now, configurable policy later
      this._additionalImports.push(
        `import { withRetryPolicy } from 'typescript-graphql-request-with-retries';`
      );
    }
...
return `
          ${o.node.name.value}(variables${optionalVariables ? '?' : ''}: ${
            o.operationVariablesTypes
          }): Promise<${o.operationResultType}> {
            return withRetryPolicy(() => client.request<${
              o.operationResultType
            }>(${doc}, variables));
          }
`;
```

## delegate all retry logic to a generic wrapper the consumer can specify

```
export type RetryWrapper = <T>(action: () => Promise<T>) => Promise<T>;
...

          return `
          ${o.node.name.value}(variables${optionalVariables ? '?' : ''}: ${
            o.operationVariablesTypes
          }): Promise<${o.operationResultType}> {
            return withRetries(() => client.request<${
              o.operationResultType
            }>(${doc}, variables));
          }
`;
        }
      })
      .map(s => indentMultiline(s, 2));

    return `export function getSdk(client: GraphQLClient, withRetries: RetryWrapper) {
  return {
${allPossibleActions.join(',\n')}
  };
}`;
```

I think this could easily be layered onto the relatively simple existing typescript-graphql-request template and i'd be happy to submit and test.

dotansimha:
@andycmaj This sounds like a nice idea! Do you want to create a PR for this? it could be turned on with a configuration (something like `pollyRetryPolicy: true`)

andycmaj:
[...]

PR #3613, opened 2020-03-10
[typescript-graphql-request] add ability to pass custom wrapper to generated sdk functions, eg. retryWrapper (#3574)

andycmaj:
# Summary

this extends graphql-typescript-request template to allow users to pass a custom wrapper for generated sdk actions. 

the intention is to allow things like adding retry policies or request logging/profiling to the generated sdk calls.

the intention is to make a small, non-breaking change that can implement something like middleware.

related #3574 

# Usage

## default usage

by default, no extra config is needed and no breaking changes to generated code will be made.

`getSdk(client)` can still be called as before, and the wrapper will default to the identity function. in other words, the wrapper will just be a passthrough to the original graphql client request (NO-OP).

## passing a custom wrapper

if the consumer wishes to customize request handing behavior, they can pass a `SdkFunctionWrapper`-typed function as the second parameter of `getSdk`:

```
  // example of adding a client timing wrapper
  const client = new GraphQLClient('');
  const clientTimingWrapper: SdkFunctionWrapper = async <T>(action: () => Promise<T>): Promise<T> => {
    const startTime = new Date();
    const result = await action();
    console.log('request duration (ms)', new Date() - startTime)
    return result;
  }

  const sdk = getSdk(client, clientTimingWrapper);
```

## passing a retry wrapper

This example uses [polly-js](https://github.com/mauricedb/polly-js) to define a failure retry wrapper. (this is the use case for which i originally intended to add this capability)

```
const withRetries: RetryWrapper = <T>(action: () => Promise<T>) =>
  polly()
    .handle((err: Error) => {
      warn('GraphqlClient:NetworkError', err);
      return err.message.includes('connect ETIMEDOUT');
    })
    .waitAndRetry(3)
    .executeForPromise(info => {
      if (info.count === 3) {
        error('GraphqlClient:MaxRetries', null, {
          ...info,
          action: action.toString(),
        });
      } else if (info.count > 0) {
        warn('GraphqlClient:RetryingCall', null, {
          ...info,
          action: action.toString(),
        });
      }

      return action();
    });

const sdk = getSdk(client, withRetries);
```

# TODO

- [x] Documentation updates. No no configuration has been added but would probably be a good idea to add examples to the docs. Agree?
- [x] Additional tests if possible to execute wrappers. @dotansimha is there a way I can actually EXECUTE generated code in the context of a unit test?


dotansimha:
@andycmaj this is awesome, thank you so much! 

I think we really need to add documentation for this plugin, is there a chance you can do it as part of this PR? 

Regarding execution, I think you should be able to take the output of plugins during a test, and run `eval` to run it, I think it might work :D

andycmaj:
> I think we really need to add documentation for this plugin, is there a chance you can do it as part of this PR? 

on it

> Regarding execution, I think you should be able to take the output of plugins during a test, and run eval to run it, I think it might work 🗡 

will give it a try thanks for the ptr!

andycmaj:
ok @dotansimha 

i was able to execute the generated results using `ts.transpileModule(output, { compilerOptions: { module: ts.ModuleKind.CommonJS } });`

i had to add [`nock`](https://www.npmjs.com/package/nock) as a `devDependency` in order to mock the underlying `graphql-request` calls when testing middleware.

the caveat here is that i also had to increase test timeout for the 2 added tests... the typescript transpilation was pushing it a bit above the 5 second limit.

lemme know what you think.

PS. docs are also updated with examples.

andycmaj:
hmmm not sure why the CI tests take so long, but i saw a timeout on node 13. 

```
  ● graphql-request › sdk › Should execute middleware passed to getSdk

    : Timeout - Async callback was not invoked within the 10000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 10000ms timeout specified by jest.setTimeout.Error:

      189 |     });
      190 | 
    > 191 |     it('Should execute middleware passed to getSdk', async () => {
          |     ^
      192 |       let duration = 0;
      193 |       const middleware: SdkFunctionWrapper = async (action: () => Promise<any>): Promise<any> => {
      194 |         const start = new Date();

      at new Spec (../../../../node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (tests/graphql-request.spec.ts:191:5)

Test Suites: 1 failed, 1 total
Tests:       2 failed, 3 passed, 5 total
Snapshots:   4 passed, 4 total
Time:        79.586s
Ran all test suites.
error Command failed with exit code 1.
```

this test suite only took 30s locally... 
i wonder if all tests are being run in parallel?

andycmaj:
ok should be ready to go now.

dotansimha:
This is awesome ❤️ Thank you so much @andycmaj !

andycmaj:
thanks for being responsive and awesome... this is one of the best OSS experiences i've had @dotansimha

packages/plugins/typescript/graphql-request/src/index.ts
```diff
@@ -5,6 +5,9 @@ import { GraphQLRequestVisitor } from './visitor';
 import { extname } from 'path';
 import { RawGraphQLRequestPluginConfig } from './config';
 
+export type SdkFunctionWrapper = <T>(action: () => Promise<T>) => Promise<T>;
+export const defaultWrapper: SdkFunctionWrapper = sdkFunction => sdkFunction();
```
SimenB: thoughts on injecting this identify function into the generated files instead of exporting it? It currently adds a runtime dependency on `@graphql-codegen/typescript-graphql-request`

SimenB: oh, already fixed by #3667 👍

Issue #3616, opened 2020-03-10
Java Enum valueOfLabel parameter has to explicitly casted to String


ducthienbui97:
**Describe the bug**
Related to #3584 
Introduced in #3585 
Sorry my MR introduced the bug
For
```graphql
input SearchUser {
  sort: ResultSort
}

enum ResultSort {
  ASC
  DESC
}
```
Generated java is
```java
public SearchUserInput(Map<String, Object> args) {
  if (args != null) {
    if (args.get("sort") instanceof ResultSort) {
      this._sort = (ResultSort) args.get("sort");
     } else {
       this._sort = ResultSort.valueOfLabel(args.get("sort"));
    }
 }
}
```
However, `args.get("sort")` is of type `Object` while parameter of `ResultSort.valueOfLabel` is `String`. So `args.get("sort")`  has to be explicitly casted to `String`
It has to be 
```java
this._sort = ResultSort.valueOfLabel((String) args.get("sort"));
```
instead
**To Reproduce**

1. My GraphQL schema:

```graphql
# Put your schema here
input SearchUser {
  sort: ResultSort
}

enum ResultSort {
  ASC
  DESC
}
```

2. My GraphQL operations:

```graphql
# Put your operations here
```

3. My `codegen.yml` config file:

```yml
# Put your YML here
```

**Expected behavior**

**Environment:**

- OS: Ubuntu 
- `@graphql-codegen/java`: 1.13.0 
- NodeJS: 12.16.1

**Additional context**

dotansimha:
Fixed in `v1.13.1` 🎉

PR #3617, opened 2020-03-10
fix(java): Cast args to String for valueOfLabel

ducthienbui97:
`Enum.valueOfLabel` accept a String as parameter. `args.get("input")` return type `Object` thus, can not be passed to `Enum.valueOfLabel`. It has to explicitly casted to String before passing to the function.
Related #3616


dotansimha:
Thank you @ducthienbui97 !

Issue #3538, opened 2020-02-21
Generated interface resolvers seem to be too narrowly typed


onethread:
**Describe the bug**
It’s possible I’m using the generated resolver types incorrectly, but I’m following Apollo’s javascript example for resolving graphql interfaces, and I’d expect to able to do something similar, but in typescript with graphql-code-generator; i.e., a resolver for an interface type should be able to determine what concrete type of the interface should be returned.

E.g. in js: 

```javascript
const resolvers = {
  Book: {
    __resolveType(parent, context, info){
      if(parent.classes){
        return 'TextBook';
      }

      if(parent.colors){
        return 'ColoringBook';
      }

      return null;
    },
  },
  …
};
```

Unfortunately, the generated typescript resolver doesn’t allow me to check what possible types it can be. `parent` is typed to `Book` which has no information about subtypes.  The typescript equivalent is below. 

If there’s a different pattern I should be using, please let me know! 

**To Reproduce**

1. My GraphQL schema:

```graphql
interface Book {
  title: String
}

type TextBook implements Book {
  title: String
  cost: Float
}

type ColoringBook implements Book {
  title: String
  colors: [String]
}

```

2. My `codegen.yml` config file:

```yml
schema: "src/schema.graphql"
generates:
  src/types/server-graphql-types.ts:
    plugins:
      - typescript
      - typescript-resolvers:
        useIndexSignature: true
```

3. Generated code:

```typescript
export type ResolversParentTypes = {
  Query: {};
  Book: Book;
  TextBook: TextBook;
  ColoringBook: ColoringBook;
  …
};
```

```typescript
export type BookResolvers<
  ContextType = any,
  ParentType extends ResolversParentTypes[“Book”] = ResolversParentTypes[“Book”]
> = {
  __resolveType: TypeResolveFn<
    "TextBook" | "ColoringBook",
    ParentType, // Because Resolvers uses the default BookResolver type, ParentType is `ResolversParentTypes[“Book”]`
    ContextType
  >;
  title?: Resolver<
    ResolversTypes[“String”],
    ParentType,
    ContextType
  >;
};

export type Resolvers<ContextType = any> = {
  Book?: BookResolvers;
}

```

**Expected behavior**

I would expect to be able to do this: 

```typescript
const bookResolver: BookResolvers = {
  __resolveType(parent, _context, _info) {
      if (parent.__typename === "TextBook")
        return "TextBook";
      else if (parent.__typename === "ColoringBook") {
[...]

PR #3618, opened 2020-03-10
fix(resolvers): use types that implement an interface in __resolveType as first argument

kamilkisiela:
not interface itself

```diff
  /** Mapping between all available schema types and the resolvers types */
  export type ResolversTypes = {
    String: ResolverTypeWrapper<Scalars['String']>,
    Boolean: ResolverTypeWrapper<Scalars['Boolean']>,
-   Book: ResolverTypeWrapper<Book>,
+   Book: ResolversTypes['ColoringBook'] | ResolversTypes['TextBook'],
    TextBook: ResolverTypeWrapper<TextBook>,
    Float: ResolverTypeWrapper<Scalars['Float']>,
    ColoringBook: ResolverTypeWrapper<ColoringBook>,
  };

  /** Mapping between all available schema types and the resolvers parents */
  export type ResolversParentTypes = {
    String: Scalars['String'],
    Boolean: Scalars['Boolean'],
-   Book: Book,
+   Book: ResolversParentTypes['ColoringBook'] | ResolversParentTypes['TextBook'],
    TextBook: TextBook,
    Float: Scalars['Float'],
    ColoringBook: ColoringBook,
  };
```

```diff
  /** Mapping between all available schema types and the resolvers types */
  export type ResolversTypes = {
    String: ResolverTypeWrapper<Scalars['String']>,
    Boolean: ResolverTypeWrapper<Scalars['Boolean']>,
-   Book: ResolverTypeWrapper<Book>,
+   Book: ResolversTypes['ColoringBook'] | ResolversTypes['TextBook'],
    TextBook: ResolverTypeWrapper<TextBook>,
    Float: ResolverTypeWrapper<Scalars['Float']>,
    ColoringBook: ResolverTypeWrapper<ColoringBook>,
  };

  /** Mapping between all available schema types and the resolvers parents */
  export type ResolversParentTypes = {
    String: Scalars['String'],
    Boolean: Scalars['Boolean'],
-   Book: Book,
+   Book: ResolversParentTypes['ColoringBook'] | ResolversParentTypes['TextBook'],
    TextBook: TextBook,
    Float: Scalars['Float'],
    ColoringBook: ColoringBook,
  };
```

```diff
const resolvers: Resolvers = {
  Book: {
    __resolveType(parent, context, info) {
-    // parent is ResolversParentTypes['Book']
+    // parent is ResolversParentTypes['TextBook'] | ResolversParentTypes['ColoringBook']
+
+    // previously `parent.__typename` didn't exist
+    if (parent.__typename === 'TextBook') {
+      return 'TextBook';
+    }
+
+    if (parent.__typename === 'ColoringBook') {
+      return 'ColoringBook';
+    }
+
+    return null;
    },
  },
};
```

Related #3538


zomble:
Great change.

This appears to have a side effect that when an interface doesn't have an implementing type it no longer is omitted in the generated types.

This presents an issue when using https://graphql-modules.com/ and a module defines an interface for others to use, but inside the interface defining module the generated types no longer have that interface.

Module A:
  interface Example

Module B:
  type Fruit implements Example

Before this version generating types for Module A (in isolation) would include definitions for `Example`.

After this change, only Module B does as it has an implementing type of `Example` (which now resolves to Fruit).

@kamilkisiela @dotansimha thoughts? I am aware that from a graph client perspective of a schema, one would not ever care about an interface with out implementing types.

PR #3628, opened 2020-03-12
Fixed issue with fragment names when using near-operation-file + dedupeOperationSuffix

wassim-k:
In v1.13.0, there's an issue where the fragment imports generated by `near-file-operation` preset do not match the fragment spreads generated in `client-side-base-visitor`.
This is due to the fact that the logic used in the preset and the plugin does not match.
This PR is an attempt to unify the logic for generating fragment names, fragment variables names (fragment doc) and general operation naming.


dotansimha:
This is great! I really wanted to refactor and change the behaviour to make it more aligned. Thank you so much for doing that @wassim-k !

wassim-k:
You're very welcome @dotansimha

3nvi:
It would be awesome if we could release a bugfix version to fix that

nodkz:
@wassim-k amazing PR now with a new 1.13.1 version all works as I want!
Thanks a lot!!!

PR #3629, opened 2020-03-12
[typescript-graphql-request] Enable to pass HTTP headers to GraphQL client

morishin:
This patch makes it possible to pass HTTP request headers to GraphQL client via [`GraphQLClient#setHeaders`](https://github.com/prisma-labs/graphql-request/blob/0b467b8410a97526559a419b6cbaf4f5639a4f30/src/index.ts#L79).

### Motivation

I would like to pass some headers such as `Authorization` and `Cookie` from Client to an API Server through GraphQL Server.

Client (Web Browser) -----> GraphQL Server -----> Another API Server

### DIFF
--- a/packages/plugins/typescript/graphql-request/src/visitor.ts
+++ b/packages/plugins/typescript/graphql-request/src/visitor.ts
@@ -29,6 +29,7 @@ export class GraphQLRequestVisitor extends ClientSideBaseVisitor<RawGraphQLReque
     }
 
     this._additionalImports.push(`import { SdkFunctionWrapper, defaultWrapper } from '@graphql-codegen/typescript-graphql-request';`);
+    this._additionalImports.push(`import { Headers } from 'graphql-request/dist/src/types';`);
   }
 
   protected buildOperation(node: OperationDefinitionNode, documentVariableName: string, operationType: string, operationResultType: string, operationVariablesTypes: string): string {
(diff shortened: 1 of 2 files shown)


andycmaj:
@morishin did you already see that you can configure the `GraphqlClient` before passing it to `getSdk`?

```
const client = new GraphQLClient(process.env.GRAPHQL_URL, {
  headers: {
    'x-hasura-admin-secret': process.env.HASURA_GRAPHQL_ADMIN_SECRET,
  },
});

const sdk = getSdk(client);
```

this will apply the passed headers to every request the sdk makes via the client.

do you have a reason for wanting to pass in different headers into the actual query calls?

morishin:
Thank you for your comment!

>did you already see that you can configure the GraphqlClient before passing it to getSdk?

>do you have a reason for wanting to pass in different headers into the actual query calls?

Yes.
it is OK if a header value is constant throughout all requests. But I think it is not enough for the case when the header value is different per client's requests.

For example, an API server wants to validate `Cookie` value to determine whether it returns 200 or 403 in the situation below.
>Client (Web Browser) -----> GraphQL Server -----> Another API Server

----

>also a warning that setting headers using setHeader in one query call will (maybe unexpectedly to other callers) set the header for all requests going forward with that client instance.

You're right 😓 
Maybe it is necessary to reset headers after every request or to recreate GraphQLClient on every request (sounds bad for performance).

andycmaj:
> Maybe it is necessary to reset headers after every request or to recreate GraphQLClient on every request (sounds bad for performance).

yeah i'd agree with that, @morishin. I'd likely choose recreating the client for each request.
it doesn't seem to be expensive. pretty trivial stuff going on in the constructor:

https://github.com/prisma-labs/graphql-request/blob/0b467b8410a97526559a419b6cbaf4f5639a4f30/src/index.ts#L9

so yeah seems like keeping it simple here would be the safest bet. 🤷‍♂

dotansimha:
Exactly as you said :) 
You can set the headers when you create your client, and if you need to change it you don't have to create a new client, you can use `client.setHeaders` or `client.setHeader` to modify that :)

Issue #3638, opened 2020-03-14
[typescript-operations] preResolveTypes generate incorrect optional types


standy:
**Describe the bug**

Optional types should be marked with `?`

Current behavior:
```ts
export type CommentQuery = {
  __typename?: 'Query';
  currentUser: Maybe<{ __typename?: 'User'; login: string; html_url: string }>;
  entry: Maybe<{
    __typename?: 'Entry';
    id: number;
    createdAt: number;
    commentCount: number;
    postedBy: { __typename?: 'User'; login: string; html_url: string };
    comments: Array<Maybe<{ __typename?: 'Comment' } & CommentsPageCommentFragment>>;
    repository: { __typename?: 'Repository'; description: Maybe<string>; open_issues_count: Maybe<number>; stargazers_count: number; full_name: string; html_url: string };
  }>;
};
```

Expected behavior:
```ts
export type CommentQuery = {
  __typename?: 'Query';
  currentUser?: Maybe<{ __typename?: 'User'; login: string; html_url: string }>;
  entry?: Maybe<{
    __typename?: 'Entry';
    id: number;
    createdAt: number;
    commentCount: number;
    postedBy: { __typename?: 'User'; login: string; html_url: string };
    comments: Array<Maybe<{ __typename?: 'Comment' } & CommentsPageCommentFragment>>;
    repository: { __typename?: 'Repository'; description?: Maybe<string>; open_issues_count?: Maybe<number>; stargazers_count: number; full_name: string; html_url: string };
  }>;
};
```

**To Reproduce**
There is an example in `dev-test` folder:
https://github.com/dotansimha/graphql-code-generator/blob/master/dev-test/githunt/types.preResolveTypes.ts#L177

dotansimha:
Fixed in `v1.13.1` 🎉

ekfn:
Hi there. Are this optional? Or why this expected behavior? @dotansimha

ekfn:
i can opt out this behavior via `avoidOptionals: true`, but i'm using this syntax:
```
avoidOptionals:
  inputValue: false
  object: true
```
and it seems like i can't opt out this behavior via this syntax

ekfn:
answer is here https://github.com/dotansimha/graphql-code-generator/pull/3657
so this is works for me:
```
avoidOptionals:
  object: true
  field: true
```

Issue #3463, opened 2020-02-07
TypeScript operations queries doesn't mark fields optional


stephenh:
Given a schema like:

```graphql
type Child {
  parent: Parent
}
```

The default `typescript-operations` output for `Child` is:

```typescript
export type Child = {
  parent?: Maybe<Parent>;
}
```

Which is fine, however, if I write a query:

```graphql
 query Foo {
  child(...) { parent { name } }
}
```

The generated type is:

```typescript
export type FooQuery = {
  { __typename?: 'Query' } &
  { child: (
    { __typename?: "Child" } &
    & parent: (
      { __typename? "Parent" } &
      Pick<Parent, "name">
    )
  )}
}
```

(I'm butchering the nested types a bit.)

Note that `parent` here in the query is `parent` and not `parent?`.

This inconsistency between the top-level type and the query type means I can't do:

```typescript
const child: Child = newFactoryMethodForChild();
assertTestWithFooQueryResult({ query: { child }));
```

(Again being somewhat loose with syntax.)

Because there is a type error that `Child`'s `parent` is optional (`Parent | null | undefined`) but the query's munged-together anonymous type wants `parent` to be `Parent | null`.

This is internally inconsistent.

Note that if I use `avoidOptionals: true`, it turns the top-level `Child` into `parent: Maybe<Parent>` and so now my "factory method that makes `Child`s" and "unit test that handles foo query results" structurally matches and compiles fine.

However there are other downsides to `avoidOptionals` (specifically that it should never apply to input types), so it'd be nice to keep `avoidOptionals` off (on the client-side types; on the server-side types I think avoidOptionals is generally best-practice fwiw).

dotansimha:
@stephenh While dealing with the schema types - we can allow optionals and allow `undefined` because the GraphQL engine knows how to handle that.
But, if you are writing a query and ask for specific fields, the GraphQL engine will always return the field for you, and if an nullable field, it means that it could be either a value, or `null` - but the field will always be there. 
You should not use `Child` (or any base type generated by `typescript` plugin) directly as a consumer of queries - because it represents the complete type and not the selection set your query is asking for.

You can customize the output with `avoidOptionals` or `maybeValue` and change the output according to your needs.

Issue #3060, opened 2019-12-02
Nested object in fragment unexpectedly not optional


KieranHarper:
**Describe the bug**
I have an entity that has optional entities as fields, and this is reflected in the type generated for that entity. However if I use a fragment to query, then the type generated for the fragment loses the ? and no longer matches.

Unless I'm misunderstanding something, I would expect a fragment that fully specifies all the fields of an entity to exactly match the type generated for the field itself.

**To Reproduce**

1. My GraphQL schema:

```graphql
type Address {
  street: String!
  suburb: String!
}

type User {
  id: String!
  address: Address # <-- optional
}

type Query {
  user(id: String!): User
}

schema {
  query: Query
}
```

2. My GraphQL operations:

```graphql
fragment AddressFields on Address {
  street
  suburb
}

fragment UserFields on User {
  id
  address {
    ...AddressFields
  }
}

query getUser($id: String!) {
  user(id: $id) {
    ...UserFields
  }
}
```

3. My `codegen.yml` config file:

```yml
schema: "schema.graphql"
documents: "operations.graphql"
generates:
  types.ts:
    - typescript
    - typescript-operations:
        preResolveTypes: true
        dedupeOperationSuffix: true
    - typescript-react-apollo:
        reactApolloVersion: 2
        withHooks: true
        withHOC: false
        withComponent: false
```

4. Minimal example that will complain because of type mismatching:
```typescript
import { User, UserFieldsFragment } from "./types";

const userMapper = (user: User): UserFieldsFragment => {
  return user;
};
```

tsc says:
```
error TS2322: Type 'User' is not assignable to type 'UserFieldsFragment'.
Property 'address' is optional in type 'User' but required in type 'UserFieldsFragment'.
```

**Expected behavior**
For `UserFieldsFragment` to declare `address` as ?

**Environment:**

- OS: MacOS Mojave
- `@graphql-codegen/...`: 1.9.1
- NodeJS: 11.13.0

ebenoist:
Running into this one as well. Any clue where the regression was? Happy to help fix this up.

dotansimha:
I think it's an issue with `preResolveTypes`, can you please try to set it to `false` and let us know if it works?

KieranHarper:
`preResolveTypes: false` has the same issue, just with a little more output
```
[...]

Issue #2590, opened 2019-09-19
Why does typescript-operations generate every filed as mandatory with <Maybe>?


TheWWWorm:
I have a type "personAlias", and query for it 
```
personAlias {
        firstName
        lastName
        gender {
          code
          description
        }
      }
```

Type that gets generated for it looks like

```
export type PersonAliasType = {
   __typename?: 'PersonAliasType',
  firstName?: Maybe<Scalars['String']>,
  gender?: Maybe<RefDataType>,
  lastName?: Maybe<Scalars['String']>,
};
```

Notice that  gender is optional.
But in the end query it looks like

```
personAlias: (
        { __typename?: 'PersonAliasType' }
        & Pick<Types.PersonAliasType, 'firstName' | 'lastName'>
        & { gender: Types.Maybe<(
          { __typename?: 'RefDataType' }
          & Pick<Types.RefDataType, 'code' | 'description'>
        )> }
      )
```

Which makes it mandatory, which causes conflicts. Why? I have set as avoidOptionals: false

n1ru4l:
Can you share your schema please?

TheWWWorm:
@n1ru4l 
```
type PersonAliasType {
  dateOfBirth: DateTime
  firstName: String
  gender: RefDataType
  id: String!
  lastName: String
}
```

n1ru4l:
The types are Maybe types cause they could be `null` according to your schema

TheWWWorm:
But null doesn't mean that they they are optional, it just means that they can exist with value of null. You can't have no key at all

n1ru4l:
Could you please provide a minimal reproduction using the code sandbox template (https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template)?

ardatan:
@TheWWWorm In GraphQL, there are `nullable`s and `nonnullable`s. If you return `undefined` or `null`, it is considered as `null`. AFAIK, GraphQL Client returns `null` for undefined fields. But it would be great if you show your problem using the sandbox template as @n1ru4l asked.

dotansimha:
Also, note that you can try to add `avoidOptionals: true` and it will avoid using `?` while generating.

dotansimha:
Closing. @TheWWWorm feel free to re-open/comment if this is still relevant.

PR #3639, opened 2020-03-14
Fix optional types in queries

standy:
Fix for #3638
Related #3463 #3060 #2590

This MR makes all nullable fields be optional (for typescript and flow), if `avoidOptionals` option is not true. 

There is an inconsistency between the top-level types and the query types. Example from `dev-test`: 
https://github.com/dotansimha/graphql-code-generator/blob/master/dev-test/githunt/types.preResolveTypes.ts
```ts
// This is the top-level type and description is optional
export type Repository = {
  __typename?: 'Repository';
  /** Just the name of the repository, e.g. GitHunt-API */
  name: Scalars['String'];
  /** The full name of the repository with the username, e.g. apollostack/GitHunt-API */
  full_name: Scalars['String'];
  /** The description of the repository */
  description?: Maybe<Scalars['String']>;
  /** The link to the repository on GitHub */
  html_url: Scalars['String'];
  /** The number of people who have starred this repository on GitHub */
  stargazers_count: Scalars['Int'];
  /** The number of open issues on this repository on GitHub */
  open_issues_count?: Maybe<Scalars['Int']>;
  /** The owner of this repository on GitHub, e.g. apollostack */
  owner?: Maybe<User>;
};

// This is query type, and description is mandatory now
export type CommentQuery = {
    repository: {
      __typename?: 'Repository';
      description: Maybe<string>;
      open_issues_count: Maybe<number>;
      stargazers_count: number;
      full_name: string;
      html_url: string;
    };
  }>;
};
```

Same bug for `preresolveTypes: false`
Example from https://github.com/dotansimha/graphql-code-generator/blob/master/dev-test/githunt/types.ts
```ts
// This is top-level type
export type Query = {
  /** Return the currently logged in user, or null if nobody is logged in */
  currentUser?: Maybe<User>;
};

// This is query type, notice currentUser become mandatory
export type CommentQuery = { __typename?: 'Query' } & {
  currentUser: Maybe<{ __typename?: 'User' } & Pick<User, 'login' | 'html_url'>>;
}
```

My fix is actually easy. There are two methods: `formatNamedField` and `wrapTypeWithModifiers`, they should be applied always together, and `formatNamedField` should add optional mark when it needed (line [53](https://github.com/dotansimha/graphql-code-generator/pull/3639/commits/ea69ed97042a34ac9c34bb65ad0baf2e63f99258#diff-b05502b152b19c27071a4c3bffb8d02fR54))

### DIFF
--- a/dev-test/githunt/__generated__/comment-added.subscription.stencil-component.tsx
+++ b/dev-test/githunt/__generated__/comment-added.subscription.stencil-component.tsx
@@ -8,7 +8,7 @@ declare global {
   };
 
   export type OnCommentAddedSubscription = { __typename?: 'Subscription' } & {
-    commentAdded: Types.Maybe<
+    commentAdded?: Types.Maybe<
       { __typename?: 'Comment' } & Pick<Types.Comment, 'id' | 'createdAt' | 'content'> & {
           postedBy: { __typename?: 'User' } & Pick<Types.User, 'login' | 'html_url'>;
         }
--- a/dev-test/githunt/__generated__/comment.query.stencil-component.tsx
+++ b/dev-test/githunt/__generated__/comment.query.stencil-component.tsx
@@ -10,8 +10,8 @@ declare global {
   };
 
   export type CommentQuery = { __typename?: 'Query' } & {
-    currentUser: Types.Maybe<{ __typename?: 'User' } & Pick<Types.User, 'login' | 'html_url'>>;
-    entry: Types.Maybe<
+    currentUser?: Types.Maybe<{ __typename?: 'User' } & Pick<Types.User, 'login' | 'html_url'>>;
+    entry?: Types.Maybe<
       { __typename?: 'Entry' } & Pick<Types.Entry, 'id' | 'createdAt' | 'commentCount'> & {
           postedBy: { __typename?: 'User' } & Pick<Types.User, 'login' | 'html_url'>;
           comments: Array<Types.Maybe<{ __typename?: 'Comment' } & CommentsPageCommentFragment>>;
--- a/dev-test/githunt/__generated__/current-user.query.stencil-component.tsx
+++ b/dev-test/githunt/__generated__/current-user.query.stencil-component.tsx
@@ -6,7 +6,7 @@ declare global {
   export type CurrentUserForProfileQueryVariables = {};
 
   export type CurrentUserForProfileQuery = { __typename?: 'Query' } & {
-    currentUser: Types.Maybe<{ __typename?: 'User' } & Pick<Types.User, 'login' | 'avatar_url'>>;
+    currentUser?: Types.Maybe<{ __typename?: 'User' } & Pick<Types.User, 'login' | 'avatar_url'>>;
   };
 }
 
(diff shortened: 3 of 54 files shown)


dotansimha:
Hi @standy ! 

This seems like a big PR with more changes, can you please explain your changes there? 
because the bug effected `preResolveTypes` types only as your said, but it seems like all examples under `dev-test` are updated with changes, and you changed some more files? 

Thanks!

standy:
Hello, @dotansimha 

I did update my PR. And I found a few similar issues, which is fixed here.

dotansimha:
Just went through the changes in this PR. This is great!
Thank you so much @standy  :)

david-golightly-leapyear:
🚨 I think this "fix" was incorrect. It seems to break the behavior described here: https://github.com/dotansimha/graphql-code-generator/issues/3060#issuecomment-570174939

We tried upgrading to 0.13.1 and noticed a lot of client-side generated query types were now suddenly "optional". A field that is present in the client-side query should never be "optional", although its value may be `null` if it is not marked non-nullable.

So this PR introduced incorrect behavior and IMO should be reverted. Happy to discuss further - I know a lot of work went into it! It's a subtle distinction, but it's important not to confuse "client" and "server" types in implementations - they may be different, but the code generation should account for this, and marking every "nullable" field as "optional" confuses the two.

standy:
@david-golightly-leapyear 
Take a look into `avoidOptionals` option, I think this is what you  need
https://graphql-code-generator.com/docs/plugins/typescript#avoidoptionals-boolean-default-value-false

david-golightly-leapyear:
@standy I see - `avoidOptionals` looks to implement the correct behavior. I'm confused why it's not `default: true`, and indeed why the behavior in this PR would ever be desirable? Unless you're mixing your server & client types inappropriately... but IMO that should be discouraged, it goes against the grain of GraphQL.

standy:
@david-golightly-leapyear
That PR fixes inconsistent behavior (described above).

david-golightly-leapyear:
@standy I don't understand how it's inconsistent. The server ("top-level") types and the client ("query") types are often going to be inconsistent in that manner, and that seems to be _correct_ behavior.

The server has to support any possible query, so it can't ever specify the correct type for all queries.
The client _does_ know what fields it's requesting, and is in a position to say "field X is non-optional, because I'm requesting it and it should appear in the GraphQL response".

Indeed, when you look at the behavior of a GraphQL response, any field requested by the client is returned either with a value or `null`, but cruciall _never_ with `undefined` or with the field missing.

That's because `undefined` (aka the ES2019 "optional type") is not a GraphQL type, and the ES2019 "optional type" is also _not the same thing_ as what is called "optional type" in the GraphQL spec.

david-golightly-leapyear:
tl;dr at a minimum, if this PR is going to be preserved, it should also make `avoidOptionals` default: true, since this PR breaks existing correct usage (avoiding use of `undefined` in generated client types).

rizkyario:
Screenshot 2020-04-24 at 17 43 26

Is this fix available in flow plugins?
seems doesn't work here, or am I missing some config?

Issue #3525, opened 2020-02-19
FragmentDoc imports are still broken in 1.12.2


mjbcopland:
Related to #3259. 

[`config.documentMode !== 'documentNode' || config.documentMode !== 'external'`](https://github.com/dotansimha/graphql-code-generator/blob/v1.12.2/packages/presets/near-operation-file/src/fragment-resolver.ts#L21) introduced in #3436 always evaluates to `true`. This causes codegen to try to import fragment documents which don't exist. What was the intended logic of this line? Looks like maybe the `||` should be `&&`?

Also, this import does not appear to respect [custom prefix/suffix values from ClientSideBaseVisitor](https://github.com/dotansimha/graphql-code-generator/blob/v1.12.2/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts#L129-L130).

dotansimha:
@mjbcopland can you please share a reproduction for this issue? otherwise we won't be able to tell if we managed to fix that.

dotansimha:
I think this was fixed in the recent PRs. Can you please try `1.12.3-alpha-b89e11fd.107`?

zorji:
@dotansimha I've tested 1.13 and it's still an issue.

The issue is, it generates references to `FragmentDoc` regardless of whether `FragmentDoc` was generated or not.

Having `preset: near-operation-file` can reproduce this issue.

Looks like @mjbcopland pointed out where the issue is because that `if` always evaluate to `true`.

dotansimha:
@zorji I think just using near-operation-file` doesn't reproduce it, it might be something specific to your setup. We have tested `near-operation-file` with different setups and it seems to work, but we probably missed one of the edge cases. 

Can you please create a reproduction for this? with the recent alpha `1.13.1-alpha-84171f94.18`.

zorji:
@dotansimha Hi, I've created a reproduction here https://github.com/zorji/graphql-codegen-3525

Simply run `npx graphql-codegen` and check the generated file `query.types.d.ts`. Thanks.

dotansimha:
Thanks @zorji . 
Your config file seems a bit messy, I first did the following steps:

1. Use the correct configurations. Why do you need to generate operations and resolvers in the same context? is it for client-side state? 

It could simply look like that:

```yml
overwrite: true
generates:
  src/base-types.ts:
    schema: ./src/base.graphql
    plugins:
      - typescript

  src/graphql:
[...]

PR #3646, opened 2020-03-16
Use importFrom to import external fragments

mjbcopland:
Discussion on #3525

Would be good to get @micimize's thoughts on what that TODO was for and whether this is a reasonable change.

### DIFF
--- a/packages/plugins/typescript/graphql-request/tests/graphql-request.spec.ts
+++ b/packages/plugins/typescript/graphql-request/tests/graphql-request.spec.ts
@@ -183,7 +183,7 @@ async function test() {
       };
     };
 
-    it.skip(
+    it(
       'should execute default no-op middleware when none passed to getSdk',
       async () => {
         const { sdk, scope } = await perpareTranspiledModule();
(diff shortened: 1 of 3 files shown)


micimize:
It was about whether to replace it with a generated import statement per-fragment, but I think it was made irrelevant by the `generateImportStatement` used for `fragmentImportStatements`.

dotansimha:
It seems like we should have a different solution, as you suggested. 
The changes I did there yesterday was just a workaround, and I think plugins should be in charge of adding those imports. This way the import for `FragmentDoc` will be done by `ClientSideBaseVisitor` and not by the preset itself (but, the preset should let the plugin know where the fragment file is located).

wassim-k:
@dotansimha what's the status of this PR?

I'm working on a 3rd party plugin which requires fragment doc imports, unfortunately the current implementation does not work, since `DOC_VAR_PLUGINS` array cannot know about 3rd party plugins.

If you can let me know what's missing from this PR? I can attempt a different solution if you don't think this one solves the problem?

wassim-k:
I've created a PR #3771 for an improved implementation, looking forward to your feedback.

dotansimha:
Merged https://github.com/dotansimha/graphql-code-generator/pull/3771 recently, with improved implementation, so we can close this one. 

Thank you @mjbcopland , @micimize , @wassim-k for amazing work!

PR #3648, opened 2020-03-16
(urql-plugin) - add additionalTypenames functionality

JoviDeCroock:
When working with `document-based-caching` in `urql` we had the caveat that when an empty list was returned this query would never be invalidated.
This can be fixed by passing `additionalTypenames` to the `useQuery` hook so the doc-cache is aware of what the dependencies are.
Now we are passing it all the time, I could see if it's a list and only pass it then, up to you what you prefer.

[docs link](https://formidable.com/open-source/urql/docs/basics/document-caching/#adding-typenames)

I made this a draft PR so we can discuss on what standards you'd like to see, I tried copying from other plugins.

Todo:
- Fragments containing types


dotansimha:
This is cool, thanks @JoviDeCroock ! 
I'm not familiar with `urql` and all it's features, but this seems like something we can support. 
Does this hooks replaces the other hooks generated by the codegen? because it seems like the name `useXYZQuery` is the same. 

Maybe @mxstbr can help with reviewing it?

JoviDeCroock:
@dotansimha it replaces the usual `query` since we add the `memoized context`, this would if the user chooses to use `document-cache` prevent weird bugs when the return value would for instance be this:

```
query todos { id }
const result = { todos: [] }
```

now we don't know that todos has a dependency on `__typename: Todo` and it won't automatically invalidate when a `mutation` has a result of `Todo`. This makes life a bit easier on the user so it should just replace the normal query when the user chooses for this approach.

packages/plugins/typescript/urql/tests/urql.spec.ts
```diff
@@ -590,6 +590,147 @@ export function useSubmitRepositoryMutation() {
...
+      const documents = parse(/* GraphQL */ `
+        query feed {
+          feed {
+            ...MyEntry
+          }
+        }
+
+        fragment MyEntry on Entry {
+          id
+          repository {
```
JoviDeCroock: Do we support doing this

JoviDeCroock: @dotansimha it seems like we aren't sending the `fragmentDefinitions` along to the `_buildX` methods. Is this possible? If so do we expect that to happen frequently?

mxstbr:
I'll defer to @JoviDeCroock who makes urql—if he thinks we should do this, let's do it! :wink:

packages/plugins/typescript/urql/tests/urql.spec.ts
dotansimha: I think it's something you can add :)

dotansimha:
@JoviDeCroock I think we can continue with that PR. What are the missing pieces here? is there something I can do to help?

JoviDeCroock:
@dotansimha if you could point me where to do the fragments stuff that would be awesome, I'll probably have time to look at it this weekend.

jgoux:
Hi, what is blocking this PR?

JoviDeCroock:
@dotansimha I couldn't find a way to add fragments, if you could point me at an existing place that would be great. This is finished apart from that https://github.com/dotansimha/graphql-code-generator/pull/3648/files#diff-cb39cefa28181b86b615240373524d1aR715

dusty:
@JoviDeCroock I'm curious, did you ever figure out how to get the typenames from fragments?  I see the PR is still open, so I'm guessing no.  Thanks.

PR #3657, opened 2020-03-18
Added 'field' flag to avoidOptionals

wassim-k:
v1.13.1 changed the behaviour of optional fields.
All nullable fields are now optional `?` including fields inside queries and fragments.
I found that a bit confusing knowing that if a field is requested as part of a query or fragment, then the GraphQL engine will always return the field, and if it's a nullable field, then its value could be null but still present.

Adding `avoidOptionals: true` to `typescript-operations` plugin doesn't work because that also makes all the nullable properties of query and mutation variables non-optional, which is not the desired behaviour.

I've managed to fix it by introducing a new `field` flag to `avoidOptionals` config which specifically makes field values (those returned by queries and fragments) non-optional.

This works because `config.avoidOptionals.field` is being read by both `typescript` and `typescript-operations` plugins which keeps the type definitions consistent.

This also ensures that when `preResolveTypes` is `false` then the fragment definition still matches the original type that it's applying `Pick` on.

### DIFF
--- a/docs/generated-config/typescript-operations.md
+++ b/docs/generated-config/typescript-operations.md
@@ -25,6 +25,7 @@ path/to/file.ts:
    - typescript
  config:
    avoidOptionals:
+     field: true
      inputValue: true
      object: true
 ```
--- a/docs/generated-config/typescript.md
+++ b/docs/generated-config/typescript.md
@@ -24,6 +24,7 @@ path/to/file.ts:
    - typescript
  config:
    avoidOptionals:
+     field: true
      inputValue: true
      object: true
 ```
--- a/packages/plugins/other/visitor-plugin-common/src/avoid-optionals.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/avoid-optionals.ts
@@ -1,8 +1,9 @@
 import { AvoidOptionalsConfig } from './types';
 
-export const DEFAULT_AVIOD_OPTIONALS: AvoidOptionalsConfig = {
+export const DEFAULT_AVOID_OPTIONALS: AvoidOptionalsConfig = {
   object: false,
   inputValue: false,
+  field: false,
 };
 
 export function normalizeAvoidOptionals(avoidOptionals?: boolean | AvoidOptionalsConfig): AvoidOptionalsConfig {
@@ -10,11 +11,12 @@ export function normalizeAvoidOptionals(avoidOptionals?: boolean | AvoidOptional
     return {
       object: avoidOptionals,
       inputValue: avoidOptionals,
+      field: avoidOptionals,
     };
   }
 
   return {
-    ...DEFAULT_AVIOD_OPTIONALS,
+    ...DEFAULT_AVOID_OPTIONALS,
     ...avoidOptionals,
   };
 }
--- a/packages/plugins/other/visitor-plugin-common/src/index.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/index.ts
@@ -5,6 +5,7 @@ export * from './naming';
 export * from './scalars';
 export * from './enum-values';
 export * from './declaration-kinds';
+export * from './avoid-optionals';
 
 export * from './base-visitor';
 export * from './base-types-visitor';
--- a/packages/plugins/other/visitor-plugin-common/src/types.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/types.ts
@@ -53,3 +53,9 @@ export interface DeclarationKindConfig {
   interface?: DeclarationKind;
   arguments?: DeclarationKind;
 }
+
+export interface AvoidOptionalsConfig {
+  field?: boolean;
+  object?: boolean;
+  inputValue?: boolean;
+}
--- a/packages/plugins/typescript/operations/src/config.ts
+++ b/packages/plugins/typescript/operations/src/config.ts
@@ -1,5 +1,4 @@
-import { RawDocumentsConfig } from '@graphql-codegen/visitor-plugin-common';
-import { AvoidOptionalsConfig } from '@graphql-codegen/typescript';
+import { RawDocumentsConfig, AvoidOptionalsConfig } from '@graphql-codegen/visitor-plugin-common';
 
 export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
   /**
@@ -29,6 +28,7 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
    *    - typescript
    *  config:
    *    avoidOptionals:
+   *      field: true
    *      inputValue: true
    *      object: true
    * ```
--- a/packages/plugins/typescript/operations/src/visitor.ts
+++ b/packages/plugins/typescript/operations/src/visitor.ts
@@ -9,6 +9,8 @@ import {
   SelectionSetProcessorConfig,
   SelectionSetToObject,
   DeclarationKind,
+  normalizeAvoidOptionals,
+  AvoidOptionalsConfig,
 } from '@graphql-codegen/visitor-plugin-common';
 import { TypeScriptOperationVariablesToObject } from './ts-operation-variables-to-object';
 import { TypeScriptDocumentsPluginConfig } from './config';
@@ -17,7 +19,7 @@ import { TypeScriptSelectionSetProcessor } from './ts-selection-set-processor';
 import autoBind from 'auto-bind';
 
 export interface TypeScriptDocumentsParsedConfig extends ParsedDocumentsConfig {
-  avoidOptionals: boolean;
+  avoidOptionals: AvoidOptionalsConfig;
   immutableTypes: boolean;
   noExport: boolean;
 }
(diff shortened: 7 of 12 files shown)


KillerCodeMonkey:
we ran into the same issue that changing the optionals behavior is a `breacking change` in our projects.

We allowed optionals for mutations (inputs), but queries always returning null for fields, but now nullable fields are marked as optional, so many type checks are failing in our Application, because the values are not only possible `null`. Now they can be undefined as well.

I think there is no additional option "field" necessary, but a field should not be typed as `undefined` at all.

patroza:
Yes, this was indeed a bad backwards compatible breakage with no option to return to the previous behavior. Would be great this fix gets approved, or the behaviour reverted.

wassim-k:
I agree with both of you, I believe the issues mentioned in the pull request #3639 that changed the behaviour of optional fields should have been solved by making all fields for types generated by `typescript` plugin non-optional instead of the solution proposed by the PR which was to make all fields generated by `typescript-operations` plugin optional (queries, fragments, etc...)
But since the maintainers have approved that PR, the changes in this PR make it possible to opt-in for non-optional fields in both plugins instead.
Though I'm happy to change it so that's the default behaviour once we hear the maintainers' feedback.

dotansimha:
I'm not sure about this one. 
This was not a breaking change from my point of view, since it fixed a bug. The behaviour has changed, that's right, but it was incorrect from the beginning. 
The PR we merged fix that issue. The way to change the behaviour is `avoidOptionals`, but I can see why it's not complete.

I think the changes you introduced here are super helpful, thank you @wassim-k .

KillerCodeMonkey:
@dotansimha but is it really possible in graphql to return "undefined" as field type? I thought in general `null` is returned by graphql

patroza:
Exactly. I think there are important subtle differences;

- ? Marks the ability to omit a field
- | undefined says the field must be there but may have value undefined.

For any model;
- A field can be null able or non null able. There is no undefined

For Query and Mutation fields;
- if you don’t ask for a field, it will not be in the output
- if you ask for a field, it will be in the output, either null able or non null able 

For Query/Mutation variables, you are allowed to omit a nullable variable, in which case it will be treated as undefined. I really like that because it means you can make mutations that allow you to model; unchanged vs null able and non nullable

In short, in the previous version, imo avoidOptionals with object set to true, made for the most correct behavior; I even would argue that should be the default. If people want less correct behavior for some reason then they can achieve it by some configuration.

KillerCodeMonkey:
@dotansimha @patroza maybe there should a difference for typescript and typescript-operations plugins.

if you are using only the typescript plugin... it might be okay that `?` marks them as undefined, because these are the base types and we do not know  if the field would be queried. But when using typescript-operations it is clear when using  `Pick` that those fields or not undefined.

dotansimha:
@KillerCodeMonkey no, `undefined` is being transformed into `null` on backend, and not possible on queries selection set.

The issue was originally with `typescript` plugin - it was a bug that ignored some nullable fields (on the root level), but applied the optionals nested fields.
**The behaviour didn't change, we just fixed a bug.**
The `typescript` plugin is the base types, and it should have types that are compatible with `typescript-resolvers` and `typescript-operations`. It's not always easy because what nullable means for client-side is not always what it means for server-side. 

The default behaviour of codegen is to apply add optionals, **changing that will be a breaking change**. 
That's why we have a lot of configurations related to it, to allow people to control the output as they need. 
You can use `avoidOptionals` to change that, and this PR extends the changes that it can apply to the output. You can also use `preResolveTypes` to have less dependencies on `typescript` base plugin.

patroza:
@dotansimha in theory yes, I will try it out once it lands, will be happy to.

Still, I feel the most correct is avoidOptionals with object and field to true then, and that it should be the default - perhaps in some future major release. But of course as long as I can achieve my use case with some configuration and it will remain working, im fine too :)

dotansimha:
@patroza We are releasing alpha version automatically for all commits. 
You can try the changes from this PR with `1.13.2-alpha-cd194dd9.26`.

patroza:
@dotansimha @wassim-k I'm trying said alpha version, but to no avail, my query/mutation fields and fragment fields still have optionals :/

dotansimha:
@patroza can you please share you `package.json` and `codegen.yml` file?

patroza:
codegen.yml:
```yaml
overwrite: true
schema: "xxx/graphql"

config:
  scalars:
    Date: Date
    DateTime: Date
    Time: string

  # constEnums: true
  avoidOptionals:
    object: true
    inputValue: false
    field: true
  namingConvention: keep
  skipTypename: true
  preResolveTypes: true

generates:
  src/schema/generated/best-in-flights.schema.json:
    plugins:
      - "introspection"
    config:
      minify: true
  src/services/flights-api/generated/graphql.ts:
    documents:
      - "src/services/flights-api/GetFlights.graphql"

    plugins:
      - "typescript"
      - "typescript-operations"
    config:
      skipTypename: false
    hooks:
      afterOneFileWrite:
        - prettier --write
```

package.json:
```json
  "dependencies": {
    "@fineway/app": "^3.6.4",
    "@fineway/dependency-injection": "^3.6.4",
    "@fineway/tracking": "^3.6.4",
    "@fineway/utils": "^3.6.4",
    "@sentry/node": "^5.15.0",
    "accept-language-parser": "^1.5.0",
    "apollo-cache-inmemory": "^1.6.5",
    "apollo-client": "^2.6.8",
    "apollo-link-http": "^1.5.16",
    "apollo-link-scalars": "^0.1.6",
    "apollo-link-timeout": "^1.4.0",
    "apollo-server-koa": "^2.11.0",
    "cheerio": "^1.0.0-rc.3",
    "dataloader": "^2.0.0",
    "eslint-plugin-jest": "^23.8.2",
    "format-number": "^3.0.0",
    "graphql": "^14.6.0",
    "graphql-import-node": "^0.0.4",
    "graphql-iso-date": "^3.6.1",
    "graphql-tag": "^2.10.3",
    "graphql-tools-fork": "^8.9.6",
    "graphql-type-json": "^0.3.1",
    "json-stable-stringify": "^1.0.1",
    "kcors": "~2.2.2",
    "koa": "~2.11.0",
    "koa-basic-auth": "~4.0.0",
    "koa-timeout": "^0.1.1",
    "rambda": "^5.0.0",
    "react": "^16.8.6",
    "tsconfig-paths": "^3.9.0",
    "typescript": "^3.8.3",
    "universal-cookie-koa": "^4.0.3",
    "winston": "^3.2.1"
  },
  "devDependencies": {
    "@graphql-codegen/cli": "~1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/introspection": "~1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/typescript": "~1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/typescript-operations": "~1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/typescript-resolvers": "~1.13.2-alpha-cd194dd9.26",
    "@types/jest": "^25.1.4",
    "@types/koa-basic-auth": "^2.0.3",
    "@types/node": "~12",
    "@typescript-eslint/eslint-plugin": "^2.24.0",
    "@typescript-eslint/parser": "^2.24.0",
    "eslint": "^6.8.0",
    "eslint-config-prettier": "^6.10.0",
    "eslint-import-resolver-typescript": "^2.0.0",
    "eslint-import-resolver-webpack": "^0.12.1",
    "eslint-plugin-import": "^2.20.1",
    "eslint-plugin-prettier": "^3.1.2",
    "eslint-plugin-sort-destructure-keys": "^1.3.3",
    "jest": "^25.1.0",
    "nodemon": "^2.0.2",
    "prettier": "^1.19.1",
    "ts-jest": "^25.2.1",
    "ts-node": "^8.7.0",
    "ts-node-dev": "^1.0.0-pre.44",
    "ts-toolbelt": "^6.3.6",
    "wait-on": "^4.0.1"
  },
```

dotansimha:
@patroza  Don't use `~` or `^` in the dependencies of the codegen, because you'll get a different version that you specified.

```
    "@graphql-codegen/cli": "1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/introspection": "1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/typescript": "1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/typescript-operations": "1.13.2-alpha-cd194dd9.26",
    "@graphql-codegen/typescript-resolvers": "1.13.2-alpha-cd194dd9.26",
```

patroza:
@dotansimha alright, many thanks, that helped.

However there's still a problem. Query variables are now mandatory. Mutation variables on the other hand remain optional I guess because of `inputValues`, imo query variables should follow that too.

wassim-k:
@patroza just to clarify, both `inputValue` and `object` options apply equally to all operations, so there should be no difference between queries and mutations.

What you might be seeing there is the difference between `inputValue` and `object`, and I'll be honest I found the two hard to distinguish myself.

`object: true` makes all nullable [Variables](https://spec.graphql.org/June2018/#sec-Validation.Variables) for queries/mutations optional.
`inputValue: true` makes all nullable properties in [Input Object Values](https://spec.graphql.org/June2018/#sec-Input-Object-Values) optional.

For example, for this schema:

```gql
type Query {
  users(sort: SortInput): [User!]!
}

type User {
  id: String!
}

input SortInput {
  field: String!
  direction: String
}
```

`object: true` will generate `type UsersQueryVariables { sort: Maybe<SortInput> }`
`object: false` will generate `type UsersQueryVariables { sort?: Maybe<SortInput> }`

`inputValue: true` will generate `type SortInput { field: string; direction: Maybe<string>  }`
`inputValue: false` will generate `type SortInput { field: string; direction?: Maybe<string>  }`

I hope this helps, as for `typescript-resolvers` plugin, I haven't used it in the past, sorry can't help you there.

patroza:
Alright, would be great if we can match `inputValue` to QueryVariables, or have `queryValue` as a new option :/

patroza:
Hold on a minute, it seems if `object: false` and `field: true` I actually get almost the behaviour I want. Minus that on the resolver side it still adds `RequireFields<`...
Well progress I guess :)

ekfn:
@wassim-k You are my hero!

PR #3694, opened 2020-03-22
Shared Jest Cache + Shared devDeps + Incremental builds + Reverts outputs dependencies

kamilkisiela:
Cold build takes 20s, rebuilds are done within 6s.
Cached jest tests take 13s.

We use projects so later on we can introduce incremental testing based on changed files that we @dotansimha talked about some time ago.

The Guild Bot updates comments instead of creating them on each commit.


kamilkisiela:
I'll fix the CLI build tomorrow

kamilkisiela:
Later on, I'll add it to bob, since current setup is quite different from other projects. We will migrate them soon too.

PR #3699, opened 2020-03-23
Update Java plugin: Omit extra 'Input' suffix from an input's class name if the input type's name ends in Input

ataylormays:
The crux of the problem I'm trying to solve is the following: consider a schema that includes a type like 

```
input MyEntityInput {
```

In this case, the current codegen will generate a Java class called MyEntityInputInput.java. I'd like to update this to simply generate MyEntityInput.java. I've updated the code to check whether the input type from the schema ends with "Input" and omit adding the "Input" suffix if it does.

The big use case here is when we're using graphql for CRUD operations. If that's the case, then it's very likely that the read operations will necessitate a schema that includes

```
type MyEntity {
```

but that means that we'd want to include the following for writes:

```
input MyEntityInput {
```

The one thing I'm hesitant about with my change is backwards compatibility. For anyone that's been using this tool and has `FooInputInput.java` classes in their codebase, this would start generating different class names. Depending on the philosophy of this open source project, this change could either be considered a breaking change for that use case, or there could be something introduced to the config file to maintain backwards compatibility, such as allowing users to specify a suffix or omit the default suffix. Let me know what your thoughts are!

### DIFF
--- a/packages/plugins/java/java/src/visitor.ts
+++ b/packages/plugins/java/java/src/visitor.ts
@@ -62,7 +62,7 @@ export class JavaResolversVisitor extends BaseVisitor<JavaResolversPluginRawConf
       allImports.push(`java.util.stream.Collectors`);
     }
 
-    return allImports.map(i => `import ${i};`).join('\n') + '\n';
+    return allImports.map((i) => `import ${i};`).join('\n') + '\n';
   }
 
   public wrapWithClass(content: string): string {
@@ -104,7 +104,7 @@ export class JavaResolversVisitor extends BaseVisitor<JavaResolversPluginRawConf
     this._addHashMapImport = true;
     this._addMapImport = true;
     const enumName = this.convertName(node.name);
-    const enumValues = node.values.map(enumValue => (enumValue as any)(node.name.value)).join(',\n') + ';';
+    const enumValues = node.values.map((enumValue) => (enumValue as any)(node.name.value)).join(',\n') + ';';
     const enumCtor = indentMultiline(`
 public final String label;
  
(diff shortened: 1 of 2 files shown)


packages/plugins/java/java/src/visitor.ts
```diff
@@ -62,7 +62,7 @@ export class JavaResolversVisitor extends BaseVisitor<JavaResolversPluginRawConf
       allImports.push(`java.util.stream.Collectors`);
     }
 
-    return allImports.map(i => `import ${i};`).join('\n') + '\n';
+    return allImports.map((i) => `import ${i};`).join('\n') + '\n';
```
ataylormays: In this and the other arrow function updates, I didn't explicitly add the parens around the variables - not sure if this was done by a linter on the project when I committed or if it was done by some other automatic process. Let me know if I should change these back or leave them.

packages/plugins/java/java/src/visitor.ts
```diff
@@ -262,13 +262,14 @@ ${getters}
   }
 
   InputObjectTypeDefinition(node: InputObjectTypeDefinitionNode): string {
-    const name = `${this.convertName(node)}Input`;
+    const convertedName = this.convertName(node);
+    const name = convertedName.endsWith('Input') ? convertedName : `${convertedName}Input`;
```
ataylormays: This is the crux of the change - omit the "Input" suffix if the schema name already includes it

dotansimha:
Thank you @ataylormays ! 
Maybe in order to prevent the breaking change, this could be a a optional configuration?

Also, what about a use case like this?

```graphql
type User {

}

type UserInput {

}
```

this will generate two types with the same name, no?

dotansimha:
@ataylormays ?

ataylormays:
Hey @dotansimha! Sorry for the delay. I can look into making this an optional configuration. I was trying to look back at the docs for the configs but the website isn't rendering them: https://graphql-code-generator.com/docs/plugins/java#configuration Any idea when those will be back up? It'd be helpful for working on config changes.

Screen Shot 2020-04-16 at 10 58 21 AM

Regarding your use case above, this change wouldn't affect it because the change only applies to `input` types in your schema. The only time you'd have conflicting classnames is if you had something like the following:

```
type UserInput {

}

input User {

}
```

dotansimha:
Ok sounds good. And the website was fixed ;) 
Thanks!!!

Issue #3695, opened 2020-03-22
Support TypeScript's `import type` syntax and `importsNotUsedAsValues: error` config


SimenB:
**Is your feature request related to a problem? Please describe.**

The code generated has TS type errors if `importsNotUsedAsValues: error` is used in config

**Describe the solution you'd like**

An option to use `import type` instead of `import`

**Describe alternatives you've considered**

**Additional context**

adam-beck:
I've tried to just `"exclude"` the generated file in my tsconfig but I still get the error. Does anybody know what that's happening?

SimenB:
With #3706, #4436, #4437, #4447, #4448 all applied locally my work project successfully compiles with `importsNotUsedAsValues: error` config 🎉 There's probably loads of other plugins/presets that we don't use that also should be updated, but it's fairly straightforward to do

dotansimha:
Available in most plugins since `v1.17.7`. Thanks @SimenB !

yoavain-sundaysky:
I see this one is closed, but I'm still having some problems with this one.
1. Getting this 
```import { GraphQLResolveInfo, GraphQLScalarType, GraphQLScalarTypeConfig } from 'graphql';```
instead of this
```import type { GraphQLResolveInfo, GraphQLScalarType, GraphQLScalarTypeConfig } from 'graphql';```

2. When configuring `contextType` i.e 
```contextType: ../context#ApolloContext```
I'm getting:
```import { ApolloContext } from '../context';```
Instead of:
```import type { ApolloContext } from '../context';```

3. Getting this:
```import * as ApolloReactCommon from "@apollo/react-common";```
Instead of this:
```import type { BaseMutationOptions, MutationFunction, MutationResult, QueryResult } from "@apollo/react-common";```

Should I open a new ticket?

**Comment by dotansimha** [comment]
@yoavain-sundaysky please report it in a new issue. We tried to cover all use cases, but I afraid we missed some.

**Comment by f0rmat1k** [comment]
@SimenB looks like there's still a problem. I've just added importsNotUsedAsValues to my project and got:
```
Type error: This import is never used as a value and must use 'import type' because 'importsNotUsedAsValues' is set to 'error'.

> 1 | import { FieldPolicy, FieldReadFunction, TypePolicies, TypePolicy } from '@apollo/client/cache';
    | ^
```

SimenB:
https://github.com/dotansimha/graphql-code-generator/issues/3695#issuecomment-711949594

PR #3706, opened 2020-03-24
feat: support type imports

SimenB:
Very rough start on #3695. Not sure how you want to handle this. I'll probably just do the ones we need at work (which means Apollo TS), but it should be fine to expand as we go?

(is there some watch mode I can run that builds as we go? right now I do a full `yarn build` every time I change something, which seems suboptimal)

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -705,7 +705,19 @@ export class BaseResolversVisitor<
     }
 
     const defaultType = types.find(t => t.asDefault === true);
-    const namedTypes = types.filter(t => !t.asDefault);
+    let namedTypes = types.filter(t => !t.asDefault);
+
+    if (this.config.useTypeImports) {
+      if (defaultType) {
+        // default as Baz
+        namedTypes = [{ identifier: `default as ${defaultType.identifier}` }, ...namedTypes];
+      }
+      // { Foo, Bar as BarModel }
+      const namedImports = namedTypes.length ? `{ ${namedTypes.map(t => t.identifier).join(', ')} }` : '';
+
+      // { default as Baz, Foo, Bar as BarModel }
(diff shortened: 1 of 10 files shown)


SimenB:
Only reason for an option rather than being default behavior is that it requires TS 3.8. Maybe some "minimum TS version" rather than an explicit option makes more sense?

dotansimha:
This is awesome! Thank you @SimenB ! 

This could work, and could be optional via configuration. What about importing enums and other things? how is it effected?

SimenB:
Are enums imported as values anywhere? The test should probably try to compile the module we generate with `isolatedModules: true` and `importsNotUsedAsValues: error`. I don't think I know enough about your setup to add that, though

https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#a-simple-transform-function

ardatan:
For example, there are const enums used as values not types. So `import type` would break this behavior.

SimenB:
`const enum` aren't supported by `isolatedModules` regardless, so I wouldn't want them 😃

vatosarmat:
Any chances this useful feature to be merged? :smiley_cat: 
Since it is optional and disabled by default, it shouldn't break anything for anyone.

dotansimha:
@vatosarmat merging :) 
I'm merging it now, and will push some fixes soon to make sure all codebase supports that. 

Thanks @SimenB , sorry it took so long.

SimenB:
Thanks @dotansimha! I believe this was still incomplete, happy to hear you'll fix the missing parts 👍

I wanted to add some tests with `isolatedModules: true` and `importsNotUsedAsValues: error`, but didn't get around to it. I believe such tests would be quite useful to ensure no regressions. Never got around to it, though 😅

SimenB:
Tested this out now in the new release, and it works for some cases. However, it does not affect

- `from 'graphql'`
- `from '@apollo/react-common'`
- imports from `near-operation-file`'s `baseTypesPath`
- fragment imports (possibly from `near-operation-file`?)

@dotansimha My OSS backlog is really long right now, so I won't be able to work on this. 🤞 you can 😀

Issue #3712, opened 2020-03-24
plugin development documentation?


seansfkelley:
**Is your feature request related to a problem? Please describe.**

I'm trying to author a new plugin (`python-operations`) in my fork of this repo. However, while [the docs talk about the basics of plugin development](https://graphql-code-generator.com/docs/custom-codegen/write-your-plugin), I'm hung up on tooling issues that even before that. A selection of things:

- How do I install the repo properly? I used some combination of yarn/lerna to get it into a valid state, and then I did _something_, and now my builds are failing and VS Code is complaining, unable to resolve `@graphql-codegen/*` dependencies.
- What's a good development workflow for testing a new plugin on a real codebase? I matched versions with my project and tried to symlink in my new plugin, but it complained about having multiple versions of (I think) `graphql` (an error message which I greatly appreciate, but means I have some hacky workaround to test my very-much-still-in-development plugin on my project).
- Are there common unit or integration tests that cover every GraphQL language feature that plugins should attempt to support? For instance, I don't use subscriptions and am not familiar with them, but if there were even one big integration test fixture that covers the syntax and I could compare what e.g. the `typescript` plugin does for it, I'm reasonably confident I could make a passable Python equivalent.
- I'd like to reuse the `BaseDocumentsVisitor` to do `python-operations`, but there are a _lot_ of moving parts. I'm trying to learn how `SelectionSetToObject` and `BaseSelectionSetProcessor` and `OperationVariablesToObject` and `DeclarationBlock` work simultaneously and it's pretty confusing. A document that explains what their role is and how they interact with each other would be awesome, because I think there's a lot in there I can reuse, but they also appear somewhat coupled to Flow/Typescript.

**Describe the solution you'd like**

More documentation! What's on the site now is a good start, but only covers the very basics and assumes you're working in your own repo, which I'm not (partly because I'm targeting Python, so I don't want to have to set up a node package when I already have `graphql-code-generator` available, and partly because I'm interested in eventually contributing this back to the mainline).

**Describe alternatives you've considered**

Trying random stuff only works for so long. I really can't figure out how I got my repo into this state, and why the nuclear option of deleting all node_modules and starting over isn't working.

**Additional context**

Similar to https://github.com/dotansimha/graphql-code-generator/issues/951, but that one is quite old and appears to have been the impetus for writing the documentation that currently exists.

GavinRay97:
I actually asked the same question in the Discord 15 minutes ago, hopefully get some good info.

[...]

PR #3715, opened 2020-03-24
WIP: Fix JS Doc comments in Vue Apollo queries

GavinRay97:
Currently, because of the `type` declaration placed before the exported function in queries, it does not pick up the JSDoc docstring:

image

image

image

Adding `@param variables` and moving the `type` below the function makes it work properly:

image

image

I am unable to test these changes properly because I wasn't able to find documentation on the development workflow for modifying existing plugins.

I tried running the `build` script from the main repo and using the `index.cjs.js`, but during the codegen process it errors due to there being "duplicate copies of graphql library":
```
ensure there is only one instance of graphql
```

### DIFF
--- a/packages/plugins/typescript/vue-apollo/src/visitor.ts
+++ b/packages/plugins/typescript/vue-apollo/src/visitor.ts
@@ -113,6 +113,7 @@ export class VueApolloVisitor extends ClientSideBaseVisitor<VueApolloRawPluginCo
  * __use${operationName}__
  *${operationType === 'Mutation' ? mutationDescription : queryDescription}
  *
+ * ${['Query', 'Subscription'].includes(operationType) ? '@param variables Query variables' : ''}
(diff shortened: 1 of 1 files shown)


iskanderbroere:
The jsdoc bug (it being in the wrong place) is fixed in https://github.com/dotansimha/graphql-code-generator/pull/3891 because of some refactoring in the code, removing the generation of the `ReactiveFunction` type and therefore putting the jsdoc in the right place. It'd be nice if you could improve the generation of jsdoc with @params tho 😄

GavinRay97:
@iskanderbroere Amazing, thank you!

Just curious, what is the development workflow for working on plugins? I couldn't find anything in the docs or contributing documentation.

dotansimha:
@GavinRay97 We are updating it now, see https://github.com/dotansimha/graphql-code-generator/issues/3712#issuecomment-616073350 

And thank you for this PR, we'll go with https://github.com/dotansimha/graphql-code-generator/pull/3891 because it includes this as well.

PR #3718, opened 2020-03-25
feat: add generateOnlyEnums option to typescript

c0m1t:
`generateOnlyEnums` was added to `typescript` plugin `options`.  
This allows `typescript` plugin to generate only enums.

Our team use `relay` package.  `relay-compiler` generates `Query`, `Mutation` and `fragment` types based on `schema.graphql` file. So other types this library generates are not useful to us, except enums.  

The problem is whenever we have a type that uses an `enum` it generates that type again, I mean there could be definition of an `enum` in 20 different files. We always liked to have a single source of truth for enums. We even wrote a small package [`relay-enum-generator`](https://github.com/c0m1t/relay-enum-generator).  So we would really appreciate it if there were an option in your package to make this possible. I suppose other teams using `relay` probably have this issue.

### DIFF
--- a/packages/plugins/typescript/typescript/src/config.ts
+++ b/packages/plugins/typescript/typescript/src/config.ts
@@ -182,4 +182,23 @@ export interface TypeScriptPluginConfig extends RawTypesConfig {
    * ```
    */
   wrapFieldDefinitions?: boolean;
+
+  /**
+   * @name generateOnlyEnums
+   * @type boolean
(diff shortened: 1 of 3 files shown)


dotansimha:
Thank you @c0m1t ! 
We recently merged this: https://github.com/dotansimha/graphql-code-generator/pull/3760 this also does that, and generates only enums ;)

PR #3721, opened 2020-03-25
feat: futureProofEnums option

c0m1t:
`relay` users usually use [noFutureProofEnums option](https://github.com/facebook/relay/blob/da7ed10312e6f10743e646ce2e3027059fd40d73/packages/relay-compiler/bin/RelayCompilerBin.js#L101) in their projects. Although It's clear that this is not something hard to accomplish without this pull request, with a utility type like this:
`type FutureProofEnums<T> = T | '%future added value'`

I thought it would be cool to have an option to support it. For now, `futureProofEnums` works only when `enumsAsTypes` is `true`. Because it looks weird if you add `%future added value` to an `enum`.

Also:
- Moved the  `|` sign from the end of the lines, to the first to be more like prettier format.
- Refactor 2 lines of code

### DIFF
--- a/packages/plugins/typescript/typescript/src/config.ts
+++ b/packages/plugins/typescript/typescript/src/config.ts
@@ -69,6 +69,25 @@ export interface TypeScriptPluginConfig extends RawTypesConfig {
    */
   enumsAsTypes?: boolean;
   /**
+   * @name futureProofEnums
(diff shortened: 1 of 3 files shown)


dotansimha:
Super cool! Thank you @c0m1t !

PR #3726, opened 2020-03-26
Add a new plugin for C#

kurmeroli:
Hi there,

I've tried to make a  C# plugin for the graphql-code-generator. The generated code works fine with the official GraphQL.NET library.
Honestly the code might not be that pretty everywhere. But it's a first running version. And we thought it would be a pity doing this and don't share it as open source.


packages/plugins/c-sharp/apollo-dotnet/src/visitor.ts
```diff
@@ -0,0 +1,287 @@
...
+  }[] = [];
+
+  constructor(schema: GraphQLSchema, fragments: LoadedFragment[], rawConfig: ApolloDotNetRawPluginConfig, documents?: Types.DocumentFile[]) {
+    super(
+      schema,
+      fragments,
+      rawConfig,
+      {
+        sdkClass: rawConfig.sdkClass,
+        ngModule: rawConfig.ngModule,
```
ardatan: Do we need that?

ardatan:
Thank you @kurmeroli ! Amazing work.
Could you check my comment ?

packages/plugins/c-sharp/apollo-dotnet/src/visitor.ts
kurmeroli: maybe not, I will check this and let you know once I finished.
sorry not sure how early I could have a look on it. maybe during the easter weekend

kurmeroli: we did not need that, it has been removed.

ardatan: Thank you! Could you align graphql version as well? So we can get it merged.

kurmeroli: aligned to graphql 15.0.0

ardatan:
LGTM @dotansimha

dotansimha:
Thanks @kurmeroli !

clintsinger:
Hi, this is a great help. 
 
One thing I noticed is that it doesn't do Pascal Case even though that is supposed to be the default.   Shouldn't it upper case the first letter of each field? 

It doesn't even do it if explicitly set in the codegen configuration file.

PR #3729, opened 2020-03-26
Remove tsconfig.json and unused files from packages

standy:
Removed files from each package, they were doing nothing
```bash
tsconfig.json # not used anymore
.npmignore # not used, dist folder contains the whole release
.gitignore # root is enough
yarn.lock # root is enough 
.npmrc # didn't do anything
```

The reason to remove `tsconfig.json` files from packages is IDE going crazy (I'm using webstorm), and type completion is broken because of them. Other files were unused, so should be removed


dotansimha:
Thank you @standy :)

PR #3736, opened 2020-03-27
Put configuration into a codeblock to make it more readable

JoeSchr:


dotansimha:
Thanks @JoeSchr , but is there a reason to do that? 
I mean, the configuration partial uses code block where needed, and it's formatted as list. See how the result looks like: https://graphql-code-generator.com/docs/plugins/typescript-vue-apollo#configuration

JoeSchr:
On my side it looks like this:
image

instead of e.g.: https://graphql-code-generator.com/docs/plugins/typescript-mongodb
image

which I find way more readable and quicker to comprehend

dotansimha:
It had a missing close block for the code (```), I updated it in the source, and it will be in the website after next deployment. 

Thanks @JoeSchr !

JoeSchr:
👍

Issue #2192, opened 2019-07-18
Why are ReactApollo.DataProps wrapped in Partial<> for HOC Props?


bensaufley:
`typescript-react-apollo` will generate the following Props for a HOC for a query named `GetThings`:

```ts
export type GetThingsProps<TChildProps = {}> = Partial<
  ReactApollo.DataProps<GetThingsQuery, GetThingsQueryVariables>
> &
  TChildProps;
```
Why is `ReactApollo.DataProps` wrapped in `Partial`? Looking at [`withQuery`](https://github.com/apollographql/react-apollo/blob/9e4b573a5ae8ee068bc6f9229e6d84b36974da6b/packages/hoc/src/query-hoc.tsx#L20), which is what `withGetThings` wraps, `WrappedComponent` should expect `TProps & TChildProps`, where `TChildProps = DataProps<…>`. Is `data` ever _not passed in_? Is it marked as optional simply to avoid _requiring_ that the `WrappedComponent` expect `DataProps`? If so, is there a way to … not do that? The alternative is that `const { data: { loading } } = this.props` blows up because `data` is marked as optional.

Is `data` ever actually undefined? It seems like it should never be undefined.

dotansimha:
@ardatan can you please take a look?

dotansimha:
Fixed in: https://github.com/dotansimha/graphql-code-generator/pull/2297

dotansimha:
Fixed in 1.5.0 🎉

valerybugakov:
@dotansimha it seems that there might be a case when `data` will be missing. When developer specifies `config.props` it might not pass `data` down. But it will still be required because of `QueryNameProps<TChildProps>`.

Is there a way to tell TS about that using current typings?

dotansimha:
@valerybugakov Can you please explain? 
With the latest Apollo-Client, `data` might not be set (in past versions, it was `{}`, but now it could be `undefined`).

ardatan:
@valerybugakov @dotansimha We directly reference to React Apollo's typings with some generics. So we don't generate all the signature for HOCs, hooks or components. If there is a problem in Apollo's typings, I think an issue should be opened in their repo.

valerybugakov:
If developer renames `data` prop with it's own alias, for example: `patientData` then, `data` prop will not exist in the child component. Code generator output for HOCs: 

```ts
// Generated output
export type TPatientProps<TChildProps = {}> = ApolloReactHoc.DataProps<
[...]

PR #3744, opened 2020-03-29
Additional type argument to allow custom config.name for react-apollo HOCs

valerybugakov:
Fixes issue described in https://github.com/dotansimha/graphql-code-generator/issues/2192


valerybugakov:
If somebody can suggest hot to reproduce CI issue locally I will be able to fix it. Strangely `yarn build-and-test` executes without errors locally:
Screenshot 2020-03-29 at 15 46 51

ardatan:
As you can see in the example, CI takes `dev-test` diff so you need to run `yarn generate:examples` and commit :)

valerybugakov:
@ardatan thanks for a quick response, committed examples 👍

dotansimha:
Thank you @valerybugakov !

Issue #3475, opened 2020-02-11
Only export enum type


TomasBarry:
**Is your feature request related to a problem? Please describe.**

The flexibility of GraphQL queries severely limits the use of exported types in the generated file of Flow types.

Consider the following schema:

```es6
/* @flow */

/** All built-in and custom scalars, mapped to their actual values */
export type Scalars = {|
  ID: string,
  String: string,
  Boolean: boolean,
  Int: number,
  Float: number,
  ISO8601Date: any,
  ISO8601DateTime: any,
|};

export const MyEnumValues = Object.freeze({
  A: 'a', 
  B: 'b', 
  C: 'c'
});

export type MyEnum = $Values<typeof MyEnumValues>;

export type MyType = {|
  id: $ElementType<Scalars, 'ID'>,
  field: MyEnum,
|};
```

It would be fragile to type the return value of a GraphQL query with the type of `MyType` as Flow would not be able to assist us if we changed the query from:

```graphql
query {
  myType {
    id,
    field
  }
}
```

To:

```graphql
query {
  myType {
    field
  }
}
```

**Describe the solution you'd like**

We would like a mechanism to just generate types for enums and skip over GraphQL types.

**Describe alternatives you've considered**

The alternative is manually defining and maintaining enums in Flow or limiting ourselves to just importing the enums. However, if the types can be imported it is not reasonable for us to be able to police every commit from using them.

dotansimha:
@TomasBarry 
The basic `flow` plugin only generates the base types, based on your GraphQL Schema. 
If you wish to get type safety based on your queries, you should use `flow-operations` as well, and use `documents:` section to specify your GraphQL operations. 
It will generate Flow types based on your selection set, and if you change your query if will generate new types according to the fields you are asking for. 

Does that covers you use-case? I don't understand why generating only enums is related?

dotansimha:
Closing. @TomasBarry let me know if it's still relevant.

darkbasic:
@dotansimha the feature could make sense for another use case. Let's say that you want to use a `.d.ts` declaration file instead of a `.ts` one, but you also **don't** want to use `enumsAsTypes`: in that situation it could be useful to generate an additional `.ts` file with only the enums, so you can import them through your application while maintaining your declaration file.

dotansimha:
[...]

PR #3760, opened 2020-03-31
New option `onlyOperationTypes` for typescript and flow

standy:
This is a new config option `onlyOperationTypes` intended to generate only types for operations (enums, scalars, etc..., https://github.com/dotansimha/graphql-code-generator/issues/3475#issuecomment-597020849 may be related)

Non-operation types are incorrect to use for the client-side. You can't get a value of the type if you don't have an operation for that. But since non-operation types are not marked, my coworkers accidentally using them all the time. That option will get rid of that

When using `preResolveTypes` you probably don't need all base types, so now you can skip them and significantly reduce the size of types file. For example, in `dev-test/star-wars` 384 lines dropped to 183 with `preResolveTypes` + `onlyOperationTypes`
[dev-test/star-wars/types.ts](https://github.com/dotansimha/graphql-code-generator/blob/167fad8fe27b3e24fbad526ef85a5f0b37a57bca/dev-test/star-wars/types.ts)
[dev-test/star-wars/types.preResolveTypes.onlyOperationTypes.ts](https://github.com/dotansimha/graphql-code-generator/blob/167fad8fe27b3e24fbad526ef85a5f0b37a57bca/dev-test/star-wars/types.preResolveTypes.onlyOperationTypes.ts)

I think probably `onlyOperationTypes` + `preResolveTypes` should be default for client-side types on the next major release. It makes types smaller and avoids incorrect usage.

### DIFF
--- a/dev-test/codegen.yml
+++ b/dev-test/codegen.yml
@@ -85,6 +85,15 @@ generates:
     plugins:
       - typescript
       - typescript-operations
+  ./dev-test/githunt/types.preResolveTypes.onlyOperationTypes.ts:
+    schema: ./dev-test/githunt/schema.json
+    documents: ./dev-test/githunt/**/*.graphql
+    config:
+      preResolveTypes: true
+      onlyOperationTypes: true
+    plugins:
+      - typescript
+      - typescript-operations
   ./dev-test/githunt/types.flatten.preResolveTypes.ts:
     schema: ./dev-test/githunt/schema.json
     documents: ./dev-test/githunt/**/*.graphql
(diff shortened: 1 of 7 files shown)


standy:
@dotansimha please could you take a look on this MR.
I really want this feature for all my projects. Any advice and suggestions will be greatly appreciated

dotansimha:
@standy Sorry for the delay, I will look into it soon :)

dotansimha:
This is awesome, thank you so much @standy . 
I agree with you, in the next version we are probably change the default behaviour to that.

xiaoyu-tamu:
@standy Looks like input type are missing?

https://github.com/dotansimha/graphql-code-generator/blob/167fad8fe27b3e24fbad526ef85a5f0b37a57bca/dev-test/star-wars/types.preResolveTypes.onlyOperationTypes.ts#L31

havenchyk:
amazing guys, thank you so much!

PR #3764, opened 2020-03-31
Simplify generate examples script

standy:
This PR moves script for generating examples from shell to npm scripts. 
It's not so easy to run a shell script on windows, but since it is a one-liner, can be safely moved into `package.json`

This will help people with no shell to contribute to the project


ardatan:
Thank you @standy !

PR #3771, opened 2020-04-02
Improved near-operation-file fragment imports

wassim-k:
The current implementation for `near-operation-file` fragment imports was only intended as a workaround to solve issues with unused imports in `typescript-operations` plugin.

This PR is the improved implementation which makes it possible for 3rd party plugins to generate fragment imports correctly.


dotansimha:
This is PERFECT! Thank you so much @wassim-k !

PR #3851, opened 2020-04-10
Run init command by yarn

ppsirius:
We should run 'graphql-codegen init' command by yarn or npx, otherwise, we get an error like this:
"command not found: graphql-codegen"


dotansimha:
Thanks @ppsirius !

Issue #3882, opened 2020-04-17
Field wrapper doesn't wrap types in array


alex-knyazev:
When I set fieldWrapper I expect that Types in Arrays will be wrapped too

**To Reproduce**

My `codegen.yml` config file:

```yml
    config:
      skipTypename: true
      wrapFieldDefinitions: true
      fieldWrapperValue: 'T | { valueOf(): T }'
      maybeValue: T | undefined
      typesPrefix: API_
      namingConvention:
        typeNames: pascal-case#pascalCase
        enumValues: camel-case#camelCase
```

My GraphQL type:

```javascript

const CustomersPayload = new GraphQLObjectType({
  name: 'CustomersPayload',
  fields: () => ({
    edges: {
      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(Customer))),
    },
  }),
})
```

**Generated type**
```
export type API_CustomersPayload = {
  edges: FieldWrapper<Array<API_Customer>>
}
```

**Expected behavior**
Should generate 
```
export type API_CustomersPayload = {
  edges: Array<FieldWrapper<API_Customer>>
}
```
**Environment:**

- `@graphql-codegen/packages/typescript/typescript`: 
- NodeJS:

dotansimha:
Fixed in v1.13.4

ibratoev:
It seems the fix for this scenario broke another. I am using the following fieldWrapperValue: `T | Promise<T> | (() => Promise<T>)` and I get the error `Type '() => Promise< API_Customer[]>' is missing the following properties from type 'FieldWrapper< API_Customer >[]`.

ibratoev:
Maybe we need additional options how to wrap arrays...

raysuelzer:
@ibratoev  I think that might be the a good solution, a similar issue bug was introduced here relating to the only the base type of the Array being used when wrapping .   #5453

kgoggin:
I landed here with the same issue as @ibratoev - I'd like to be able to define a resolver as a function so that I can lazy fetch an array of things, but with the change above that doesn't work anymore.

Is there any reason why the `FieldWrapper` type can't be applied to _both_ the array and the inner type? I'm not a TS expert but as far as I can tell there wouldn't be an issue with the above example turning out like:

```
export type API_CustomersPayload = {
  edges: FieldWrapper<Array<FieldWrapper<API_Customer>>>
}
```

[...]

PR #3883, opened 2020-04-17
Add field wrapper for list type item [plugins/typescript]

alex-knyazev:
Related to issue:  https://github.com/dotansimha/graphql-code-generator/issues/3882

Add wrapping in field wrapper in list item element

### DIFF
--- a/packages/plugins/typescript/typescript/src/visitor.ts
+++ b/packages/plugins/typescript/typescript/src/visitor.ts
@@ -120,8 +120,22 @@ export class TsVisitor<
   }
 
   FieldDefinition(node: FieldDefinitionNode, key?: number | string, parent?: any): string {
-    const typeString = this.config.wrapFieldDefinitions ? `FieldWrapper<${node.type}>` : ((node.type as any) as string);
-    const originalFieldNode = parent[key] as FieldDefinitionNode;
+    let typeString;
+    const originalFieldNode = parent[key];
+
+    if (!this.config.wrapFieldDefinitions) {
+      typeString = node.type;
+    } else {
+      // if field is a list - we should add field resolver in list item
+      if (node.type.toString().startsWith('Array')) {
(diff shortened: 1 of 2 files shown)


dotansimha:
Thank you @alex-knyazev ! 
I think these changes are good, but it seems like it handles `string` a bit too much. I think maybe we can improve that and solve that before having this `Maybe...` string that we need to process.

alex-knyazev:
> Thank you @alex-knyazev !
> I think these changes are good, but it seems like it handles `string` a bit too much. I think maybe we can improve that and solve that before having this `Maybe...` string that we need to process.

alex-knyazev:
> Thank you @alex-knyazev !
> I think these changes are good, but it seems like it handles `string` a bit too much. I think maybe we can improve that and solve that before having this `Maybe...` string that we need to process.

I tried to do it more beautiful , but I always added this logic to inputs. Can you suggest where to see?

dotansimha:
@alex-knyazev I think this should be moved into `base-types-visitor.ts` and then it will be more compatible. I'm trying to move it there, into `FieldDefinition` or `NamedType` visitor functions, and see if it works. I'll update very soon ;)

dotansimha:
Thank you @alex-knyazev :) I used your tests and applied a fix for it here: https://github.com/dotansimha/graphql-code-generator/pull/3938

BTW, I'm pretty sure that we need to use `Maybe<Wrapper<Type>>` and not `Wrapper<Maybe<Type>>`, what do you think?

alex-knyazev:
> Thank you @alex-knyazev :) I used your tests and applied a fix for it here: #3938
> 
> BTW, I'm pretty sure that we need to use `Maybe<Wrapper<Type>>` and not `Wrapper<Maybe<Type>>`, what do you think?

I think yes, you right

PR #3886, opened 2020-04-17
feat: support @export directive

### LINKED ISSUE #2771: Remove variables from types when they are exported by the @export directive

Imagine you have a query:

```graphql
query getCurrentUserPosts($userId: ID!, $topic: String!) {
  currentUser @client {
    id @export(as: "userId")
  }
  getPostsByUser(id: $userId, topic: $topic) {
    id
    title
    text
  }
}
```

As of right now, if you run codegen on this, it would generate queries that would still require both `topic` and `userId` as variables. Even though we export `userId` by querying it from our client state.

Given the query above, we could strip out `userId` as a variable based on the occurence of their `@export` directive

andreialecu:
Ensures that variables exported via the `@export` directive are not part of the generated typescript variables types.

Closes #2771

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-documents-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-documents-visitor.ts
@@ -194,7 +194,8 @@ export class BaseDocumentsVisitor<
 
     const selectionSet = this._selectionSetToObject.createNext(operationRootType, node.selectionSet);
     const visitedOperationVariables = this._variablesTransfomer.transform<VariableDefinitionNode>(
-      node.variableDefinitions
+      node.variableDefinitions,
+      node.selectionSet
     );
     const operationType: string = pascalCase(node.operation);
     const operationTypeSuffix = this.getOperationSuffix(name, operationType);
--- a/packages/plugins/other/visitor-plugin-common/src/variables-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/variables-to-object.ts
@@ -1,4 +1,4 @@
-import { Kind, TypeNode, VariableNode, NameNode, ValueNode } from 'graphql';
+import { Kind, TypeNode, VariableNode, NameNode, ValueNode, SelectionSetNode, SelectionNode } from 'graphql';
 import { indent, getBaseTypeNode } from './utils';
 import { NormalizedScalarsMap, ConvertNameFn, ParsedEnumValuesMap } from './types';
 import { BaseVisitorConvertOptions } from './base-visitor';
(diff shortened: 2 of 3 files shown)


ardatan:
I think this is Apollo Client specific behavior so it would be better to have an option like `excludeExportedTypes` or something like that.

andreialecu:
Could do that. Does the CI put the build somewhere? I see a Publish Canary step.

Was wondering what I could use to be able to use the updated package without publishing the built files myself.

andreialecu:
@ardatan fwiw, the behavior only applies if there is a directive that matches exactly the name `export` and also has an `as` value that matches one of the operation variables exactly. It's unlikely this would catch anything it isn't supposed to.

ardatan:
@andreialecu I see but this is still something Apollo specific :) We don't prefer to have specific behaviors  for generic plugins. So having it under an option would be better. Canary publishing doesn't work in this PR because the branch is under your fork repository. After we merge this PR, we will have a canary version on master.

andreialecu:
I understand, feel free to take over and improve the code as you see fit. `Allow edits from maintainers` is enabled.

It was a quick and dirty implementation to get me unblocked.

andreialecu:
There's a small problem that happens when all `NON_NULL_TYPE` variables are deleted by the functionality of this PR.

`isVariablesRequired` will be `true` here even though no variables are required:
https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/react-apollo/src/visitor.ts#L201-L216

So react components will complain that `skip` is a required property. `<SomeQueryComponent skip={false}...` gets around it.

I'm not sure what the most elegant way to deal with it would be.

dotansimha:
Thank you @andreialecu . 
As @ardatan said, this behaviour is specific to Apollo, while codegen plugins are independent and not specific to any framework. 

We are thinking on implementing support for schema transformations (see https://github.com/dotansimha/graphql-code-generator/issues/3939)

So this PR might turn into something like: 
```
transforms:
  - strip-apollo-export-directive
```

I'm closing this, and keeping the issue open. I know it's not ideal, but we are really trying to keep the codegen agnostic, and make sure it's goes by the spec of GraphQL.

ToledoNicola:
hi, is it possible to have a detailed example of how to create a custom schema loader to support the @export directive?
thank you
this is my query
`query Dashboard() {
  currentUserId @client @export(as: "userId")
  getPatient(id: $userId) {
    id
    name
    doctor {
      id
      name
      surname
      available
    }
    nextSessions {
      id
      day
      hour
      paid
      dayOfWeek
    }
  }
}
`
and this is the error
`6:33:39] → Syntax Error: Expected "$", found ")".
[16:33:39] Generate libs/data-access/src/lib/generated/generated.ts [failed]
[16:33:39] → Syntax Error: Expected "$", found ")".`

PR #3923, opened 2020-04-24
Preset for GraphQL-Modules (v1) integration

kamilkisiela:
This PR introduces a new preset for creating encapsulated types and resolvers signature for the new upcoming version of `graphql-modules` (v1).


dotansimha:
@kamilkisiela We should also support `resolversComposition` and generate typings for the possible keys of it per each module (with wildcard support).

LucienLee:
Will graphql module preset support federation mode?

dotansimha (changes requested):
Changes required for merging this PR:
- Please add support for `encapsulateModuleTypes: "prefix" | "namespace" | "none"` (default = `prefix`). This will make sure that generated files  
- @kamilkisiela this preset introduces a new concept for the baseTypes, because it generates it by itself, and not just expecting it to exists. I really like that. But - we need to make sure we can pass additional plugins  to it. I tried to add `add` plugin for `eslint-disable` directive but it only applies it to the module types. So we need to make sure to fix that. 
- I noticed some changes in the API of `graphql-modules` v1 beta that, like `Middleware` - I fixed it but not sure if there are more issues related to the recent changes there.

dotansimha:
@kamilkisiela @ardatan added support for encapsulation (default: `none`). Personally - I think `namespace` is the nicest way to have it. 
Also fixed a bug I found with external scalars.

dotansimha (approved):

dotansimha:
@kamilkisiela @ardatan Noticed some issues with scalars as well. It was assuming `export type MyScalar = Types.ScalarName`  would work, and it doesn't, since scalars are located under `Scalars` type (so it should be `Types.Scalars[ScalarName]`. 
I did a workaround by exporting scalars as regular types.

PR #3926, opened 2020-04-25
Fix optional types in queries [flow]

rizkyario:
Related https://github.com/dotansimha/graphql-code-generator/pull/3639
Fix for https://github.com/dotansimha/graphql-code-generator/pull/3639#issuecomment-619090429

This PR makes all nullable fields to be optional in flow.
Currently there is an inconsistency between the top-level types and the query types.

example: https://github.com/dotansimha/graphql-code-generator/blob/master/dev-test/githunt/flow.flow.js

```js
// top-level types
export type Repository = {|
  __typename?: 'Repository',
  name: $ElementType<Scalars, 'String'>,
  full_name: $ElementType<Scalars, 'String'>,
  // description is optional field
  description?: ?$ElementType<Scalars, 'String'>,
  html_url: $ElementType<Scalars, 'String'>,
  stargazers_count: $ElementType<Scalars, 'Int'>,
  open_issues_count?: ?$ElementType<Scalars, 'Int'>,
  owner?: ?User,
|};
// query types
export type CommentQuery = {
        repository: {
          ...{ __typename?: 'Repository' },
          ...$Pick<
            Repository,
           // description is mandatory field now
            {| description: *, open_issues_count: *, stargazers_count: *, full_name: *, html_url: * |}
          >,
        },
      |},
    },
  |},
};
```

### DIFF
--- a/dev-test/githunt/flow.flow.js
+++ b/dev-test/githunt/flow.flow.js
@@ -176,7 +176,7 @@ export type OnCommentAddedSubscriptionVariables = {
 export type OnCommentAddedSubscription = {
   ...{ __typename?: 'Subscription' },
   ...{|
-    commentAdded: ?{
+    commentAdded?: ?{
       ...{ __typename?: 'Comment' },
       ...$Pick<Comment, {| id: *, createdAt: *, content: * |}>,
       ...{|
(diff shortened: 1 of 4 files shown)


dotansimha:
Thank you @rizkyario ! 
Am I missing something here? I saw that the optional mark you added is after the field name (`commentAdded?: ?{`), but just `commentAdded: ?{` is not enough? 
From the Flow docs, this is how you mark field as optional: https://flow.org/en/docs/types/maybe/

rizkyario:
optional field mark should be put after the field name indeed, here is the Flow docs: https://flow.org/en/docs/types/objects/#toc-optional-object-type-properties

it's consistent with the top-level types where it's  `description?: ?$ElementType<Scalars, 'String'>, `

dotansimha:
Got it :) Thank you @rizkyario !

Issue #3929, opened 2020-04-26
Java scalar default to 'any' instead of 'Object'


ducthienbui97:
**Describe the bug**
 

**To Reproduce**

<!--
#############
#############
PLEASE READ THE FOLLOWING:
You can use this template to create a live sandbox of the issue: 
https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template
Make sure sure fork this template, and run `yarn generate` in the terminal. 
Please make sure the Codegen and plugins version under `package.json` matches yours. 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~ ISSUES WITH A LIVE REPRODUCTION WILL BE ANSWERED QUICKLY ~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#############
#############
-->

1. My GraphQL schema:

```graphql
# Put your schema here
scalar Date

schema {
  query: Query
}

type Query {
  me: User!
  user(id: ID!): User
  allUsers: [User]
  search(term: String!): [SearchResult!]!
  myChats: [Chat!]!
}

enum Role {
  USER,
  ADMIN,
}

interface Node {
  id: ID!
}

union SearchResult = User | Chat | ChatMessage

type User implements Node {
  id: ID!
  username: String!
  email: String!
  role: Role!
}

type Chat implements Node {
  id: ID!
  users: [User!]!
  messages: [ChatMessage!]!
}

type ChatMessage implements Node {
  id: ID!
  content: String!
  time: Date!
  user: User!
}

```

2. My GraphQL operations:

```graphql
# Put your operations here
```

3. My `codegen.yml` config file:

```yml
# Put your YML here
generates:
  src/main/java/my/app/generated/Resolvers.java:
    plugins:
      - java-resolvers
```

**Expected behavior**
[...]

PR #3930, opened 2020-04-26
Fix: set java default class for scalar to Object

ducthienbui97:
related #3929 
For scalars that have not been configured with class mapping, the default behavior is to set it to `any` which works for typescript but in java default type should be `Object`.

I have updated the function `buildScalars` to accept another parameter : defaultScalarType. It's set to `any` by default, but for java, it's set to `Object`. It's likely that other package also need to be updated.


dotansimha:
Perfect! Thank you so much @ducthienbui97 !

PR #3932, opened 2020-04-26
docs(README): fix init command

glenwinters:
The CLI is installed with `yarn add -D @graphql-codegen/cli`. Since it's not global, `graphql-codegen init` will fail with a "command not found" error. A few lines down, it's suggested to not install it globally, so we need a different solution.

I added [npx](https://www.npmjs.com/package/npx) in front since that will execute the CLI in `node_modules/.bin`.


dotansimha:
Thanks @glenwinters ! 
I'm working on improving it now as well.

Issue #3899, opened 2020-04-20
[question] generate typeDefs


guiaramos:
Hey guys, thanks for your amazing project.

I would like to know if there is any plugin for that generates typeDefs like:

```js
export const typeDefs = gql`
  extend type Query {
    isLoggedIn: Boolean!
    cartItems: [ID!]!
  }

  extend type Launch {
    isInCart: Boolean!
  }

  extend type Mutation {
    addOrRemoveFromCart(id: ID!): [ID!]!
  }
`;
```

I am using Apollo Local state management, therefore I want to generate the typeDefs automatically.

Thanks!

[ref document](https://www.apollographql.com/docs/tutorial/local-state/)

guiaramos:
One solution is to use: https://github.com/ardatan/graphql-import

marceloverdijk:
I had basically the same question.
My codegen yml looks like:
```
overwrite: true
schema: "./graphql/schema.graphql"
..
```

It would be useful to generate the typeDefs automatically based on this.

marceloverdijk:
I used a custom plugin like:

`typescript-typedefs.js:`

```
const { printSchema } = require('graphql');

module.exports = {
    plugin: (schema, documents, config) => {
        return [
            'import gql from "graphql-tag";',
            '',
            'export const typeDefs = gql`',
            printSchema(schema),
            '`;',
            '',
        ].join('\n');
    },
};
```

and then added the custom plugin in my codegen.yml:

```
overwrite: true
schema: "./graphqls/*.graphqls"
documents: null
generates:
  src/generated/graphql.ts:
    plugins:
      - "typescript"
      - "typescript-resolvers"
      - "typescript-typedefs.js" 

```

Maybe useful to add something like this to `graphql-code-generator`  so other people can use out of the box?

dotansimha:
Closing, as @marceloverdijk suggested a quick plugin implementation if someone needs that.

igrek8:
Make sure to use `printSchemaWithDirectives` ⚠️ from `@graphql-toolkit/common`. Otherwise, the final schema will not have directives which may result in problems. For example, no `@auth` directives will be applied.

ardatan:
@igrek8 That's right but `graphql-toolkit` has been deprecated in favor of `graphql-tools`. So `@graphql-tools/utils` has the same function :)

shepherdjerred:
[...]

PR #3941, opened 2020-04-28
PR for #3899 providing a TypeScript TypeDefs plugin

marceloverdijk:
Hi @dotansimha,

Here is a PR for the Typedefs plugin.
I created it as a draft for now as the test is failing and I'm stuck.

The test is created from https://graphql-code-generator.com/docs/custom-codegen/contributing/ and looks similar as other tests. 
But for some reason `expect(result.content).toBeSimilarStringTo(..);` in the test is complaining `TypeError: expect(...).toBeSimilarStringTo is not a function`. I just don't see what I'm doing wrong unfortunately.

### DIFF
--- a/packages/plugins/typescript/typedefs/jest.config.js
+++ b/packages/plugins/typescript/typedefs/jest.config.js
@@ -1,0 +1,1 @@
+module.exports = require('../../../../jest.project')({	dirname: __dirname });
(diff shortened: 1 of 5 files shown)


ardatan:
We already have a plugin that prints the schema in SDL so we can improve it that outputs a code file with a typeDefs export if file extension is TS or JS.
For example, Fragment Matcher plugin works in similar way;
https://graphql-code-generator.com/docs/plugins/fragment-matcher#configuration

marceloverdijk:
OK if there is an easier way to integrate in a existing plugin that would be nice as well.
Then this one can be closed.

Issue #3484, opened 2020-02-12
Do we have way to generate code for the Type which is the response object


kvsravindrareddy:
I have the below schema 
Step 1: 

```graphql
type Query {
  bookById(id: ID): Book 
}

type Book {
  id: ID
  name: String
  pageCount: Int
  author: Author
}

type Author {
  id: ID
  firstName: String
  lastName: String
}
```

Step 2: Below code generated with the code generator.

```java
package my.app.generated;

import java.util.Map;

public class Types {
  
  
  public static class QueryBookByIdArgs {
    private Object _id;
  
    public QueryBookByIdArgs(Map<String, Object> args) {
      if (args != null) {
        this._id = (Object) args.get("id");
      }
    }
  
    public Object getId() { return this._id; }
  }
}
```
Step 3: Is there any way generate code for the Book and Author type as well?

Thanks in advance.

dotansimha:
@kvsravindrareddy Can you please share the complete output, codegen version, codegen config and what do you think we should generate? 

Thanks!

dotansimha:
@kvsravindrareddy any update?

dotansimha:
Closing. @kvsravindrareddy feel free to comment if it's still relevant.

kliakos:
I actually had the same question. I would like to be able to generate classes for all the types. In this example, I would expect the below classes:

```java
  public static class Author {
    private String ID;
    private String firstName;
    private String lastName;

    public String getID() {
      return ID;
    }

    public void setID(final String ID) {
      this.ID = ID;
    }

    public String getFirstName() {
      return firstName;
    }

    public void setFirstName(final String firstName) {
      this.firstName = firstName;
    }

    public String getLastName() {
      return lastName;
    }

    public void setLastName(final String lastName) {
      this.lastName = lastName;
    }
  }
  
  public static class Book {
    private String ID;
    private String name;
    private Integer pageCount;
    private Author author;

    public String getID() {
      return ID;
    }

    public void setID(final String ID) {
      this.ID = ID;
    }

    public String getName() {
      return name;
    }

    public void setName(final String name) {
      this.name = name;
    }

    public Integer getPageCount() {
      return pageCount;
    }

    public void setPageCount(final Integer pageCount) {
      this.pageCount = pageCount;
    }

    public Author getAuthor() {
      return author;
    }

    public void setAuthor(final Author author) {
      this.author = author;
    }
  }
```

PR #3993, opened 2020-05-09
feat: add generation of kotlin classes for types

ShyykoSerhiy:
Added generation of Kotlin data classes for object types. Probably connected with the https://github.com/dotansimha/graphql-code-generator/issues/3484, but this is only for Kotlin and not Java.


dotansimha:
Nice!! Thank you @ShyykoSerhiy !

kliakos:
Is it possible somehow to support `interface` as well? Currently, the fields that return an interface are set as `Any`.

Urigo:
@kliakos could you create a new issue about your feature request?
Best if you could even add a test demonstrating this
Thanks!

Issue #3275, opened 2020-01-10
Plugin typescript does not export a valid JS object with "plugin" function.


ashubham:
**Describe the bug**
 
When I try to follow the getting started from the docs, I get the error:
`Plugin typescript does not export a valid JS object with "plugin" function.`

**To Reproduce**
1. My codegen.yml
```
overwrite: true
schema: "./schema.graphql"
documents: null
generates:
  ./types.ts:
    plugins:
      - "typescript"
```
1. My GraphQL schema:

```graphql
type Query {
  resultSet(searchText: String!): ResultSet
  board(id: ID!): Board
}

type Mutation {
  createBoard: Board
  addPhotosToBoard(input: AddPhotosToBoardInput!): AddPhotosToBoardPayload
  updateBoardLayout(input: UpdateBoardLayoutInput!): UpdateBoardLayoutPayload
}

type ResultSet {
  photos: [Photo]!
}

type Photo {
  id: ID!
  owner: String!
  secret: String!
  server: Int!
  farm: Int!
  title: String!
  ispublic: Int!
  isfriend: Int!
  isfamily: Int!
}

enum TileSize {
  SMALL
  MEDIUM
  LARGE
}

type Tile {
  photo: Photo!
  size: TileSize!
}

type Board {
  id: ID!
  tiles: [Tile!]!
}

input AddPhotosToBoardInput {
  id: ID!
  photos: [PhotoInput!]!
}

input PhotoInput {
  id: ID!
  owner: String!
  secret: String!
  server: Int!
  farm: Int!
  title: String!
}

type AddPhotosToBoardPayload {
  board: Board
}

input UpdateBoardLayoutInput {
  id: ID!
  tileChanges: [TileChange!]!
}

input TileChange {
  id: ID!
  size: TileSize
}

type UpdateBoardLayoutPayload {
  board: Board
}
```

**Expected behavior**

Types should be generated.

**Environment:**

- OS: Mac OSX 
- `@graphql-codegen/...`: 1.11.2
- NodeJS: 10.11.0

This is related to #1703 which was closed without a solution.

dotansimha:
@ashubham can you please provide a reproduction in codesandbox/repo? We couldn't reproduce this issue. 

Please make sure you have `@graphql-codegen/typescript` installed.

doronnac:
Had the same issue, this was because the quick start section did not mention `@graphql-codegen/typescript` installation. I can contribute to the docs if you'd like.

[...]

PR #3996, opened 2020-05-09
In quickstart we need typescript plugin setup

worldsayshi:
Clarifying the quickstart a tiny bit: https://github.com/dotansimha/graphql-code-generator/issues/3275


ardatan:
@worldsayshi GraphQL Codegen is not only for TypeScript. We have plugins for Java, C# etc.

PR #3999, opened 2020-05-10
Fix #3998: Fix nested input suffix reduplication

### LINKED ISSUE #3998: Remove implicit adding of `Input` suffix
https://github.com/dotansimha/graphql-code-generator/blob/f7d17324e7b24efbf46815139964ae042dd34657/packages/plugins/java/kotlin/src/visitor.ts#L125

Hi, I am developing a Java / Kotlin solution. And my problem is that I can't use `Input` endings in my schema for the `input` types because `codegen` utility translates `input` type names into `SomeNameInputInput`. 
So, generally speaking, it duplicates the `Input` ending. 

Can you make this optional? Or add some logic that checks "if type ends with `Input` word, then do not add additional `Input` word"?

Steps to reproduce:
1) Open `https://graphql-code-generator.com/`
2) Select `Java --> Types (Enum & Input)` in dropdown
3) Put this in the textarea on the left side

```graphql
input CreateUserInput {
   userName: CreateUserNameInput
}

input CreateUserNameInput {
  given: String!
  family: String!
}
```
4) Find `CreateUserNameInputInput` in the right side textarea

meiblorn:
Fix for the #3998 

BREAKING: 
- I've changed `ID` default scalar's type for the `Kotlin` language from `String` to `Any` — it was done to achieve consistency with the `Java` language: it uses `Object` type (not `java.lang.String` as it was in `Kotlin`)

Except this breaking change above, I hadn't change previously written tests, so there should not be any regression.

### DIFF
--- a/packages/plugins/java/java/src/visitor.ts
+++ b/packages/plugins/java/java/src/visitor.ts
@@ -62,7 +62,7 @@ export class JavaResolversVisitor extends BaseVisitor<JavaResolversPluginRawConf
       allImports.push(`java.util.stream.Collectors`);
     }
 
-    return allImports.map((i) => `import ${i};`).join('\n') + '\n';
+    return allImports.map(i => `import ${i};`).join('\n') + '\n';
   }
 
   public wrapWithClass(content: string): string {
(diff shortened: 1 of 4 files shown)


dotansimha:
Great catch. Thank you @meiblorn :)

PR #4007, opened 2020-05-12
Fix __typename indentation if immutableTypes is false

marcoreni:
if `config.immutableTypes` is false generated code contains an extra space before __typename:

Expected result:
```ts
export type MyType = {
  __typename?: 'Entry';
  /** Information about the repository from GitHub */
  repository: Repository;
}
```

Actual result:
```ts
export type MyType = {
   __typename?: 'Entry';
  /** Information about the repository from GitHub */
  repository: Repository;
}

```

This PR removes the extra space.


dotansimha:
Awesome! thank you @marcoreni :)

Issue #3486, opened 2020-02-12
Typescript interface property definitions end in commas instead of semicolons


Tackgnol:
It seems like the typescript interfaces are generated in a wrong way, they have commas after property declaration, I would expect semicolons as per TS specification [see here](https://www.typescriptlang.org/docs/handbook/interfaces.html)

1. My GraphQL schema:
Can't share :/

2. My GraphQL operations:
Can't share :/

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: ./graphQL/*.graphql
generates:
  types.ts:
    plugins:
      - typescript
config:
  typesPrefix: I
  declarationKind: 'interface'
```

**Expected behavior**
Interfaces in typescript are defined as: 
`[property name]: type;` not `,`
https://www.typescriptlang.org/docs/handbook/interfaces.html
Am I missing a setting somewhere? Can't find it in the documentation... 

**Environment:**

- OS: 
Windows 10
- `@graphql-codegen/...`: 
Can't share
- NodeJS: 
10.14

ggomma:
Have same issue :(

dotansimha:
I think this was fixed in: https://github.com/dotansimha/graphql-code-generator/pull/3505

Tackgnol:
Awesome! Yeah seems so :) 
Do we have an ETA on the release?

marcoreni:
I think this is still an issue with ResolversType and ResolversParentType

eg:
```ts
/** Mapping between all available schema types and the resolvers parents */
export type ResolversParentTypes = ResolversObject<{
  Query: {},
  ID: Scalars['ID'],
  Context: Context,
  String: Scalars['String'],
  [...]
}>;
```

```ts
export type MyTypeResolvers<ContextType = TContext, ParentType extends ResolversParentTypes['MyType'] = ResolversParentTypes['MyType']> = ResolversObject<{
  id: Resolver<Maybe<ResolversTypes['ID']>, ParentType, ContextType>,
  [...]
}>;
```

dotansimha:
@marcoreni Not all plugins supports `declarationKind`. PRs are always welcome.

dotansimha:
Fixed in `v1.13.0` for `typescript` plugin

PR #4009, opened 2020-05-12
Fix punctuation in typescript-resolvers

marcoreni:
This is a followup to https://github.com/dotansimha/graphql-code-generator/issues/3486 for typescript-resolvers . 

I chose not to implement declarationKind because it would have required a heavier refactoring (because of the `useIndexSignature` option), but I made the plugin use the `getPunctuation` function so that semicolons are now correctly used.


ardatan:
Thank you @marcoreni !

PR #4012, opened 2020-05-12
Add useExplicitTyping to fragment matcher config

Andrew-Talley:
This is my first time contributing to this repo, so apologies in advance if I need to change something – please just let me know what needs changing, and I'll be happy to do it. 

I was using the fragment matcher for a project that didn't use Apollo Client at all. I was doing some custom logic on objects based on if it inherited from a particular interface. The fragment matcher was really useful – I was storing typenames, so I could just check `fragmentMatcher.possibleTypes.[INTERFACE].includes(object.__typename)`, and it worked like a charm.

But, there wasn't any typescript support, since the `possibleTypes` property had a generic `{ [key: string]: string[] }` type. So, this pull request add more explicit types. Specifically – it just defined the type by the content actually created. I tested it in my local repository, and it doesn't seem to produce any type errors, and it does suggest the names of each fragment in VSCode.

For now, I just add it is as a configuration option, leaving it off by default. If other people would find this useful, and if it wouldn't create any issues, it may be worth making this the default behavior to limit the number of options for a fairly simple plugin, and I'd be happy to make that change. But for now, I figured it was best to err on the side of leaving the current behavior in place.

### DIFF
--- a/packages/plugins/other/fragment-matcher/src/index.ts
+++ b/packages/plugins/other/fragment-matcher/src/index.ts
@@ -58,6 +58,23 @@ export interface FragmentMatcherConfig {
    * ```
    */
   apolloClientVersion?: 2 | 3;
+  /**
+   * @name useExplicitTyping
+   * @type boolean
+   * @description Create an explicit type based on your schema. This can help IDEs autofill your fragment matcher. This is mostly useful if you do more with your fragment matcher than just pass it to an Apollo-Client.
+   * @default false
+   *
+   * @example
+   * ```yml
+   * generates:
+   * path/to/file.ts:
+   *  plugins:
+   *    - fragment-matcher
+   *  config:
+   *    useExplicitTyping: true
+   * ```
+   */
+  useExplicitTyping?: boolean;
   federation?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


dotansimha:
Awesome. Thank you @Andrew-Talley . I tried to check if it's possible to extract the actual type from a statically defined `Record`, but it seems like it's not possible, so we can go with this solution.

Thanks!

CarsonF:
Wouldn't it be possible to define this via the `typeof` the result object. What was probably missing the previous attempt was the `as const` on the result, which changes how TS views the type of the object literal.
```tsx
const result = {
  possibleTypes: {
    ...
  },
} as const; // necessary to get correct type with typeof

export type PossibleTypesResultData = typeof result;
```

Andrew-Talley:
@CarsonF I think that's right – initially, I did exactly that. My one hesitation was that it then converts everything within `result` to `readonly`, which I worry could throw type errors if people tried to modify it (unlikely, perhaps, but possible), and that it might be incompatible with functions that expect non-readonly types. Since you still get the same type either way, I figured this was the simpler approach from a user perspective.

CarsonF:
Hmmm that's a good point. I wonder if this config option `useExplicitTyping` could be used in conjunction with `immutableTypes` to do what I've specified.

Issue #3949, opened 2020-04-29
Add operation export name for "typescript-operations"


aspirisen:
**Is your feature request related to a problem? Please describe.**
Currently in order to use `.gql` file you need to generate `.d.ts` file by `typescript-graphql-files-modules` but the module definitions are wildcarted and have nothing with real placement of the `gql` file. Also, you need types for the query and its variables by `typescript-operations` and usually they are placed along with `gql` file. It would be good to combine them.

**Describe the solution you'd like**
Add to `typescript-operations` in the emit the const export per each operation such as:
```export const GetClients: import("graphql").DocumentNode;```

Using preset `near-operation-file` you can put operation definitions along with its `gql` file. Then you need to set `extension` to be `".gql.d.ts"` by that you can import `gql` file in `ts` files.

In result you can import everything at once:
```
import {
  GetClients,
  GetClientsQuery,
  GetClientsQueryVariables,
} from "./GetClients.gql";
```

eddeee888:
This looks like it's done in https://github.com/dotansimha/graphql-code-generator/pull/4020
Closing for now, please let me know if this should stay open

PR #4020, opened 2020-05-14
Add operation export name

aspirisen:
This PR is relative to this issue https://github.com/dotansimha/graphql-code-generator/issues/3949

It allows you to add operation name to output file.

### DIFF
--- a/packages/plugins/typescript/operations/src/config.ts
+++ b/packages/plugins/typescript/operations/src/config.ts
@@ -89,4 +89,23 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
    */
   noExport?: boolean;
   globalNamespace?: boolean;
+    /**
+   * @name addOperationExport
+   * @type boolean
+   * @description Add const export of the operation name to output file. It will allow you to get everything with one import: ```import { GetClient, GetClientQuery, GetClientQueryVariables, } from "./GetClient.gql";```.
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @aspirisen ! 

This should work only with `.d.ts` output, right? it won't work with `.ts` extension? because some configuration flags might effect that as well.

Also, have you checked `graphql-let` (https://github.com/piglovesyou/graphql-let#readme) package? it should does that automatically as far as I know.

aspirisen:
@dotansimha  Yes, it would work only in `d.ts`files, but it is up to user because there is a flag for that. But if user uses only types this flag will help him to import everything from one file.

I checked `graphql-let` and there is one issue that stops me using it `https://github.com/piglovesyou/graphql-let/issues/60`

dotansimha:
If you need to have a more strict path in the module declaration, you can use this configuration: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/graphql-files-modules/src/index.ts#L6-L25 (it's not in the docs, we need to update it). 

I afraid that the name you set for this variable might conflict with other plugins, we need to check that.

aspirisen:
@dotansimha what variable do you mean, the const export in `d.ts`?

dotansimha:
I think we can get this one merged. Thank you @aspirisen , sorry it took so long!

PR #4022, opened 2020-05-15
Fix typescript-resolvers doc

anault:
tejfsHJ


aryzing:
Perhaps you should squash all 3 commits into a single one

ardatan:
Thank you @aryzing !

PR #4050, opened 2020-05-19
Add links to typescript-oclif plugin

ryanwersal:
I wasn't able to find the documentation through the website for the oclif plugin and realized that it didn't get included in the "All Plugins" page or the sidebar. This change adds it to both. I also made a couple tweaks to the oclif docs itself and noted that the page got rather heavily reformatted. It looks like the formatting aligns better with the other pages' formatting so I kept it for this PR.

Thanks for the project, I'm quite excited to have a foundation to build out some more generation plugins for and not have to maintain them by hand! 👍


dotansimha:
Thank you @ryanwersal 😍

PR #4070, opened 2020-05-23
fix(time): config and default message

bboure:
This PR fixes the following small issues:

- Change default message to "Generated on", which is better English
- when passing config as an object, `format` should not be required.


dotansimha:
That's perfect, thank you @bboure

PR #4122, opened 2020-05-28
Docs: Correct hook usage in typescript-urql

petetnt:
Instead of the `HOC`:ish name, a `use`-prefixed function is generated


ardatan:
Thanks @petetnt !

PR #4134, opened 2020-05-29
fix: don't write duplicate fragments to output

### LINKED ISSUE #3990: Duplicate TypeScript fragment imports with React Apollo
**Describe the bug**

It looks like it's the code generated by the `typescript-operations` plugin. I'm just using some fragments in some queries, like:

```typescript
import gql from "graphql-tag"
import { entryFieldsFragment, folderFieldsFragment, projectFieldsFragment } from "graphql/fragments"

const someMutation = gql`
  mutation Bla(
    $projectUuid: UUID!
    $folderUuid: UUID!
  ) {

  ...

  ${entryFieldsFragment}
  ${folderFieldsFragment}
`
```

I've not duplicate the imports in this file, but the fragments are used multiple times within this file in different queries. Interestingly though, it's duplicating an import for a fragment I'm only using once in the file.

**To Reproduce**

https://codesandbox.io/s/nifty-night-j4mhx

This one took some reproducing, but it does give a lot of clues about what's going on. In the above CodeSandbox I've used the same config we have locally and created a dummy schema. I've also mimicked the relevant part of our folder structure.

What I've observed there is that the generated files' imports seem to be somehow linked if the input filename is the same (e.g. `users.ts` in 2 different folders, each with GraphQL queries in - the output seems linked for both of those files with the same name). If I import a fragment in one, but not the other it imports the fragment in both generated files. The issue seems to come in when the imports aren't identical in both files (not in the CodeSandbox I've just changed the order of those imports and that's confused it into generating duplicate imports).

In my local exmaple, I have 2 shared fragment imports in both separate source files (like the user.ts files in the CodeSandbox) and 1 different fragment. The generated result looks like both concatenated (without de-duplication). It's also worth noting, the particular imports that are duplicated in my local environment are in the same order in both files, there are just additional fragment imports. I think either scenario leads to this duplicate import result.

It looks like in 1.13.2 the generated import format changed and the bug was introduced with that, seems like the `___Doc` imports were split out then.

1. My GraphQL schema:

Closed source, sorry!

2. My GraphQL operations:

Closed source, sorry!

```yml
overwrite: true
schema:
- "../management/schema.graphqls"
documents:
- "src/**/*.{ts,tsx}"
- "!src/**/*.generated.{ts,tsx}"
generates:
  src/types.ts:
    plugins:
    - typescript
    config:
      scalars:
        Date: string # TODO: Might not be the right type to use?
        UUID: string
  src/:
    preset: near-operation-file
    presetConfig:
      extension: .generated.ts
      baseTypesPath: types.ts
    plugins:
    - "typescript"
    - "typescript-operations"
    - "typescript-react-apollo"
    - "fragment-matcher"
    config:
      withComponent: false
      withHooks: true
      withHOC: false
      scalars:
        Date: string # TODO: Might not be the right type to use?
        UUID: string

```

**Expected behavior**

* No duplicate imports in the generated code.
* Generated source code using near-operation-file shouldn't be affected by non-nearby operations (i.e. if I import a fragment in one folder, it shouldn't show up in another folder's generated code).

**Environment:**

- OS: MacOS 10.15.4
- `@graphql-codegen/...`: v1.13.5
- NodeJS: v12.16.3 (lts/erbium)

smeijer:
I don't think there is much to explain, #3990 explains it pretty well. `typescript-operations` is writing fragments multiple times to the out file.

I've filtered the output string, and not the objects before it because that resulted in other errors.

This small fix, works quite nice.

---

related #3990

### DIFF
--- a/packages/plugins/typescript/operations/src/index.ts
+++ b/packages/plugins/typescript/operations/src/index.ts
@@ -32,7 +32,7 @@ export const plugin: PluginFunction<TypeScriptDocumentsPluginConfig, Types.Compl
     leave: visitor,
   });
 
-  let content = visitorResult.definitions.join('\n');
+  let content = Array.from(new Set(visitorResult.definitions)).join('\n');
 
   if (config.globalNamespace) {
     content = `


dotansimha:
Thanks @smeijer !

PR #4144, opened 2020-05-31
[WIP] Python support

seansfkelley:
Per https://github.com/dotansimha/graphql-code-generator/issues/3712#issuecomment-636436999, opening this as a PR for visibility and contributions. I don't have any time at the moment to continue authoring this, unfortunately. That means it will likely idle for quite a while, though I'm happy to pitch in with reviewing and testing improvements made by others. I would really love to have this feature! If the PR gets too stale for the maintainer's tastes, please feel free to close it.

The rest of this comment is adapted from https://github.com/dotansimha/graphql-code-generator/issues/3712#issuecomment-636403564.

In it's current state, this is a hodgepodge of half-baked ideas and me trying to understand the types of things I'm asking about in https://github.com/dotansimha/graphql-code-generator/issues/3712.

I did end up generating some promising types. The basic design I was going for was:

- Uses [Pydantic](https://pydantic-docs.helpmanual.io/). Love this library.
- The "global" type generator would only output scalars, enumerations and input types.
- The "operations" type generator would define purpose-built types according to the operation in question (this is different than the Typescript generator, which generates all types in the global space, and then uses `Pick`/`Partial`/whatever to select subsets of them for operations).

This made code that, when used with the `near-operations-file` preset, looked like (from memory here -- sorry if this doesn't match with what it does exactly):

```py
# Assuming a sibling file defines some query:
# query getStuff(input: MyInputType) {
#   fooBar { # of type FooBar!
#     baz # of type String
#   }
# }

from wherever.you.set.global.types.to import GraphqlBaseModel

class GetStuff(GraphqlBaseModel):
  class FooBar(GraphqlBaseModel):
    baz: Optional[str]
  
  foo_bar: FooBar
```

A couple things to notice about this that I liked:

- Though the selection set for any given type (e.g. `FooBar`) may vary in different queries, it always takes the same name when generated. This is because you get a nice behavior where the enclosing class acts as a namespace (like `GetStuff.FooBar`) which makes it abundantly clear both what type the selection set is on and that this class definition is only for this operation.
- Pydantic has "alias generators" that make it super easy to translate between camelCase GraphQL and snake_case Python so it's idiomatic in all usages. (That's what the `GraphqlBaseModel` import is doing here, along with a couple of other behaviors like disallowing extraneous fields and disallowing mutations.


ardatan:
This looks awesome! @seansfkelley 
How can we help you to make this ready to merge ? :)

seansfkelley:
I don't have any plans to try to get this merged; it was only intended for visibility and to demonstrate some approaches to design and implementation.

I'm going to close this to avoid any further confusion. It'll still serve its documentary purpose while closed.

PR #4158, opened 2020-06-03
Adding option for numeric typescript enums

### LINKED ISSUE #4157: [TypeScript plugin] Supporting enums generated with integer values instead of strings

Generating from schema enums uses the keys as values when it should use integers.

Given the following enum in a schema:
```graphql
enum Status {
  RUNNING
  IDLE
  ERROR
}
```
The generated typescript enum I would expect is
```ts
enum Status {
  RUNNING, // 0
  IDLE, // 1
  ERROR // 2
}
```
However the actual output is:

```ts
enum Status {
  RUNNING = "RUNNING",
  IDLE = "IDLE",
  ERROR = "ERROR"
}
```

- Adding a configuration option to not transform enum values to their keys could be a solution

- I have tried all the enum configuration options currently supported in the typescript plugin but can't find one that helps with this.

**Additional context**

 This makes it difficult to do frontend conditional checks against enum values. E.g. I'm checking that the returned `Status` from a query is of a certain enum value. In my case the one on the graphql response will be an integer, however the generated enum value is now a string – so checking equality is not possible.

theomjones:
Covers #4157 

- [x] Added option to types
- [x] Implement option functionality
- [x] Add tests

### DIFF
--- a/packages/plugins/typescript/typescript/src/config.ts
+++ b/packages/plugins/typescript/typescript/src/config.ts
@@ -68,6 +68,21 @@ export interface TypeScriptPluginConfig extends RawTypesConfig {
    */
   enumsAsTypes?: boolean;
   /**
+   * @description Controls whether to preserve typescript enum values as numbers
+   * @default false
+   *
+   * @exampleMarkdown
+   * ```yml
+   * generates:
+   * path/to/file.ts:
+   *  plugins:
+   *    - typescript
+   *  config:
+   *    numericEnums: true
+   * ```
+   */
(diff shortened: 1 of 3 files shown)


dotansimha:
Thank you @theomjones ! 
Maybe the code there could be improved a bit? it seems a bit duplicated

theomjones:
@dotansimha - thanks, I've extracted some dupe code out into a function. Let me know if you need anything more

dotansimha:
That's perfect, thank you @theomjones :)

PR #4189, opened 2020-06-08
Fix incorrectly renamed config key

mattleff:
Refs #3109. Currently the actual key is `baseTypesPath` so this is just a docs fix.


dotansimha:
Thank you @mattleff !

PR #4207, opened 2020-06-10
JSDoc - Added support for @deprecated directive

CarloPalinckx:
This PR adds support for the `@deprecated` directive by appending a textual warning to the `@property` of a `@typedef`. Sadly JSDoc does not allow you to use the actual `@deprecated` annotation on properties but only on symbols. Would have been cool to have added actual IDE warnings and compiler warnings... ah well 🤷‍♂️


dotansimha:
This is great :) Thank you @CarloPalinckx

PR #4215, opened 2020-06-12
Info about required C# library + example code

cklam2:
Had some hard time figuring out for which GraphQL library it generates the C# code for. The author [mentioned](https://github.com/dotansimha/graphql-code-generator/pull/3726) GraphQL.Net which seems to be the wrong one. Then tried GraphQL for .NET, also the wrong one. 

Finally figured out that it requires [GraphQL.Client](https://www.nuget.org/packages/GraphQL.Client/). Would really be useful if there is a reference to that library in the docs. Not sure if example codes are allowed, added it anyway, helpful for starters.

### DIFF
--- a/website/docs/plugins/c-sharp-operations.md
+++ b/website/docs/plugins/c-sharp-operations.md
@@ -2,6 +2,20 @@
 id: c-sharp-operations
 title: C# Operations
 ---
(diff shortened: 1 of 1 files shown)


dotansimha:
Thank you so much @cklam2 !

PR #4220, opened 2020-06-12
Remove faux prompt because it gets copied to the clipboard

erikras:
The "copy to clipboard" functionality is great, but it's picking up the `$`'s. I think it's pretty self explanatory that the commands are for the command line without the `$`'s there. This seems easier than trimming the string in the clipboard utility.


dotansimha:
You are right, thank you @erikras !

Issue #4089, opened 2020-05-25
Federation: Limit resolver parent types to contain only fields from @key and @requires directives


jakeblaxon:
**Is your feature request related to a problem? Please describe.**
With federation, we now have the ability to determine exactly which parent fields are guaranteed to be available to a resolver. In theory, these are only the fields specified in the `@key` directive for the parent type and the `@requires` directive for the field. (Note: this only applies to the fields of extended types).

**Describe the solution you'd like**
Using this information, we should generate unique parent types for each field's resolver such that they only contain the subset of fields specified by these directives. This would give developers better type safety, since they'll be able to know when they're trying to use a parent field that's not part of the `@key` or `@requires` directives.

**Describe alternatives you've considered**
We could instead just make all the other parent fields that are not in these directives optional, which would force developers to handle their null cases. But that would obscure the nullability information provided for those fields in the schema, and it would still incorrectly indicate that those fields may not be null in some cases.

**Additional context**
For example, consider the following federated type:
```graphql
extend type Book @key(fields: "isbn") {
    isbn: String! @external
    title: String! @external
    numberOfPages: Int! @external
    author: Author @requires(fields: "title")
    publisher: Publisher @requires(fields: "numberOfPages")
}

type Author {
    name: String!
}

type Publisher {
    name: String!
}
```

The parent type for the `Book.author` resolver that currently gets generated contains all fields found in the `Book` type, but since `numberOfPages` isn't included in the `@key` or `@requires` directives, this field should be excluded from the parent type. The `publisher` field should also be excluded from the parent type.

Likewise for the `Book.publisher` resolver, but the parent type shouldn't include `title` and `author` in this case.

ardatan:
@jakeblaxon PRs are welcome!

dotansimha:
Fixed in `v1.17.4`

PR #4232, opened 2020-06-14
Limit ParentType to only contain fields from the @key and @requires directives

jakeblaxon:
This PR seeks to address #4089. It modifies the `ParentType` for resolvers of federated types to only contain fields from the `@key` and `@requires` directives, since these fields are the only ones guaranteed to be available by federation.

This is my first PR with this project, so it's a little rough. My goal was to lay a foundation to solve this issue, but certain edge cases may not be fully addressed (such as unwrapped types and type mappings) and may break backwards compatibility. I furthermore only focused on adding this feature to Typescript, so it remains unimplemented in Flow for now. I think it would be best if we make this an experimental feature, but I'm unsure of the best way to do so. Please feel free to make any recommendations/modifications you see necessary. Thanks!


jakeblaxon:
I don't intend for this PR to be merged yet, but I do want to open it up for discussion. Since this has a high risk of breaking backward compatibility, I think it'd be best to offer it as an experimental feature for now. What would be the best way to do this? Should I add it as a config option or maybe move this feature to its own plugin altogether?

packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -71,6 +71,22 @@ export type NewStitchingResolver<TResult, TParent, TContext, TArgs> = {
...
+            : Maybe<T[K]> extends T[K] ?
+              NonNullable<T[K]> extends (infer U)[] ?
+                Maybe<RecursivePick<NonNullable<U>, S[K]>>[]
+                : Maybe<RecursivePick<NonNullable<T[K]>, S[K]>>
+              : T[K] extends (infer U)[] ?
+                RecursivePick<U, S[K]>[]
+                : RecursivePick<T[K], S[K]>;
+      },
+      keyof T & keyof S
+    >;`);
```
kamilkisiela, line 89: Maybe we could split it into 2-3 types to make it more readable?

dotansimha:
This sounds like a very nice solution for that. @jakeblaxon can you please fix @kamilkisiela 's suggestions? And do you think we need more test cases for this change? maybe add tests with GraphQL `interface` and `type` to make sure it's compatible with that?

packages/plugins/typescript/resolvers/src/index.ts
jakeblaxon: Good suggestion. I split up the types to be a bit more readable. Can you take another look?

dotansimha:
@jakeblaxon looks like a great progress :) What are the remaining tasks for this PR? 

@kamilkisiela can you please take a look again?

packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -72,6 +72,13 @@ export type NewStitchingResolver<TResult, TParent, TContext, TArgs> = {
       context: TContext,
       info: GraphQLResolveInfo
     ) => Promise<TResult> | TResult;`);
+
+    defsToInclude.push(`
+      type scalarCheck<T, S> = S extends true ? T : nullableCheck<T, S>;
+      type nullableCheck<T, S> = Maybe<T> extends T ? Maybe<listCheck<NonNullable<T>, S>> : listCheck<T, S>;
+      type listCheck<T, S> = T extends (infer U)[] ? nullableCheck<U, S>[] : GraphQLRecursivePick<T, S>;
+      export type GraphQLRecursivePick<T, S> = { [K in keyof T & keyof S]: scalarCheck<T[K], S[K]> };
+    `);
```
kamilkisiela, line 81: <3

dotansimha: Maybe just change it to be pascal case? (for example, `ScalarCheck` instead `scalarCheck`) to it will aligned with other helpers we have in this repo.

jakeblaxon:
I apologize for the delay on this. I've updated the types to pascal case. The last few remaining tasks I want to add are tests for interface and union types (and any other cases that you suggest) and documentation for how to use the `GraphQLRecursivePick` type (in case you want to use it as a param or return type in a custom method). The only other thing I would want to consider is adding a flag to enable/disable this specific feature as experimental. I'm fine with not including that, though, if you don't think it's necessary. @dotansimha what are your thoughts?

dotansimha:
Sounds good, thanks @jakeblaxon ! 

Not sure about the need for a feature flag, because this is actually a bug fix and anyway it's better than what we have now. @kamilkisiela what do you think?

jakeblaxon:
@dotansimha I think this PR is in a good state so I've opened it up. I added an interface test, but I'm skipping it for now since this PR doesn't support them. However, I think it's better to just get this in now since it works for types and is non-breaking for interfaces. I ended up not adding a test for unions since they don't really make sense in this context. I also didn't add documentation for the `GraphQLRecursivePick` type, but I think it's easy enough to figure out by looking at the generated types that use it. I wasn't sure where the best place was to put it. If you think there's a good spot where I can add it though, please let me know and I can get that in too!

jakeblaxon:
@dotansimha is there anything preventing us from merging this PR? I noticed this has been open for a while.

eurostar-fennec-cooper:
I could really use this fix, too. Would love to see this merged 👍

dotansimha:
@jakeblaxon No reason. You got @kamilkisiela 's approval so we good :) 
Thank you so much!

dotansimha:
@jakeblaxon  @eurostar-fennec-cooper can you please try the latest alpha? `1.17.4-alpha-48e09323.0`

eurostar-fennec-cooper:
@dotansimha 1.17.4-alpha-48e09323.0 works perfectly/resolves the issue for me.

dotansimha:
Great! Thank you for the feedback, @eurostar-fennec-cooper ! 
We will have a stable release soon :)

Issue #4226, opened 2020-06-12
export Exact type only once


xiaoyu-tamu:
**Is your feature request related to a problem? Please describe.**
```yml
generates:
    preset: near-operation-file
    presetConfig:
      extension: .graphql.ts
      baseTypesPath: common/generated.ts
    plugins:
      - typescript-operations
      - typescript-react-apollo
```

I'm using near-operation-file plugin, After upgrading to 1.15.3, all generated files have 

`export type Exact<T extends { [key: string]: any }> = { [K in keyof T]: T[K] };`. 

I like to export all generated files in index.ts file, so other component can import it without nested import path.

```ts
// index.ts
export * from './a/a.generated.ts';
export * from './b/b.generated.ts';
```

But the problem is they all export type Exact.
```
Module './document.graphql' has already exported a member named 'Exact'. 
Consider explicitly re-exporting to resolve the ambiguity.ts(2308)
```

Can we reuse this type instead of define it in every generated files?

**Describe the solution you'd like**

Define Exact type near Maybe Type, other plugins use this like how they use Maybe.

```
// generated by near-operation-type plugin

import * as Types from 'some_base_file_base';

export type SomeQueryVariables = Types.Exact<{
  listingId: Types.Maybe<Types.Scalars['Int']>;
}>;

```
**Describe alternatives you've considered**

Do not export it.
```
`type Exact<T extends { [key: string]: any }> = { [K in keyof T]: T[K] };`. 

```

**Additional context**

dotansimha:
You are right, @xiaoyu-tamu . Thank you for reporting this!

dotansimha:
I guess the only way to avoid this, it to put the `Exact` type near `Maybe` (in `typescript` plugin). 

I've changed it in this PR: https://github.com/dotansimha/graphql-code-generator/pull/4235

@xiaoyu-tamu can you please try the alpha version `1.15.4-alpha-622ea4d0.0`?

xiaoyu-tamu:
@xiaoyu-tamu Yes, it works

dotansimha:
Fixed in v1.15.4

PR #4235, opened 2020-06-15
fix(typescript): duplicate Exact definition when used with presets

dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/4226


seansfkelley:
Just ran into this issue, and so happy to see that you _just_ fixed it. Thank you! What's your normal release schedule/when can I expect to see a non-alpha version of this published?

dotansimha:
@seansfkelley this fix is available since v1.15.4

PR #4239, opened 2020-06-15
JSDoc: Removed incorrect results for documents containing operations

CarloPalinckx:
This plugin does not yet support operations as output, however when your documents contain operations it breaks the output by casting an object to string. Resulting in a `[object Object]` in the output. Which could turn out to be trouble when ran with a bundler. This PR removes that behaviour and makes it so that the plugin no-op's when encountering a operation or fragment.


dotansimha:
Thanks @CarloPalinckx !

dotansimha:
@CarloPalinckx can you please check the CI? it seems to fail :(

CarloPalinckx:
@dotansimha should be fine now 🙂

Issue #4214, opened 2020-06-12
[c-sharp] Properties is missing ? (nullable) when graphtype is allowing null and the same for deprecated


madsrode:
**Describe the bug**
When a field is marked as nullable and/or deprecated is it correctly generated in ex. typescript, but in C# is both missing from the generated file.

**To Reproduce**
Sandbox link: https://codesandbox.io/s/lingering-paper-0zysu

1. My GraphQL schema:

```graphql
type Query {
    user(id: ID!): User!
}

type User {
    id: ID!
    username: String!
    email: String!
    nullableFloat: Float @deprecated(reason: "Just gone.")
}

```

2. My GraphQL operations:

```graphql
query user {
    user(id: 1) {
        id
        username
        email
        nullableFloat
    }
}

```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  types.ts:
    plugins:
      - typescript
  Gen.cs:
    plugins:
      - c-sharp

```

**Expected behavior**
Gen.cs file with User class should look like this: (with ? at the type of nullableFloat and Obsolete attribute above the field.

`  #region User
  public class User {
    #region members
      
            [JsonProperty("id")]
            public string id { get; set;}
          
    
            [JsonProperty("username")]
            public string username { get; set;}
          
    
            [JsonProperty("email")]
            public string email { get; set;}
          
    
            [JsonProperty("nullableFloat")]
            [Obsolete("Just gone.")]
            public float? nullableFloat { get; set;}
          
    #endregion
  }
  #endregion`

**Environment:**

- OS: 
- `@graphql-codegen/...`: 
- NodeJS:

cklam2:
Indeed. Also probably due to the missing nullable marker I see quite some warnings in the generated types file. For example
```
if (this.hasPhone != null) {
   d["hasPhone"] = hasPhone;
}
```

Warning shown for this excerpt:
```
bool User.hasPhone { get; set; }
The result of the expression is always 'true' since a value of type 'bool' is never equal to 'null' of type 'bool?'
```

cklam2:
Attempting to solve this + some other improvements:
- Nullable markers for fields that use value type and are optional (so no support yet for nullable reference types)
- Null check only for nullable value types and reference types
- Mark field obsolete when the @deprecated decorator found
- Comment header. Currently only for fields, hopefully can do it for all fields/classes, etc
- Boolean => bool

[...]

PR #4242, opened 2020-06-16
C# plugin improvements

cklam2:
Solves #4214. 

Additionally added some more improvements:
- Nullable value types
- Initial value support
- Class and field documentation in summary block
- Deprecated fields have DeprecatedAttribute
- Required input fields have JsonRequiredAttribute
- Generic implementation to compose input object
- Code in namespace indented and wrapped

There's one incompatibility though: the generated C# method `getInputObject` is renamed to `GetInputObject` just to follow the general C# naming convention. If breaking compatibility is a no-go to you, it can be reverted.

### DIFF
--- a/packages/plugins/c-sharp/c-sharp/src/common/c-sharp-declaration-block.ts
+++ b/packages/plugins/c-sharp/c-sharp/src/common/c-sharp-declaration-block.ts
@@ -1,9 +1,10 @@
-import { transformComment, indentMultiline } from '@graphql-codegen/visitor-plugin-common';
+import { indentMultiline } from '@graphql-codegen/visitor-plugin-common';
 import { StringValueNode, NameNode } from 'graphql';
+import { transformComment } from './common';
 const stripIndent = require('strip-indent');
 
 export type Access = 'private' | 'public' | 'protected';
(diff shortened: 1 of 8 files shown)


dotansimha:
This is great! Thank you @cklam2 ! 
Any chance you can add some unit tests to make sure we are not breaking these changes in the future?

cklam2:
@dotansimha
Yep, already added some unit tests (locally) for the fixes but they are not complete yet. Eg missing tests for existing code. Ok to do another pull request once they are done? Or you want it be included in this one?

dotansimha:
I think if tests are missing for the existing code, we can improve it in a future PR. Let me know when you push the tests for the modified code, and we'll progress on that.

Thank you!

cklam2:
Well, glad that I extended the unit tests cause found some regressions in the initial values part and the array composition. Also refactored some of the code, hence all those changes.

Also committed the unit test for the c-sharp plugin. Works locally. Only not sure if I need to commit a `jest.config.js` or need to change `package.json`. @dotansimha can you check if the unit test needs some configuration in the project? 

Rebased the branch to latest upstream master.

dotansimha:
This is great, thank you so much @cklam2 !

Issue #4221, opened 2020-06-12
[c-sharp-operations] suffixes for query, mutation and subscription don't work


cklam2:
**Describe the bug**
According to the documentation from [c-sharp-operations](https://graphql-code-generator.com/docs/plugins/c-sharp-operations) you can use the `querySuffix`, `mutationSuffix` and `subscriptionSuffix` to customize the suffix value (default GQL) but setting a value in codegen.yml doesn't have any effect on the output.

**To Reproduce**
1. Go to https://graphql-code-generator.com/
2. Select `C# operations` from the dropdown list
3. Append `querySuffix` to codegen.yml:

```yml
config:
  querySuffix: Query
generates:
  src/main/c-sharp/my-org/my-app/Operations.cs:
    plugins:
      - c-sharp-operations
```

**Expected behavior**
The generated `getFindUserGQL`  should be `getFindUserQuery`.
Same happens with `mutationSuffix` and `subscriptionSuffix`.

**Environment:**
GraphQL Code Generator live environment 

and

Windows 10 
```
  "devDependencies": {
    "@graphql-codegen/c-sharp": "^1.15.2",
    "@graphql-codegen/cli": "1.15.3",
    "@graphql-codegen/c-sharp-operations": "^1.15.3",
    "graphql": "^15.1.0"
  }
```

dotansimha:
Fixed in v1.15.4

PR #4243, opened 2020-06-16
Fix operation name suffix

cklam2:
Fixes the suffix bug as described in #4221 

Comparison was against Pascal Cased operation type string while operation type is always camel case, regardless of the operation type's "alias".


dotansimha:
Thank you @cklam2 !

PR #4275, opened 2020-06-23
dont generate fragments because they already come from typescript-rea…

auryn31:
Fix issue with fragments.
If there are some, typescript-react-apollo and typescript-react-offix will generate them and create uncompilable code


dotansimha:
Makes sense ;) Thanks @auryn31 !

PR #4284, opened 2020-06-24
C sharp operation request and summary

cklam2:
Some more improvements in the c# operations plugin:

Summary header for methods with a list of required and optional input variables so that you don't have to check the gql for arguments. Note that there may be some gaps, eg incorrect casing for enum types but since it's in comment block it's not that big of an issue. Can improve it later when we know the corner cases.

Added `Request()` that accepts optional variable arg values. That way you can pass variables to the request in one go. Example:
```
var request = UserGQL.Request(new
{
  id = 3,
  name = "Minion"
});
```

I moved the existing "get{ServiceName}" methods like `getUserGql()` a bit to the background. Could not mark it with the ObsoleteAttribute cause even after setting isError = false the VS compiler is still treating warnings as errors by default. Cause I didn't want to upset existing users I used a <remark> block instead.


packages/plugins/c-sharp/c-sharp-operations/src/visitor.ts
```diff
@@ -126,6 +138,26 @@ export class CSharpOperationsVisitor extends ClientSideBaseVisitor<
...
+  private _gqlInputSignature(variable: VariableDefinitionNode): { signature: string; required: boolean } {
+    const typeNode = variable.type;
+    const innerType = getBaseTypeNode(typeNode);
+    const schemaType = this._schema.getType(innerType.name.value);
+
+    const name = variable.variable.name.value;
+    const baseType = !isScalarType(schemaType) ? innerType.name.value : this.scalars[schemaType.name] || 'object';
+
+    const isArray =
+      typeNode.kind === Kind.LIST_TYPE ||
```
dotansimha, line 150: Would this work with `field: [[[String]]]`? 🤔

dotansimha: Never mind, it's just for indication.

dotansimha:
Thanks @cklam2 ❤️

packages/plugins/c-sharp/c-sharp-operations/src/visitor.ts
cklam2: Good question. I will check that later and do another pull request if changes are needed.

PR #4287, opened 2020-06-24
Fix bug with flow-operations codegeneration and flattenGeneratedTypes config

ganemone:
This fixes a bug where fragment types are not generated when `flattenGeneratedTypes` config is provided for flow operations.

### DIFF
--- a/dev-test/githunt/flow.flow.js
+++ b/dev-test/githunt/flow.flow.js
@@ -1,5 +1,7 @@
 /* @flow */
 
+// @flow
+
 /** All built-in and custom scalars, mapped to their actual values */
 export type Scalars = {|
   ID: string,
@@ -167,7 +169,6 @@ export type SubscriptionCommentAddedArgs = {|
   repoFullName: $ElementType<Scalars, 'String'>,
 |};
 
-// @flow
 type $Pick<Origin: Object, Keys: Object> = $ObjMapi<Keys, <Key>(k: Key) => $ElementType<Origin, Key>>;
 
 export type OnCommentAddedSubscriptionVariables = {
--- a/dev-test/githunt/flow.flow.js
+++ b/dev-test/githunt/flow.flow.js
@@ -167,6 +167,7 @@ export type SubscriptionCommentAddedArgs = {|
   repoFullName: $ElementType<Scalars, 'String'>,
 |};
 
+// @flow
 type $Pick<Origin: Object, Keys: Object> = $ObjMapi<Keys, <Key>(k: Key) => $ElementType<Origin, Key>>;
 
 export type OnCommentAddedSubscriptionVariables = {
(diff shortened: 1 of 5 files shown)


dev-test/githunt/flow.flow.js
```diff
@@ -1,5 +1,7 @@
 /* @flow */
 
+// @flow
```
dotansimha, line 3: Is it really need? We have this in the 1st line, no? (as `/* @flow */`)

packages/plugins/flow/operations/src/index.ts
```diff
@@ -10,25 +10,38 @@ export const plugin: PluginFunction<FlowDocumentsPluginConfig> = (
   config: FlowDocumentsPluginConfig
 ) => {
   const documents = config.flattenGeneratedTypes ? optimizeOperations(schema, rawDocuments) : rawDocuments;
-  const prefix = `type $Pick<Origin: Object, Keys: Object> = $ObjMapi<Keys, <Key>(k: Key) => $ElementType<Origin, Key>>;\n`;
 
-  const allAst = concatAST(documents.map(v => v.document));
+  const prefix = config.preResolveTypes
```
dotansimha, line 14: Great catch :D

dev-test/githunt/flow.flow.js
ganemone: yes - if you are using flow-operations plugin alone (as is with near-operation preset)

packages/plugins/flow/operations/src/index.ts
```diff
@@ -10,25 +10,38 @@ export const plugin: PluginFunction<FlowDocumentsPluginConfig> = (
...
+    : `type $Pick<Origin: Object, Keys: Object> = $ObjMapi<Keys, <Key>(k: Key) => $ElementType<Origin, Key>>;\n`;
+
+  let allAst = concatAST(documents.map(v => v.document));
+  const includedFragments = concatAST(rawDocuments.map(v => v.document)).definitions.filter(
+    d => d.kind === Kind.FRAGMENT_DEFINITION
+  );
+
+  // Fragments get removed when the types are optimized by relay
+  // We still want to export the types of the fragments included in the documents
+  if (config.flattenGeneratedTypes) {
```
ardatan, line 25: Maybe we should use `includeFragments` instead?
https://github.com/ardatan/graphql-tools/blob/master/packages/relay-operation-optimizer/src/index.ts#L15

ganemone: Updated to use that approach. Good find!

matikrk:
Hey
@ganemone @dotansimha 
This change causing bug with flow
Right now we have 2 definitions of flow

In the generated file, you can find:
```js
/* @flow */


// @flow
```

When we have smth like this we got error:
`Flow: Unexpected `@flow` declaration. Only one per file is allowed.` 
and then whole generated file is incorrect

ganemone:
Is this a flow lint error? AFAIK there is no problem with having multiple flow comments in a single flow file

Issue #4273, opened 2020-06-22
[typescript] non nullable graphql inputs with default values are not optional as ts types


bzvikler:
**Describe the bug**
Currently if a property within the gql schema is set as non nullable, but has a default value assigned to it, the generated typescript property will be set as required.

**To Reproduce**

- navigate to this code sandbox: https://codesandbox.io/s/wandering-forest-efbol?file=/types.ts
- notice that while the input of `App` has a default for `users`, it is still required in the generated ts type

1. My GraphQL schema:

```graphql
input App {
    id: ID!
    displayName: String
    users: [String!]! = ["Hey", "hello"]
}
```

2. Generated type

```typescript
export type App = {
  id: Scalars['ID'];
  displayName?: Maybe<Scalars['String']>;
  users: Array<Scalars['String']>;
};
```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-operations

```

**Expected behavior**
If there is a provided default value, the generated ts type should allow for a value of `undefined`, as this is when the default value would be applied.

**Environment:**
- running on the code sandbox

**Additional context**
The issue arose when I was trying to integrate with a gql schema that indicated an array type with a default value. I wanted the default value, but due to this issue, I had to explicitly enter that value, as opposed to just omitting the type and letting the default handle it.

dotansimha:
Fixed in `@graphql-codegen/typescript@1.17.10`

mhart:
Hmmm, this has caused a bunch of incompatible changes in our codebase from a patch release.

Making these optional might make sense on a client, but not on a server – because they're guaranteed to have values. What's the best way to get around this issue?

dotansimha:
@mhart I think you are right, can you please report this as separate issue? we'll look into it :) maybe this should be configurable.

PR #4296, opened 2020-06-26
Nested array, interface and more

cklam2:
* Added nested array support (eg [[[String]]])
* Dropped assigning default values
* Mark fields with default value optional, even when it had a ! (see #4273)
* Generate C# interface from gql interface
* If request doesn't have input args, don't include the variable arg in Request
* Solved circular dependency issue
* Typo: <remark> => <remarks>


dotansimha:
Amazing work. Thank you so much @cklam2 !

PR #4319, opened 2020-07-03
feat: allow hooks to be functions if used from node

### LINKED ISSUE #4219: Allow config hooks to be functions if used cli programatically

I would like to execute a function when a watch is triggered.
I could do so with using the `onWatchTriggered` hook, but this only executes commands.
It is a very big overhead to spawn a new node process if this happens often and a function would help here to run the thing in process.

I would like to add a function to the hook configuration.

KnisterPeter:
This adds the ability to use hook functions instead of only run
executabels as hooks.
This option could only be used if integrating graphql-codegenerator
directly from nodejs.

Related #4219


dotansimha:
This is awesome! Thank you so much @KnisterPeter !

Issue #3992, opened 2020-05-08
init does not work out of the box


theogravity:
I executed `npx graphql-codegen init` for just a backend server

I have a `schema.graphql` file i've built, and the generated `codegen.yml` reads like this:

```
overwrite: true
schema: "schema.graphql"
documents: null
generates:
  src/generated/graphql.ts:
    plugins:
      - "typescript"
      - "typescript-resolvers"
  ./graphql.schema.json:
    plugins:
      - "introspection"
```

when I run the generate command, I get

```
  ✖ src/generated/graphql.ts
    Unable to find template plugin matching 'typescript-resolvers'
    Install one of the following packages:
    - @graphql-codegen/typescript-resolvers

...

  ✖ ./graphql.schema.json
    Unable to find template plugin matching 'introspection'
    Install one of the following packages:
    - @graphql-codegen/introspection
...
```

the init script added to my `package.json` (I have run `npm i` to make sure it's installed):

```
    "@graphql-codegen/typescript": "1.13.5",
    "@graphql-codegen/typescript-resolvers": "1.13.5",
    "@graphql-codegen/introspection": "1.13.5"
```

so I'm not clear what's missing here.

node version `12.16.1`, using npm `6.13.4`

dotansimha:
Hi @theogravity ,
Thank you for reporting this. I'll update the docs for getting started, because we can't really use `npx` at the moment, because `npx` installs the codegen globally, so it means that `graphql` needs to be installed globally as well, which causes a lot more issues after.

PR #4331, opened 2020-07-06
Remove npx connand from quickstart docs

BrockReece:
I just got tripped up by the same issue as in #3992
I thought I would be a good citizen and remove this from the docs so it doesn't confuse anyone else?


dotansimha:
Thank you @BrockReece !

PR #4355, opened 2020-07-12
Fixing Issue #4437 - Failing tests for unhandledRejection

asafh:
In regards to issue #4437 I've made two fixes to the angular tests: (i) awaiting all validateTypeScript calls as to not hide failures and (ii) fixing the test that was (silently) failing by updating its gql mutation/subscription to fit the actual schema.

Other than that, this PR also includes a setup file for jest which makes sure the execution fails if there's an unhandledRejection so that they may not stay hidden.

### DIFF
--- a/dev-test/setup.js
+++ b/dev-test/setup.js
@@ -1,0 +1,3 @@
+process.on('unhandledRejection', (err) => {
+    fail(err);
+});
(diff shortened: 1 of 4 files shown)


dotansimha:
Great catch. Thank you so much @asafh !

PR #4391, opened 2020-07-18
Add note to near-operation-file-preset docs

danielrearden:


dotansimha:
Thanks @danielrearden ! 
I think maybe we can solve that? 🤔  I recall that we have some piece of code in codegen that adds the expected output files to the list of ignored globs. I guess it's based on `generates` section, which is not relevant by itself for the ignore list (because it's just `./` or similar and not specific file).

danielrearden:
@dotansimha I was holding off on merging this PR until I confirmed with you that it wasn't in fact fixable. If it is something we can fix, I can circle back to it this week and puzzle it out.

dotansimha:
This is what I had in mind: 

The signature of presets at the moment is defined like that:  https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-helpers/src/types.ts#L238-L240

Maybe we can change it to:

```
 export type OutputPreset<TPresetConfig = any> = {
    buildGeneratesSection: (options: PresetFnArgs<TPresetConfig>) => Promisable<GenerateOptions[]>;
    expectedOutputFiles?: (options: PresetFnArgs<TPresetConfig>): string[]
  };
```

The `expectedOutputFiles` can be an extraction of the logic we have in the presets, example: https://github.com/dotansimha/graphql-code-generator/blob/master/packages/presets/near-operation-file/src/index.ts#L152-L163

Then, we can all it here (when it's available, so we won't break the existing API by adding a required field):  https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/codegen.ts#L167

And get a list of expected output files from that preset. This way we can pass it to `context.loadSchema` and `context.loadDocuments` (https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/codegen.ts#L197) and from there to the config loader and add it to the list of ignored files (or, negative glob) while passing it to graphql-tools. 

What do you think @danielrearden ?

dotansimha:
Merging as-is for now, we can improve that later. Thanks @danielrearden !

PR #4394, opened 2020-07-19
Update import paths for @apollo dependencies

koenpunt:
Not all types are exported from `@apollo/client`. So instead of using that for all imports, I've updated the imports to point to the correct location.


dotansimha:
Thanks @koenpunt ! I'll review it (and the other PR) soon
We are thinking a lot how to do that, since we need to make sure to have good support for Apollo 3, generate modern code (hooks) and disable old methods (HOC/components), without making it too difficult for developers to do.

dotansimha:
Thanks @koenpunt , I opened a PR with all changes required to have complete support (and deprecated all old ones). 
We'll make it available soon as v2 of that plugin: https://github.com/dotansimha/graphql-code-generator/pull/4487

PR #4395, opened 2020-07-19
Change default value of withComponent to false

koenpunt:
Since render prop components are deprecated:

Ref: https://www.apollographql.com/docs/react/api/react/components/


dotansimha:
Thanks @koenpunt , I opened a PR with all changes required to have complete support (and deprecated all old ones). 
We'll make it available soon as v2 of that plugin: https://github.com/dotansimha/graphql-code-generator/pull/4487

dotansimha:
I fixed it in https://github.com/dotansimha/graphql-code-generator/pull/4487 , also v3 of Apollo-Client and Hooks should be generated by default now. can you please try the alpha version? it's `1.17.8-alpha-f79b3113.0`

PR #4401, opened 2020-07-21
Add context to plugins and custom loaders

danielrearden:
This exposes a `pluginContext` object that's accessible to plugins, presets and custom loaders. This provides a way to pass arbitrary data from custom loaders (and presets) to plugins.

### DIFF
--- a/packages/graphql-codegen-core/src/execute-plugin.ts
+++ b/packages/graphql-codegen-core/src/execute-plugin.ts
@@ -11,6 +11,7 @@ export interface ExecutePluginOptions {
   outputFilename: string;
   allPlugins: Types.ConfiguredPlugin[];
   skipDocumentsValidation?: boolean;
+  pluginContext?: { [key: string]: any };
 }
 
 export async function executePlugin(options: ExecutePluginOptions, plugin: CodegenPlugin): Promise<Types.PluginOutput> {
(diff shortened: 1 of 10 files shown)


packages/graphql-codegen-core/src/execute-plugin.ts
```diff
@@ -33,10 +34,19 @@ export async function executePlugin(options: ExecutePluginOptions, plugin: Codeg
 
   const outputSchema: GraphQLSchema = options.schemaAst || buildASTSchema(options.schema, options.config as any);
   const documents = options.documents || [];
+  const pluginContext = options.pluginContext || {};
 
   if (plugin.validate && typeof plugin.validate === 'function') {
     try {
-      await plugin.validate(outputSchema, documents, options.config, options.outputFilename, options.allPlugins);
+      // FIXME: Sync validate signature with plugin signature
```
danielrearden, line 41: Would be good to sync the signatures for v2

website/docs/getting-started/schema-field.md
```diff
@@ -172,8 +172,10 @@ If your schema has a different or complicated way of loading, you can point to a
 
 ```yml
 schema:
-  - ./my-schema-loader.js:
-      noPluck: true
+  - http://localhost:3000/graphql:
```
danielrearden, line 175: Reverted the change to the docs here. While CodeFileLoader can be used here with the `noPluck` option, it doesn't actually work with functions (see [here](https://github.com/ardatan/graphql-tools/blob/master/packages/loaders/code-file/src/exports.ts#L60-L88))

PR #4417, opened 2020-07-23
Fix plugin Typescript mongoDB: generate optional type for @map 

xiaoliu-heng:
```gql
name: String @column @map(path: "fullName")
```
output before: 
```js
fullName: Maybe<string>
```
fixed output
```js
fullName?: Maybe<string>
```


dotansimha:
That's perfect! Thank you so much @xiaoliu-heng !

PR #4435, opened 2020-07-27
Test upload of canary versions to gh artifacts

dotansimha:


dotansimha:
It seems like GitHub Actions artifacts are not available for direct consumption using npm/yarn. The links are temporary, and not predicable. Skipping this :(

PR #4436, opened 2020-07-27
feat: support type imports in `near-operation-file` preset

SimenB:
For #3695.

I'll try to find the time to go through the others I need for my own project, but it seems we might need to update most plugins and presets, which I don't have the time for. Hopefully others will just fix the ones they need themselves as we go 🙂


dotansimha:
That's awesome, thank you for this great work. @SimenB  ❤️

PR #4437, opened 2020-07-27
feat(typescript-resolvers): use `import type` when importing from graphql

SimenB:
Another one for #3695. For my own use case I _believe_ I'm just missing `react-apollo`. However, I'm not sure when it's safe to do `import type` and when not to there - help would be appreciated 🙂 This is however enough for the stuff we have server side 🎉


dotansimha:
Great :) We are close to make it compatible with all plugins ;) Thanks @SimenB !

PR #4448, opened 2020-07-28
feat(react-apollo): generate type imports

SimenB:
Again for #3695. Didn't turn out too bad! Sorta weird, and maybe not correct, but these changes works at least for my work project


dotansimha:
This is great! Thank you for amazing work @SimenB !

PR #4461, opened 2020-07-30
fix: use 'unknown' instead of 'any' for type Exact

mjbcopland:
Resolves _"Unexpected any. Specify a different type."_ from `@typescript-eslint/no-explicit-any`.

___Why am I linting generated files?___

1. It's useful when [using ESLint as a lifecycle hook](https://graphql-code-generator.com/docs/integrations/prettier#eslint).
2. It's useful to prevent `any` being used for custom scalar types.


dotansimha:
Perfect, thank you @mjbcopland !

PR #4462, opened 2020-07-30
fix: do not import the unused DocumentNode for typed-document-node

mjbcopland:
The new(ish?) typed-document-node plugin includes an unused import of the default DocumentNode from ClientSideBaseVisitor. One issue with this is it triggers ESLint's no-unused-vars rule.

This PR moves the custom DocumentNode import to the base class with added customisability similar to the `gql` import (e.g. it should now be possible to use [graphql-typed](https://www.npmjs.com/package/graphql-typed) should you wish) and removes the unused import.


dotansimha:
This is awesome! Thank you @mjbcopland ! 
Great idea to alias the import and use `as` for importing it with the same name. It seems like this will solve these 2 issues:
- https://github.com/dotansimha/graphql-typed-document-node/issues/30
- https://github.com/dotansimha/graphql-typed-document-node/issues/25

PR #4502, opened 2020-08-04
fix: apollo v3 should not use type import

SimenB:
Otherwise output is

```ts
import type * as Apollo from '@apollo/client';
import * as Apollo from '@apollo/client';
```

Which both complains about duplicate identifiers and using type import as value


SimenB:
@dotansimha should I target the `release-pipeline` branch?

dotansimha:
@SimenB `master` is great, I will merge the other PR in a few minutes. Thanks !

PR #4507, opened 2020-08-04
Support Apollo Angular 2.0 and add `serviceProvidedIn` option

kamilkisiela:
Related #4513 and kamilkisiela/apollo-angular#1383


dotansimha:
@kamilkisiela Let's merge this when `apollo-angular` v2 is available. We can also change the default to be `v2` and have a `major` bump.

kamilkisiela:
LGTM!

PR #4520, opened 2020-08-05
🐛: double quotes gql queries are not correctly escaped in c-sharp operations plugin

afitzek:
If a double quotes for a constant string are used in a graphql
query these are escaped in c-sharp code as `"""` in the
verbatim string, but in a verbatim string the escape sequence
for a `"` is `""`.

Change-type: patch
Signed-off-by: Andreas Fitzek <andreas.fitzek@gmail.com>


dotansimha:
That's great, thank you @afitzek !

Issue #4529, opened 2020-08-06
Mappers are not used for interface types


toriningen:
**Describe the bug**
I have been migrating my code from `typescript-resolvers@1.9.1` to `typescript-resolvers@1.17.7`, and have noticed that v1.9.1 applies mappers to interfaces (which is behavior that I would expect), while 1.17.7 does not — instead, it expands types implementing interface, and it breaks my types.

I expect `ResolversParentTypes['User']` to be `User: UserParent` in accordance to `mapper` config (where it's set as `User: ./parents#UserParent`), but v1.13.1 and later versions output `User: ResolversParentTypes['Human'] | ResolversParentTypes['Bot']` instead.

I have tracked down the breaking change to happen between v1.13.0 and v1.13.1, in this particular commit — https://github.com/dotansimha/graphql-code-generator/commit/cbd87933dd79685935b8bb1db724340a1166dc4e#diff-9ecae63f92a8c9dcfce805bfbe64220aR426-R442 — it puts `graphql.isInterfaceType` check before `isMapped` check, so it never happens for interface types.

**To Reproduce**

1. Follow the link: https://codesandbox.io/s/compassionate-hooks-whqs9?file=/types.ts:3877-3946
2. Run `yarn && yarn generate` from terminal;
3. Observe `types.ts`, line 122, you will see it's `User: ResolversParentTypes['Human'] | ResolversParentTypes['Bot']`, which is **incorrect** result, because mapper isn't honored;
4. In `package.json`, roll version back to `"@graphql-codegen/typescript-resolvers": "1.13.0"`;
5. Run `yarn && yarn generate` from terminal again;
6. Observe `types.ts`, line 122, you will see it's `User: UserParent`, which is desired result;

Sample config I use for case reproduction:

1. My GraphQL schema:

```graphql
type Query {
    user(id: ID!): User!
}

interface User {
    id: ID!
    username: String!
}

type Human implements User {
    id: ID!
    username: String!
    email: String!
}

type Bot implements User {
    id: ID!
    username: String!
    token: String!
}
```

2. My GraphQL operations:

```graphql
query user {
    user(id: 1) {
        id
    }
}
```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      mappers:
        User: ./parents#UserParent
```

**Expected behavior**
[...]

PR #4531, opened 2020-08-06
Allow mappers to be used over interfaces

toriningen:
This attempts to fix the issue mentioned in https://github.com/dotansimha/graphql-code-generator/issues/4529 by <s>introducing additional configuration options to bring back</s> defaulting to pre-1.13.1 behavior.


dotansimha:
Thank you @toriningen ! 

Regarding interfaces:
I think the default behaviour should be that interface mappers should use the types for the mappers (like it does today), but if a mapper is specified explicitly for an `interface`, it should use it instead. This way we can even avoid the need for the new plugin configuration, and still support both use cases. (because it's possible to use types implementing the interface, or even override it if you wish).

Regarding enums: 
Can you please explain the issue? You issue doesn't explain what's missing for enums. We recently had a major work on the integration of the resolvers plugin, mappers and the `enumValues` field. So now if you have mappers for enums, it should be applied, and also if you are using `enumValues`. 

Anyway, I think this PR should only handle the interface issue, and not both, since we are trying to limit PRs to a single change at a time (I know it's annoying sometimes, but it's easier for us to review and test those).

toriningen:
@dotansimha, thanks for a feedback!

**Regarding interfaces:** if I understand it correctly, what you propose is to make pre-1.13.1 behavior the default again — e.g. specifying mapper explicitly on interface type would just use the mapper type. I have considered it, but because I have discovered this issue by migrating my code from older version of `graphql-codegen`, I thought it would be poor experience to break somebody's else code which has been written after 1.13.1 with current behavior in mind after they upgrade.

Because now specifying mapper for interface is a no-op, would it make sense to show a warning during code generation, if a mapper is specified for an interface? E.g. something along the lines of "Warning: Mapper `./foo-model#FooParent` was specified for type `Foo`, which is an interface. This does not affect generated code output. Remove the mapper to discard this warning, or enable `applyMappersOverInterfaces` config option to use it (will become enabled by default in 1.x.x)."

Or if you think it's safe enough to just go for it, then no problem, I will just remove this config option completely and make it the new default — in a hope that no one has included a seemingly no-op instruction in their `codegen.yml`.

**Regarding enums:** to be honest, I've included it just for the sake of completeness, because that breaking change in 1.13.1 also prevented enum mappers from being used. Though, I haven't tried to use any enum mappers in my code base myself, and if there is already an option to configure enum mappers explicitly, then I think it's safe to just delete this config option altogether and keep existing behavior.

------

So, to keep things isolated, I'd only limit this PR to interfaces. Should I go with a warning and keep `applyMappersOverInterfaces: false` as default, or make `applyMappersOverInterfaces: true` a default and remove this configuration option altogether?

toriningen:
Anyway, to keep things moving, I've reduced the patch to only affect interfaces, and made it default behavior to use explicit mappers over interfaces, and reverted related configuration options.

dotansimha:
I think it's safe enough to assume that developer who are using mappers for implementing types are not using mappers for the base interface, especially because it's not working at the moment. 
To be honest, it feels for me more like a bug than a feature, that's why i think it should be the default behaviour, without any configuration.

Thank you so much @toriningen !

Issue #4493, opened 2020-08-03
TypeScript Resolvers - Apollo Federation Broken


rherber:
**Describe the bug**
Upgraded from 1.16.3 to 1.17.7 and all resolvers resolving a federated type broke.

**To Reproduce**
Attempt to compile TypeScript files using generated types and resolvers.

1. My GraphQL schema:

Schema from service 1
```graphql
 extend type Query {
    getRoomType(id: String!): RoomType
    getRoomTypes(destination: String!): [RoomType]
  }

  type RoomType @key(fields: "id") @key(fields: "operaCode operaPropertyCode") {
    propertyId: String!
    operaCode: String!
    property: Property!
    id: String!
    name: String!
    active: Boolean
    operaPropertyCode: String @deprecated(reason: "Not intended for client use.")
  }

  extend type Property @key(fields: "id") {
    id: String! @external
    roomTypes: [RoomType!]
  }
```

Schema from service 2
```graphql
 extend type Query {
    getProperty(id: String!): Property
    getProperties(destination: String = "all"): [Property!]!
  }

  type Property @key(fields: "id") @key(fields: "operaPropertyCode operaRoomTypeCode") {
    id: String!
    propertyId: String!
    path: String!
    detailPageBrandURL: String
    detailPageCorpURL: String
    operaPropertyCode: String
    name: String!
    images: Images
    shortName: String
    region: String!
    regionName: String
    timezone: String
    websiteUrl: String
    emailAddress: String
    phoneNumber: PropertyPhoneNumber!
    propertyAddress: PropertyAddress!
    digitalKeyVendor: String @deprecated(reason: "digitalKeyVendor is now at the room level instead of property.")
    operaRoomTypeCode: String @deprecated(reason: "Not intended for client use.")
  }

  type PropertyPhoneNumber {
    generalLabel: String
    generalNumber: String
    reservationsLabel: String
    reservationsNumber: String
    additionalPhone1Label: String
    additionalPhone1Number: String
    cabanasNumber: String
  }

  type PropertyAddress {
    streetAddress: String
    addressLocality: String
    addressRegion: String
    postalCode: String
    addressCountry: String
  }
```

2. My GraphQL operations:

N/A. Issue is on TypeScript compilation

3. My `codegen.yml` config file:

```yml
generates:
  # Property Service
  packages/gql-types/src/schema/property/schemaTypes.generated.ts:
    schema:
      - ./services/property/src/schema.ts
    config:
      federation: true
[...]

PR #4542, opened 2020-08-11
fix(resolvers/federation): only transform federated parent types when they contain @external directive

jakeblaxon:
This PR seeks to fix the issue raised in https://github.com/dotansimha/graphql-code-generator/issues/4493

This updates the `transformParentType` function to only transform parent types if one of their fields contains an @external directive, or if the field is named `__resolveReference`. This essentially translates to whether the type extends a type defined in a remote schema, or whether the type is a base type.

The objective is to target only the resolvers that the gateway will call from the `_entities` query in a federated schema. The gateway only guarantees that fields from the `@key` or `@requires` directives will be supplied in the parent type, so we limit the parent type to only contain these fields. If the parent type is a base type, however, then its field resolvers (other than `__resolveReference`) will be called from elsewhere, so we therefore go with the default behavior which is to assume all fields will be available.

### DIFF
--- a/packages/plugins/typescript/resolvers/tests/federation.spec.ts
+++ b/packages/plugins/typescript/resolvers/tests/federation.spec.ts
@@ -139,8 +139,8 @@ describe('TypeScript Resolvers Plugin + Apollo Federation', () => {
       }
 
       extend type Address {
-        street: String!
-        zip: Int!
+        street: String! @external
+        zip: Int! @external
       }
     `;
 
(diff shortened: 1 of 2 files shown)


dotansimha:
Thank you @jakeblaxon , this is great! Can you please add a changeset? Using `yarn changeset`, and bump the resolvers package with a `patch`.

dotansimha:
Never mind, merged, because I want to make sure it will be in the next release ;) 
I will add the changeset :) 
Thanks @jakeblaxon !

PR #4551, opened 2020-08-11
fix: @graphql-codegen/typescript-graphql-request peerDependencies of graphql-tag 

andykenward:
There is a typo of `peerDependenxcies` and `graphql-tag` is needed to use `@graphql-codegen/typescript-graphql-request`.


dotansimha:
Great catch. Thanks @andykenward !

PR #4552, opened 2020-08-11
fix(jsdoc): Correctly placed description in typedef

CarloPalinckx:
Turns out that you are not supposed to add an `@description` annotation under an `@typedef` annotation or it will end being invalid


dotansimha:
Perfect, thank you @CarloPalinckx !

PR #4579, opened 2020-08-14
Remove unused code in c-sharp plugins

jorrit:
While preparing another change I noticed some code in the c-sharp plugin that seems to be a left over from a fork. This PR removes this code.

This PR also fixes the package description of the c-sharp-operations package.


dotansimha:
Perfect! Thank you @jorrit !

PR #4581, opened 2020-08-17
Update typescript-apollo-angular.md

moemode:
Not showing newbies which observable implementation to use might lead to them getting a hard to interpet typescript error:
Type 'Observable<unknown>' is missing the following properties from type 'Observable<UsdRatesGQL>': [Symbol.observable], map, filter, reduce, and 2 more.ts(2740)


dotansimha:
Thank you @moemode !

PR #4597, opened 2020-08-19
[typescript] Added `disableComments` to remove ignore all descriptions

TheOnlyBeardedBeast:
Extended typescript generation, added `disableComments` option


dotansimha:
Hi @TheOnlyBeardedBeast ! 
I think we might be able to implement this with minimal changes. 
One of the advantages of AST visiting is the ability to modify the AST and transform it. So I think removing comments can be done on a simpler form, with less code changes.


```ts
import { parse, visit } from "graphql";

const exampleAst = parse(`type Query { """this is a commit""" test: String }`);
const withoutComments = visit(exampleAst, { 
  leave: {
    FieldDefinition: node => ({ ...node, description: null }),
    ObjectTypeDefinition: node => ({ ...node, description: null }),
    // ... 
  }
});
```

What do you think?

TheOnlyBeardedBeast:
I like it, it would be easier to implement and maybe reusable for other plugins this way, I will take a look at it. @dotansimha

dotansimha:
@TheOnlyBeardedBeast that's great! Let me know what it's ready for review ;) Thanks!

dotansimha:
@TheOnlyBeardedBeast update? :)

TheOnlyBeardedBeast:
@dotansimha Hi, sorry for the late work, I had some business trips etc. I created an transformSchemaAST, I let the config abstract, if multiple transforms would be needed in the future. This function allows to transform the schema and the AST and return their transformed version.
```typescript
transformSchemaAST(schema: GraphQLSchema, config: { [key: string]: any })
```
Also the config flag was changed as requested.

TheOnlyBeardedBeast:
@dotansimha any news? Do you require any changes?

ardatan:
Thank you very much! Sorry it took so long to merge that :)

TheOnlyBeardedBeast:
@ardatan Sure, no problem, I understand, maintaining a project takes time. I am happy that it was merged 🙂

PR #4601, opened 2020-08-20
docs: remove $ from code blocks

ruisaraiva19:
This fixes the issue with the code blocks copy to clipboard functionality.

When a user clicks on the `Copy` button, it will copy the bash command prefixed with `$`.

**Currently:**
CleanShot 2020-08-20 at 23 12 18

**This PR:**
CleanShot 2020-08-20 at 23 15 44


dotansimha:
Thank you @ruisaraiva19 ! This is great! 😍

PR #4613, opened 2020-08-23
[typescript-vue] Update composition api changes

angad777:
createComponent has been changed to defineComponent


dotansimha:
Thank you @angad777 ! 
Can you please explain what versions of Vue has changed? because this seems like a breaking change, so I would like to consider to bump `major` version for that plugin release.

angad777:
> Thank you @angad777 !
> Can you please explain what versions of Vue has changed? because this seems like a breaking change, so I would like to consider to bump `major` version for that plugin release.

Welcome mate @dotansimha 

The change for `createComponent` to `defineComponent` was introduced in version `@vue/composition-api": "^0.5.0"`

FYI : this is only if people have opted in to using the new vue composition api.

Thanks

dotansimha:
I see, so we need a major bump. Can you please create a changeset that matches it? (run `yarn changeset` and it will guide you through the rest of the flow)

angad777:
> I see, so we need a major bump. Can you please create a changeset that matches it? (run `yarn changeset` and it will guide you through the rest of the flow)

Sure, no worries mate. I've added changeset. Thanks

dotansimha:
Thank you @angad777 !

Issue #4623, opened 2020-08-25
Can we make flow resolver plugin support es6 classes?


gyDBD:
graphql-js documents that we can use ES6 classes to represent GraphQL types and add resolver methods on.
https://graphql.org/graphql-js/object-types/
Our resolvers are something like this:  (example taken from the docs linked above)
```
// Construct a schema, using GraphQL schema language
var schema = buildSchema(`
  type RandomDie {
    numSides: Int!
    rollOnce: Int!
    roll(numRolls: Int!): [Int]
  }

  type Query {
    getDie(numSides: Int): RandomDie
  }
`);

// This class implements the RandomDie GraphQL type
class RandomDie {
  constructor(numSides) {
    this.numSides = numSides;
  }

  rollOnce() {
    return 1 + Math.floor(Math.random() * this.numSides);
  }

  roll({numRolls}) {
    var output = [];
    for (var i = 0; i < numRolls; i++) {
      output.push(this.rollOnce());
    }
    return output;
  }
}

// The root provides the top-level API endpoints
var root = {
  getDie: function ({numSides}) {
    return new RandomDie(numSides || 6);
  }
}
```
And we tried to use `graphql-codegen` to auto generated flow type for our resolvers. 
Here's the codegen.yml looks like:
```
overwrite: true
schema: "schema.graphql"
documents: null
generates:
    src/generated/types.js:
        plugins:
            - "flow"
            - "flow-resolvers"
        config:
            contextType: Context
```
[...]

PR #4636, opened 2020-08-25
add support for resolvers written with es6 classes

gyDBD:
Improvement for this issue: https://github.com/dotansimha/graphql-code-generator/issues/4623
This pr include these changes:
Add two new config for flow-resovlers plugin:
1. `customResolverFn:string` let people to custom the Resolver Function. (so that we can change `type Resolver` into 3 arguments(args, context, info), because ES6 class resolvers methods only contain 3 argument . see https://github.com/apollographql/apollo-server/issues/1996)
2. `resolverReadOnly:boolean` resolver functions can be read only. (Flow provide [read-only ](https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only)vs writable properties under interface)  

Add a new config under base resolvers:
`resolverDeclarationKind:DeclarationKind` let people to define the resolver object type. (so that we can change resolvers output from `type` to `interface`, cause to flow type a ES6 class, we need interface. see https://flow.org/en/docs/types/interfaces/) 
Note: My change won't affect any existing plugin and its config behavior. 
## Testing Done: 
created unit test and it passed.
yarn build, yarn test passed. 
Also, linked the local package to my local graphql service and tested, got expected results.

### DIFF
--- a/packages/plugins/flow/resolvers/src/index.ts
+++ b/packages/plugins/flow/resolvers/src/index.ts
@@ -1,8 +1,8 @@
 import { printSchemaWithDirectives } from '@graphql-tools/utils';
-import { RawResolversConfig } from '@graphql-codegen/visitor-plugin-common';
 import { Types, PluginFunction, addFederationReferencesToSchema } from '@graphql-codegen/plugin-helpers';
 import { parse, printSchema, visit, GraphQLSchema } from 'graphql';
 import { FlowResolversVisitor } from './visitor';
+import { FlowResolversPluginConfig } from './config';
 
 /**
  * @description This plugin generates resolvers signature based on your `GraphQLSchema`.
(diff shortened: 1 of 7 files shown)


gyDBD:
Hi @dotansimha, as we discussed in https://github.com/dotansimha/graphql-code-generator/issues/4623, I would like to add a Flow config to support resolvers written with ES6 classes. Not sure how to assign reviewers here, could you take a look when you have time? Or do you know whom else I should reach out to? Thanks!

magicmark:
@gyDBD This is an awesome initiative, thanks for starting this, this is looking very exciting!!!

I'm wondering if this could be configured on a per type basis - similar to mappers? I'm wondering what happens for when a codebase has a mix of ES6 classes and object literals. (I could imagine using object literals for smaller, less complex types)

packages/plugins/flow/resolvers/src/index.ts
```diff
@@ -46,12 +46,21 @@ export const plugin: PluginFunction<RawFlowResolversConfig, Types.ComplexPluginO
...
 
-  const header = `export type Resolver<Result, Parent = {}, Context = {}, Args = {}> = (
+  // ES6 class resolvers methods only contain 3 argument (args, context, info). see https://github.com/apollographql/apollo-server/issues/1996
+  const resolverType = config.supportES6Classes
+    ? `export type Resolver<Result, Parent = {}, Context = {}, Args = {}> = (
+  args: Args,
+  context: Context,
+  info: GraphQLResolveInfo
+) => Promise<Result> | Result;`
+    : `export type Resolver<Result, Parent = {}, Context = {}, Args = {}> = (
```
magicmark, line 56: this is a lot for my eyes to process as a single ternary block

I'll gingerly suggest splitting this up, and using a proper if statement? this would add some whitespace to break this up, and may improve the readability

gyDBD: Thanks! fixed it, have much less code to read now.

dotansimha:
Thank you @gyDBD and @magicmark ! 
I'll review this during this week :)

gyDBD:
@dotansimha Hi Dotan, would you mind take a look at this pr when you have time? Thanks!

gyDBD:
no longer need this feature

PR #4643, opened 2020-08-26
Fixes for plugin `typescript-compatibility`

dotansimha:
- [x] Fix issues with `__typename` (Remove `RequiredField` and use optional/non-optional typename for discrimination)
- [x] Replace `[0]` with `[number]`.
- [x] Fix issues with nested arrays.

### DIFF
--- a/.changeset/ninety-trees-wave.md
+++ b/.changeset/ninety-trees-wave.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-compatibility': major
+---
+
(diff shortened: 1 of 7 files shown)


devinjameson:
Can some please clarify why the types of things that seem to be arrays are instead typed as individual elements within that array? For example:

`export type Comments = NonNullable<NonNullable<NonNullable<CommentQuery['entry']>['comments']>[number]>;`

I would expect this to be:

`export type Comments = NonNullable<NonNullable<NonNullable<CommentQuery['entry']>['comments']>[]>;`

based on the pluralization of "Comments"`.

Issue #4653, opened 2020-08-27
Type 'void' is not assignable to type 'object'.Vetur(2769)


DarkLite1:
**Describe the bug**
I don't know if this issue is related to the graphql-codegenerator, Vetur, typescript, @vue/appolo-composable or me. So sorry if this is the wrong place to ask. I don't seem to be the only one experiencing this issue. You can find other users on [Quasar Framework Discord](https://discord.com/channels/415874313728688138/596276596319453207).

In our Vue project we try to consume the generated code from `useAllaccountsQuery` like described in the docs. The query functions fine and is working as expected but Vetur shows an error message and tab completion (iintellisense) is not working. 

```typescript
const { result, loading } = useAllAccountsQuery()
const allAccounts = useResult(result) 
```
The return type of `allAccounts` is:
```
const allAccounts: Readonly<Ref<void | Readonly<AllAccountsQuery>>>
```
But for one reason or another we get the following error message from Vetur and also no intellisense when trying to access properties of `account`. Although the Apollo plugin is setup correcty

![image](https://user-images.githubusercontent.com/10828570/91407901-44073f80-e843-11ea-9191-033e90b71cf6.png)

When we try this:
```typescript
const { result, loading } = useAllAccountsQuery()
const allAccounts = <Readonly<AllAccountsQuery>>(<unknown>useResult(result))
```
We get the following:
![image](https://user-images.githubusercontent.com/10828570/91409006-128f7380-e845-11ea-8b0c-b1b789e0cae1.png)

**To Reproduce**

<!--
#############
#############
🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 

PLEASE READ THE FOLLOWING: 
You can use this template to create a live sandbox of the issue: 
https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template
Make sure sure fork this template, and run `yarn generate` in the terminal. 
Please make sure the Codegen and plugins version under `package.json` matches yours. 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #4660, opened 2020-08-28
Added an extra example with useResult

DarkLite1:
After some struggle to understand it better I opened an issue here and I wanted to share the solution with others.
https://github.com/dotansimha/graphql-code-generator/issues/4653


dotansimha:
That's perfect. Thank you @DarkLite1 !

PR #4682, opened 2020-09-02
Improve support for unnamed queries

chriskrycho:
Previously, `ClientSideBaseVisitor#OperationDefinition` aborted eagerly if the node name was not set, since the `getName` function invoked from `BaseVisitor#getOperationSuffix` did not correctly handle scenarios where the node name was unset. This caused the `typed-document-node` plugin not to be able to generate a `DocumentNode` export for unnamed queries.

Related dotansimha/graphql-typed-document-node#40

### DIFF
--- a/.changeset/late-spoons-vanish.md
+++ b/.changeset/late-spoons-vanish.md
@@ -1,0 +1,8 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+'@graphql-codegen/typescript-react-apollo': patch
+'@graphql-codegen/typescript-react-offix': patch
+'@graphql-codegen/typescript-urql': patch
+---
+
+Support unnamed queries in operation visitors
--- a/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
@@ -222,7 +222,7 @@ export class BaseVisitor<TRawConfig extends RawConfig = RawConfig, TPluginConfig
     operationType: string
   ): string {
     const { omitOperationSuffix = false, dedupeOperationSuffix = false } = this.config as { [key: string]: any };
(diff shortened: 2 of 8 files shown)


chriskrycho:
I'd really love to write a test for this. I do not present see where or how I should add that—please point me to it and I will!

dotansimha:
@chriskrycho thank you! 

Regarding tests: please make sure to add some, just to make sure we didn't break anything. Usually, changes made in the plugins-common packages are tested as functions, and if you wish to have a complete test flow, you can use the typescript plugins and add plugins there, it will reflect your changes easily there (similar to integration tests).

Also, can you please add a changeset? (Run` yarn changeset`, it will guide you) Do you think there is anything breaking in your PR? or is it just a bug fix?

chriskrycho:
@dotansimha thanks for getting back to me. This is basically just a bug fix. I'll see if I can't get a test added in the next couple hours here!

While I'm at it, we would *mildly prefer* that these kinds of anonymous queries end named something like `FooDocument` when they're in a module locally named `Foo`, but can work with it if the result is simply `Document`. My initial exploration suggested that making the module information available all the way through to the relevant places would be a pretty significant change—does that match your understanding?

chriskrycho:
@dotansimha updated with a new test that does the end-to-end work to handle this along with a changeset!

chriskrycho:
Friendly ping to take another look at this, @dotansimha!

dotansimha:
Thank you @chriskrycho , this looks good :)

chriskrycho:
Thanks, @dotansimha!

chriskrycho:
@dotansimha once I tried to integrate this into my app, I caught a related issue: the `TypedDocumentNode` support doesn't correctly understand that it needs to to use the type names (as generated by the base visitor) through to the `Document` type. Can you point me in the right direction for what I need to do in *that* plugin to make it do the right thing?

PR #4433, opened 2020-07-26
[WIP] Use Yarn v2

dotansimha:
Checklist to get this merged:
- [x] Fix issues related to strict dependencies
- [x] Fix and support PnP 
- [x] Fix repo setup  
- [ ] Wait for https://github.com/renovatebot/renovate/issues/5230
- [ ] Fix release flow and maybe use https://yarnpkg.com/features/release-workflow  (blocked; needs *working* support for `--force` or support for marking all packages as "changed")

PR #4688, opened 2020-09-03
fix(cli): import relative to user config

merceyz:

`@graphql-codegen/cli` is trying to require user provided packages from itself making it rely on hoisting which does not guarantee that the packages are placed in an accessible location and when using PnP the resolution is rejected because `@graphql-codegen/cli` doesn't declare them.

Fixes https://github.com/yarnpkg/berry/issues/1599#issuecomment-686629356 cc @xenoterracide

**How did you fix it?**

Use `createRequire` to resolve packages relative to the user config

### DIFF
--- a/packages/graphql-codegen-cli/src/codegen.ts
+++ b/packages/graphql-codegen-cli/src/codegen.ts
@@ -15,8 +15,21 @@ import { getPresetByName } from './presets';
 import { debugLog } from './utils/debugging';
 import { printSchemaWithDirectives } from '@graphql-tools/utils';
 import { CodegenContext, ensureContext } from './config';
+import path from 'path';
+// eslint-disable-next-line
+import { createRequire, createRequireFromPath } from 'module';
 
-export const defaultLoader = (mod: string) => import(mod);
+const makeDefaultLoader = (from: string) => {
+  if (!path.extname(from)) {
+    from = path.join(from, '__fake.js');
+  }
+
+  const relativeRequire = (createRequire || createRequireFromPath)(from);
+
+  return (mod: string) => {
+    return import(relativeRequire.resolve(mod));
+  };
+};
 
 export async function executeCodegen(input: CodegenContext | Types.Config): Promise<Types.FileOutput[]> {
   function wrapTask(task: () => void | Promise<void>, source: string) {
@@ -73,8 +86,9 @@ export async function executeCodegen(input: CodegenContext | Types.Config): Prom
   async function normalize() {
     /* Load Require extensions */
     const requireExtensions = normalizeInstanceOrArray<string>(config.require);
+    const loader = makeDefaultLoader(context.cwd);
     for (const mod of requireExtensions) {
-      await import(mod);
+      await loader(mod);
     }
 
     /* Root plugin  config */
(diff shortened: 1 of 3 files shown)


merceyz:
@dotansimha I see you have a partial fix for this in https://github.com/dotansimha/graphql-code-generator/pull/4433 but it's using the PnP API directly, but that wont fix it for workspaces using `node_modules` so this should be preferred

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -16,7 +16,7 @@ import { debugLog } from './utils/debugging';
 import { printSchemaWithDirectives } from '@graphql-tools/utils';
 import { CodegenContext, ensureContext } from './config';
 
-export const defaultLoader = (mod: string) => import(mod);
+const makeDefaultLoader = (from: string) => (mod: string) => import(require.resolve(mod, { paths: [from] }));
```
merceyz, line 19: Removed the export since it's only used in this file

packages/graphql-codegen-cli/src/presets.ts
```diff
@@ -1,14 +1,12 @@
 import { DetailedError, Types } from '@graphql-codegen/plugin-helpers';
-import { resolve } from 'path';
 
 export async function getPresetByName(
   name: string,
   loader: Types.PackageLoaderFn<{ preset?: Types.OutputPreset; default?: Types.OutputPreset }>
 ): Promise<Types.OutputPreset> {
   const possibleNames = [`@graphql-codegen/${name}`, `@graphql-codegen/${name}-preset`, name];
-  const possibleModules = possibleNames.concat(resolve(process.cwd(), name));
```
merceyz, line 9: The default loader loads relative to the user config and any custom loader should handle this themselves, though perhaps the loader should also be given the config file location

dotansimha:
That's great! Thank you so much @merceyz ! 
You are right, I was trying to solve that partially in the Yarn v2 branch, but you solution should fix that in a better way. 
Can you please take a look at the CI? It seems to be broken.

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -15,8 +15,21 @@ import { getPresetByName } from './presets';
...
+// eslint-disable-next-line
+import { createRequire, createRequireFromPath } from 'module';
 
-export const defaultLoader = (mod: string) => import(mod);
+const makeDefaultLoader = (from: string) => {
+  if (!path.extname(from)) {
+    from = path.join(from, '__fake.js');
+  }
+
+  const relativeRequire = (createRequire || createRequireFromPath)(from);
```
merceyz, line 27: What node version is `@graphql-codegen/cli` supposed to support?

merceyz (commented):
> Can you please take a look at the CI? It seems to be broken.

@dotansimha Finally got some time to take a look at it, fixed :+1:

packages/graphql-codegen-cli/src/codegen.ts
dotansimha: Node>10 is great, so I think we are fine :)

dotansimha:
Thank you so much @merceyz , this is great :)

n1ru4l:
Hey @merceyz, I am investigating native ESM support right now, and this change prevents the loading of ESM modules. Do you have any recommendations for handling this situation for both ESM and CommonJS?

The `relativeRequire.resolve` call will always resolve the CommonJS module path. So to me, it seems like we need to detect within `codegen.ts` whether we are within an ESM or CommonJS context.

merceyz:
I suppose you could try `import.meta.resolve` and if that fails fall back to `relativeRequire.resolve`.
https://nodejs.org/docs/latest-v16.x/api/esm.html#importmetaresolvespecifier-parent.

Issue #4719, opened 2020-09-08
typescript-apollo-angular ignores importOperationTypesFrom


jbarrus:
**Describe the bug**
https://graphql-code-generator.com/docs/plugins/typescript-apollo-angular lists `importOperationTypesFrom` as a valid config but it is ignored

**To Reproduce**

```
schema: "schema.graphql"
generates:
  types.ts:
    plugins:
      - "typescript"
  operations.ts:
    documents: "document.graphql"
    config:
      documentMode: graphQLTag
    presetConfig:
      baseTypesPath: types.ts
    plugins:
      - "typescript-operations"
      - "typed-document-node"
  angular.ts:
    documents: "document.graphql"
    config:
      documentMode: external
      importOperationTypesFrom: "Operations"
      importDocumentNodeExternallyFrom: "./document.generated"
    plugins:
      - "typescript-apollo-angular"
```

https://codesandbox.io/s/hungry-saha-4t2pq?file=/angular.ts

**Additional Context**
I'm trying to generate types and operations in one lib and angular services in another to segregate the angular dependency (i.e. I want to be able to import operations from non-angular code and not pull in angular as a dependency). The examples for this are somewhat limited, so I may be doing this wrong, but I think I have the right config but the "importOperationTypesFrom" doesn't seem to be implemented in the typescript-apollo-angular plugin.
https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/apollo-angular/src/visitor.ts#L263

As a note, this does appear to be implemented in the typescript-apollo-react plugin:
https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/react-apollo/src/visitor.ts#L442

jbarrus:
Sorry to pollute this issue, I can submit another, but I'm having another problem with this approach as well - when documentMode is set to external for the angular.ts output, the document names do not match in angular.ts.

angular.ts example service when generated in the same file as operations
```
export const UserDocument = gql`
    query user {
  user(id: 1) {
    id
    username
    email
  }
}
    `;

  @Injectable({
    providedIn: 'root'
  })
  export class UserGQL extends Apollo.Query<UserQuery, UserQueryVariables> {
    document = UserDocument;
    
    constructor(apollo: Apollo.Apollo) {
      super(apollo);
    }
  }
```

when imported from a separate file using documentMode external
```
  @Injectable({
    providedIn: 'root'
  })
[...]

PR #4765, opened 2020-09-18
modify apollo angular visitor to allow external operation definitions related to #4719

jbarrus:
This is a work in progress. I'm trying to get my types, operations, and angular services to all be in separate files (with angular generated in a separate lib within my project). I couldn't get this working without these code changes. #4719 is related, but more was required.

If there is a better way, great. If not and this works then I'll complete the tests and whatever is needed for the change.

Example codegen.yml
```yml
schema: "schema.graphql"
generates:
  lib/generated/src/lib/types.ts:
    plugins:
      - "typescript"
  libs/generated/src/lib/:
    preset: near-operation-file
    documents: "libs/generated/**/*.graphql"
    config:
      documentMode: graphQLTag
    presetConfig:
      baseTypesPath: types.ts
    plugins:
      - "typescript-operations"
      - "typed-document-node"
  libs/generated-angular/src/lib/graphql-services.ts:
    documents: "libs/generated/**/*.graphql"
    config:
      documentMode: external
      importOperationTypesFrom: "Operations"
      importDocumentNodeExternallyFrom: "@myproject/generated"
    plugins:
      - "typescript-apollo-angular"
```

### DIFF
--- a/.changeset/four-ants-lick.md
+++ b/.changeset/four-ants-lick.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-apollo-angular': minor
+---
+
+add support for importing operations from external file in angular-apollo plugin
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -443,6 +443,11 @@ export class ClientSideBaseVisitor<
               `import * as Operations from './${this.clearExtension(basename(this._documents[0].location))}';`
             );
           } else {
+            if (!this.config.importDocumentNodeExternallyFrom) {
+              // eslint-disable-next-line no-console
+              console.warn('importDocumentNodeExternallyFrom must be provided if documentMode=external');
+            }
+
             this._imports.add(
               `import * as Operations from '${this.clearExtension(this.config.importDocumentNodeExternallyFrom)}';`
             );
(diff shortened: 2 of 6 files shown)


dotansimha:
Thank you @jbarrus ! 
You are right, it's currently not fully supported. 
Can you please add a changeset to your PR (Run `yarn changeset` and it will guide you). Also - can you please add tests and fix CI for your changes?

dandean:
Hey there – I would love to see this patch land. Anything I can do to help out?

jbarrus:
I'm a little fuzzy on how the options all work together, but it seems it will only work correctly if all three options are set - 
```
documentMode: external
importOperationTypesFrom: "Operations"
importDocumentNodeExternallyFrom: "@myproject/generated"
```

Does it make sense to throw an error if `importOperationTypesFrom` is set and the other two are not?

dotansimha:
> importOperationTypesFrom

I'm not sure actually, maybe just print a warning or something? Or we can make sure the docs are simpler and easier to understand. 
The `importOperationTypesFrom` configured where the types are coming from, that means it effects the TS types only, that are coming from `typescript`. 
While `documentMode: external` + `importDocumentNodeExternallyFrom` effects the pre-compiled GraphQL operations (`DocumentNode`) behaviour, and this once comes from typescript-apollo-angular.

Maybe we can warn if `importOperationTypesFrom` is set, and `documentMode === external`, and `importDocumentNodeExternallyFrom` is missing?

jbarrus:
I added some warnings and some tests. It seem all three of these options must be used in tandem if `near-file-operations` is not used. The `console.warn` requires setting an eslint ignore - is there a different recommendation?

dotansimha:
> I added some warnings and some tests. It seem all three of these options must be used in tandem if `near-file-operations` is not used. The `console.warn` requires setting an eslint ignore - is there a different recommendation?

This is great :) We don't have a logger standard so using console.log+eslint-ignore in this case is great. 
Thank you so much, merging :)

Issue #4766, opened 2020-09-20
Better support for loading plugins when using Yarn v2


mrtnbroder:
**Describe the bug**
 

When trying to use external plugins like `graphql-codegen-apollo-next-ssr` with yarn v2, the cli is trying to look for ` @graphql-codegen/graphql-codegen-apollo-next-ssr` which causes an error with yarn v2.

```
 Unable to load template plugin matching 'graphql-codegen-apollo-next-ssr'.
                  Reason:
                    @graphql-codegen/cli tried to access @graphql-codegen/graphql-codegen-apollo-next-ssr, but
 it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

    Required package: @graphql-codegen/graphql-codegen-apollo-next-ssr (via "@graphql-codegen/graphql-codegen-
apollo-next-ssr")
    Required by: @graphql-codegen/cli@virtual:2e4a546c98da4032db6b07cf4e91653b3f8547d84f83ed1956838abce54d5b79
47b8134ef943904cc7e53db132963a47bc02791f768239a1502cedcc9d502dd1#npm:1.17.8 (via
/usr/local/app/.yarn/$$virtual/@graphql-codegen-cli-virtual-16408af03c/0/cache/@graphql-codegen-cli-npm-1.17.8
-a51601b2d4-5d03c4b055.zip/node_modules/@graphql-codegen/cli/)
```

**To Reproduce**

upgrade to yarn v2, use third-party plugin that is not starting with `@graphql-codegen/*` and get the error.

One way to fix this issue would be to not try to prefix plugin names with `@graphql-codegen/` and take the name of the plugin in the `codegen.yml` as-is.

From:
```yml
graphql/schema.ts:
    plugins:
      - typescript
      - typescript-operations
```

To:
```yml
graphql/schema.ts:
    plugins:
      - @graphql-codegen/typescript
      - @graphql-codegen/typescript-operations
```

dotansimha:
@mrtnbroder can you please try latest version of the cli/core? we fixed the issues related to Yarn v2 compatibility. (also, please refer to my comment on your PR. Thanks!)

mrtnbroder:
Hi @dotansimha 
Thanks, I just tried it out. I am on `1.17.9` for `@graphql-codegen/cli`, but I am still getting the same error for third-party plugins.

```
  ✔ Parse configuration
  ❯ Generate outputs
[...]

PR #4767, opened 2020-09-20
fix: try to resolve by name first

mrtnbroder:
fixes an issue with yarn v2 #4766


dotansimha:
Thank you @mrtnbroder , but I think it might cause issues. 
Mainly because we have some plugins, such as `typescript` that might conflicts with that change (it will load `typescript` instead of the codegen plugin with the same name). I don't want to introduce this kind of breaking change at the moment. 

Does this helps? https://github.com/dotansimha/graphql-code-generator/pull/4688 I'm not really sure.

mrtnbroder:
#4688 sounds like the correct solution, I was in a rush 👍

PR #4790, opened 2020-09-24
Make export types configurable in typescript-graphql-files-modules

balazsorban44:
I recently started migrating from `@apollo/client` to `react-query`, and instead of using a package like `graphql-request`, I am planning to simply use a `fetch` wrapper similar to this: https://github.com/tannerlinsley/react-query/discussions/301#discussioncomment-74886

For that, I would like to generate types for named exports from my `.gql` files. The plugin
`@graphql-codegen/typescript-graphql-files-modules` almost makes it possible, but it can only output `DocumentNode` types. I added a config option, to make it possible to use strings as types.

The feature can be expanded/generalized if there is demand for it. I am open for discussion.

Here is a link where I try to explain the problem a bit better:
https://stackoverflow.com/questions/64050929/named-export-multiple-graphql-operation-from-gql-files

A link to a GitHub gist with code example:
https://gist.github.com/balazsorban44/352d5295f604fb4274bfc934937737a1

### DIFF
--- a/packages/plugins/typescript/graphql-files-modules/src/index.ts
+++ b/packages/plugins/typescript/graphql-files-modules/src/index.ts
@@ -3,7 +3,7 @@ import { Types, PluginFunction, PluginValidateFn } from '@graphql-codegen/plugin
 import { GraphQLSchema, OperationDefinitionNode } from 'graphql';
 
 /**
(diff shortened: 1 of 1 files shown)


dotansimha:
Great, thank you @balazsorban44 !

PR #4796, opened 2020-09-27
Typescript: Make non nullable input field with default value optional

### LINKED ISSUE #4273: [typescript] non nullable graphql inputs with default values are not optional as ts types

ArnaudBarre:
Related #4273

Similar to #2018, but for inputs. The fixes didn't apply to inputs because input don't use the argument transformer. Maybe it should be the real fix, but that will impact all plugins and I don't know the codebase enough to try it!

### DIFF
--- a/.changeset/tender-steaks-do.md
+++ b/.changeset/tender-steaks-do.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript': patch
+---
+
+Make non nullable input field with default value optional
--- a/packages/plugins/typescript/typescript/src/visitor.ts
+++ b/packages/plugins/typescript/typescript/src/visitor.ts
@@ -153,7 +153,8 @@ export class TsVisitor<
   InputValueDefinition(node: InputValueDefinitionNode, key?: number | string, parent?: any): string {
     const originalFieldNode = parent[key] as FieldDefinitionNode;
     const addOptionalSign =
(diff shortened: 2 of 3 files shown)


dotansimha:
Great catch, thank you so much @ArnaudBarre !

PR #4818, opened 2020-09-29
fix(cli): handle missing dependencies correctly

merceyz:

When the CLI is trying to require dependencies dynamically it checks for module not found errors by scanning the error message to see if a package is missing, instead of checking the error code.

Related https://github.com/dotansimha/graphql-code-generator/issues/4766

**How did you fix it?**

Check `code !== 'MODULE_NOT_FOUND'` instead of searching the error message

### DIFF
--- a/.changeset/twenty-dingos-remain.md
+++ b/.changeset/twenty-dingos-remain.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': patch
+---
+
+Check the error code instead of the error message to determine if a package wasn't found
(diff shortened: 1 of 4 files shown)


mrtnbroder:
Just tried this out @merceyz , but this doesnt fix the issue. Yarn throws before it reaches that point.

merceyz:
Interesting, do you have a repro I can test it on?

mrtnbroder:
@merceyz here you go: https://github.com/mrtnbroder/graphql-codegen-yarn-error

merceyz:
Your patch checks `err.code === "MODULE_NOT_FOUND"` but it should check `err.code !== "MODULE_NOT_FOUND"` https://github.com/mrtnbroder/graphql-codegen-yarn-error/blob/69fb4ca40ee5ac6d0e3f8fbd9d0aea01598cdc05/patches/%40graphql-codegen/cli%401.17.9.diff#L10

mrtnbroder:
@merceyz whoops. That actually fixed it. My bad! 😅  good to go with this PR 👍

Issue #4820, opened 2020-09-29
Data class generation broken in @graphql-codegen/kotlin since v.1.10.0


danielw93:
The merge Request #3156 broke generation of Kotlin data classes.

**To Reproduce**
* Define an input `TestInput` with a String property in schema.graphqls
* Run graphql-codegen
* Try to instantiate the `TestInput` data class inside a DataFetcher by calling `val input = TestInput(env.arguments)
** This fails with a compiler error: 

> Type mismatch.
> Required: String
> Found: (Mutable)Map<String!, Any!>!

Solution: 
The change from #3156 should be reverted as it can not work. 
I am not sure what the author of the PR meant when they said:"In data class in Kotlin, usually constructors are not needed.". This is simply not the case.

1. My GraphQL schema:

```graphql
type Test {
    content: String!
}

input TestInput {
    content: String!
}
```

2. Inside Kotlin Fetcher:

```kotlin
 val setTest: DataFetcher<Test> = DataFetcher { env ->
    val content = TestInput(env.arguments).content // <-- this can not compile
    // setter logic 
}
```

3. My `codegen.yml` config file:

```yml
schema: src/main/resources/graphql/schema.graphqls
generates:
  build/generated/source/graphql/main/com/test/test/graphql/generated/Types.kt:
    plugins:
      - kotlin
    config:
      className: Types
      withTypes: true
      package: com.test.test.graphql.generated
      listType: List
      namingConvention:
        enumValues: keep
```

**Expected behavior**
The call `TestInput(env.arguments)` in the kotlin fetcher code should be able to compile. 
This works perfectly in @graphql-codegen/kotlin v.1.9.1 but breaks in v.1.10.0 

**Environment:**

- OS: 20.04.1 LTS
- `@graphql-codegen/kotlin`: >= 1.10.0

danielw93:
Can I do anything more to help move this along? 
Maybe I can find another Kotlin developer that can take a look at this as well and see if my assessment of the situation is correct?

dotansimha:
Fixed in `@graphql-codegen/kotlin@1.17.9`

PR #4822, opened 2020-09-29
Revert "[kotlin] remove constructor method from generated code (#3156)"

danielw93:
This reverts commit b0582584a9beeedf72d416582d713d7099753906.

Fix for Issue #4820

### DIFF
--- a/packages/plugins/java/kotlin/src/visitor.ts
+++ b/packages/plugins/java/kotlin/src/visitor.ts
@@ -95,7 +95,7 @@ export class KotlinResolversVisitor extends BaseVisitor<KotlinResolversPluginRaw
 
     return `${comment}enum class ${enumName}(val label: String) {
 ${enumValues}
-
+        
   companion object {
     @JvmStatic
     fun valueOfLabel(label: String): ${enumName}? {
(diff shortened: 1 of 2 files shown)


dotansimha:
Reverting this is the only solution? I believe there should be other options to fix that. 
@omuomugin can you please take a look?

danielw93:
Hey, thanks for the quick reply!
I'm almost certain that the changes by @omuomugin could never have worked because they violate Kotlin's type system.
However I'm a bit intrigued that no one else seems to have run into these problems before, during the past half year...you'd think with >3000 weekly downloads for graphql-codegen/kotlin others would have reported this problem

danielw93:
For easier review, here is a Repl to illustrate the problem: https://repl.it/@xwalled/CodegenCompilerProblem

Lines 20-22 is what versions of graphql-codegen/kotlin prior to 1.10.0 would have created from the query definition in lines 9-11.
Line 29 is what the same query definition gets converted to, as of v.1.10.0. 

To see the desired behavior Run the Repl once (no output is shown, but we get no compiler errors).

To see the error behavior, comment out lines 20-22 and uncomment line 29. Then run the Repl again.
=> you'll get a compiler error from line 36, where we try to instantiate the `QueryItemArgs` 
     from the graphql arguments we received.
 > error: type mismatch: inferred type is Map<String, Any> but Long was expected

dotansimha:
Got it. Thank you @danielw93 :)

Issue #4815, opened 2020-09-28
Use non-polling for watch mode by default


mintchkin:
**Is your feature request related to a problem? Please describe.**

The current implementation of watch mode explicitly enables the [`usePolling` option of chokidar](https://github.com/dotansimha/graphql-code-generator/blob/80a1e6726c42e5a74d2717befb0834d94b515df4/packages/graphql-codegen-cli/src/utils/watcher.ts#L97), which has a much higher resting cpu footprint than the native filesystem watcher. Polling is technically supported in more cases, but the [documentation suggests](https://nodejs.org/api/fs.html#fs_availability) that native file watch is supported on most common platforms.

**Describe the solution you'd like**

Set `usePolling` to false by default (the default value in chokidar anyway), and either gracefully fall back on platforms that don't support it or allow users an option to override it if they need to.

**Describe alternatives you've considered**

You can override the the option in chokidar through [an environment variable named `CHOKIDAR_USEPOLLING`](https://github.com/paulmillr/chokidar#performance), but IMO it would be more ideal to optimize for the common case and potentially advertise that option to users who need polling for some reason (or support a more direct way to override it via the cli).

dotansimha:
Thank you @mintchkin , I think it could help a lot. Any chance you want to create a PR for this? :) I think we might be able to expose some of the chokidar configuration, and allow developers to override it if needed.

mintchkin:
@dotansimha opened #4823, let me know if that's in line with what you imagined

dotansimha:
Available in `@graphql-codegen/cli@1.18.0`

PR #4823, opened 2020-09-29
Add watch polling configuration

mintchkin:
ref: #4815 

- Change usePolling default from `true` to `false`
- Add `watchPolling` config file option to configure polling

Let me know if there's a better option for the config option or if it should be dropped in favor of a cli flag.

### DIFF
--- a/packages/graphql-codegen-cli/src/utils/watcher.ts
+++ b/packages/graphql-codegen-cli/src/utils/watcher.ts
@@ -88,19 +88,14 @@ export const createWatcher = (
         }
       });
 
-    let pollingInterval = 100;
-    if (typeof config.watchPolling === 'object') {
-      pollingInterval = config.watchPolling.interval;
-    }
-
     watcher = chokidar.watch(files, {
       persistent: true,
       ignoreInitial: true,
       followSymlinks: true,
       cwd: process.cwd(),
       disableGlobbing: false,
-      usePolling: !!config.watchPolling,
-      interval: pollingInterval,
+      usePolling: config.watchConfig?.usePolling,
+      interval: config.watchConfig?.interval,
       depth: 99,
       awaitWriteFinish: true,
       ignorePermissionErrors: false,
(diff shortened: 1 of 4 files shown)


packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -88,15 +88,19 @@ export const createWatcher = (
         }
       });
 
+    let pollingInterval = 100;
+    if (typeof config.watchPolling === 'object') {
```
dotansimha, line 92: Maybe we can have a function to parse the `config.watchConfig` object and apply defaults?

packages/utils/plugins-helpers/src/types.ts
```diff
@@ -414,6 +414,18 @@ export namespace Types {
...
      */
     watch?: boolean | string | string[];
+    /**
+     * @description Allows overriding the behavior of watch to use stat polling over native file watching support.
+     *
+     * You can specify a boolean to toggle polling on or off (default: false) or an interval option to configure polling.
+     *
+     * For more details: https://graphql-code-generator.com/docs/getting-started/development-workflow#watch-mode
+     */
+    watchPolling?:
```
dotansimha, line 424: Maybe it can be `watchConfig`? and then `{ pollingInterval: number, usePolling: boolean }` (or, same as it's defined in chokidar).

dotansimha (commented):
That's perfect, thank you @mintchkin ! 

Can you please add a changeset file? (run `yarn changeset` and it will guide you thought the process) 
I think this PR can bump the minor flag.

packages/graphql-codegen-cli/src/utils/watcher.ts
mintchkin: Since the config structure's now identical to chokidar's and I don't think there's any reason to maintain different default values for them, I just inlined the values where we create the watcher

dotansimha:
That's perfect. Thank you @mintchkin !

PR #4181, opened 2020-06-07
fix `Maybe` only refers to a type, but is being used as a value here

leejh3224:
Fix the issue related to the comment below.

**comment**
Thanks @borremosch @dotansimha!
Though if you have something like this (tested with 1.8.2-alpha-6ee6e064.78 ): 
```ts
type MyType{
  myProperty: [SomeOtherType]
}
```
Will result in:
```ts
@TypeGraphQL.Field(type => \[**Maybe**\<SomeOtherType>], { nullable: true })
myProperty!: Maybe<Array<Maybe\<SomeOtherType>>>;
```
So ts will say: 'Maybe' only refers to a type, but is being used as a value here.

_Originally posted by @groege in https://github.com/dotansimha/graphql-code-generator/pull/2177#issuecomment-544915163_

PR #4835, opened 2020-09-30
Fix type graphql nullable types

### LINKED ISSUE #4663: `Maybe` only refers to a type, but is being used as a value here (only arrays)
**Describe the bug**
When generating types for type-graphql, I got type of typescript, like: ```@TypeGraphQL.Field(type => [Maybe<CalendarEvent_Bool_Exp>])``` instead of  ```@TypeGraphQL.Field(type => [CalendarEvent_Bool_Exp])```

```typescript
@TypeGraphQL.InputType()
export class CalendarEvent_Bool_Exp {

  @TypeGraphQL.Field(type => [Maybe<CalendarEvent_Bool_Exp>], { nullable: true })
  _and!: Maybe<Array<Maybe<CalendarEvent_Bool_Exp>>>;
...
```
**To Reproduce**
link: [codesandbox](https://codesandbox.io/s/graphql-code-generator-type-graphql-maybe-error-in-arrays-j5uym
)

My GraphQL schema:

```graphql
type mutation_root {
  # delete data from the table: "calendarEvent"
  delete_calendarEvent(
    # filter the rows which have to be deleted
    where: calendarEvent_bool_exp!
  ): calendarEvent_mutation_response
}

type calendarEvent {
  id: Int!
  eventId: Int!
  hallId: Int
  isWorking: Boolean!
}

# response of any mutation on the table "calendarEvent"
type calendarEvent_mutation_response {
  # number of affected rows by the mutation
  affected_rows: Int!

  # data of the affected rows by the mutation
  returning: [calendarEvent!]!
}

# Boolean expression to filter rows from the table "calendarEvent". All fields are combined with a logical 'AND'.
input calendarEvent_bool_exp {
  _and: [calendarEvent_bool_exp]
  _not: calendarEvent_bool_exp
  _or: [calendarEvent_bool_exp]
}
...
```

My `codegen.yml` config file:

```yml
schema:
  - ./schema.graphql
generates:
  ./hasura.ts:
    plugins:
      - typescript-type-graphql
```

**Expected behavior**
expecting correct type mapping
like this:
```typescript
@TypeGraphQL.InputType()
export class CalendarEvent_Bool_Exp {

  @TypeGraphQL.Field(type => [CalendarEvent_Bool_Exp], { nullable: true })
  _and!: Maybe<Array<Maybe<CalendarEvent_Bool_Exp>>>;
...
```

**Environment:**

- OS: MacOS
- `@graphql-codegen/cli`: 1.17.8 
- `@graphql-codegen/typescript-type-graphql`: 1.17.8 
- NodeJS: 12.18.1

borremosch:
:warning: This pull request includes the changes in #4833. Unfortunately there was no other way to do it, as #4833 includes a refactored way of setting TypeGraphQL decorator options

This merge requests is an improvement over #4181 and fixes #4663. In GraphQL arrays, there are several nullable types:

- Items in the array may be null, (`[String]!`)
- The array itself may be null, (`[String!]`)
- Both the array and items in the array may be null (`[String]`)
- Neither the array nor the items in it may be null (`[String!]!]`)

So far, `@graphql-codegen/typescript-type-graphql` did not support either of the cases where the items in the array can be null. This led to uncompilable code, which #4181 attempted to fix. However, there were two problems with this merge request:

- It only fixed fields definitions, but not input value definitions, leaving the bug in #4663
- It removes the compilation error, but actually outputs the incorrect, non-nullable item type

TypeGraphQL supports nullable list types through the options `{ nullable: 'items' }`, and  `{ nullable: 'itemsAndList' }`, which this pull request will properly set on the field and input field decorators.

### DIFF
--- a/.changeset/moody-pots-move.md
+++ b/.changeset/moody-pots-move.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-type-graphql': patch
+---
+
+Fixed decorators for nullable type-graphql array types
--- a/.changeset/rude-scissors-wait.md
+++ b/.changeset/rude-scissors-wait.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-type-graphql': patch
+---
+
+GraphQL descriptions are now set in their corresponding TypeGraphQL decorator options
--- a/packages/plugins/typescript/type-graphql/src/visitor.ts
+++ b/packages/plugins/typescript/type-graphql/src/visitor.ts
@@ -1,9 +1,4 @@
-import {
-  transformComment,
-  indent,
-  DeclarationBlock,
-  AvoidOptionalsConfig,
-} from '@graphql-codegen/visitor-plugin-common';
+import { indent, DeclarationBlock, AvoidOptionalsConfig } from '@graphql-codegen/visitor-plugin-common';
 import { TypeGraphQLPluginConfig } from './config';
 import autoBind from 'auto-bind';
 import {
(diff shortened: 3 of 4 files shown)


borremosch:
Branch has become unmergable. Will create a new branch

Issue #4831, opened 2020-09-30
typescript-type-graphql plugin should have option to filter types to annotate


borremosch:
The @graphql-codegen/typescript-type-graphql plugin generates types in a very similar fashion to @graphql-codegen/typescript. However, it will annotate all types with TypeGraphQL decorators. The plugin currently annotates *all* types, which leads to two problems:

1. It easily leads to duplicate type definitions, which are prohibited in GraphQL. Especially when the number of interdependencies between several GraphQL services increases, there is a high risk of re-exposing the same types twice, immediately leading to non-compiling code.
2. As all annotated types are re-exposed by TypeGraphQL, types may be exposed externally when they are not supposed to.

**Describe the solution you'd like**
It would be nice if the typescript-type-graphql plugin would have a filter option that would add decorator only to the types that match the filter. When the filter option is not supplied it would work similar to how it works at the moment.

**Describe alternatives you've considered**
One of the alternatives to using graphql-codegen to pass types downstream would be to use Apollo Federation. However, Apollo Federation as of yet does not support subscriptions, which is an essential part of our infrastructure. Also, Apollo Federation is only meant for federation, while there are other conceivable situations where passing types downstream is desirable.

dotansimha:
Fixed in `@graphql-codegen/typescript-type-graphql@1.18.1` (thanks @borremosch )

PR #4836, opened 2020-09-30
Types that must be TypeGraphQL decorated can now be specified

borremosch:
Related #4831. New `decorateTypes` config option can now be used to specify types that should be decorated.


borremosch:
I am having some trouble with the generated types. I will try another approach and submit a new pull request

Issue #4216, opened 2020-06-12
GraphQLError: Unknown field '__type' on type 'Query' when using flattenGeneratedTypes


paales:
**Describe the bug**

**To Reproduce**

Create a GraphQL file that is valid GraphQL:
```
query FilterInputTypes {
  __type(name: "ProductAttributeFilterInput") {
    inputFields {
      name
      type {
        name
      }
    }
  }
}
```

See the error while generating:
https://codesandbox.io/s/distracted-yonath-hhbqi?file=/package.json

```
$ yarn codegen
$ graphql-codegen --config ./codegen.yml
  ✔ Parse configuration
  ❯ Generate outputs
    ❯ Generate types.ts
      ✔ Load GraphQL schemas
      ✔ Load GraphQL documents
      ✖ Generate
        → Unknown field '__type' on type 'Query'.


 Found 1 error

  ✖ document.graphql
    GraphQLError: Unknown field '__type' on type 'Query'.
        at createUserError (/sandbox/node_modules/relay-compiler/lib/core/CompilerError.js:35:1
0)
        at GraphQLDefinitionParser._transformField (/sandbox/node_modules/relay-compiler/lib/co
re/RelayParser.js:779:13)
        at /sandbox/node_modules/relay-compiler/lib/core/RelayParser.js:614:23
        at Array.map (<anonymous>)
        at GraphQLDefinitionParser._transformSelections (/sandbox/node_modules/relay-compiler/l
ib/core/RelayParser.js:610:36)
        at GraphQLDefinitionParser._transformOperation (/sandbox/node_modules/relay-compiler/li
b/core/RelayParser.js:584:27)
        at GraphQLDefinitionParser.transform (/sandbox/node_modules/relay-compiler/lib/core/Rel
ayParser.js:351:21)
        at parseDefinition (/sandbox/node_modules/relay-compiler/lib/core/RelayParser.js:327:17
)
        at /sandbox/node_modules/relay-compiler/lib/core/RelayParser.js:154:18
        at eachWithCombinedError (/sandbox/node_modules/relay-compiler/lib/core/CompilerError.j
s:93:9)
        at RelayParser.transform (/sandbox/node_modules/relay-compiler/lib/core/RelayParser.js:
151:5)
        at /sandbox/node_modules/relay-compiler/lib/core/RelayParser.js:92:19
        at Object.run (/sandbox/node_modules/relay-compiler/lib/core/GraphQLCompilerProfiler.js
:55:30)
[...]

PR #4839, opened 2020-09-30
#4216: remove broken isTypeOf call (always undefined in graphql-tools v6)

benmosher:
This change passes all tests (I ran `yarn test` locally) plus one new test, and "fixes" the TypeError from later comments in issue #4216.

"fixes" is in quotes because I honestly don't understand exactly what is happening here, so I am not sure what the possible impact might be. And I would guess that this line wasn't ever covered by tests to begin with -- I believe it is triggered by fragments against interfaces directly, possibly co-interacting with the union.

I did look at the `package.json` in the commit when this line was created, and it was using graphql-tools v4, so I am suspicious this may be a consequence of a breaking change in graphql-tools between v4 and v6.

I was unable to find anything replacing it or any upgrade notes in the graphql-tools migration guide.

I created a new test from a minimal repro derived from my actual code. It does reproduce the TypeError from #4216 if line 105 of `selection-set-to-object.ts` is reverted.

The resulting types look good to me but definitely would like more eyes on it!

Thanks much for this excellent tool.

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -102,7 +102,7 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
           this._appendToTypeMap(types, typeOnSchema.name, fields);
           this._appendToTypeMap(types, typeOnSchema.name, spreadsUsage[typeOnSchema.name]);
           this._collectInlineFragments(typeOnSchema, inlines, types);
-        } else if (isInterfaceType(typeOnSchema) && parentType.isTypeOf(typeOnSchema, null, null)) {
+        } else if (isInterfaceType(typeOnSchema)) {
           this._appendToTypeMap(types, parentType.name, fields);
           this._appendToTypeMap(types, parentType.name, spreadsUsage[parentType.name]);
           this._collectInlineFragments(typeOnSchema, inlines, types);
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -102,7 +102,7 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
           this._appendToTypeMap(types, typeOnSchema.name, fields);
           this._appendToTypeMap(types, typeOnSchema.name, spreadsUsage[typeOnSchema.name]);
           this._collectInlineFragments(typeOnSchema, inlines, types);
-        } else if (isInterfaceType(typeOnSchema)) {
+        } else if (isInterfaceType(typeOnSchema) && parentType.getInterfaces().includes(typeOnSchema)) {
           this._appendToTypeMap(types, parentType.name, fields);
           this._appendToTypeMap(types, parentType.name, spreadsUsage[parentType.name]);
           this._collectInlineFragments(typeOnSchema, inlines, types);
--- a/.changeset/chilled-trees-act.md
+++ b/.changeset/chilled-trees-act.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+---
+
+remove broken isTypeOf call for expanding fragments with flattenGeneratedTypes = true
--- a/packages/plugins/typescript/operations/tests/__snapshots__/ts-documents.spec.ts.snap
+++ b/packages/plugins/typescript/operations/tests/__snapshots__/ts-documents.spec.ts.snap
@@ -443,6 +443,59 @@ export type UserQueryQuery = (
 "
 `;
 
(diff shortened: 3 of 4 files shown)


benmosher:
FWIW, this also works:
```        } else if (isInterfaceType(typeOnSchema) && parentType.getInterfaces().includes(typeOnSchema)) {```
and it seems like maybe it has the same meaning as the original code? though I am not sure that this predicate is ever needed.

**Comment by dotansimha** [comment]
That's great, thank you @benmosher :) 
@n1ru4l can you please give it a look? I think you are the last person who worked on that piece of code... :)

File: packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts [hunk 1]
```diff
@@ -102,7 +102,7 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
           this._appendToTypeMap(types, typeOnSchema.name, fields);
           this._appendToTypeMap(types, typeOnSchema.name, spreadsUsage[typeOnSchema.name]);
           this._collectInlineFragments(typeOnSchema, inlines, types);
-        } else if (isInterfaceType(typeOnSchema) && parentType.isTypeOf(typeOnSchema, null, null)) {
```
n1ru4l, line 105: I think this `isTypeOf` check is there to ensure that only fragments that can be applied to the `parentType` are added to the selection set. AFAIK according to the GraphQL spec it is possible to apply fragments on a type where it does not make sense, so this check should prevent generating incorrect types.

But it also seems like the isTypeOf method is actually optional. https://github.com/graphql/graphql-js/blob/711425eaf3baa3efa5da10e11ed76f58ec149ad7/src/type/definition.js#L729

I think the correct way would be to use the getInterfaces method and check whether the typeOnSchema is included in that array. https://github.com/graphql/graphql-js/blob/711425eaf3baa3efa5da10e11ed76f58ec149ad7/src/type/definition.js#L762

@benmosher could you try that?

n1ru4l (changes requested):

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
n1ru4l: See as a reference: https://codesandbox.io/s/staging-brook-g1nrs?fontsize=14&hidenavigation=1&theme=dark

It spreads Avatar on the User type, however, execution does not complain 🤷

dotansimha:
@benmosher  let me know if you need any help with this PR. I really want to have it in our upcoming release!

benmosher:
Thanks for the feedback. Added back the check to the predicate as `parentType.getInterfaces().includes(typeOnSchema)` as discussed. Sorry for the lag, I missed the first few notifications!

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -102,7 +102,7 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
...
+        } else if (isInterfaceType(typeOnSchema) && parentType.getInterfaces().includes(typeOnSchema)) {
```
n1ru4l, line 105: @benmosher Could you please add a test case for the functionality? In general for an operation that spreads a fragment on a type where it does not belong.

Also since interfaces can implement interfaces, it seems as we would also have to use recursion as the getInterfaces function on parentType could also return an interface that implements another interface.

```graphql
interface A {
  a: String!
}

interface B implements A {
  a: String!
  b: String
}

type C implements B {
  a: String!
  b: String
  c: String!
}

type Query {
  foo: C
}
```

```graphql
query {
  foo {
    ... on A {
      a
    }
  }
}
```

benmosher: I will if I get a chance, but I'm not sure when that will be. also, is this syntax valid? I recently added an interface to another interface and got a warning in my editor that was essentially `Type 'C' must also implement A` and had to update my analogous concrete types to read as `type C implements B & A {`.

I believe the specification bears this out: https://github.com/graphql/graphql-spec/pull/373/files#diff-30a69c5a5eded8e1aea52e53dad1181e6ec8f549ca2c50570b035153e2de1c43R1090

n1ru4l: @benmosher That's fair! I wasn't aware of that detail.

I will see whether I can add a test for the spread tomorrow :)

dotansimha: I added this test. The output seems fine: `{ foo?: Maybe<{ __typename?: 'C' }> }`. GraphQL allow you to spread on incorrect types, so that's fine :) 
Also, did a rebase for this PR. Looks great and we can go ahead and merge it.

dotansimha:
@benmosher I couldn't push a rebase to your branch, i assume because it's a fork. No reason to keep this PR hanging. 
I created a new PR (with the original commits, just rebased and added the missing test): https://github.com/dotansimha/graphql-code-generator/pull/5039 

Thank you @benmosher and @n1ru4l for your work ❤️

PR #4840, opened 2020-09-30
Types that must be TypeGraphQL decorated can now be specified

### LINKED ISSUE #4831: typescript-type-graphql plugin should have option to filter types to annotate

borremosch:
Resolves #4831. New `decorateTypes` config option can now be used to specify types that should be decorated.

### DIFF
--- a/.changeset/nasty-squids-press.md
+++ b/.changeset/nasty-squids-press.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-type-graphql': minor
+---
+
+Types that must be decorated may now be specified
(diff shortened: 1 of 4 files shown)


borremosch:
Branch has become unmergable. Will create a new pull request

Issue #4859, opened 2020-10-05
Codegen fails when base folder name ends with something looking like a file extension


adambrgmn:
**Describe the bug**

This one really got me stumbling :D I kept getting an error saying `Unable to find template plugin matching 'typescript'` even though everything had been installed properly. And I have used this project before, without issues.

But after some searching and messing with the code I found the culprit – [`createRequire`](https://github.com/dotansimha/graphql-code-generator/blob/398b094b3f7ee6aa93db547a14bef88d39bbf15a/packages/graphql-codegen-cli/src/codegen.ts#L22).

Before the require is created it runs `path.extname(from)` to determine if `from` is pointing towards a file or a folder. But this check gives a false positive if the folder the project is located within ends with something similar to an extension name, e.g. `mysite.com`.

**To Reproduce**

1. Setup any project or revisit and earlier
2. Change the root folder name (the folder where `package.json` is located) to end with e.g. `.com` or `.whatever`
3. Get into the now renamed folder and run your codegen script (e.g. `yarn generate` or `npm run generate`)

The last step should fail complaining that the first plugin in the plugins array can't be found. Remove the extension name from the folder to once again make it work.

**Expected behavior**
The expected behavior is that this should not fail. Any kind of folder name should be accepted.

**Environment:**

- OS: MacOS
- `@graphql-codegen/cli`: 1.17.10
- NodeJS: 12.18.3

**Additional context**
I have found a solution that might work and it requires minimal changes to the `makeDefaultLoader` function:

```diff
const makeDefaultLoader = (from: string) => {
-  if (!path.extname(from)) {
+  if (fs.statSync(from).isDirectory()) {
    from = path.join(from, '__fake.js');
  }

  const relativeRequire = (createRequire || createRequireFromPath)(from);

  return (mod: string) => {
    return import(relativeRequire.resolve(mod));
  };
};
```

I would be happy to supply a PR for this if you like!

adambrgmn:
I took the liberty of submitting a PR #4860

mastorm:
Heres a minimal repro: https://github.com/mastorm/graphql-code-generator-error-repro

do yarn install && yarn gen in both folders

alexander9306:
> 
> 
> **Describe the bug**
> 
> This one really got me stumbling :D I kept getting an error saying `Unable to find template plugin matching 'typescript'` even though everything had been installed properly. And I have used this project before, without issues.
> 
[...]

PR #4860, opened 2020-10-05
Use stat instead of extname when creating require

adambrgmn:
This PR will make use of `fs.statSync` instead of `path.extname` when creating the custom `require` function. The issue with the previous version is discussed in #4859. But to summarize the previous version creates issues when the folder the script is running from has a name ending with something that might be mistaken for a file extension – for example `mysite.com` (`path.extname('mysite.com') === '.com'`).

One could argue if we even need to do the check since `makeDefaultLoader` is always passed `context.cwd` which I guess is always a directory? We could probably just do with `from = path.join(from, '__fake.js')`. Please let me know if you think I should remove the check completely.

Related #4859

### DIFF
--- a/.changeset/rotten-flies-sort.md
+++ b/.changeset/rotten-flies-sort.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': patch
+---
+
+Use `fs.statSync` when creating custom require instead of `path.extname`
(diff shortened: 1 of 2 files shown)


openscript:
Ran into the same issue and this PR fixed it.

dotansimha:
This is great. Thank you @adambrgmn !

openscript:
Would be great to have a release for this soon, so my colleagues don't need to relay on a local build of it.

PR #4869, opened 2020-10-07
Fix graphql request import plugin

Gigiz:


dotansimha:
Great catch, thank you @Gigiz !

Dechowmedia:
Hi @dotansimha sorry to be a bother :) 
When is this going to be released? :) - Had to update the generated SDK for now (which is doable) just would want to do this without having to invoke a patching codemod (as I won't be able to build otherwise)

dotansimha:
@Dechowmedia it's available now :P `@graphql-codegen/typescript-graphql-request@2.0.2`

PR #4881, opened 2020-10-10
docs: fix typo in import-types example

IlyaSemenov:
The current example is incorrect, the PR fixes that.

(Sorry about the EOL change creeping in, but that's odd that it wasn't there in the first place, as most sane editors always add it. Don't you want to enforce it? And perhaps adopt `.editorconfig`?)


dotansimha:
Thank you @IlyaSemenov ! 
I'm all in with adding `.editorconfig` if you think it's useful. Didn't have the chance to try it for real :)

PR #4882, opened 2020-10-10
Allow the vue-apollo plugin to write code to .tsx files

cdrx:
Allowing the vue-apollo plugin to write to .tsx files means that piglovesyou/graphql-let will work out of the box with the the `typescript-vue-apollo` plugin.

Without it writing to a `.tsx` file throws this error: **Plugin "vue-apollo" requires extension to be ".ts"!**


dotansimha:
That's great, thank you @cdrx !

PR #4927, opened 2020-10-17
feat(types-import): support useTypeImports

cevr:
This PR adds support for the `useTypeImports` config when using the `types-import` preset.


dotansimha:
Great! Thank you @cevr !

PR #4930, opened 2020-10-18
fix(vue-apollo): keep up with @vue/apollo-composable

aki77:
keep up with vuejs/vue-apollo@81475da

### DIFF
--- a/.changeset/shiny-rockets-judge.md
+++ b/.changeset/shiny-rockets-judge.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-vue-apollo': minor
+---
+
+keep up with vuejs/vue-apollo@81475da
--- a/dev-test/githunt/types.vueApollo.ts
+++ b/dev-test/githunt/types.vueApollo.ts
@@ -464,19 +464,11 @@ export const CurrentUserForProfileDocument = gql`
  */
 export function useCurrentUserForProfileQuery(
   options:
-    | VueApolloComposable.UseQueryOptions<CurrentUserForProfileQuery, CurrentUserForProfileQueryVariables>
-    | VueCompositionApi.Ref<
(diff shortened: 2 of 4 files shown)


dotansimha:
Thank you @aki77  :) Is that a breaking change in vue-apollo? Do we need to notify users that this change only supported from a specific version of vue-apollo?

aki77:
Thanks @dotansimha !

This is not a breaking change for users using vue-apollo alone.
But for users using @graphql-codegen/typescript-vue-apollo, this is a breaking change.
So I think we should notify them that @vue/apollo-composable@^4.0.0-alpha.12 is the target.

dotansimha:
Got it. I wonder if this needs to be merge right now, since `@vue/apollo-composable` is still alpha, and I don't think it's a good idea to bump major version for a codegen plugin, based on an alpha API that might change later.

aki77:
I'm fine with minor upgrades.
Because as a codegen plugin it doesn't introduce any breaking changes to the API.
To begin with, only the alpha version of @vue/apollo-composable exists.
The best option is to keep up with the latest version of @vue/apollo-composable.

nickreynke:
Good work. Would love to see this merged anytime soon.

I tried to work with Vue3 in combination with Apollo Client and AWS AppSync but got into trouble setting it up using the generated types from codegen.

quentinus95:
This would unlock the possibility to upgrade existing Vue apps that are using this codegen lib to Apollo Client 3. I think it's worth merging and releasing this ASAP.

P4sca1:
There is another PR which fixes this bug along with other improvements and fixes to the vue apollo plugin: https://github.com/dotansimha/graphql-code-generator/pull/5007

aki77:
@P4sca1 Good work!

@dotansimha I think the PR for #5007 should be merged.

dotansimha:
Great, thank you all for your amazing contributions. We'll get https://github.com/dotansimha/graphql-code-generator/pull/5007 merged now :)

PR #4931, opened 2020-10-18
Add errors-only flag to graphql-codegen-cli

maurobalbi:
When watching multiple processes in parallel, the output from the graphql-cli can be quite verbose. This PR allows the --errors-only flag to be passed to the cli which suppresses all other output. 

There are some things that might be questionable: for example the nested ternary, even though I think in this case it's actually quite clear to understand. However I'm happy for any feedback, since I'm quite new to the open source world.

### DIFF
--- a/.changeset/brown-eels-chew.md
+++ b/.changeset/brown-eels-chew.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/cli': minor
+'@graphql-codegen/plugin-helpers': minor
+---
+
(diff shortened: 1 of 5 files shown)


dotansimha:
Awesome work, thank you @maurobalbi ! 
Can you please add a changeset to this PR? (run `yarn changeset` and it will guide you) This PR should bump minor for cli package.

maurobalbi:
Ok, added. I also bumped the plugin-helpers since the types.ts file was also affected.

PR #4932, opened 2020-10-18
Fixed unquoted numeric enum identifiers

### LINKED ISSUE #4834: Enum members should be quoted if numeric
See the screenshot below. The problem is easily solved by quoting the problematic members.

image

borremosch:
Related #4834


dotansimha:
That's awesome, thank you so much @borremosch !

PR #4945, opened 2020-10-20
Fixes #4944, fragment import statement generation for flow

### LINKED ISSUE #4944: Fragment type imports are not generated when using Flow / Flow operations plugin
**Describe the bug**
👋 Firstly, thanks for building and maintaining this tool — I'm sure this saves a lot of time and effort for a lot of people. 

While trying out the `nearest-file-output` `preset`, I noticed that type imports for fragments weren't generated for query type files. It seems like this might only be an issue with Flow, as the flow plugin doesn't have the same steps for reading `fragmentImports` as the typescript one.

**To Reproduce**

See this minimal repro -
https://codesandbox.io/s/codegen-flow-bug-kvi73

When I run `yarn codegen`, the `src/index.generated.js` file uses fragments for building its type, but has a missing import. This makes the generated file invalid.

**Expected behavior**
Type imports for fragments are generated whenever a fragment is imported.

**Environment:**

- OS: Any
- `@graphql-codegen/...`: 1.18.0
- NodeJS: 12.3

pastelsky:
Related: #4944

The typescript visitors already do this — but looks like the flow implementations were missed out.

### DIFF
--- a/.changeset/hot-pants-vanish.md
+++ b/.changeset/hot-pants-vanish.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/flow-operations': patch
+---
+
+Fixes generation of type imports for fragments
--- a/packages/plugins/flow/operations/src/index.ts
+++ b/packages/plugins/flow/operations/src/index.ts
@@ -30,12 +30,14 @@ export const plugin: PluginFunction<FlowDocumentsPluginConfig> = (
     ...(config.externalFragments || []),
   ];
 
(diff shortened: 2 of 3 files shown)


dotansimha:
This is awesome, thank you @pastelsky . Can you please add a simple unit test for this? just to be sure the output is not creating new issues, and to be sure we are not breaking this in the future.

dotansimha:
Rebased, we can get this merged. Thank you @pastelsky !

PR #4960, opened 2020-10-21
Fixed DocumentNodeVariable

alexkharech:


dotansimha:
Hi @alexkharech , thank you for this PR. Can you please share more background about this issue? what does this fix? do we have an issue for tracking the progress on that?

alexkharech:
@dotansimha 
These are two commits to different plugins:
`graphql-request` - will add import type prefixes if used
`vue-apollo` - adds **documentVariableSuffix** and names of imported operations according to general parameters, it doesn't work in current version

This fixes compatibility issues with importing types from another module 
```yml
preset: import-types
presetConfig:
      typesPath: ./types
```

dotansimha:
Got it, thank you @alexkharech. 
Can you please separate this PR to two different PRs? we are trying to minimize PR and changes to the bare minimum, and in this case it seems like the changes to the 2 plugins are unrelated?

dotansimha:
@alexkharech I'm closing this one, let me know if it needs to be reopened (see my latest comment). Thanks!

PR #4966, opened 2020-10-22
as is types path for near-operation-file docs

PabloSzx:
I was configurating `graphql-codegen` for a monorepo project and I needed an absolute types path, but the docs didn't specify anything about using absolute paths, for example, for node_modules reference, and I found while looking at the source code about this undocumented feature

https://github.com/dotansimha/graphql-code-generator/blob/master/packages/presets/near-operation-file/src/index.ts#L148


dotansimha:
This is awesome, thank you @PabloSzx . It also closes this: https://github.com/dotansimha/graphql-code-generator/issues/4935

PR #4967, opened 2020-10-22
GenerateOptions: dont require config & documents

ludwigschubi:
These options aren't required in [this tutorial](https://graphql-code-generator.com/docs/getting-started/programmatic-usage) but they are in the code, is this correct? I get the respective ts-error when trying out the example

### DIFF
--- a/.changeset/khaki-mugs-ring.md
+++ b/.changeset/khaki-mugs-ring.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/plugin-helpers': patch
+---
+
+Unrequire documents and config for GenerateOptions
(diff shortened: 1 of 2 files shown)


dotansimha:
Yeah, you are right. @ludwigschubi can you please add a changeset for this PR? (Run `yarn changeset`)

ludwigschubi:
@dotansimha done

dotansimha:
Thanks @ludwigschubi .
I think @ardatan has some thoughts regarding this change? can you please share it here?

ardatan:
@dotansimha I think `config` can be `{}` and `documents` can be an empty array for now and we can update tutorials. I am afraid this is a breaking change because if we don't do null/undefined checks inside any of plugins and stuff, different kind of weird errors can pop up in the future.

ludwigschubi:
I am also using the workaround already and I agree that it is quite a minor thing, however I want to mention that when I did the tutorial this was a bit confusing... will close this issue then 👍

PR #4970, opened 2020-10-23
typesafe c-sharp operations

aokht:
Hi! Thank you for the great library.

I'd like to generate C# operation code.
Currently, there are no class definitions for the request/response parameter. So we need to define parameter classes by ourselves.

This p-r provides the option to generate the request/response parameters for operations so that we can execute typesafe query/mutation/subscription operations without defining parameter classes by ourselves.

I attached a diff below, from the example in https://graphql-code-generator.com/

Thank you for reviewing.

```
--- a/codegen.yml
+++ b/codegen.yml
 generates:
   Operations.cs:
     plugins:
-      - c-sharp-operations
+      - c-sharp-operations:
+          typesafeOperation: true
```
```
--- a/Operations.cs
+++ b/Operations.cs
@@ -1,3 +1,7 @@
+using Newtonsoft.Json;
+using GraphQL;
+using GraphQL.Client.Abstractions;
+
 namespace GraphQLCodeGen {

     public class FindUserGQL {
@@ -30,7 +34,43 @@ namespace GraphQLCodeGen {
           username
           role
         }";
+
+      public class Variables {
+
+        [JsonProperty("userId")]
+        public string userId { get; set; }
+
+      }

+      public class Response {
+
+        public class UserSelection {
+
+          [JsonProperty("id")]
+          public string id { get; set; }
+
+          [JsonProperty("username")]
+          public string username { get; set; }
+
+          [JsonProperty("role")]
+          public Role role { get; set; }
+
+        }
+
+        [JsonProperty("user")]
+        public UserSelection user { get; set; }
+
+      }
+
+      public static System.Threading.Tasks.Task<GraphQLResponse<Response>> SendQueryAsync(IGraphQLClient client, Variables variables, System.Threading.CancellationToken cancellationToken = default) {
+        return client.SendQueryAsync<Response>(Request(variables), cancellationToken);
+      }
+
+    }
+
+    public enum Role {
+      USER,
+      ADMIN
     }

 }
```

### DIFF
--- a/.changeset/new-beans-obey.md
+++ b/.changeset/new-beans-obey.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/c-sharp-operations': minor
+---
+
+typesafe c-sharp operations
--- a/packages/plugins/c-sharp/c-sharp-operations/src/config.ts
+++ b/packages/plugins/c-sharp/c-sharp-operations/src/config.ts
@@ -58,4 +58,15 @@ export interface CSharpOperationsRawPluginConfig extends RawClientSideBasePlugin
    * ```
    */
   subscriptionSuffix?: string;
+  /**
+   * @description Allows to generate operation methods with class definitions for request/response parameters
+   * @default false
+   *
+   * @exampleMarkdown
+   * ```yml
+   * config:
+   *   typesafeOperation: true
+   * ```
+   */
+  typesafeOperation?: boolean;
 }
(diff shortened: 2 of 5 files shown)


dotansimha:
Thank you so much @aokht . 
TBH, I'm not familiar enough with GraphQL and C#. @cklam2  you are the last one who worked on these plugins, any chance you can take a look at this PR?

rus-art (approved):

Issue #4911, opened 2020-10-15
Double // @flow comment added to generated file


FezVrasta:
**Describe the bug**
 

Hi, when I enable `flow` and `flow-operations` plugins at the same time, two `// @flow` headers are added to the `graphql.js` file.

**To Reproduce**

I tried to edit the template but it doesn't seem to work with the Flow plugins.

https://codesandbox.io/s/wizardly-paper-wsqgl?file=/document.graphql

1. My GraphQL schema:

```graphql
Not relevant for the issue
```

2. My GraphQL operations:

```graphql
Not relevant for the issue
```

3. My `codegen.yml` config file:

```yml
overwrite: true
schema: 'allSchema.graphql'
generates:
  ./src/generated/graphql.js:
    documents: 'queries/*.js'
    plugins:
      - flow
      - flow-operations
```

**Expected behavior**

**Environment:**

- OS: macOS Catalina
- `@graphql-codegen/flow`: same as codesandbox
- `@graphql-codegen/flow-operations`: same as codesandbox
- NodeJS: 14.5.0

**Additional context**

dotansimha:
I think this was introduced in: https://github.com/dotansimha/graphql-code-generator/pull/4609  
Does it's breaking anything @FezVrasta ? or it's just semantics?

FezVrasta:
It errors in my create-react-app project

dotansimha:
@FezVrasta  build error or lint error? As far as I can tell, the Flow compiler works perfectly fine with that...

FezVrasta:
![image](https://user-images.githubusercontent.com/5382443/96447775-bb899780-1212-11eb-8564-bdbaaccde23e.png)

sam-rad:
Having the the same issue. This is where [flow throws](https://github.com/facebook/flow/blob/master/src/typing/errors/error_message.ml#L2828):

```ml
| MultipleFlowAttributes ->
  [
    text "Unexpected ";
    code "@flow";
    text " declaration. Only one per ";
    text "file is allowed.";
  ]
```

sam-rad:
[...]

PR #4971, opened 2020-10-24
Remove extra `// @flow` declaration from flow operations

sam-rad:
Related: https://github.com/dotansimha/graphql-code-generator/issues/4911

`// @flow` declaration is removed from [flow operations](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/flow/operations/src/index.ts#L38) since [flow plugin](https://github.com/dotansimha/graphql-code-generator/blob/c0f7e16f2e9ddaab61421643da7e4851cadccf97/packages/plugins/flow/flow/src/index.ts#L10) already prepends this to the final output.

Not sure if this affects other part of the lib though.

### DIFF
--- a/.changeset/fair-spoons-drum.md
+++ b/.changeset/fair-spoons-drum.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/flow': patch
+'@graphql-codegen/flow-operations': patch
+---
+
+Consolidate `// @flow` declarations to avoid duplicates
--- a/dev-test/githunt/flow.flow.js
+++ b/dev-test/githunt/flow.flow.js
@@ -1,7 +1,5 @@
 // @flow
 
-// @flow
-
 /** All built-in and custom scalars, mapped to their actual values */
 export type Scalars = {|
   ID: string,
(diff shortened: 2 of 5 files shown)


dotansimha:
@sam-rad 
Can you please add a changeset? (run `yarn changeset`), please bump patch.

Also, I think the cause of this issue is the fact that we have different definitions of the flow comments, in different plugins. 

See:
* https://github.com/dotansimha/graphql-code-generator/blob/18befe146c24963943788d5c0b9675398e5c9399/packages/plugins/flow/flow/src/index.ts#L14
* https://github.com/dotansimha/graphql-code-generator/blob/abe7e6d54aa33fe4e843f396cd14b22e4c74b4bf/packages/plugins/flow/operations/src/index.ts#L38

This thing is - the `prepend` statements are managed as `Set` to avoid duplicates (see https://github.com/dotansimha/graphql-code-generator/blob/ac067ea088011cd43a145aa1b021f76f3603b1eb/packages/graphql-codegen-core/src/codegen.ts#L133-L139), so if both lines will equal, it won't be generated as duplicate, even if both returns it. So can you please fix that and just make sure both flow comments are the same?

sam-rad:
Cool, thanks for the pointers. I'll look into it today.

sam-rad:
@dotansimha This should be ready now. I also removed the extra space after the `@flow` and before the `\n`.

Issue #4700, opened 2020-09-04
Generate and verify `@apollo/client@3` `keyFields` configuration


SimenB:
`@apollo/client@3` comes with a new config option called `typePolicies` passed to the cache's constructor, which can be used (amongst many different awesome things) to specify `keyFields` for each GraphQL type in your schema, so that the cache knows how to normalize the entries properly.

@benjamn talks about it in this great talk: https://youtu.be/n_j8QckQN5I, and specifically what I'm requesting is the code generation mentioned briefly here: https://youtu.be/n_j8QckQN5I?t=1347.

My hope for this potential new codegen plugin is:

- there would have some sort of directive (`@id`? `@keyField`?) that is applied to types in the schema
- the plugin would generate the config object with `keyFields` (and possibly other stuff in the future) defined, which would be splatted into the cache constructor
- throw an error if we request a type in. a query without specifying all the `keyFields` (i.e. fields with the `@id` directive). This last one would be _really_ awesome as we'd know when we define queries that we request all the relevant fields to properly normalize the cache, avoiding runtime errors/warnings

Docs: https://www.apollographql.com/docs/react/caching/cache-configuration/#generating-unique-identifiers

/cc @hwillson @benjamn would love your thoughts on this 🙂

dotansimha:
That could be great! Thank you @SimenB ! 
Moving to Discussions (we track new plugins requests there).

PR #4973, opened 2020-10-25
Added new plugin for generating apollo-client v3 types helpers

dotansimha:
Including:
- [x] keyFields generated with all possible GraphQL types / interface and their fields.

Related:
- https://github.com/dotansimha/graphql-code-generator/discussions/4700


packages/plugins/typescript/apollo-client-helpers/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      const fieldPolicyVarName = `${typeName}FieldPolicy`;
+
+      perTypePolicies.push(
+        `export type ${keySpecifierVarName} = (${fieldsNames
+          .map(f => `'${f}'`)
+          .join(' | ')} | ${keySpecifierVarName})[];`
+      );
+
+      perTypePolicies.push(`export type ${fieldPolicyVarName} = {
+${fieldsNames.map(fieldName => `\t${fieldName}?: FieldPolicy<any> | FieldReadFunction<any>`).join(',\n')}
```
SimenB, line 43: would it be possible to get a more correct type than `any` here?

packages/plugins/typescript/apollo-client-helpers/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+\t\tfields?: ${fieldPolicyVarName},
+\t}`,
+      };
+    }
+
+    return prev;
+  }, {} as Record<string, string>);
+
+  const rootContent = `export type TypedTypePolicies = TypePolicies & {${Object.keys(typedTypePolicies)
+    .map(typeName => `\n\t${typeName}?: ${typedTypePolicies[typeName]}`)
```
SimenB, line 62: would it be possible to get an option for removing the `?` here, making it so we _have_ to specify all types?

packages/plugins/typescript/apollo-client-helpers/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      );
+
+      perTypePolicies.push(`export type ${fieldPolicyVarName} = {
+${fieldsNames.map(fieldName => `\t${fieldName}?: FieldPolicy<any> | FieldReadFunction<any>`).join(',\n')}
+};`);
+
+      return {
+        ...prev,
+        [typeName]: `{
+\t\tkeyFields?: false | ${keySpecifierVarName} | (() => undefined | ${keySpecifierVarName}),
```
SimenB, line 49: and option for forcing `keyFields` to be defined would be sweet 👍

SimenB:
Yay, this is awesome! Thoughts on the directives and stuff I mentioned in #4700 and possible failing if I query for a type but do not include all of its keys?

packages/plugins/typescript/apollo-client-helpers/src/index.ts
SimenB: except for fields defined in `schema`:

```graphql
schema {
  query: Query
  mutation: Mutation
}
```

https://graphql.org/learn/schema/#the-query-and-mutation-types

packages/plugins/typescript/apollo-client-helpers/src/index.ts
SimenB: except for possibly if one of the fields is named `id` as that's automatically picked up by apollo? not sure

packages/plugins/typescript/apollo-client-helpers/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  schema: GraphQLSchema,
+  config: ApolloClientHelpersConfig
+): Types.ComplexPluginOutput {
+  const typeMap = schema.getTypeMap();
+  const perTypePolicies: string[] = [];
+  const typedTypePolicies = Object.keys(typeMap).reduce((prev, typeName) => {
+    const type = typeMap[typeName];
+
+    if (!typeName.startsWith('__') && (isObjectType(type) || isInterfaceType(type))) {
+      const fieldsNames = Object.keys(type.getFields()).filter(f => !f.startsWith('__'));
```
SimenB, line 32: it is allowed to specify nested keys (see the `Book` example using `["title", "author", ["name"]]`) here: https://www.apollographql.com/docs/react/caching/cache-configuration/#generating-unique-identifiers. Unless I'm doing something weird, that produces a type error here

packages/plugins/typescript/apollo-client-helpers/src/index.ts
dotansimha: I was thinking about that, and it might be tricky, I think this eventually needs the actual type right? Or am I wrong?

SimenB: yep `typeName` above. Maybe imported via a `baseTypesPath` option like the `near-operation-file` preset?

dotansimha: Yeah, but that's for later ;)

packages/plugins/typescript/apollo-client-helpers/src/index.ts
dotansimha: Working on those fixes here: https://github.com/dotansimha/graphql-code-generator/pull/4979

packages/plugins/typescript/apollo-client-helpers/src/index.ts
dotansimha: Are you sure it's an actual issue? I can see that it works perfectly: 
image

Which TS version are you using

packages/plugins/typescript/apollo-client-helpers/src/index.ts
dotansimha: > Yay, this is awesome! Thoughts on the directives and stuff I mentioned in #4700 and possible failing if I query for a type but do not include all of its keys? A directive could be used to generate the config (or at least their `keyFields`) instead of just verifying the type 🙂

@SimenB I'm trying to avoid usage of directives / changes to the schema to manipulate codegen behaviour...

SimenB:
> > Yay, this is awesome! Thoughts on the directives and stuff I mentioned in #4700 and possible failing if I query for a type but do not include all of its keys? A directive could be used to generate the config (or at least their `keyFields`) instead of just verifying the type 🙂
> 
> @SimenB I'm trying to avoid usage of directives / changes to the schema to manipulate codegen behaviour...

I totally understand that. Any other idea on how to enforce queries always fetch the necessary `keyFields`?

PR #4979, opened 2020-10-26
Use ac3 as default for fragment matcher plugin, and a few more AC3 fixes

dotansimha:


SimenB:
🎉

Using both of these options _and_ specifying `Query` makes `keyFields` mandatory in `Query`

image

(we currently specify `Query` to use `fields`)

Issue #3836, opened 2020-04-08
@skip, @include directives do not result in fields becoming optional in the generated code from typescript-operations


helios1138:
right now, there is no difference in typescript being generated by `typescript-operations` for
```gql
query DynamicTest {
  viewer {
    id
    me
    ...userFragment
  }
}
```
and
```gql
query DynamicTest($include: Boolean!) {
  viewer {
    id
    me @include(if: $include)
    ...userFragment @include(if: $include)
  }
}
```
expected result: field `me` and all the fields from `userFragment` should become optional (`T | undefined`) in the typings, since they are no longer guaranteed to be present when query executes

Cellule:
I have 2 fragments fetching similar fields, but one has lots of `@include(id: $somevar)`  with different vars.
I want to use typings to figure out which variables should be true to ensure the fragment with conditional results in equivalent fields being fetched.
However, because props with `@include` are not optional, I'm stick listing them all manually which can be error prone (especially when changing the fragments over time)

Fi1osof:
This make me sad :(

@helios1138 +100

rathpc:
Any work being actively done to address this issue?

ardatan:
Currently no. Maybe, you can start with a PR with some initial work so we can support you.

dannycochran:
For reference, here's a PR in the apollo codegen for this:
https://github.com/apollographql/apollo-tooling/pull/1854

dotansimha:
There is some WIP for this issue here: https://github.com/dotansimha/graphql-code-generator/pull/5017

dotansimha:
https://github.com/dotansimha/graphql-code-generator/pull/5017 seems to be ready, merging soon. This feature will be available in next release :) Awesome work, @gilgardosh , thank you!

dotansimha:
Available in `@graphql-codegen/typescript-operations@1.17.10` and `@graphql-codegen/flow-operations@1.18.3`.

Thanks @gilgardosh !

Cellule:
I've updated all the plugins I'm using 

```json
"@graphql-codegen/cli": "^1.19.2",
"@graphql-codegen/fragment-matcher": "^2.0.0",
"@graphql-codegen/typescript": "^1.17.11",
"@graphql-codegen/typescript-operations": "^1.17.10",
"@graphql-codegen/typescript-resolvers": "^1.17.11",
```

[...]

Issue #5002, opened 2020-10-29
Incorrect error "Unable to find parent type! Please make sure you operation passes validation" when using near-operation-file and @include


asfernandes:
**Describe the bug**
Incorrect validation error when using `near-operation-file`.

**To Reproduce**
Easy way to reproduce: go to https://graphql-code-generator.com/ and set "Operations types (near-operation-file)".

With "Operations types" it works.

Use this `operation.graphql`:

```graphql
query chats {
  myChats {
    ...ChatFields
  }
}

fragment ChatFields on Chat {
  id
  ...@include(if: true) {
      id
      messages {
          id
      }
  }
}
```

**Environment:**

- OS: Windows

- package.json
```
    "@graphql-codegen/add": "^2.0.1",
    "@graphql-codegen/cli": "1.19.0",
    "@graphql-codegen/introspection": "1.18.0",
    "@graphql-codegen/near-operation-file-preset": "^1.17.11",
    "@graphql-codegen/typed-document-node": "^1.17.9",
    "@graphql-codegen/typescript": "1.17.11",
    "@graphql-codegen/typescript-operations": "1.17.8",
    "@graphql-codegen/typescript-react-apollo": "2.0.7",
```

- NodeJS: 10.18.1

dotansimha:
Hi @asfernandes ,
We are currently working on adding support for `@skip` / `@include` directives, see: https://github.com/dotansimha/graphql-code-generator/pull/5017 
Closing (duplicate of https://github.com/dotansimha/graphql-code-generator/issues/3836)

asfernandes:
> Hi @asfernandes ,
> We are currently working on adding support for `@skip` / `@include` directives, see: #5017
> Closing (duplicate of #3836)

The issue I have here (specific with near-operation-file) seems broad than what is described in #3836 which talks only about the types. Is this issue tracked there as well?

dotansimha:
I see. Let me give this a look. I this your actual error is coming from the fact that the inline fragment you are using isn't named, while codegen assumes it must be.

Changing the fragment to this:

```graphql

fragment ChatFields on Chat {
  id
  ... on Chat @include(if: true) {
      id
      messages {
          id
      }
  }
}
```

Should fix that.

dotansimha:
@asfernandes I think I managed to fix it in: https://github.com/dotansimha/graphql-code-generator/pull/5036 

[...]

PR #5017, opened 2020-11-02
Support for `@skip` and `@include` directives in `typescript-operations`

gilgardosh:
Related: https://github.com/dotansimha/graphql-code-generator/issues/3836

TODO:
- [x] Reproduce with a failing test
- [x] Apply a fix in `PreResolveTypes` processor
- [x] Apply a fix in default Flow processor + tests for `flow-operations`
- [x] Add a changeset
- [x] Apply a fix in default TS processor (`Pick`)
- [x] Make sure it works with presets as well (because of the need for a new TS helper to solve that issue with `Pick`)


dotansimha:
@gilgardosh please also note this one: https://github.com/dotansimha/graphql-code-generator/issues/5002 , this is also a valid use case of `@skip` and `@include` and we might need to address that.

packages/plugins/other/visitor-plugin-common/src/selection-set-processor/base.ts
```diff
@@ -1,7 +1,8 @@
 import { ScalarsMap, ConvertNameFn } from '../types';
-import { GraphQLObjectType, GraphQLInterfaceType, GraphQLOutputType } from 'graphql';
+import { GraphQLObjectType, GraphQLInterfaceType, GraphQLOutputType, DirectiveNode } from 'graphql';
 
 export type PrimitiveField = string;
+export type PrimitiveWithFlagsField = { makeNullable: boolean; fieldName: string };
```
dotansimha, line 5: Maybe instead of `makeNullable`, let's call it `isConditional`?

packages/plugins/flow/operations/src/flow-selection-set-processor.ts
```diff
@@ -73,7 +73,7 @@ export class FlowWithPickSelectionSetProcessor extends BaseSelectionSetProcessor
 
   transformPrimitiveFields(
     schemaType: GraphQLObjectType | GraphQLInterfaceType,
-    fields: PrimitiveField[]
+    fields: PrimitiveWithFlagsField[]
```
dotansimha, line 76: You can just change the `PrimitiveField` type to match your needs. This way we can spot other places where this change it needed.

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -383,7 +383,10 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
       ...(typeInfoField ? this._processor.transformTypenameField(typeInfoField.type, typeInfoField.name) : []),
       ...this._processor.transformPrimitiveFields(
         parentSchemaType,
-        Array.from(primitiveFields.values()).map(field => field.name.value)
+        Array.from(primitiveFields.values()).map(field => ({
+          makeNullable: this.resolveDirectives(field.directives),
```
dotansimha, line 387: Maybe we can move `resolveDirectives` method to common library or plugin-helpers or something? 
And maybe we can change the name to `hasConditionalDirectives`?

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -512,4 +515,8 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
       suffix: typeName ? `_${typeName}_${suffix}` : suffix,
     });
   }
+
+  protected resolveDirectives(directives: any): boolean {
+    return directives.length > 0;
```
dotansimha, line 520: Here we need to check the actual directives, right?

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -512,4 +515,8 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
       suffix: typeName ? `_${typeName}_${suffix}` : suffix,
     });
   }
+
+  protected resolveDirectives(directives: any): boolean {
```
dotansimha, line 519: Avoid `any` :)

dotansimha (changes requested):

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
gilgardosh: sure. not implemented yet.

dotansimha (approved):

PR #4983, opened 2020-10-26
fix(cli): resolve modules provided through the -r flags relative to the cwd

merceyz:
**What's the problem this PR addresses?**

The CLI is requiring modules passed through the `-r` flag from itself and not the users project.

Related https://github.com/dotansimha/graphql-code-generator/issues/4947

**How did you fix it?**

Resolve all modules provided to the `-r` flag relative to the CWD

PR #5032, opened 2020-11-04
fix(cli): resolve from folders correctly

merceyz:

I broke the tests in https://github.com/dotansimha/graphql-code-generator/pull/4983

**How did you fix it?**

Use `require.resolve` to resolve from the cwd


dotansimha:
Missed that :D thank you @merceyz :)

Issue #4161, opened 2020-06-03
Future proof unions


ArnaudBarre:
**Is your feature request related to a problem? Please describe.**

I want to avoid making assumptions on the fact that my graphQL unions are stable.

With this schema:

```graphql
query {
  notifications: [Notification!]!
}

union Notification =  MessageNotification | PostNotification

type NewMessage {
  author: User!
}

type PostNotification {
  author: User!
}
```

TS allow me to write 
```ts 
notifications.map(n => n.author.someField)
```
which could breaks at runtime if the schema was update with:

```graphql
union Notification =  MessageNotification | PostNotification | CallNotification

type NewMessage {
  caller: User!
}
```

**Describe the solution you'd like**

An optional codegen parameter to append `{ __typename: "FutureValue" }` to all unions (I don't have ideas when using codegen without __typename  )

**Describe alternatives you've considered**

For now I use this patch. But this is not a long-term solution

```patch
diff --git a/node_modules/@graphql-codegen/visitor-plugin-common/index.cjs.js b/node_modules/@graphql-codegen/visitor-plugin-common/index.cjs.js
index 2b5cde2..b15e7d0 100644
--- a/node_modules/@graphql-codegen/visitor-plugin-common/index.cjs.js
+++ b/node_modules/@graphql-codegen/visitor-plugin-common/index.cjs.js
@@ -2434,7 +2434,7 @@ class SelectionSetToObject {
     }
     transformSelectionSet() {
         const grouped = this._buildGroupedSelections();
-        return Object.keys(grouped)
+        const values = Object.keys(grouped)
             .map(typeName => {
             const relevant = grouped[typeName].filter(Boolean);
             if (relevant.length === 0) {
@@ -2446,9 +2446,10 @@ class SelectionSetToObject {
             else {
                 return `( ${relevant.join(' & ')} )`;
             }
-        })
-            .filter(Boolean)
-            .join(' | ');
+        }).filter(Boolean);
+        return values.length > 1
+          ? values.concat('{ __typename: "FutureValue" }').join(" | ")
+          : values[0];
     }
     transformFragmentSelectionSetToTypes(fragmentName, fragmentSuffix, declarationBlockConfig) {
         const grouped = this._buildGroupedSelections();
@@ -2479,7 +2480,7 @@ class SelectionSetToObject {
                 .export()
                 .asKind('type')
                 .withName(this.buildFragmentTypeName(fragmentName, fragmentSuffix))
[...]

PR #5041, opened 2020-11-04
Adds optional futureProofUnions for unions like relay

skk2142:
This fixes this issue: https://github.com/dotansimha/graphql-code-generator/issues/4161

### DIFF
--- a/packages/plugins/typescript/typescript/src/visitor.ts
+++ b/packages/plugins/typescript/typescript/src/visitor.ts
@@ -22,6 +22,7 @@ import {
   InputValueDefinitionNode,
   GraphQLSchema,
   isEnumType,
+  UnionTypeDefinitionNode,
 } from 'graphql';
 import { TypeScriptOperationVariablesToObject } from './typescript-variables-to-object';
 
@@ -30,6 +31,8 @@ export interface TypeScriptPluginParsedConfig extends ParsedTypesConfig {
   constEnums: boolean;
   enumsAsTypes: boolean;
   futureProofEnums: boolean;
+  futureProofUnions: boolean;
+  nonOptionalTypename: boolean;
   enumsAsConst: boolean;
   numericEnums: boolean;
   onlyOperationTypes: boolean;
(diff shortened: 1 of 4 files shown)


packages/plugins/typescript/typescript/src/visitor.ts
```diff
@@ -30,6 +31,8 @@ export interface TypeScriptPluginParsedConfig extends ParsedTypesConfig {
   constEnums: boolean;
   enumsAsTypes: boolean;
   futureProofEnums: boolean;
+  futureProofUnions: boolean;
+  nonOptionalTypename: boolean;
```
dotansimha, line 35: Can you please explain why `nonOptionalTypename` is needed?

packages/plugins/typescript/typescript/src/visitor.ts
```diff
@@ -123,6 +128,39 @@ export class TsVisitor<
     return `Maybe<${super.ListType(node)}>`;
   }
 
+  UnionTypeDefinition(node: UnionTypeDefinitionNode, key: string | number | undefined, parent: any): string {
+    if (this.config.onlyOperationTypes) return '';
+    let withFutureAddedValue: string[] = [];
+    if (this.config.futureProofUnions) {
+      if (this.config.addTypename) {
+        if (this.config.nonOptionalTypename) {
```
dotansimha, line 136: Maybe you can de-duplicate this? If `this.config.nonOptionalTypename` only effects the `?` sign, it can be inlined.

packages/plugins/typescript/typescript/src/visitor.ts
```diff
@@ -123,6 +128,39 @@ export class TsVisitor<
...
+          withFutureAddedValue = [
+            this.config.immutableTypes ? `{ readonly __typename: "%other" }` : `{ __typename: "%other" }`,
+          ];
+        } else {
+          withFutureAddedValue = [
+            this.config.immutableTypes ? `{ readonly __typename?: "%other" }` : `{ __typename?: "%other" }`,
+          ];
+        }
+      } else {
+        withFutureAddedValue = ['{}'];
```
dotansimha, line 146: The `__typename` thing isn't really supported on unions, that's why I originally asked about that flag. 
Adding this configuration just because of this new feature might cause confusion for users. Scheme types doesn't have `__typename`, it's something that being resolved only on the operation level.
Maybe this flow should be applied on `typescript-operations` and not here?

dotansimha (changes requested):
Thank you @skk2142 ! 
Please note my comments. Also, can you please add a changeset? (run `yarn changeset` and it will guide you)

packages/plugins/typescript/typescript/src/visitor.ts
skk2142: removed!

packages/plugins/typescript/typescript/src/visitor.ts
skk2142: We don't need this anymore.

skk2142 (commented):
I removed the addOptionalTypename option as you suggested.

ardatan:
Thanks @skk2142 ! Merged in https://github.com/dotansimha/graphql-code-generator/commit/34b8087efceeb40eb612b19f963ab365afb4f09f

PR #5050, opened 2020-11-05
feat(plugin: graphql-requests): Support the request header parameter …

moeHaydar:
A new feature is being supported by graphql-request library: Support passing custom headers per each request method.
This PR is to support the extra parameter than can be passed to the generated code in the plugin.

graphql-request: PR: prisma-labs/graphql-request#228

### DIFF
--- a/.changeset/flat-readers-guess.md
+++ b/.changeset/flat-readers-guess.md
@@ -1,0 +1,7 @@
+---
+'@graphql-codegen/typescript-graphql-request': major
+---
+
+Support passing custom headers per each request method.
+
+NOTE: This version of this plugin requires you to update to graphql-request > 3.4.0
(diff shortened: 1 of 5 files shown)


dotansimha:
Thank you @moeHaydar ! Let's wait until it's available in `graphql-request` and then we can merge it here as well.

moeHaydar:
@dotansimha PR merged there :) we are all good to merge here.

dotansimha:
Great, thanks! Waiting for a release. Is it a breaking change? in terms of users who are using previous versions of this library? (I mean old graphql-request and the changes for this PR)

dotansimha:
Just an update: It's still in `next` version (`3.4.0-next.1`) and not in a stable release, so this PR is still pending.

dotansimha:
Ok so `3.4.0` is out. We can get this merged :)

PR #5066, opened 2020-11-11
feat(typescript-resolvers): Added `internalResolversPrefix` flag to allow modifications to `resolveType` and `isTypeOf` resolvers

cometkim:
Hi @dotansimha, I recently started to use [mecurius](https://github.com/mercurius-js/mercurius) instead of Apollo Server and they use `resolveType` instead of `__resolveType` as a field name for abstract type resolver.

See https://github.com/mercurius-js/mercurius/commit/fd9040b772eedc2d50d4bce196c7ff6745d3742a

I think it will be nice if the field name is customizable. and here is how :)

appreciate your work.


cometkim:
Well I want to use isTypeOf also without prefix. Would it be better to be `resolveTypeFieldName` and `isTypeOfFieldName`?

dotansimha:
This is great. Thank you so much @cometkim !  2 small notes:
1. What about `__isTypeOf`? it should also be without `__`? 
2. Can you please add a changeset? (run `yarn changeset`)

dotansimha:
> Well, I want to use `isTypeOf` also without prefix. Would it be better to be `resolveTypeFieldName` and `isTypeOfFieldName`?

Maybe it could be `prefixInternalTypeResolver` or something similar? @kamilkisiela ideas on that name? :D

dotansimha:
Thank @cometkim .
I changed the name of the configuration flag to `internalResolversPrefix` and applied it to both `isTypeOf` and `resolveType`. Also, added the missing `changeset`. 

Thank you!

dotansimha:
Available in `@graphql-codegen/typescript-resolvers@1.18.0`

cometkim:
Thank you @dotansimha !

PR #5071, opened 2020-11-11
java plugin changes

tzachs:
Hi,

I've made several changes to java plugin to, (A) make it compile, (B) use best practice.
If you have any question please feel free to contact me 😃 

Thanks!
Tzach

### DIFF
--- a/.changeset/wet-islands-promise.md
+++ b/.changeset/wet-islands-promise.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/java': patch
+---
+
+Fixed error compiliation, changed to empty constructor and using setters
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ out.txt
 tsconfig.tsbuildinfo
 .yarn
 .yarnrc
+recompile.sh
--- a/packages/plugins/java/java/src/config.ts
+++ b/packages/plugins/java/java/src/config.ts
@@ -5,6 +5,19 @@ import { RawConfig, EnumValuesMap } from '@graphql-codegen/visitor-plugin-common
  */
 export interface JavaResolversPluginRawConfig extends RawConfig {
   /**
+   * @description Customize the class members prefix. The default is empty (this might be a breaking change from the default that was _)
+   *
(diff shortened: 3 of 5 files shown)


dotansimha:
Thank you @tzachs :)

hannesj:
@tzachs What was the reason to switch from a constructor parameter to individual setters? The args shape should be always correct based on the schema, which is used to validate the resolver arguments. Now we must duplicate the code, which was previously autogenerated for each argument.

tzachs:
@hannesj ,

1. If I've created a regression in the code than first I'm sorry
2. If I remember correct, the constructor received a map of params in order to init the data structure.
To use this map, the user must remember the name of each param it wants to init, thus making hard to use and prone to errors in case the variable in the data structure changes.
3. I'd be happy to see your use case and understand what is needed in order to make the solution more generic.

hannesj:
In my understanding you shouldn't create the map by had, but just use the `environment.getArguments()`as follows. The reason is that the map uses the names from the schema, and graphql-java validates that the input arguments are as specified in the schema. Here the `QueryTypeNearestArgs` is a generated class, and the idea is that you don't have to remember how the schema looks, and you can be assured that the fields in the args can be used directly.

```java
  @Override
  public DataFetcher<Connection<PlaceAtDistance>> nearest() {
    return environment -> {
      QueryTypeNearestArgs args = new QueryTypeNearestArgs(environment.getArguments());
      InputFiltersInput filterByIds = args.getFilterByIds();
      ...
```

tzachs:
@hannesj 
I'll try working on this ASAP, sorry for the late response...

tzachs:
@hannesj https://github.com/dotansimha/graphql-code-generator/pull/6012 should solve the problem. Please let me know if it did (or not)
@dotansimha / @ardatan may you please merge https://github.com/dotansimha/graphql-code-generator/pull/6012? thanks :)

tzachs:
@hannesj https://github.com/dotansimha/graphql-code-generator/pull/6012 was merged to master. Please check it fixed the problem and LMK. Thanks :)

Issue #5070, opened 2020-11-11
If Query operation not has non nullable variable, generated code with typescript-vue-apollo has a compile error


fossamagna:
**Describe the bug**
If query operation not has non nullable variable, parameter `variables` has question mark. it cause, generated code by `typescript-vue-apollo` has a compile error.

**To Reproduce**
Reproduce here: https://codesandbox.io/s/typescript-vue-apollo-repro-q7t1d?file=/document.graphql
```sh
yarn install
yarn generate
yarn build
```

1. My GraphQL schema:

```graphql
type Query {
    user(id: ID!): User!
    allUsers(limit: Int): [User]
}

type User {
    id: ID!
    username: String!
    email: String!
}
```

2. My GraphQL operations:

```graphql
query allUsers($limit: Int) {
  allUsers(limit: $limit) {
    id
    username
  }
}
```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  src/types.ts:
    plugins:
      - typescript
      - typescript-operations
      - typescript-vue-apollo
    config:
      vueCompositionApiImportFrom: vue
```

**Expected behavior**
Completion to compilation with no error.

**Environment:**

- OS: macOS
- `@graphql-codegen/typescript-vue-apollo`: 2.2.0 
- NodeJS: v14.9.0

**Additional context**

dotansimha:
Fixed in `@graphql-codegen/typescript-vue-apollo@2.2.1`.

PR #5076, opened 2020-11-12
fix(vue-apollo): Use empty object as default value to variables parameter instead of question mark when variables of query is all optional

fossamagna:
Related #5070 
A test has been added to cover this.


dotansimha:
Awesome. Thank you @fossamagna ! :)

Issue #5112, opened 2020-11-21
[typescript] Non-null default fields are incorrectly optional


jun-sheaf:
**Describe the bug**
 
Default values are handled incorrectly.

**To Reproduce**
Suppose we have the schema
```graphql
input SomeInput {
  test: String! = "test"
}
```
Then `graphql-codegen` will optionalize the argument `test`.

**Expected behavior**

The above should not be optional as `String!` is never `null`.

**Environment:**

- OS: Mac OS Catalina
- `@graphql-codegen/...`: Latest as of issue date
- NodeJS: Latest as of issue date

ardatan:
If you provide a default value for an input field, it means that input field should be optional on client side. You are able not to provide a value for that because the server is able to handle it with the given default value.

jun-sheaf:
Test cannot be set nullable on client side since test is a non-nullable type.

ardatan:
I was talking about this behavior so in this case `bar` is optional there even if it is `String!`.
https://codesandbox.io/s/late-cache-npyqj?file=/src/index.js

jun-sheaf:
Yes, I am aware, but as I stated, `bar` is not actually `null` for the resolver. [See here](https://codesandbox.io/s/stoic-chaum-un1z1?file=/src/index.js:0-1110)

Thus it would be better to have an `avoidOptionals` option for `defaultValue` for those who generate server types.

dotansimha:
Merged and published. Thanks @jun-sheaf!

PR #5113, opened 2020-11-21
feat(typescript): adds a `defaultValue` flag to all `avoidOptionals` configs

jun-sheaf:
This PR adds a `defaultValue` flag to the `avoidOptionals` configuration in all plugins. The behavior is as follows:
- Upon turning on the flag, all defaulted fields (in `input`s or arguments) will strictly obey the type of that field.
   - e.g. `input TestInput { test: String! = "test" }` will transpile to `type TestInput { test: Scalars['String'] }`.
- For `input`s, if the `inputValue` flag in `avoidOptionals` is on, then `defaultValue` is not necessary.
- For arguments, if the `objectValue` is `true`, then `defaultValue` is ignored. (to keep with current behavior)

Related #5112

### DIFF
--- a/.changeset/chilled-cows-fry.md
+++ b/.changeset/chilled-cows-fry.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript': minor
+---
+
+See https://github.com/dotansimha/graphql-code-generator/issues/5112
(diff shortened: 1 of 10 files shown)


dotansimha:
Thank you @jun-sheaf , looks good. Just to make sure- the default output is the same as it was before, right? And it's changing only by specifying this new config flag?

jun-sheaf:
Yes, the tests ensure this is the case.

dotansimha:
That's great, thank you @jun-sheaf !

Issue #4888, opened 2020-10-12
Types for input Lists do not support coercion


eightypop:
**Describe the bug**
 
According to the [graphql spec](https://spec.graphql.org/draft/#sel-FAHjBJFCAACE_Gh7d) input lists can also take a single value of the array's type. However, the types generated do not allow this. This causes the ts to generate type errors with valid queries.

**To Reproduce**
Define an argument that accepts a List
In the operation use the list variable as a single value.

Live repro: https://codesandbox.io/s/wizardly-tereshkova-gqv23?file=/types.ts

<!--
#############
#############
🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 

PLEASE READ THE FOLLOWING: 
You can use this template to create a live sandbox of the issue: 
https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template
Make sure sure fork this template, and run `yarn generate` in the terminal. 
Please make sure the Codegen and plugins version under `package.json` matches yours. 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~ ISSUES WITH A LIVE REPRODUCTION WILL BE ANSWERED QUICKLY ~~~~~~~
~~~~~~ ISSUES WITHOUT REPRODUCTION MIGHT GET CLOSED WITHOUT RESPONSE ~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#############
#############
-->

1. My GraphQL schema:

```graphql
type Query {    
    search(term: [String!]!): [User!]
}

type User {
    id: ID!    
}

```

2. My GraphQL operations:

```graphql
query user($term: [String!]! ) {
[...]

PR #5115, opened 2020-11-23
Support for input lists coercion on `Variables` types

gilgardosh:
related: [#4888](https://github.com/dotansimha/graphql-code-generator/issues/4888)


dev-test/test-schema/resolvers-root.ts
```diff
@@ -150,14 +150,14 @@ export type QueryRootResolvers<
   ContextType = any,
   ParentType extends ResolversParentTypes['QueryRoot'] = ResolversParentTypes['QueryRoot']
 > = {
-  allUsers?: Resolver<Array<Maybe<ResolversTypes['User']>>, ParentType, ContextType>;
+  allUsers?: Resolver<Array<Maybe<ResolversTypes['User']>> | Maybe<ResolversTypes['User']>, ParentType, ContextType>;
```
dotansimha, line 153: This should be effected here? I'm not sure.

dotansimha (changes requested):

dev-test/test-schema/flow-types.flow.js
```diff
@@ -111,7 +111,11 @@ export type ResolversParentTypes = {
 };
 
 export type QueryResolvers<ContextType = any, ParentType = $ElementType<ResolversParentTypes, 'Query'>> = {
-  allUsers?: Resolver<Array<?$ElementType<ResolversTypes, 'User'>>, ParentType, ContextType>,
+  allUsers?: Resolver<
```
dotansimha, line 114: Shouldn't get modified as well, right?

packages/plugins/flow/flow/tests/__snapshots__/flow.spec.ts.snap
```diff
@@ -126,9 +126,9 @@ export type MyType = {|
 export type MyTypeFooArgs = {|
```
dotansimha, line 126: I'm not 100% sure that we should modify `Args`. I think only `variables` should be effected. 

Let's say you have an argument defined for field, if you define it as `[String]`, your Args will always be `string[]`. But, on the level of consumption, where you use variables, you should be able to use `string | string[]`. 

What do you think?

dotansimha (changes requested):

dev-test/test-schema/flow-types.flow.js
gilgardosh: It shouldn't. same issue accured on TS resolvers.
Fixed.

packages/plugins/flow/flow/tests/__snapshots__/flow.spec.ts.snap
gilgardosh: I actually don't know if GraphQL support coercion on `Args`
we should try :)

dotansimha: Let's just apply it for `Variables` for now. I think `Args` will always be the way you define it.

packages/plugins/typescript/mongodb/src/visitor.ts
```diff
@@ -23,6 +22,7 @@ import {
   InterfaceTypeDefinitionNode,
   UnionTypeDefinitionNode,
 } from 'graphql';
+import { wrapTypeNodeWithModifiers } from '../../../other/visitor-plugin-common/src/utils';
```
dotansimha, line 25: Please fix this import, it should take it from `@graphql-codegen/visitor-plugin-common`

.changeset/fresh-balloons-talk.md
```diff
@@ -0,0 +1,5 @@
+---
```
dotansimha, line 1: Please update your `changeset`, since `flow-operations` has also changed.

dotansimha: Also, `visitor-plugin-common` has changed.

dotansimha (changes requested):

PR #5121, opened 2020-11-24
Improve 'dotenv' Integration documentation

shiferlior:
I didn't understand how to use it, this documentation help to understand it.

### DIFF
--- a/website/docs/getting-started/require-field.md
+++ b/website/docs/getting-started/require-field.md
@@ -37,3 +37,29 @@ Specifying `-r` using a CLI flag will load your `require.extension` _before_ loa
 If you wish to use [dotenv](https://github.com/motdotla/dotenv) to load environment variables, you can install `dotenv` from npm and then to use the `require` cli flag to preload the `dotenv` require extensions: `-r dotenv/config`.
 
 It will make sure to load your `.env` file before executing the codegen and loading your `.yml` file.
+
+#### For example:
+
+local.env file:
+```ENV
+    SCHEME_PATH=http:localhost:8080/graphql
+```
+    
+  codegen.yml file:
+```yml
+    schema: ${SCHEMA_PATH}
+    documents: ./src/**/*.graphql
+    generates:
+      ./src/types.ts:
+        plugins:
+          - typescript
+          - typescript-operations
+```
+   package.json file:
+   ```json
+   {
+    "scripts": {
+      "generate": "graphql-codegen --require dotenv/config --config codegen.yml DOTENV_CONFIG_PATH=local.env"
+    }
+   }
+   ```
--- a/website/docs/getting-started/require-field.md
+++ b/website/docs/getting-started/require-field.md
@@ -1,6 +1,6 @@
 ---
 id: require-field
-title: "`require` field"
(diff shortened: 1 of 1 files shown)


website/docs/getting-started/require-field.md
```diff
@@ -37,3 +37,29 @@ Specifying `-r` using a CLI flag will load your `require.extension` _before_ loa
...
+      ./src/types.ts:
+        plugins:
+          - typescript
+          - typescript-operations
+```
+   package.json file:
+   ```json
+   {
+    "scripts": {
+      "generate": "graphql-codegen --require dotenv/config --config codegen.yml DOTENV_CONFIG_PATH=local.env"
```
ardatan, line 62: Using 'cross-env' would be better because some OSes don't support this kind of env definition.

website/docs/getting-started/require-field.md
```diff
@@ -37,3 +37,29 @@ Specifying `-r` using a CLI flag will load your `require.extension` _before_ loa
 If you wish to use [dotenv](https://github.com/motdotla/dotenv) to load environment variables, you can install `dotenv` from npm and then to use the `require` cli flag to preload the `dotenv` require extensions: `-r dotenv/config`.
 
 It will make sure to load your `.env` file before executing the codegen and loading your `.yml` file.
+
+#### For example:
+
+local.env file:
+```ENV
+    SCHEME_PATH=http:localhost:8080/graphql
```
ardatan, line 45: I think this should be 'SCHEMA_PATH'

website/docs/getting-started/require-field.md
```diff
@@ -37,3 +37,29 @@ Specifying `-r` using a CLI flag will load your `require.extension` _before_ loa
 If you wish to use [dotenv](https://github.com/motdotla/dotenv) to load environment variables, you can install `dotenv` from npm and then to use the `require` cli flag to preload the `dotenv` require extensions: `-r dotenv/config`.
 
 It will make sure to load your `.env` file before executing the codegen and loading your `.yml` file.
+
+#### For example:
+
+local.env file:
```
ardatan, line 43: This example shows how to provide custom env file. Maybe we should mention.

ardatan (changes requested):

website/docs/getting-started/require-field.md
dotansimha: @ardatan it's not set from env, so it's fine not to use cross-env in the specific one.

dotansimha, line 62: Please change `DOTENV_CONFIG_PATH=local.env` to `dotenv_config_path=local.env`. I know both works the same, but uppercase feel like an environment variable,

dotansimha (commented):
Thank you! This is great :) 
Can you please go over the review notes?

website/docs/getting-started/require-field.md
ardatan: Oh that's my mistake :) You're right!

dotansimha:
Thanks @shiferlior . I did some fixes and now it's ready for merge.

PR #5128, opened 2020-11-25
adds useImplementingTypes flag support

### LINKED ISSUE #5126: Add `useImplementingInterfaces` flag for typescript plugin

Currently interfaces are translated to typescript types using the interface name.
Schema:
```
type Foo {
   m: [Bar!]!
}

interface Bar {
   foo: Foo!
}

type Blub implements Bar {
   ...
}

type Blabla implements Bar {
   ...
}
```

This will generate following types:
```typescript
export type Bar = {
  foo: Foo;
};

export type Blabla = Bar & {
  __typename?: 'Blaba';
  foo: Foo;
  ...
};

export type Blub = Bar & {
  __typename?: 'Blaba';
  foo: Foo;
  ...
};

export type Foo = {
  __typename?: 'Foo';
  m: Array<Bar>;
};
```

The problem here is that I want to store the actual implementations of the interface in my array `m` on type `Foo` and not (per definition non-existent) instances of an interface.

Add a `useImplementingInterfaces` flag to codegen.yml config.
When set, the generated type `Foo` from above will look like this:
```typescript
export type Foo = {
  __typename?: 'Foo';
  m: Array<Blub | Blabla>;
};
```

No alternatives currently available.

maaft:
Related: #5126

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -19,6 +19,7 @@ import {
   DirectiveNode,
   Kind,
   GraphQLEnumType,
+  GraphQLObjectType,
 } from 'graphql';
 import flatMap from 'array.prototype.flatmap';
 import { BaseVisitor, ParsedConfig, RawConfig } from './base-visitor';
@@ -42,6 +43,7 @@ import {
 } from './utils';
 import { OperationVariablesToObject } from './variables-to-object';
 import { parseEnumValues } from './enum-values';
+import { schema } from 'packages/plugins/typescript/resolvers/tests/common';
 
 export interface ParsedTypesConfig extends ParsedConfig {
   enumValues: ParsedEnumValuesMap;
@@ -51,6 +53,7 @@ export interface ParsedTypesConfig extends ParsedConfig {
   enumPrefix: boolean;
   fieldWrapperValue: string;
   wrapFieldDefinitions: boolean;
+  useImplementingInterfaces: boolean;
 }
 
 export interface RawTypesConfig extends RawConfig {
@@ -174,6 +177,22 @@ export interface RawTypesConfig extends RawConfig {
    * ```
    */
   onlyOperationTypes?: boolean;
+  /**
+   * @description This will cause the generator to use the implementing types of an interface as the type for a field.
+   * @default false
+   *
+   * @exampleMarkdown
+   * ## Override all definition types
+   * ```yml
+   * generates:
+   * path/to/file.ts:
+   *  plugins:
+   *    - typescript
+   *  config:
+   *    useImplementingInterfaces: true
+   * ```
+   */
+  useImplementingInterfaces?: boolean;
 }
 
 export class BaseTypesVisitor<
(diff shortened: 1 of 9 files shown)


packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -174,6 +176,22 @@ export interface RawTypesConfig extends RawConfig {
    * ```
    */
   onlyOperationTypes?: boolean;
+  /**
+   * @description This will cause the generator to use the implementing types of an interface as the type for a field.
```
dotansimha, line 180: `When a GraphQL interface is used for a field, this flag will use the implementing types, instead of the interface itself.`

packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -577,6 +596,24 @@ export class BaseTypesVisitor<
 
     const schemaType = this._schema.getType(node.name as any);
 
+    if (this.config.useImplementingInterfaces) {
```
dotansimha, line 599: Can you please add here:
```
// TODO: Move this to a better place, since we are using this logic in some other places as well.
```

packages/plugins/typescript/typescript/tests/typescript.spec.ts
```diff
@@ -2678,4 +2678,71 @@ describe('TypeScript', () => {
...
+      testSchema,
+      [],
+      {
+        useImplementingInterfaces: true,
+      } as any,
+      { outputFile: 'graphql.ts' }
+    )) as Types.ComplexPluginOutput;
+
+    // Type should be Array<RedPill|GreenPill> and not Pill
+    expect(output.content).toBeSimilarStringTo(`
```
dotansimha, line 2727: Can you please add here:

```ts
 expect(output.content).toMatchSnapshot();
```

dotansimha (changes requested):

dotansimha:
@maaft please add a changeset for your changes. Run `yarn changeset` and it will guide you. 
In this case, we need to bump the `visitor-plugin-common`, `typescript`, `flow` packages with `minor`.

maaft (commented):
I will also change the flag to `useImplementingTypes` which is more correct I think.

packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -577,6 +596,24 @@ export class BaseTypesVisitor<
...
+      const allTypesMap = this._schema.getTypeMap();
+      const implementingTypes: string[] = [];
+
+      for (const graphqlType of Object.values(allTypesMap)) {
+        if (graphqlType instanceof GraphQLObjectType) {
+          const allInterfaces = graphqlType.getInterfaces();
```
ardatan, line 605: Does this recursively handles interfaces that implement another interface?

maaft: Good point. I'll add a unit test and check

maaft: I added a test which should cover this case and it finishes without error. But please have another look, maybe I've missed something

packages/plugins/typescript/typescript/tests/typescript.spec.ts
maaft: I just added this here but honestly I don't know what it does so you might want to take another look.

dotansimha:
Thank you @maaft .
I did an additional review for this, and it seems good. I moved the implementation to be in `typescript` plugin and not in the common library, since it will effect other plugins as well, that might no fully support it yet. 

Merging soon :)

PR #5138, opened 2020-11-26
fix: Fix passing paths with empty space

glennreyes:
Small fix that prevents commands in hooks from breaking when passing a path with a whitespace to the command, such as `prettier --write /Users/johndoe/my project/index.ts`.


glennreyes:
I noticed that this solution might not work on every use case. Suggestions on how to improve this are welcome.

dotansimha:
Thanks @glennreyes ! 
I'm not sure what to do in this case, since dealing with args in exec/spawn might be tricky :/ spaces might break args. 

Have you tried running `prettier --write "/Users/johndoe/my project/index.ts"` ?

dotansimha:
Closing. @glennreyes let me know if it needs to be reopened :)

PR #5197, opened 2020-12-04
[react-query] Respect typesPrefix

FredyC:
I noticed that `typesPrefix` was used for generated hook functions. That is not aligned with other plugins. Besides, that config entry should be applied by definition to types only, not to runtime things.


dotansimha:
Thanks @FredyC . Can you please just rebase and i'll get this merged? :) thanks

FredyC:
@dotansimha Done

PR #5235, opened 2020-12-10
Enable flattening typed document nodes

fehnomenal:
I wanted to completely flatten fragments.

### DIFF
--- a/.changeset/happy-kids-walk.md
+++ b/.changeset/happy-kids-walk.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typed-document-node': patch
+---
+
+Enable flattening typed document nodes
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @fehnomenal 
Can you please share more about the use-case? I just want to understand why is it needed and how to documents it. 
Also, can you please add a `changeset`? (`yarn changeset` and it will guide you)

fehnomenal:
Honestly, I can't really remember what the problem was but I can no longer reproduce it. Nevertheless, I think with this option the plugin is more in line with the other

fehnomenal:
Ahh, I remember now. I use https://github.com/clentfort/urql-custom-scalars-exchange to convert ISO datetime strings to Date objects and this library does not play well with fragments. So I had the choice of patching either library and this inlining was way easier.

But inlining only the operations was not enough because now the typed documents referenced the generated types for the fragments, which no longer existed due to being inlined.

dotansimha:
Got it, thanks @fehnomenal !

PR #5249, opened 2020-12-12
Update typescript-react-query.md with typescript example of custom fetcher

mthines:


website/docs/plugins/typescript-react-query.md
```diff
@@ -172,4 +172,31 @@ Your `myFetcher` should be in the following signature:
 type MyFetcher<TData, TVariables> = (operation: string, variables?: TVariables): (() => Promise<TData>)
 ```
 
+#### Usage example
+```tsx
+export const fetchData = <TData, TVariables>(query: string, variables?: TVariables): (() => Promise<TData>) => {
+  return async () => {
+    const res = await fetch('https://graphql.organiced.app/v1/graphql', {
```
mthines, line 179:
```suggestion
    const res = await fetch('https://api.url', {
```

mthines:
I realized I added our API url in the example, so I've removed it.

dotansimha:
Awesome, thank you @mthines !

PR #5255, opened 2020-12-13
Fixes mistaken MODULE_NOT_FOUND error for missing dependency

lucasconstantino:
When the referenced plugin/preset throws a `MODULE_NOT_FOUND` error within itself, `@graphql-codegen/cli` mistakenly assumes the plugin/preset itself wasn't found, resulting in a very misleading error message.

This shouldn't happen when using external packages as plugins/presets, but I hit this issue when using a local file, within which I had forgotten to install a dependency - in my case, `ramda`.

### DIFF
--- a/.changeset/loud-clouds-lick.md
+++ b/.changeset/loud-clouds-lick.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': patch
+---
+
+Fix wrong MODULE_NOT_FOUND for missing dependencies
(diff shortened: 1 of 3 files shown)


dotansimha:
Thank you @lucasconstantino , this is great. Can you please add a `changeset` for your changes? (run `yarn changeset` and it will guide you).

lucasconstantino:
@dotansimha done ✅

PR #5261, opened 2020-12-14
Introduce replaceKinds to use graphql.Kind instead of strings

kamilkisiela:
```diff
- {"kind" : "Document", definitions: []}
+ {"kind" : Kind.DOCUMENT, definitions: []}
```

It should save up some bundle size

### DIFF
--- a/.changeset/brown-terms-confess.md
+++ b/.changeset/brown-terms-confess.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/visitor-plugin-common': minor
+---
+
+Introduce replaceKinds option to replace kinds as strings with Kind of graphql-js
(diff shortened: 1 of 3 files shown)


.changeset/brown-terms-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-codegen/visitor-plugin-common': minor
```
dotansimha, line 2: @kamilkisiela please change to `patch`. I know it's not ideal, but it's the best way to workaround conflicts later.

kamilkisiela: Oh, ok. I'm testing it still so let's keep the pr open or as a draft

PR #5265, opened 2020-12-15
feat(apollo-client-helpers): Add TypePolicies merge field

jeroenvisser101:
Adds the merge field ([see API](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#fieldpolicy-api-reference)) for generated TypePolicies.

We could also, instead, use something like `Omit<FieldPolicy, "fields" | "keyFields"> & { fields?: ..., keyFields?: ... }>`, that way, when Apollo adds other fields, we still only override the ones that we type specifically.

wdyt?

### DIFF
--- a/.changeset/twelve-papayas-scream.md
+++ b/.changeset/twelve-papayas-scream.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-apollo-client-helpers': patch
+---
+
+Extend TypePolicy for types in TypedTypePolicies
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @jeroenvisser101 ! this is great. I think using `Omit` might be simpler here?
Also, can you please add a changeset to your changes? (run `yarn changeset` and it will guide you, in this case you can bump `patch`).

jeroenvisser101:
@dotansimha will do, and yes, I think it will be simpler, will update PR and add changeset, thanks for your input :)

dotansimha:
That's awesome, thank you @jeroenvisser101 !

PR #5278, opened 2020-12-16
fix: hooks can be functions and need to be normalized

KnisterPeter:
If a hook is a single function, it is filtered out here, but arrays for
functions would be kept.


dotansimha:
Great catch, thank you @KnisterPeter !

PR #5284, opened 2020-12-17
docs: generates.ts -> generated.ts

alecf:
this confounded me when trying to set up my config until I realized it was just a docs error


dotansimha:
Great catch, thanks :)

PR #5286, opened 2020-12-18
docs: fix typescript-type-graphql plugin name

cncolder:
fix plugin name so that we can generating `type-graphql` codes.


dotansimha:
Awesome, thank you @cncolder !

Issue #5263, opened 2020-12-14
Nested interfaces results in fragment fields getting picked too widely


voxpelli:
## To Reproduce

Sandbox link: https://codesandbox.io/s/sad-galileo-ljeo2?file=/types.ts

1. Create an interface `B`that extends an existing interface `A`
2. Do a GraphQL query where you use a (inline) fragment for `B` where you pick some fields specific to `B`
3. Generate the types for that query
4. There will be a `Pick<>` that tries to pick `B` fields from types only implementing `A`.

## Describe the bug

If one has a schema with:

```graphql
interface Entity {
  id: ID!
}

interface NamedEntity implements Entity {
  id: ID!
  name: String!
}
```

<details>
<summary>paired with the rest of the fields needed to have a complete schema</summary>

```graphql
type Query {
    entity(id: ID!): Entity!
}

interface Entity {
  id: ID!
}

interface NamedEntity implements Entity {
  id: ID!
  name: String!
}

type Session implements Entity {
  id: ID!
  data: String!
}

type User implements NamedEntity & Entity {
  id: ID!
  name: String!
}
```
</details>

And then does a query like:

```graphql
query entity {
    entity(id: 1) {
        id
        ... on NamedEntity {
            name
        }
    }
}
```

Then the generated type will contain the following erroneous code:

```typescript
export type EntityQuery = (
  { __typename?: 'Query' }
  & { entity: (
    { __typename?: 'Session' }
    & Pick<Session, 'name' | 'id'>
  ) | (
    { __typename?: 'User' }
    & Pick<User, 'name' | 'id'>
  ) }
);
```

Where it tries to pick a `NamedEntity` field `name` from the non-`NamedEntity` type `Session`, `Pick<Session, 'name' | 'id'>`, rather than correctly just pick `id`.

After checking back and forth I believe that this is caused by one interface extending another and that they somehow gets mixed up then.

dotansimha:
Thank you for reporting it @voxpelli . It seems like we somehow collects fields incorrectly in this use-case. 
I assume your intended output is: 
```
export type EntityQuery = (
  { __typename?: 'Query' }
  & { entity: (
    { __typename?: 'Session' }
    & Pick<Session, 'id'>
  ) | (
    { __typename?: 'User' }
    & Pick<User, 'name' | 'id'>
  ) }
);

```

right?

voxpelli:
[...]

PR #5290, opened 2020-12-18
fix(visitor-plugin-common): correctly handle fragments with an interface type condition that are spread on an interface field

n1ru4l:
Related #5263


dotansimha:
This is awesome, thank you @n1ru4l ❤️

PR #5291, opened 2020-12-18
encapsulation with generated definition files for graphql-modules

### LINKED ISSUE #5260: [experimental-graphql-modules]: option to create d.ts files per module generating
I saw in docs i need to import my namespaced typings for a module. I think we can improve this. So not exporting but simply declaring.

```
        presetConfig:
            filename: generated-types/module-types.d.ts
```

Then the file should not be:

```
import * as Types from '../../../generated-types/_graphql';
import * as gm from 'graphql-modules';

export namespace ProductModule {
```

but

```
declare namespace ProductModule {

  import * as Types from "../../../generated-types/_graphql";
  import * as gm from "graphql-modules";
```

Then i dont need to import the type, i can simply use it:
```
import { ProductDataloader } from '../providers/product.dataloader';

export const resolvers: ProductModule.Resolvers = {
    Query: {
        product: async (
            _: any,
            { id }: any,
            { injector }: GraphQLModules.Context,
```

What do you think? You like to explicitly import the namespace, or just to consume it?

maapteh:
when using encapsulation and having filenames ending with .d.ts then declare the namespace instead of exporting it.

related #5260


dotansimha:
Thank you @maapteh .
You don't have to add `shouldDeclare` in this one, since you can use the output type name inside the plugin function, so you can tell if it's a `d.ts` or not. (see https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-helpers/src/types.ts#L544-L554) 
Also, can you please add a `changeset` file? (Run `yarn changeset` and it will guide you..)

dotansimha:
I added the missing changeset. Thank you so much @maapteh !

PR #5302, opened 2020-12-21
Strip trailing whitespace

greim:
Auto-generated files in my project have some trailing whitespace. This change removes them.

### DIFF
--- a/.changeset/tiny-phones-act.md
+++ b/.changeset/tiny-phones-act.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+---
+
+Strip trailing whitespace from some output
--- a/packages/plugins/flow/flow/tests/__snapshots__/flow.spec.ts.snap
+++ b/packages/plugins/flow/flow/tests/__snapshots__/flow.spec.ts.snap
@@ -228,8 +228,8 @@ export type Scalars = {|
 |};
 
 export const myenumvalues = Object.freeze({
-  a: 'A', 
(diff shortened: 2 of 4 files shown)


dotansimha:
Thank you @greim , can you please share more details on this one? what's the cause for those whitespaces? because maybe it's an option to remove it at the source.

greim:
Hi @dotansimha

> what's the cause for those whitespaces?

As far as I can tell they're artifacts of how strings are concatenated.

> because maybe it's an option to remove it at the source.

I tried this, but simply removing the space character—while it fixed formatting in some places—caused bad formatting to pop up in other places. Simply post-processing the output seemed safer and simpler.

greim:
> you can add this to your `codegen.yml`

I'll try this, thanks @dctalbot!

MaximS:
One place where trailing spaces appear is [typescript/apollo-angular/src/visitor.ts:287](https://github.com/dotansimha/graphql-code-generator/blob/a9c52da3fb82a571e9f951944e7ca1ed8938b6ec/packages/plugins/typescript/apollo-angular/src/visitor.ts#L287)
Here
```
    const content = `
....
    ${this._namedClient(node)}
^^^^
```
if namedClient is empty the leading 4 spaces are output.

dotansimha:
Closing. We are not able to put effort on removing whitespaces. PRs are always welcome of course! 
You can integrate prettier with codegen, and it will remove most of these. Also, you can ignore lint / checks on generated files if it's an issue for you.

Issue #5238, opened 2020-12-11
@skip, @include directives are sometimes ignored


deftomat:
**Describe the bug**
 

`@skip` and `@include` directives are ignored when they are applied on "object".

Apollo server handles them correctly with no issue. 

Also, the example in GraphQL docs is basically the same: https://graphql.org/learn/queries/#directives

**Doesn't work - `chat` is required:**
```
query User($withChat: Boolean!) {
  user {
    name
    chat @include(if: $withChat) {
      unread
    }
  }
}
```

**Work as expected - `unread` is optional:**
```
query User($withChat: Boolean!) {
  user {
    name
    chat {
      unread @include(if: $withChat)
    }
  }
}
```

**Environment:**

```@graphql-codegen/typescript-operations@1.17.12```

dotansimha:
We can close this one, @gilgardosh fixed it :)

PR #5326, opened 2020-12-28
@skip, @include directives for objects

gilgardosh:
Related #5238

### DIFF
--- a/packages/plugins/flow/operations/tests/flow-documents.spec.ts
+++ b/packages/plugins/flow/operations/tests/flow-documents.spec.ts
@@ -1220,17 +1220,23 @@ describe('Flow Operations Plugin', () => {
 
         type User {
           id: String!
-          name: String
-          address: String!
+          name: String!
+          address: Address!
+        }
+
+        type Address {
+          city: String!
         }
       `);
 
       const ast = parse(/* GraphQL */ `
-        query user($showAddress: Boolean!) {
+        query user($showAddress: Boolean!, $showName: Boolean!) {
           user {
             id
-            name
-            address @include(if: $showAddress)
+            name @include(if: $showName)
+            address @include(if: $showAddress) {
+              city
+            }
           }
         }
       `);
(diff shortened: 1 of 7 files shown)


packages/plugins/flow/operations/tests/flow-documents.spec.ts
```diff
@@ -1251,9 +1257,13 @@ describe('Flow Operations Plugin', () => {
       expect(result).toBeSimilarStringTo(`
       export type UserQueryVariables = {
         showAddress: $ElementType<Scalars, 'Boolean'>,
+        showName: $ElementType<Scalars, 'Boolean'>,
       };
 
-      export type UserQuery = { user: $MakeOptional<$Pick<User, { id: *, name?: *, address: * }>, { address: * }> };
+      export type UserQuery = { user: ({
```
gilgardosh, line 1263: @dotansimha Please make sure that's the wanted outcome

dotansimha:
Thanks @gilgardosh , what about `preResolveTypes`? can you please add a test for it?

dotansimha:
Thanks @gilgardosh . as discussed offline - please clean it a bit and then we can merge :)

dotansimha:
Awesome work. thank you @gilgardosh !

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -372,14 +372,15 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
...
 
       linkFields.push({
         alias: field.alias ? this._processor.config.formatNamedField(field.alias.value, selectedFieldType) : undefined,
-        name: this._processor.config.formatNamedField(field.name.value, selectedFieldType),
+        name: this._processor.config.formatNamedField(field.name.value, selectedFieldType, isConditional),
         type: realSelectedFieldType.name,
         selectionSet: this._processor.config.wrapTypeWithModifiers(
           selectionSet.transformSelectionSet().split(`\n`).join(`\n  `),
-          selectedFieldType
+          isConditional ? realSelectedFieldType : selectedFieldType
```
raysuelzer, line 383: Breaks arrays as this strips the interable wrapper (Array<T>) from the base type when in a directive. 

EG if people is a list: 
```
people @include(if: $withPeople) { 
 id
}
```
Result is that in typescript the output for People is no longer Maybe<Array<Person>> but just Maybe<Person>

PR #5392, opened 2021-01-13
Remove`request` from `@graphql-codegen/cli`.

wtgtybhertgeghgtwtg:
Please tell me if I am wrong, but it does not seem to be used.  Additionally, it is deprecated and Yarn gives a warning to that effect when installing `@graphql-codegen/cli`.


dotansimha:
You are right, great catch. 
thank you @wtgtybhertgeghgtwtg

wtgtybhertgeghgtwtg:
Thank you for merging.

PR #5412, opened 2021-01-16
[Typescript-mongodb] AvoidOptionals config in additionalFields

mtperessoni:
Hello guys!
Today i was working with typescript-mongodb and i faced with a "problem" in generated types. When i was using additionalFields in @entity directive i couldn't add a field as optional.
I realized that it is possible to add a field as optional just by adding optional chaining to the path. The problem with this is that the avoidOptionals config is not being considered.
So I made the fix, now it is possible to use the avoidOptionals with additionalFields. I added a test and also made a fix in the dos to help users who have the same problem.


dotansimha:
Thank you @mtperessoni , awesome work. Worth mentioning [this comment](https://github.com/dotansimha/graphql-code-generator/discussions/5393#discussioncomment-293264) in the context of this plugin.

PR #5414, opened 2021-01-17
fix(plugin: graphql-requests): Fix argument types in auto-generated methods

remew:
`GraphQLClient`'s `request` method accepts `Dom.RequestInit['headers']` parameter (it is same as `HeadersInit`).
but auto-generated methods accept `Headers` parameter only.
so I can not pass an object to that methods.
I fixed it.

### DIFF
--- a/.changeset/swift-rules-do.md
+++ b/.changeset/swift-rules-do.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-graphql-request': patch
+---
+
+fix(plugin: graphql-requests): Fix argument types in auto-generated methods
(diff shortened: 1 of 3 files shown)


dotansimha:
Thank you @remew , great catch!

lilred:
I'm on 3.0.1 and I'm still facing this issue. `graphql-request` version 3.4.0.

remew:
@lilred it is my careless mistake.
I will fix it and create new pull request as soon as possible.

workaround is replacing all `HeadersInit` to `Headers`  in auto-generated `.ts` files (or just use 3.0.0).

lilred:
I used 3.0.0 with the additional plugin `- add: import {Headers} from 'request';` and it works for me.

PR #5419, opened 2021-01-19
docs: remove back tick from description

smhmd:
I'm not sure if the intention was to link somewhere else.


dotansimha:
Thank you!

Issue #5334, opened 2020-12-29
expose react-query Query Keys to allow direct interactions with the cache


Merott:
**Is your feature request related to a problem? Please describe.**

react-query uses [Query Keys](https://react-query.tanstack.com/guides/query-keys) to manage caching. Methods like [setQueryData](https://react-query.tanstack.com/reference/QueryClient#queryclientsetquerydata) and [getQueryData](https://react-query.tanstack.com/reference/QueryClient#queryclientgetquerydata) need this key.

`typescript-react-query` automatically generates an array-based key, using the query names and variables. Sweet! 💯

The problem is that the key isn't exported from the generated file, which means we have to hard-code it in our own source files, and that can be fragile.

**Describe the solution you'd like**

Not sure what the best solution is here. Here's an idea though… Given this:

```ts
export const useCourses = <TData = CoursesData, TError = unknown>(
  variables?: CoursesVariables,
  options?: UseQueryOptions<CoursesData, TError, TData>,
) =>
  useQuery<CoursesData, TError, TData>(
    ['Courses', variables],
    fetcher<CoursesData, CoursesVariables>(CoursesDocument, variables),
    options,
  )
```

Can we attach a `getKey` method to the hook, like this?

```ts
useCourses.getKey = (variables?: CoursesVariables) => [
  'Courses',
  variables,
]
```

That way, instead of doing this in my React code:

```ts
// no strict typing, and hardcoded 'Courses' 👎
const COURSES_QUERY_KEY = ['Courses', { someVar: myVar }]
```

I can do this:

```ts
// strictly typed variables, and no hardcoded strings 👍
const COURSES_QUERY_KEY = useCourses.getKey({ someVar: myVar })
```

**Describe alternatives you've considered**

The above solution works, but we have to attach something to the hook itself, which isn't ideal.

The alternative would be to expose a method and accept the hook as an input parameter. Specifying strict typings for this would be quite tricky though.

<details>
<summary>See an example...</summary>

```ts
const COURSES_QUERY_KEY = getKey(useCourses, { someVar: myVar })

export const getKey = (
  // specifying the types will be complex -- union types?
  query,
  variables,
) => {
  switch (query) {
    case useCourses: {
      return ['Courses', variables]
    }

    case useAnotherQuery: {
      return ['AnotherQuery', variables]
    }

    default: {
      throw new Error('That was unexpected!')
    }
  }
}
```

</details>

---

[...]

PR #5423, opened 2021-01-19
Expose query keys for react-query (#5334)

andrei-datcu:
As suggested in https://github.com/dotansimha/graphql-code-generator/issues/5334

for each `useTestQuery` you'll get an ```useTestQuery.getKey = (variables: TestQueryVariables) => ['test', variables];```
This option can be turned on by setting `exposeQueryKeys: true` in config


dotansimha:
This is awesome work! Thank you so much @andrei-datcu !

Issue #5360, opened 2021-01-05
Enums are generated incorrectly for value with ' 


retyui:
**Describe the bug**

TS codegen don't escape `'` symbols in enum values (see sandbox)

```ts
export enum Foo {
  DoNot = 'Don't work' // invalid ts syntax
}
```

**To Reproduce**

https://codesandbox.io/s/billowing-platform-x7l8v?file=/types.ts

1. My GraphQL schema:

```js
const schema = new GraphQL.GraphQLSchema({
  query: new GraphQL.GraphQLObjectType({
    name: "Query",
    fields: {
      foo: {
        type: new GraphQL.GraphQLEnumType({
          name: "Foo",
          values: {
            DoNot: {
              value: "Don't work" // Hello am a buggy case
            }
          }
        })
      }
    }
  })
});
```

3. My `codegen.yml` config file:

```js
const config = {
  filename: outputPath,
  schemaAst: schema,
  plugins: [
    // Each plugin should be an object
    {
      typescript: {
        declarationKind: "interface"
      }
    }
  ],
  pluginMap: {
    typescript: typescriptPlugin
  }
}
```

**Expected behavior**

Should generate valid ts enum

```
export enum Foo {
  DoNot = 'Don\'t work' // the sterilizer should escape ' symbol
  DoNot = "Don't work" // or use " symbol for strings that includes ' symbols
}
```

**Environment:**

- OS: linux
- `@graphql-codegen/...`: latest
- NodeJS: 14

**Additional context**

dotansimha:
Fixed in: https://github.com/dotansimha/graphql-code-generator/pull/5432

dotansimha:
Fix available in `@graphql-codegen/typescript@1.20.1`

PR #5432, opened 2021-01-21
[typescript/flow] Escape enum values defined in GraphQLSchema correctly

dotansimha:
Related: https://github.com/dotansimha/graphql-code-generator/issues/5360

### DIFF
--- a/.changeset/silly-carpets-walk.md
+++ b/.changeset/silly-carpets-walk.md
@@ -1,0 +1,7 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+'@graphql-codegen/flow': patch
+'@graphql-codegen/typescript': patch
+---
+
+Correctly escape enum values defined in the GraphQLSchema object
--- a/packages/plugins/other/visitor-plugin-common/src/enum-values.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/enum-values.ts
@@ -3,6 +3,10 @@ import { GraphQLSchema, isEnumType, GraphQLEnumType } from 'graphql';
 import { DetailedError } from '@graphql-codegen/plugin-helpers';
 import { parseMapper } from './mappers';
 
+function escapeString(str: string) {
+  return str.replace(/\\/g, '\\\\').replace(/\n/g, '\\n').replace(/'/g, "\\'");
+}
+
 export function parseEnumValues({
   schema,
   mapOrStr = {},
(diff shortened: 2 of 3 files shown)


cometkim:
Hello @dotanSimha It seems that there is a type error not detected by TypeScript here.

https://github.com/cometkim/gatsby-plugin-typegen/issues/143

dotansimha:
Thanks @cometkim , working on a fix :)

dotansimha:
@cometkim should be fixed by https://github.com/dotansimha/graphql-code-generator/commit/7cc37da85feeaf1aee26fde49472d8136b0be2d2 , i'll release it in a few minutes :)

dotansimha:
@cometkim should be fixed in `@graphql-codegen/flow@1.18.4`, can you please try?

cometkim:
@dotansimha Is this change only affect flow? I guess @tony is using TypeScript.

@tony Can you please help to confirm this is fixed? or can you share your project configuration? I can't reproduce the issue in the latest version of the gatsby starter.

tony:
@cometkim I may not be able to get to this until tomorrow (same time)

When I do, Will there be a new version of `gatsby-plugin-typegen` to try again? Or should I pin `@graphql-codegen/flow`?

cometkim:
https://github.com/cometkim/gatsby-plugin-typegen/issues/143#issuecomment-767611740

dotansimha:
> @cometkim I may not be able to get to this until tomorrow (same time)
> 
> 
> 
> When I do, Will there be a new version of `gatsby-plugin-typegen` to try again? Or should I pin `@graphql-codegen/flow`?

Should fix both :)

cometkim:
@dotansimha Another issue reported on the same path. `[object Object]` appears in the stacktrace :thinking: 

https://github.com/cometkim/gatsby-plugin-typegen/issues/143#issuecomment-767681411

cometkim:
Hmm, I see...

https://github.com/dotansimha/graphql-code-generator/blob/49242c20cd985addba5f74442d435a527bdbe163/packages/plugins/other/visitor-plugin-common/src/types.ts#L21-L23

The definition of `EnumTypesMap` is telling that the value could be a number. I think this might be the case:

```js
new GraphQLEnumType({
  name: 'RGB',
  values: {
    RED: { value: 0 },
    GREEN: { value: 1 },
    BLUE: { value: 2 }
  }
});
```

The graphql-codegen core API does not restrict the inclusion of user-created enum objects in the input schema. We cannot assume that there will always be only strings.

cometkim:
https://graphql.org/graphql-js/type/#graphqlenumtype

> GraphQL serializes Enum values as strings, however internally Enums can be represented by any kind of type, often integers.

@dotansimha Can we provide meaningful artifacts even if it's not a string? Otherwise, I think it would be the correct behavior to ignore them all.

PR #5435, opened 2021-01-21
[React Query] Allow the custom fetcher to lazily receive mutation variables

andrei-datcu:
Currently the generated hooks use the custom fetcher in the following way:

query hooks: They invoke the fetcher inline. I.e you could **use react hooks in your fetcher**
mutation hooks: They create a lambda function that will invoke your fetcher at a later time, when react query performs your mutation. i.e you cannot use react hooks in your fetcher.

**Wait, why would you want to use react hooks in your fetcher?**
My usecase is rather simple: I have to inject some headers and the fetch url at runtime. So in order to not pass them to the queries and mutations everywhere in the code, I could keep them in a `ReactContext` and retrieve them through `useContext` in the fetcher.

**The solution**
This PR extends the `fetcher` option and allows one to pass an object with the following interface:
```
{
   func: string; // this is the custom identifier eg ./myFile#useMyFetcher
   lazyVariables?: bool // this specifies that useMyFetcher will not receive the query variables itself. *The async function it returns will receive those variables instead*
}
```
Omitting `lazyVariables` is equivalent to just passing the custom fetcher string and getting the old functionality.

If `lazyVariables` is `true`, then fetcher should look like this:

```
function useMyFetcher<TData, TVariables>(document: string) {
  // it's now safe to call any react hook here
  // ex:
  // const url = useContext(URLContext);
   return async (variables?: TVariables) {
      await fetch(...);
   }
}
```

### DIFF
--- a/.changeset/wicked-onions-complain.md
+++ b/.changeset/wicked-onions-complain.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-react-query': minor
+---
+
+Allow fetcher to receive variables lazily so it can use react hooks
--- a/packages/plugins/typescript/react-query/src/config.ts
+++ b/packages/plugins/typescript/react-query/src/config.ts
@@ -1,7 +1,7 @@
 import { RawClientSideBasePluginConfig } from '@graphql-codegen/visitor-plugin-common';
 
 export type HardcodedFetch = { endpoint: string; fetchParams?: Record<string, any> };
-export type CustomFetch = { func: string; lazyVariables?: boolean } | string;
+export type CustomFetch = { func: string; isReactHook?: boolean } | string;
 
 /**
  * @description This plugin generates `React-Query` Hooks with TypeScript typings.
--- a/packages/plugins/typescript/react-query/src/config.ts
+++ b/packages/plugins/typescript/react-query/src/config.ts
@@ -1,6 +1,7 @@
 import { RawClientSideBasePluginConfig } from '@graphql-codegen/visitor-plugin-common';
 
 export type HardcodedFetch = { endpoint: string; fetchParams?: Record<string, any> };
+export type CustomFetch = { func: string; lazyVariables?: boolean } | string;
 
 /**
  * @description This plugin generates `React-Query` Hooks with TypeScript typings.
@@ -28,7 +29,7 @@ export interface ReactQueryRawPluginConfig
    * - `file#identifier` - You can use custom fetcher method that should implement the exported `ReactQueryFetcher` interface. Example: `./my-fetcher#myCustomFetcher`.
    * - `graphql-request`: Will generate each hook with `client` argument, where you should pass your own `GraphQLClient` (created from `graphql-request`).
    */
-  fetcher?: 'fetch' | HardcodedFetch | 'graphql-request' | string;
+  fetcher?: 'fetch' | HardcodedFetch | 'graphql-request' | CustomFetch;
 
   /**
    * @default false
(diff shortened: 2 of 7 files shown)


dotansimha:
Thank you @andrei-datcu , a few notes:
1. Can you please rebase this PR? we added a few other changes, and it conflicts now.
2. Can you please add a changeset? (run `yarn changeset` and it will guide you)
3. Do we need to expect a breaking change here in mutations? what's the effect of these changes?

andrei-datcu:
Resolved conflicts and pushed the changeset.
@dotansimha: To answer your last question:
As long as the *options remain unchanged, the output remains unchanged*. So clients using the old `fetcher: "./myFile#myFetcher"` option will not see any difference in the generated code.

Whenever one uses `fetcher: {func: "./myFile#myFetcher", lazyVariables: true}` the generated code will look differently both for mutations and queries.

1) Queries
no lazy variables:
```
export const useTestQuery = <
          TData = TTestQuery,
          TError = unknown
        >(
          variables?: TTestQueryVariables,
          options?: UseQueryOptions<TTestQuery, TError, TData>
        ) =>
        useQuery<TTestQuery, TError, TData>(
          ['test', variables],
          myCustomFetcher<TTestQuery, TTestQueryVariables>(TestDocument, variables),
          options
        );
```

lazyVariables: true

```
export const useTestQuery = <
          TData = TTestQuery,
          TError = unknown
        >(
          variables?: TTestQueryVariables,
          options?: UseQueryOptions<TTestQuery, TError, TData>
        ) =>
        useQuery<TTestQuery, TError, TData>(
          ['test', variables],
          useCustomFetcher<TTestQuery, TTestQueryVariables>(TestDocument).bind(null, variables),
          options
        );
```

2) Mutations
no lazy variables:

```
export const useTestMutation = <
        TError = unknown,
        TContext = unknown
      >(options?: UseMutationOptions<TTestMutation, TError, TTestMutationVariables, TContext>) =>
      useMutation<TTestMutation, TError, TTestMutationVariables, TContext>(
        (variables?: TTestMutationVariables) => myCustomFetcher<TTestMutation, TTestMutationVariables>(TestDocument, variables)(),
        options
      );
```

lazyVariables: true

```
export const useTestMutation = <
        TError = unknown,
        TContext = unknown
      >(options?: UseMutationOptions<TTestMutation, TError, TTestMutationVariables, TContext>) =>
      useMutation<TTestMutation, TError, TTestMutationVariables, TContext>(
        useCustomFetcher<TTestMutation, TTestMutationVariables>(TestDocument),
        options
      );
```

dotansimha:
Sounds good, thank you for explaining it, @andrei-datcu ❤️ 
Any chance I can ask you for another rebase? (we are going to have a very interesting version of `react-query` plugin with many new features!)
We also need to document this new fetcher here:  https://github.com/dotansimha/graphql-code-generator/blob/master/website/docs/plugins/typescript-react-query.md

Thank you so much!

andrei-datcu:
@dotansimha Rebased and updated the documentation. While writing it, I did have some second thoughts with the parameter's name (`lazyVariables`). Do you think a more explicit name (eg. `isReactHook: true`) would make it more discoverable? `lazyVariables`'s advantage is its generality, yet I struggle to find out another use case for it apart from the hooks scenario.
I could change the name, your call.

dotansimha:
> @dotansimha Rebased and updated the documentation. While writing it, I did have some second thoughts with the parameter's name (`lazyVariables`). Do you think a more explicit name (eg. `isReactHook: true`) would make it more discoverable? `lazyVariables`'s advantage is its generality, yet I struggle to find another use case for it apart from the hooks scenario.
> I could change the name, your call.

I think `isReactHook` might be a better fit here? Just because it feels like more related to the actual behaviour change.

andrei-datcu:
@dotansimha Renamed it to `isReactHook`. Please let me know how it looks now :)

ardatan:
Thank you @andrei-datcu ! This seems a cool feature that adds more flexibility!

yassinebridi:
This is such a nice feature to have, thank you.

PR #5442, opened 2021-01-23
Documentation: Provide a dotenv example

jonathanstanley:
This extends the documentation on `dotenv` and adds a reminder that sensitive keys may be output into the generated code.


dotansimha:
Thank you @jonathanstanley ,
Can you please explain the note you added: `The env values will be saved in the generated code output.  Be careful not to commit code with sensitive keys.` ? We don't use the environment variables for the generated output, it's only for parsing the YAML config file.

jonathanstanley:
Ah, yes good question.  

In my case, it was a plugin that was reading the value and adding it to the output .ts to create a hard-coded fetcher with the `APIKEY` in the header.  Or if someone were to create a custom plugin, the possibility for those parsed dotenv values to find their way into the output might be overlooked.

It should probably be written as `may be` instead of `will be`.

ardatan:
Thanks @jonathanstanley !

Issue #5402, opened 2021-01-15
Using flow plugin produces `Maybe` type


ardok:
**Describe the bug**
```
    "@graphql-codegen/add": "^2.0.2",
    "@graphql-codegen/cli": "^1.20.0",
    "@graphql-codegen/flow": "^1.18.2",
    "@graphql-codegen/flow-operations": "^1.18.5",
```

```
schema: ./.graphql/schema.json
documents:
  - src/queries/*.js
generates:
  ./src/queries/types.js:
    plugins:
      - add:
          content:
            - '/* eslint-disable */'
            - '// @generated'
      - flow
      - flow-operations
    config:
      scalars:
        BigInt: number
      typesPrefix: Gql_
```

```
// this got generated, throwing error for flowtype
type $MakeOptional<T, K: Object> = $Diff<T, K> & $ObjMapi<$Rest<T, K>,<SubKey>(k: SubKey) => Maybe<$ElementType<T, SubKey>>>;
```

Flowtype error:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ types.js:35:48

Cannot resolve name Maybe.

     32│   <Key>(k: Key) => $ElementType<Origin, Key>
     33│ >;
     34│ type $MakeOptional<T, K: Object> = $Diff<T, K> &
     35│   $ObjMapi<$Rest<T, K>, <SubKey>(k: SubKey) => Maybe<$ElementType<T, SubKey>>>;
     36│ export type Gql_UserQueryVariables = {};
     37│
     38│ export type Gql_UserQuery = {
```

**To Reproduce**
https://codesandbox.io/s/inspiring-sara-50zc9?file=/types.js

```
yarn generate
yarn flow
```

1. My GraphQL schema:

```graphql
type Query {
    user(id: ID!): User!
}

type User {
    id: ID!
    username: String!
    email: String!
}
```

2. My GraphQL operations:

```graphql
None
```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  types.js:
    plugins:
      - add:
          content:
            - '/* eslint-disable */'
            - '// @generated'
      - flow
      - flow-operations
    config:
      scalars:
        BigInt: number
      typesPrefix: Gql_
```

**Expected behavior**
[...]

PR #5448, opened 2021-01-25
fix(flow-operation): remove irrelevant `Maybe` definition

gilgardosh:
Related: #5402


dotansimha:
@gilgardosh thanks! 
1. Please fix ci
2. Add changeset

PR #5461, opened 2021-01-27
New Plugin: Urql Introspection

ardatan:


packages/plugins/other/urql-introspection/src/index.ts
```diff
@@ -0,0 +1,103 @@
+import { extname } from 'path';
+
+import { PluginFunction, PluginValidateFn, Types } from '@graphql-codegen/plugin-helpers';
+import { GraphQLSchema } from 'graphql';
+import { getIntrospectedSchema, minifyIntrospectionQuery } from '@urql/introspection';
+
+/**
+ * @description This plugin generates an introspection file but only with Interfaces and Unions, based on your GraphQLSchema.
+ *
+ * If you are using `urql` and your schema contains `interface` or `union` declaration, it's recommended to use provide an introspection to enable Schema Awareness.
```
kitten, line 10: The only thing I've noticed is that this comment may be from a different plugin? We could probably just leave a link to the schema awareness docs page here but since it includes more than just interfaces and unions I'm concerned it may just be a little confusing ✌️

ardatan: Yes you're right. The PR is not ready yet in terms of documentation :) I'll make it better.

ardatan: @kitten Do you think it is ok now ? :)

ardatan: @kitten Do you think it is fine now? Let me know if you want something specifically in the docs :)

PR #5502, opened 2021-02-05
Added support for multi-search

tvvignesh:
@Urigo @ardatan Only a very small change is needed to be made to enable multi-search on the client side. 

Have added the client side changes to this PR which works across multiple Guild websites.

In addition to it, I have also added a README to the website folder to document running the website locally

**NOTE:** [The indexer](https://github.com/tvvignesh/algolia-docsearch-indexer) has to be run manually or on schedule after setting it up in a server for updating the search results in  the index as the docs on the relevant websites are changed.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -31,3 +31,9 @@ jobs:
           FOLDER: website/build
           CLEAN: true
           SINGLE_COMMIT: true
+      - name: Index Docs
+        uses: darrenjennings/algolia-docsearch-action@master
+        with:
+          algolia_api_key: ${{ secrets.ALGOLIA_API_KEY }}
+          algolia_application_id: ${{ secrets.ALGOLIA_APPLICATION_ID }}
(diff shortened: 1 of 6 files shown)


ardatan:
@tvvignesh Awesome! Do you think we can run this indexer in the website workflow of CI?

tvvignesh:
@ardatan Yup. We can if we want to run it on changes rather than on schedule. Essentially what needs to be done is, a docker container to be run on changes.

One thing though is that if the indexing ends up happening frequently (cause it can be triggered by multiple repos), we may end up exceeding the API quota from algolia.

Once the approach is decided, I can send a new PR for the same.

ardatan:
@tvvignesh I think having it in website deployment workflow is fine! Could you add that :)

tvvignesh:
@ardatan Sure. Will try setting up a github actions pipeline on my end and update this PR tomorrow.

PR #5509, opened 2021-02-07
allow csharp plugin to emit c# 9 record types instead of classes

oising:
Add an `emitRecords: true` to config, and it will spit out the immutable syntax (init-only) for c# 9 records, e.g. `record MyType(int Age, string Name) : <interfacelist>` etc. Note that the fields/parameters are PascalCased, as they should be in this case since they are exposed as properties.

```yaml
schema:
  - ./types.graphql
  - ./units.graphql
generates:
  ./src/types.cs:
    plugins:
    - c-sharp
config:
    namespaceName: MyNamespace.Types
    emitRecords: true
    scalars:
        ISO8601DateTime: DateTime
```

~~Oh, and I squeaked in a fix for enforcing Pascal Case on C# public properties too. EDIT: which broke tests expecting lower case property names, oops. Will fix tomorrow.~~

input:

```graphql
type Account {
  id: ID!
  login: String!
  lastLogin: ISO8601DateTime
  locations: [Location!]
  disabled: Boolean!
}
```

example output (for records)

```c#
#region Account
    public record Account(string Id, string Login, DateTime LastLogin, List<Location> Locations, bool Disabled) {
      #region members
      [JsonProperty("id")]
      public string Id { get; init; } = Id;
    
      [JsonProperty("login")]
      public string Login { get; init; } = Login;
    
      [JsonProperty("lastLogin")]
      public DateTime LastLogin { get; init; } = LastLogin;
    
      [JsonProperty("locations")]
      public List<Location> Locations { get; init; } = Locations;
    
      [JsonProperty("disabled")]
      public bool Disabled { get; init; } = Disabled;
      #endregion
    }
    #endregion
```
(p.s. I dislike regions, but the class generator uses them already)

### DIFF
--- a/.changeset/four-pillows-explain.md
+++ b/.changeset/four-pillows-explain.md
@@ -1,0 +1,20 @@
+---
+'@graphql-codegen/c-sharp': patch
+---
+
+This release adds support to optionally emit c# 9 records instead of classes.
+
+To enable this, add `emitRecords: true` to your codegen yaml or json configuration file. Example:
+
+```yaml
+schema: './types.graphql'
+generates:
+  ./src/types.cs:
+    plugins:
+    - c-sharp
+config:
+    namespaceName: My.Types
+    emitRecords: true
+    scalars:
+      DateTime: DateTime
+```
--- a/.changeset/four-pillows-explain.md
+++ b/.changeset/four-pillows-explain.md
@@ -1,5 +1,5 @@
 ---
-'@graphql-codegen/c-sharp': patch
+'@graphql-codegen/c-sharp': minor
 ---
 
 This release adds support to optionally emit c# 9 records instead of classes.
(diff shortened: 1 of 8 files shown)


oising:
It seems to work -- hah! that doesn't happen to me too often. Tests are the next thing to do -- there are obvious tests in the plugin's tests folder -- I guess that's it?

oising:
Ok -- just realized that interfaces are not cased correctly for records (using PascalCase) and also the interfaces should be getters only (because records are get init / immutable)

ardatan:
Thank you @oising !
@rus-art What do you think about these changes?

oising:
> Thank you @oising !
> @rus-art What do you think about these changes?

I think I can do some more clean up to remove dupe code between records/classes but I had an itch to scratch!

ardatan:
@oising To be honest, I don't know much about C# and GraphQL :) If you feel this can be done later, we are ok to merge this now.

oising:
> @oising To be honest, I don't know much about C# and GraphQL :) If you feel this can be done later, we are ok to merge this now.

Sounds good to me! I think I covered all of the edge cases for records (immutable types for .net) -- I'll keep an eye on the issues and if you tag me, I'll respond and address anything that comes up.

ardatan:
Thank you very much @oising !

oising:
> Thank you very much @oising !

You're most welcome -- thank you for writing such a nice tool!

Issue #5289, opened 2020-12-18
GraphQL Modules preset fails when there are multiple schema entries in config


tvvignesh:
**Describe the bug**
I use `graphql-modules` preset with codegen and it works well. It fails when my `.graphqlrc.yml` file has multiple entries for schema. That is,

**This works:**

```yml
config:
      assumeValidSDL: true
      assumeValid: true
    schema:
      # - './packages/core/gql/directive.gql'
      # - './packages/core/gql/core.gql'
      - './packages/modules/**/schema/*.gql'
    extensions:
      codegen:
        generates:
          ./packages/modules/:
            schema: ./packages/modules/*/schema/*.gql
            preset: graphql-modules
 ```

**This doesn't:**

```yml
config:
      assumeValidSDL: true
      assumeValid: true
    schema:
      - './packages/core/gql/directive.gql'
      - './packages/core/gql/core.gql'
      - './packages/modules/**/schema/*.gql'
     extensions:
      codegen:
        generates:
          ./packages/modules/:
            schema: ./packages/modules/*/schema/*.gql
            preset: graphql-modules
```

I get this error when running codegen:

<img width="773" alt="1" src="https://user-images.githubusercontent.com/1165845/102632483-302a7500-4175-11eb-96f3-ad2f1d38f82b.PNG">

The files do exist in the right locations and are valid.

I am not sure if this is an error or is expected. Have separated service and module codegen into 2 separate projects to get this working now. In the above case, there are some GQL files which does not make sense within modules, so I have it outside the modules and was doing codegen for both in the same project causing the error.

**Deps:**

```
"@graphql-codegen/add": "^2.0.2",
"@graphql-codegen/cli": "^1.19.4",
"@graphql-codegen/graphql-modules-preset": "^1.1.0",
"@graphql-codegen/typescript": "^1.19.0",
"@graphql-codegen/typescript-operations": "^1.17.12",
"@graphql-codegen/typescript-resolvers": "^1.18.0",
```

**CC:** @dotansimha

tvvignesh:
**UPDATE:** Its working well after splitting it into multiple projects. If this is not a bug, feel free to close it. Thanks.

dotansimha:
@tvvignesh any chance for a minimal reproduction? :)

tvvignesh:
[...]

PR #5523, opened 2021-02-09
Fixed issue with preset breaking when pattern doesn't match (Fixes #5…

### LINKED ISSUE #5: Add CLI

tvvignesh:
- Applied check for relativePath before using those for extracting the module directories

Related: 
https://github.com/dotansimha/graphql-code-generator/issues/5289


packages/presets/graphql-modules/src/utils.ts
```diff
@@ -139,14 +139,19 @@ export function groupSourcesByModule(sources: Source[], basePath: string): Recor
   const grouped: Record<string, Source[]> = {};
 
   sources.forEach(source => {
-    // PERF: we could guess the module by matching source.location with a list of already resolved paths
-    const mod = extractModuleDirectory(source.name, basePath);
+    const normalizedFilepath = normalize(source.name);
```
Urigo, line 142: this is a bit duplicated?

Maybe I'm wrong but is there a chance in `extractModuleDirectory` that would save this duplication and we could just call that?

tvvignesh: You're right. I thought of moving it to a separate function but ignored it later since it did not make sense for me to make a separate function for 3 lines of code. But maybe it makes sense over time. Will update this PR.

Issue #5433, opened 2021-01-21
Add `baseOptions` customization for `typescript-react-apollo` plugin.


dmitrif:
**Is your feature request related to a problem? Please describe.**
In a codebase that uses multiple GraphQL endpoints, it is common to use `context.clientName` to direct the query to the correct ApolloLink. For example:

```
const client = new ApolloClient({
  link: ApolloLink.split(
    operation => operation.getContext().clientName === "third-party",
    thirdPartyLink, // <= apollo will send to this if context.clientName is "third-party"
    myLink // <= otherwise will send to this
  )
  // other options
});
```

From a codegen perspective, the third party queries are often delimited with a custom gql tag and extracted by configuring `pluckConfig`. The generated hooks / HOC's however currently do not have a way to provide a default context / baseOptions. Current generation example:

```
export function useArticleQuery(baseOptions: ApolloReactHooks.QueryHookOptions<Types.ArticleQuery, Types.ArticleQueryVariables>) {
        return ApolloReactHooks.useQuery<Types.ArticleQuery, Types.ArticleQueryVariables>(ArticleDocument, baseOptions);
      }
```

**Describe the solution you'd like**
It'd be great if there was a way to supply an object to `pluginOptions` such as:
```
config: {
  baseOptions: {
    context: {
      clientName: 'third-party'
    }
  }
}
```

and have the generated code extend the base options for HOC's / Hooks. Example generated:
```
export type const defaultBaseOptions = {
  context: {
    clientName: 'third-party'
  }
}

export function useArticleQuery(baseOptions: ApolloReactHooks.QueryHookOptions<Types.ArticleQuery, Types.ArticleQueryVariables>) {
       const options = deepMerge(defaultBaseOptions, baseOptions);
        return ApolloReactHooks.useQuery<Types.ArticleQuery, Types.ArticleQueryVariables>(ArticleDocument, options);
      }
```

**Describe alternatives you've considered**
1. Doing an AST walk post graphql-codegen run and adding the above.
2. Ignoring the incovinience.

Thank you!

trompx:
On my side I have a custom fetcher (react-query) which takes as first parameter the context. I am looking for a way to have that context passed as the first parameter in the query hooks generated by codegen (not currently possible as per https://graphql-code-generator.com/docs/plugins/typescript-react-query#using-custom-fetcher).

PR #5530, opened 2021-02-10
Add baseOptions customization for typescript-react-apollo

luxer066:
Hi 👋! 

With my team made an adaptation to be able to pass a context name to the generator this helps us to have multiple apollo connections and generate ready-to-use queries pointing to their context.

We follow the recommendation of Dmitri Farkov made in issues [#5433](https://github.com/dotansimha/graphql-code-generator/issues/5433)

Works correctly by passing: 

`generates:

    config:
      defaultBaseOptions: {
        context: { clientName: "yourClientName" },
      }
`

### DIFF
--- a/dev-test/githunt/types.reactApollo.customSuffix.tsx
+++ b/dev-test/githunt/types.reactApollo.customSuffix.tsx
@@ -4,6 +4,7 @@ export type Maybe<T> = T | null;
 export type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
 export type MakeOptional<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]?: Maybe<T[SubKey]> };
 export type MakeMaybe<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]: Maybe<T[SubKey]> };
+const defaultOptions = {};
 /** All built-in and custom scalars, mapped to their actual values */
 export type Scalars = {
   ID: string;
@@ -355,9 +356,10 @@ export function useOnCommentAddedSubscription(
     OnCommentAddedSubscriptionVariables
   >
 ) {
+  const options = { ...defaultOptions, ...baseOptions };
   return Apollo.useSubscription<OnCommentAddedSubscriptionMyOperation, OnCommentAddedSubscriptionVariables>(
     OnCommentAddedDocument,
-    baseOptions
+    options
   );
 }
 export type OnCommentAddedSubscriptionHookResult = ReturnType<typeof useOnCommentAddedSubscription>;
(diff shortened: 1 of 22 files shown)


packages/plugins/typescript/react-apollo/src/visitor.ts
```diff
@@ -329,14 +331,19 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
...
     this.imports.add(this.getApolloReactHooksImport(false));
 
+    const options = this.rawConfig.defaultBaseOptions
+      ? JSON.stringify(this.rawConfig.defaultBaseOptions)
+      : 'baseOptions';
     const hookFns = [
       `export function use${operationName}(baseOptions${
         hasRequiredVariables && operationType !== 'Mutation' ? '' : '?'
       }: ${this.getApolloReactHooksIdentifier()}.${operationType}HookOptions<${operationResultType}, ${operationVariablesTypes}>) {
+        const defaultBaseOptions = ${JSON.stringify(this.rawConfig.defaultBaseOptions)}
```
ardatan, line 341: Maybe we should merge `baseOptions` parameter from the function and defined `defaultBaseOption` like;
```ts
, { ...defaultBaseOptions, ...baseOptions}
```

luxer066: Perfect @ardatan ! There we add your recommendation and fix the tests

ardatan: Thank you! And also I think `const defaultBaseOptions` can be a global value because it will be duplicated in every single hook function in this way. Could you fix that as well? Then we can merge the PR :)

ardatan: And also could you fix the tests :) Thanks again!

ardatan:
I wrote a few more comments :) Thank you very much!

luxer066:
Hi @ardatan ! How's it going?

We believe that now the code looks good

What do you think?

ardatan:
Great! Thank you @luxer066

dev-test/githunt/types.reactApollo.customSuffix.tsx
```diff
@@ -355,9 +356,10 @@ export function useOnCommentAddedSubscription(
     OnCommentAddedSubscriptionVariables
   >
 ) {
+  const options = { ...defaultOptions, ...baseOptions };
```
SimenB, line 359: could we drop this when `defaultOptions` is empty (or `baseOptions` is not passed)? extra spreads for every single render is unnecessary and breaks any and all `useMemo` optimization

nkovacic: This line  basically overrides default options from the Apollo client, so it would be great if the suggestion from @SimenB would be implemented

ardatan: Would you create a PR for this change?

Urigo: could someone move this into a new issue?

PR #5540, opened 2021-02-11
Add generates.preset

JeffBaumgardt:
Adds `generates.preset` documentation to highlight some of the presets available such as near-operation-file.

When I was setting up for the first pass it wasn't clear to me that all the documents no matter where in the repo would just end up in a single file.
We might want to look into adding a step in the cli tool to offer presets as an option. Not sure how that would work though.


dotansimha:
This totally makes sense. Thank you @JeffBaumgardt !

PR #5617, opened 2021-02-25
Removed the unnecessary call to print in the graphql-request plugin

### LINKED ISSUE #5614: graphql-request's unnecessary `print` function call
The `request` function of the [graphql-request](https://github.com/prisma-labs/graphql-request) library accepts either a `DocumentNode` or a `string` as the query. If a `DocumentNode` is passed, it will convert it into a `string` by calling graphql's `print` function internally:
https://github.com/prisma-labs/graphql-request/blob/a8d99f5cdbe57786ecb8d99c88175599608d2fc6/src/index.ts#L207-L211
So there is no need here for us to call the `print` function to convert the query to a `string` before passing it to the `request` function.
https://github.com/dotansimha/graphql-code-generator/blob/4ea2ccd0f7dcb9486145bd7abec6804f6d756f26/packages/plugins/typescript/graphql-request/src/visitor.ts#L103

I was about to create a PR for this but I read that you said you should always post an issue before creating a PR. So I'm waiting for your response.

AradAral:
Related #5614


AradAral:
Sorry, I should've modified the tests too.
Will do.

AradAral:
I messed up my fork. See #5618

PR #5645, opened 2021-03-03
Umbrella PR - federation resolvers plugin refactor

kamilkisiela:
Yes we already support Federation in `typescript-resolvers` but it kind of broken... and we want to fix it!

## Umbrella PR

Treat this PR as a base for smaller, atomic PRs. Here's a [doc on Notion](https://www.notion.so/Federation-typescript-resolvers-3b47e825ff9a43b7a970c28e8784a6f70) explaining what, how and why.

## We need help!

We want to get there quickly and if there's a chance someone from the community can help us, it would be awesome! 😎 

A list of tasks is available as jest tests (or todos) in `packages/plugins/typescript/resolvers/tests/federation.spec.ts` file but also here:

- [x] should add `__resolveReference` to objects with `@key`
- [x] `__resolveReference` should use a union of `@key` directives
- [ ] only fields specified by `@requires` should be available in parent object
- [x] do not create resolvers for `@external` fields
- [ ] should accept a `reference`
- [ ] should accept a `reference` based on union of `@key` directives
- [x] should NOT accept a `reference` on an external type without `@key`
- [x] should NOT accept a `reference` on an external type (with `@extends`) without `@key`
- [x] should NOT accept a `reference` on a local (non-external) type
- [ ] support `__resolveObject`
- [ ] should mapper apply to `__resolveReference` as output?
- [ ] should mapper apply to `__resolveReference` as input?
- [ ] mappers should not collide with references - both accepted
- [ ] check if Federation allows to resolve a custom object and if so then what data goes to other services - is it the custom object or an object matching GraphQL type.

## How to help

You can help in two ways:
- create a test to make sure we follow federation spec
- implement a part of Federation into typescript-resolvers plugin
- discuss with us the implementation and tests
- make sure we follow Federation spec

if you eager to help, please mention @kamilkisiela (me) or @dotansimha in a comment and let's get things done!

### DIFF
--- a/.changeset/giant-ravens-return.md
+++ b/.changeset/giant-ravens-return.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-resolvers': minor
+---
+
+Some fixes and attempts to make it better
--- a/packages/plugins/typescript/resolvers/tests/federation.spec.ts
+++ b/packages/plugins/typescript/resolvers/tests/federation.spec.ts
@@ -1,8 +1,25 @@
-import '@graphql-codegen/testing';
+import { compileTs } from '@graphql-codegen/testing';
 import { parse } from 'graphql';
 import { codegen } from '@graphql-codegen/core';
+import { plugin as tsPlugin } from '@graphql-codegen/typescript';
 import { plugin } from '../src';
 import { TypeScriptResolversPluginConfig } from '../src/config';
+import { JsxEmit, ScriptTarget } from 'typescript';
+
+function compile(content: string) {
+  compileTs(
+    content,
+    {
+      jsx: JsxEmit.None,
+      target: ScriptTarget.ES2018,
+      allowJs: false,
+      skipLibCheck: true,
+    },
+    /* tsx */ false,
+    /* playground */ false,
+    /* flag to enable diagnostics*/ true
+  );
+}
 
 function generate({ schema, config }: { schema: string; config: TypeScriptResolversPluginConfig }) {
   return codegen({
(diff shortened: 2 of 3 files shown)


skk2142:
Hi @kamilkisiela thanks for working on this! In the todo list, `do not create resolvers for @external fields` is not necessarily correct. The resolver for that field is called. It becomes necessary to resolver that field when I have a mapper. For instance, in the schema below, if I have a custom mapper for `User`, say `{ userId: string }` and I am resolving the `reviewer` field, I will return `{ userId: "some-id"}` from the resolver function for `reviewer`. To resolve the correct federation entity for `User`, I need to return the `id` field of `User` even though it is an external field. Does that sound correct?

```graphql
extend type User @key(fields: "id") {
  id: ID @external
}
type Review {
  id: ID!
  reviewer: User
}
type Query {
   review(id: ID!): Review
}
```

skk2142:
Also, I've been doing some experiment with federation. I think the `__resolveReference` method should return the mapper for the entity. The parent object of the `__resolveReference` resolver should be an object with fields specified in `fields` of the entity's key.

skk2142:
Hi @kamilkisiela We are a heavy user of apollo federation and we are encountering many issues with this library for typescript. I'd love to talk about how we can help to expedite apollo federation support.

dotansimha:
> Hi @kamilkisiela We are a heavy user of apollo federation and we are encountering many issues with this library for typescript. I'd love to talk about how we can help to expedite apollo federation support.

Hi! We would love to start a discussion about that. I think this PR is the ideal place for that :)

Urigo:
Hi @skk2142, we would still love your help and feedback.
Would you like to start the discussion here?
If you want, we are also open to schedule a call

skk2142:
HI @Urigo would love to explain the issues. Call would be easier and after that may be I can add more details as needed.

charlypoly:
Putting this PR aside and will come back later to this concern.

PR #5699, opened 2021-03-12
feat: configuration of default scalar type

### LINKED ISSUE #5274: Support configuration of default scalar type for custom scalars

When generating typescript code, I would prefer that the `Scalars` type mapping for custom scalars default to `unknown` rather than `any` (better type safety; forces explicit typecasting; makes missing custom scalar type overrides more obvious). It's conceivable that there may be similar desires when generating code in other languages too.

A new config option for the typescript code generator that allows me to set the default scalar type to `unknown`. This could possibly be a string config that technically supports any type to be specified, or a boolean config specific to typescript code generation only that restricts the options to `any` or `unknown`. 

I think a string option might be more preferable to generally support this option for other languages. Even within typescript, it may be desirable to use `never` as the default type to more strictly require that custom scalars be given type mappings.

Hackishly post-processing the output of the code generator to find/replace `any` with `unknown`. Doesn't seem worth the effort to do it properly.

**Additional context**

I have poked around the source code and found that there is a `buildScalars()` method in `packages/plugins/other/visitor-plugin-common/src/utils.ts` with a `defaultScalarType` parameter that defaults to `'any'`. I don't understand the structure of this project well enough to figure out the "right" way to make the change myself, but I would recommend starting by making the `defaultScalarType` required (no default value; because `any` is not a good default for ALL languages), then following the compiler errors to every usage of `buildScalars()` and make a case-by-case decision on whether to pass a hard-coded value for `defaultScalarType`, or require it to be passed down from the next level up in the call stack, grabbed from config, etc.

andreialecu:
## Description

Allows defining the scalar type to use for unknown scalars, instead of `any`.

Additionally, allows setting a config value `strictScalars: true`, which will result in the generator throwing an error if it encounters a scalar that wasn't defined in the `scalars` config field.

Related #5274

## Type of change

N/A

## How Has This Been Tested?

Added unit tests.

## Checklist:

## Further comments

I chose to create a new wrapper method for `buildScalars` and keep the changes minimal. The reason for the new method is to ensure that there are no breaking changes. I wasn't sure if `buildScalars` is exported and used by other external plugins.

### DIFF
--- a/.changeset/wild-rules-press.md
+++ b/.changeset/wild-rules-press.md
@@ -1,0 +1,14 @@
+---
+'@graphql-codegen/c-sharp': minor
+'@graphql-codegen/c-sharp-operations': minor
+'@graphql-codegen/java': minor
+'@graphql-codegen/kotlin': minor
+'@graphql-codegen/java-resolvers': minor
+'@graphql-codegen/visitor-plugin-common': minor
+'@graphql-codegen/typescript-compatibility': minor
+'@graphql-codegen/typescript-mongodb': minor
+'@graphql-codegen/typescript': minor
+'@graphql-codegen/near-operation-file-preset': minor
+---
+
+Added new configuration settings for scalars: `strictScalars` and `defaultScalarType`
--- a/packages/plugins/c-sharp/c-sharp-operations/src/visitor.ts
+++ b/packages/plugins/c-sharp/c-sharp-operations/src/visitor.ts
@@ -5,8 +5,8 @@ import {
   LoadedFragment,
   indentMultiline,
   getBaseTypeNode,
-  buildScalars,
   indent,
+  buildScalarsFromConfig,
 } from '@graphql-codegen/visitor-plugin-common';
 import autoBind from 'auto-bind';
 import {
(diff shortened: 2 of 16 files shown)


dotansimha:
This looks amazing. Thank you so much @andreialecu ! 
Can you please add a changeset file (run `yarn changeset` and it will guide you). I think we need to bump all touched and dependent packages. This will also allow us to have an alpha version for testing purposes before merging this.

andreialecu:
@dotansimha done.

andreialecu:
@dotansimha bump :)

raveclassic:
Hey guys, this is awesome! Looking forward to seeing this released!

dotansimha:
Hi @andreialecu ! Sorry for the delay. This looks great, merging and will cut a release very soon!

UselessPickles:
@andreialecu While you were digging around the code for this, did you see any easy opportunities for adding another strict check that fails if the `scalars` config contains entries for custom scalars that do not exist in the schema? I would expect `strictScalars: true` to be fully strict, requiring the `scalars` config to provide mappings for exactly all custom scalars in the schema. No more, no less.

Urigo:
@UselessPickles could you create a new issue for that?

PR #5724, opened 2021-03-18
fix(visitor-plugin-common): guards possible runtime type error

cometkim:
There is type error on run `graphql-mesh generate-sdk`.

```
TypeError: Cannot read property 'length' of undefined
    at hasConditionalDirectives (/Users/tim/Workspace/tmp/graphql-mesh-test/node_modules/@graphql-codegen/visitor-plugin-common/index.cjs.js:479:20)
```

image

The line 479:
image

I think we shouldn't trust TypeScript that much :/

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/utils.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/utils.ts
@@ -20,7 +20,6 @@ import {
   isListType,
   isAbstractType,
   GraphQLOutputType,
-  DirectiveNode,
 } from 'graphql';
 import { ScalarsMap, NormalizedScalarsMap, ParsedScalarsMap } from './types';
 import { DEFAULT_SCALARS } from './scalars';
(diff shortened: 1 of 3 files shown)


packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -403,7 +403,9 @@ export function getPossibleTypes(schema: GraphQLSchema, type: GraphQLNamedType):
   return [];
 }
 
-export function hasConditionalDirectives(directives: readonly DirectiveNode[]): boolean {
+export function hasConditionalDirectives(directives: readonly DirectiveNode[] | null | undefined): boolean {
+  if (!directives) return false;
```
ardatan, line 407: How about the following?
```js
const CONDITIONAL_DIRECTIVES = ['skip', 'include']; // This can be global.
return directives?.some(directive => CONDITIONAL_DIRECTIVES.includes(directive?.name?.value));
```

ardatan (changes requested):

cometkim:
@ardatan Could you review this again?

ardatan:
Thank you @cometkim ! That looks good to me!

Issue #5227, opened 2020-12-09
[typescript-resolvers] Field wrapper doesn't wrap array types


nebbles:
**Describe the bug**
The change made for #3882 is counter-intuitive to me and seems like a bug. By my understanding, a Field definition includes whether its an array or not.

Based on this change, when adding to `config`:
```yml
wrapFieldDefinitions: true
fieldWrapperValue: T | Promise<T> | (() => T) | (() => Promise<T>)
```

Then I'd *expect* this wrapper to be applied to the Type of the field

```ts
// a type like this
images: Array<GqlImageData>;

// turns into
images: FieldWrapper<Array<GqlImageData>>;
```

However as mentioned, the above change means that Arrays cannot be given a custom wrapper. The result of using the config returns

```ts
images: Array<FieldWrapper<GqlImageData>>;
```

Is this a bug or a feature request? As alluded to in the example config, it would simplify code a lot if a custom field wrapper enabled resolver functions for fields (I've read #1704 - however moving to partials removes type strictness which is useful to developers in the team).

Thanks

ibratoev:
I hit the same issue as I use the same `fieldWrapperValue`. Added a few comments in #3882 about it. In my case I get the error `Type '() => Promise<User[]>' is missing the following properties from type 'FieldWrapper<User>[]': pop, push, concat, join, and 27 more`.

hampusborgos:
I have attempted to solve this in a new PR: https://github.com/dotansimha/graphql-code-generator/pull/5753 that adds a new configuration option target to arrays (and `Maybe<>`) wrappers.

dotansimha:
Fix is available in `@graphql-codegen/typescript@1.22.0`. Thanks @hampusborgos !

Issue #4614, opened 2020-08-23
FieldWrapper wraps the final element instead of the overall field type


richiejd:
Basic example:
- expected: `FieldWrapper<Maybe<Array<Maybe<Favorite>>>>`
- actual: `Maybe<Array<Maybe<FieldWrapper<Address>>>>`

Not sure whether to call this a bug or just a missing feature, but the FieldWrapper is put inside rather than outside of the Array and Maybe wrappers.

Example:
With the config: 
```
      defaultMapper: Partial<{T}>
      fieldWrapperValue: T | Promise<T>
      wrapFieldDefinitions: true
```

Given the following graphql (note I'm contriving an example, but don't usually have nullables in an array...):

```
type User {
  addresses: [Address!]!
  orders: [Order]!
  saved: [Saved!]
  favorites: [Favorite]
  defaultAddress: Address!
  favoriteAddress: Address
}
```

I would expect the following ts:
```
export type User = {
  __typename?: "User";
  addresses: FieldWrapper<Array<Address>>;
  orders: FieldWrapper<Array<Maybe<Order>>>;
  saved?: FieldWrapper<Maybe<Array<Saved>>>;
  favorites?: FieldWrapper<Maybe<Array<Maybe<Favorite>>>>;
  defaultAddress: FieldWrapper<Address>;
  favoriteAddress?: FieldWrapper<Maybe<Address>>;
};
```

instead I get:
```
export type User = {
  __typename?: "User";
  addresses: Array<FieldWrapper<Address>>;
  orders: Array<Maybe<FieldWrapper<Address>>>;
  saved?: Maybe<Array<FieldWrapper<Address>>>;
  favorites?: Maybe<Array<Maybe<FieldWrapper<Address>>>>;
  defaultAddress: FieldWrapper<Address>;
  favoriteAddress?: Maybe<FieldWrapper<Address>>;
};
```

As you can see, I would expect support for FieldWrapper to wrap the whole thing so I can have `Promise<T> | T`

This is on version `1.17.0`

richiejd:
This same issue is actually a further issue in the resolver map, e.g. I get the following errors
```
Type '(_: {}, __: {}, { customerId }: Context) => Promise<CustomerRecord | null>' is not assignable to type 'ResolverFn<Maybe<ResolverTypeWrapper<Partial<Customer>>>, {}, Context, {}>'.
[...]

PR #5753, opened 2021-03-26
Add `entireFieldWrapperValue` configuration option, to wrap arrays.

hampusborgos:
## Description

Fixes regression introduced in: #3938, fixes feature requested in: #5227  and #4614 (issue can be discussed ether in these issues, or here).

### The problem
The change made in issue #3938 changed how `fieldWrapper` worked, so that it is no longer possible to wrap an entire array using `FieldWrapper`, rather the values INSIDE the array are wrapped.

@nebbles summed this up well in #5227:

When adding to config:

    wrapFieldDefinitions: true
    fieldWrapperValue: T | Promise<T> | (() => T) | (() => Promise<T>)

Then I'd expect this wrapper to be applied to the Type of the field

A type like this:

    images: Array<GqlImageData>;

Should turn into: 

     images: FieldWrapper<Array<GqlImageData>>;

However as mentioned, the above change (#3938) means that Arrays cannot be given a custom wrapper. The result of using the config returns

    images: Array<FieldWrapper<GqlImageData>>;

### The fix

This adds two new configuration options: `wrapEntireFieldDefinitions` and `entireFieldWrapperValue` that fixes this. So when adding to config:

    wrapEntireFieldDefinitions: true
    entireFieldWrapperValue: T | Promise<T> | (() => T) | (() => Promise<T>)

A type like this:

    images: Array<GqlImageData>;

Turns into:

     images: EntireFieldWrapper<Array<GqlImageData>>;

This can also be nested with `wrapFieldDefinitions` so that 

    wrapFieldDefinitions: true
    fieldWrapperValue: T | Promise<T> | (() => T) | (() => Promise<T>)
    wrapEntireFieldDefinitions: true
    entireFieldWrapperValue: T | Promise<T> | (() => T) | (() => Promise<T>)

Will make it so a type like this:

    images: Array<GqlImageData>;

Turns into:

     images: EntireFieldWrapper<Array<FieldWrapper<GqlImageData>>>;

## Type of change

## How Has This Been Tested?

Added a new test case for these new configuration option(s).

## Checklist:

## Further comments

It can also be considered, that this should be added to `flow` generator also. But I'm not an expert there, so requires assistance.

### DIFF
--- a/.changeset/warm-candles-teach.md
+++ b/.changeset/warm-candles-teach.md
@@ -1,0 +1,7 @@
+---
+'@graphql-codegen/visitor-plugin-common': minor
+'@graphql-codegen/typescript': minor
+'@graphql-codegen/flow': minor
+---
+
+Add entireFieldWrapperValue configuration option, to wrap arrays
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -51,6 +51,8 @@ export interface ParsedTypesConfig extends ParsedConfig {
   enumPrefix: boolean;
   fieldWrapperValue: string;
   wrapFieldDefinitions: boolean;
+  entireFieldWrapperValue: string;
+  wrapEntireDefinitions: boolean;
   ignoreEnumValuesFromSchema: boolean;
 }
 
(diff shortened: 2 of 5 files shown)


ibratoev:
Looks like a good configurable approach with good defaults. I would use that. Thanks! 👍

dotansimha:
Thank you @hampusborgos , this seems like a good direction. Merging and will release soon.

PR #5809, opened 2021-04-08
Added Github action for Declarative label declaration, Standard Labels added for The Guild 

tvvignesh:
## Description

@Urigo @dotansimha Please review if the labels make sense for The Guild repos. If yes, then I can send similar PRs for all repos.

**It does not delete any of the existing labels as of now**. If you want only the labels which are available in the yaml to be used and remaining to be deleted you can set `skip-delete: true` in the yaml. And if you want to test it you can set `dry-run: true` in the yaml

## Type of change

image

image

## Checklist:


.github/labels.yml
```diff
@@ -0,0 +1,109 @@
+# Please add the labels which are applicable to all repos in The Guild Stack here
```
dotansimha, line 1: @tvvignesh is there a way to put this in a more global place? like a repo under https://github.com/the-guild-org/ ?

tvvignesh: @dotansimha I was also wondering the same, but sadly not in this action. I noticed that it is possible in the other action though: https://github.com/marketplace/actions/label-syncer#sync-labels-on-another-repository

Can try that and see.

tvvignesh: @dotansimha  The problem though is that its possible within an org or user. GITHUB TOKEN will not be able to access other repos.

PR #5814, opened 2021-04-09
Merge imports and fix example type defs

electerious:


ardatan:
Thanks @electerious !

Issue #5859, opened 2021-04-16
Plugin csharp, nullable DateTime field is generated as mandatory


kondratov-saritasa:
Plugin csharp, nullable DateTime field is generated as mandatory

Steps to reproduce:
1. Use SDL:
```
type Test {
  dateOptional: DateTime
  dateMandatory: DateTime!
}

scalar DateTime
```
2. Use codegen.yaml:
```
schema: schema.graphql
generates:
  Model.cs:
    plugins:
      - c-sharp
    config:
      className: Model
      scalars:
        DateTime: DateTime
```
3. generate

Result:
```
using System;
using System.Collections.Generic;
using Newtonsoft.Json;
using GraphQL;

namespace GraphQLCodeGen {
  public class Model {
    
    #region Test
    public class Test {
      #region members
      [JsonProperty("dateOptional")]
      public DateTime dateOptional { get; set; } // <-- should be optional
    
      [JsonProperty("dateMandatory")]
      public DateTime dateMandatory { get; set; }
      #endregion
    }
    #endregion
  }
  
}
```

Expected result:
```
using System;
using System.Collections.Generic;
using Newtonsoft.Json;
using GraphQL;

namespace GraphQLCodeGen {
  public class Model {
    
    #region Test
    public class Test {
      #region members
      [JsonProperty("dateOptional")]
      public DateTime? dateOptional { get; set; }
    
      [JsonProperty("dateMandatory")]
      public DateTime dateMandatory { get; set; }
      #endregion
    }
    #endregion
  }
  
}

```

dotansimha:
Available in: `@graphql-codegen/c-sharp-operations@1.19.4`

PR #5860, opened 2021-04-16
Consider DateTime as builtin C# value type

kondratov-saritasa:
This code causes issues with nullability of value types. Task #5859 

```
export function isValueType(type: string): boolean {
  // Limitation: only checks the list of known built in value types
  // Eg .NET types and struct types won't be detected correctly
  return csharpValueTypes.includes(type);
}
```

I do not eliminate that limitation completly:
```
  // Limitation: only checks the list of known built in value types
  // Eg .NET types and struct types won't be detected correctly
```

But with this fix - generator more or less usable in C# projects. It is quite major blocker - if code does not support DateTime nullability.

In general DateTime - is not language primitive like ```int, bool, float``` but it is built-in value type. For CRL - there is no much difference in DateTime and Int32 (int is an alias for Int32): 
*  https://github.com/microsoft/referencesource/blob/master/mscorlib/system/datetime.cs
* https://github.com/microsoft/referencesource/blob/master/mscorlib/system/int32.cs

### DIFF
--- a/.changeset/tasty-ravens-doubt.md
+++ b/.changeset/tasty-ravens-doubt.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/c-sharp': major
+---
+
+Consider DateTime as builtin C# value type
--- a/packages/plugins/c-sharp/c-sharp/test/c-sharp.spec.ts
+++ b/packages/plugins/c-sharp/c-sharp/test/c-sharp.spec.ts
@@ -452,6 +452,25 @@ describe('C#', () => {
           public bool? boolOpt { get; set; }
         `);
       });
+
+      it('Should properly treat DateTime nullability', async () => {
+        const schema = buildSchema(/* GraphQL */ `
+          type Test {
+            dateOptional: Date
+            dateMandatory: Date!
+          }
+
+          scalar Date
+        `);
+        const result = await plugin(schema, [], {}, { outputFile: '' });
+
+        expect(result).toBeSimilarStringTo(`
+          [JsonProperty("dateOptional")]
+          public DateTime? dateOptional { get; set; }
+          [JsonProperty("dateMandatory")]
+          public DateTime dateMandatory { get; set; }
+        `);
+      });
     });
 
     describe('Array', () => {
--- a/packages/plugins/c-sharp/common/scalars.ts
+++ b/packages/plugins/c-sharp/common/scalars.ts
@@ -7,9 +7,7 @@ export const C_SHARP_SCALARS = {
   Date: 'DateTime',
 };
 
-// All native C# built-in value types
-// See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types
-export const csharpNativeValueTypes = [
+export const csharpValueTypes = [
   'bool',
   'byte',
   'sbyte',
@@ -23,4 +21,5 @@ export const csharpNativeValueTypes = [
   'ulong',
   'short',
   'ushort',
+  'DateTime',
 ];
--- a/packages/plugins/c-sharp/common/utils.ts
+++ b/packages/plugins/c-sharp/common/utils.ts
@@ -1,6 +1,6 @@
 import { Kind, TypeNode, StringValueNode } from 'graphql';
 import { indent } from '@graphql-codegen/visitor-plugin-common';
-import { csharpNativeValueTypes } from './scalars';
+import { csharpValueTypes } from './scalars';
 import { ListTypeField, CSharpFieldType } from './c-sharp-field-types';
 
 export function transformComment(comment: string | StringValueNode, indentLevel = 0): string {
@@ -26,7 +26,7 @@ function isStringValueNode(node: any): node is StringValueNode {
 export function isValueType(type: string): boolean {
   // Limitation: only checks the list of known built in value types
   // Eg .NET types and struct types won't be detected correctly
-  return csharpNativeValueTypes.includes(type);
+  return csharpValueTypes.includes(type);
 }
 
 export function getListTypeField(typeNode: TypeNode): ListTypeField | undefined {


kondratov-saritasa:
@dotansimha Could you please take a look to this one?

kondratov-saritasa:
> CI / Testing on windows-latest and Node 10 (pull_request)  Failing after 3m
Not sure it is related to PR changes.

ardatan:
Could you add some unit tests for this change to make sure the results are correct?
Before merging this PR, we need a changeset for changelogs. Could you add it with `yarn changeset`?

kondratov-saritasa:
@ardatan 
Please review last two commits. Thank you.

kondratov-saritasa:
@ardatan @dotansimha 
Anything else prevents this PR to be merged? This issue blocks me with setting up proper CI process on one of the projects, I had to generate code locally from forked version. 

Could you please take a look?

kondratov-saritasa:
@ardatan @dotansimha Guys, more than two weeks passed since your request. Could you please review this PR?

dotansimha:
> @ardatan @dotansimha Guys, more than two weeks passed since your request. Could you please review this PR?

Sure. Sorry for the delay. We are not highly familiar with C#, so we are doing our best here :) 

In this case, you are adding `DateTime` like it's a built-in scalar of GraphQL, which might cause other implications. What if someone is using `scalar Date` with a class different from `DateTime`? 

We are really trying to stick to the original spec of GraphQL, so I think assuming that developers has `scalar Date` in their schema is tricky since it's not part of the spec. 

In your case, why not adding `scalars` config and just extend the default set of scalars? You should be able to map `scalar Date` to `DateTime` with that: 

```
scalars:
  Date: DateTime
```

kondratov-saritasa:
@dotansimha 
> In this case, you are adding DateTime like it's a built-in scalar of GraphQL, which might cause other implications.

I do not add it as a scalar type - it was registered before: https://github.com/dotansimha/graphql-code-generator/blob/cdfdddff252aa9221cf4589d7f4b9f3e306bccad/packages/plugins/c-sharp/common/scalars.ts#L7
But I am not sure the intention.

> What if someone is using scalar Date with a class different from DateTime?
It should work. Do not see any problems - C# does not have builtin type `Date` he can define his own.

```
We are really trying to stick to the original spec of GraphQL, so I think assuming that developers has scalar Date in their schema is tricky since it's not part of the spec.
```
I do not see how my change might cause that - could you please clarify?

> In your case, why not adding scalars config and just extend the default set of scalars? You should be able to map scalar Date to DateTime with that:
 I shared my config in the report: https://github.com/dotansimha/graphql-code-generator/issues/5859
It includes: 
```
      scalars:
        DateTime: DateTime
```
But that does not help - generator does not recognize it properly.


May be I described the problem not clear - here are a few clarifications.

* Anything in C# code which has type DateTime - will be handled as default, built-in, value type.
* You can not change it
* Code generator - treats DateTime properties as reference types (which can be null) and do not change type to Nullable<DateTime>(if spec requires that) - this causes runtime crashes during serialization
* So I made generator threat DateTime as value type - no matter how it appeared in the pipeline. In any case if it will be in CS file it will be threated as DateTime(https://github.com/microsoft/referencesource/blob/master/mscorlib/system/datetime.cs)

Probably that problem might be solved by adjusting generator or config. But I assume it is wrong to threat DateTime this way - so I decided to add it to list of native value types. And this solves the problem.

kondratov-saritasa:
@dotansimha
> We are not highly familiar with C#, so we are really trying our best here :)

We could ask the author of original solution if this fix is ok. 

Hi @cklam2 - could you please take a look to this PR if you have time? Do you have any concerns having DateTime included to `csharpNativeValueTypes`? Thank you in advance.

cklam2:
The DateTime is not a built in value type but a struct. In C# struct is a value type which are non nullable. To mark a value type nullable you'll have to wrap it into the nullable structure, eg `Nullable<DateTime>` or `DateTime?`. So far nothing new for C# developers.

Current generation code for C# is indeed not able to determine if a type is a value type like struct, hence the remark in isValueType(). 
One workaround would be to define a custom scalar for the optional value type like this:

```
scalars:
  DateTimeNullable: DateTime?
  DateTime: DateTime
```

And then use it in your gql:
```
type Query {
  """ Converts to public DateTime? dateOptional { get; set; } """
  dateOptional: DateTimeNullable
  """ Converts to public DateTime dateMandatory { get; set; } """
  dateMandatory: DateTime
}
```

Though it's a bit nasty and in some cases even unwanted when the gql schema is common for various language implementations.

Your solution works but I'd prefer a more generic way to specify an arbitrary type as value type, including types like BigInteger, Point, Rectangle, etc. Otherwise we'd end up with a huge list of types and there'll still be no option for user to specify their custom structs as value type.

So here's my suggestion: 
Don't add the DateTime. Instead extend the config to accept a list of types from codegen.yml and update the [isValueType](https://github.com/dotansimha/graphql-code-generator/blob/6734c58055d283f051bf32f115f2308000e56431/packages/plugins/c-sharp/common/utils.ts#L26) to recognize those types and mark them as value type.

Something like:
```
treatAsValueTypes:
  DateTime
  Point
```

Regarding the `Date` in the C_SHARP_SCALARS list, I don't remember why that one is added there, guess was copied from the [Scalar Types](https://graphql.org/learn/schema/#scalar-types) page. Which is wrong cause it's a custom scalar type so it should be removed from the list. Might break compatibility for certain users though.

kondratov-saritasa:
> The DateTime is not a built in value type but a struct. 
What `built-in`  in current context and why it is important for generation? May be because it does not require `using`? But we have `using System;` in default template so it is safe to use all classes from there.

As I already noted - there is no a much difference between types you referenced in list initially and DateTime under hood. E.g.:
`sbyte`: https://docs.microsoft.com/en-us/dotnet/api/system.sbyte?view=net-5.0
`[System.CLSCompliant(false)]
public struct SByte : IComparable, IComparable<sbyte>, IConvertible, IEquatable<sbyte>, IFormattable`

`DateTime`: https://docs.microsoft.com/en-us/dotnet/api/system.datetime?view=net-5.0

> To mark a value type nullable you'll have to wrap it into the nullable structure, eg Nullable<DateTime> or DateTime?. So far nothing new for C# developers.
This is what is happing if register this type in csharpValueTypes

> Current generation code for C# is indeed not able to determine if a type is a value type like struct, hence the remark in isValueType()
Agree - but other types are not so relevant. You mentioned BigInteger and Point, may be Vector - these all are not that popular as DateTime and questionable:
* Point - which point System.Windows or System.Drawing or may be from Unity? 3D point of 2D point? 
* BiInteger - not sure it might be transferred via JSON and supported by most of severs without major customization (converting to string or so)
* Rectangle and Point is much more custom than everything else.

But DateTime is a WAY more popular, I took github spec: https://docs.github.com/en/graphql/overview/public-schema and searched for types used. 
* Bolean 468
* Float 9
* DateTime 290
* String 1829

Also DateTime has  ISO 8601 standard which are aplied for most of use cases. So my message: It should work from out of box - not require customization. So your concern - what there will be a lot of types in `csharpValueTypes`, I do not think that is the case.

> Though it's a bit nasty and in some cases even unwanted when the gql schema is common for various language implementations.
I would say it is not nasty - it looks like a hack. I am ok if we will have to use that for some specific `DateTime` with custom serialization or e.g. with TZ like TimeSpan or so - but basic usage of DateTime, should be simple. Also - please fix me if I am wrong, it does not work if you don't have the access to the server code?

> Don't add the DateTime. Instead extend the config to accept a list of types from codegen.yml and update the isValueType to recognize those types and mark them as value type.
Good idea - but I would suggest to have DateTime work without additional customization. And work from out the box. 

Main question:
How does it work for you? Do you have DateTime, how do you handle nullable cases? I suggest merge this PR to support them and start another PR with list of valueTypes in config.

cklam2:
Well, I am well aware that DateTime is used pretty frequently. 

The ones that I previously listed were just some random examples from the top of my head. You may think that none of the other types are relevant but doesn't mean that no one would use them.

Anyway, my idea was not to limit to DateTime but to introduce a way to allow users to specify any type as value type. Cause in the end it's not about popularity but whether a user can do what (s)he had in mind.

In your case you're probably only interested in DateTime so let's just merge this (@dotansimha, please merge) and later will give it a follow up.

kondratov-saritasa:
@cklam2 
Sorry - I might seem rude, but I am actually not :) 

> Well, I am well aware that DateTime is used pretty frequently.
If so how is that plugin works for your without it?

> In your case you're probably only interested in DateTime so let's just merge this (@dotansimha, please merge) and later will give it a follow up.
I fixed issue in my fork and can use it. Unfortunately I do not have time now to implement that PR you mentioned, and to be honest I am not TS expert. 

And I think that simple hotfix - worth it. This will make that plugin somehow usable - I cannot imagine real APIs without DateTime fields. Or may be one with Star Wars movies database does not have it?

cklam2:
We are using DateTime in our solution. We just don't need it to be nullable since DateTime value is key in our messages.

kondratov-saritasa:
@ardatan @dotansimha We still unable to merge this one?

dotansimha:
> @ardatan @dotansimha We still unable to merge this one?

I'm going with @cklam2 recommendation here. Merging :)

Issue #5815, opened 2021-04-09
[c-sharp-operations] losing double-precision due to wrong value type


124C41p:
The GraphQL value type "Float" refers to a double-precision floating-point value. It is translated by the plugin c-sharp-operations to the C#-type "float", which has only single precision (the correct analog would be "double"). Because of that translation error, the generated C#-Code ignores half of the precision of requested floating-point data.

I strongly suspect this bug can be easily corrected by replacing "float" with "double" in [scalars.ts, line 6](https://github.com/dotansimha/graphql-code-generator/blob/6734c58055d283f051bf32f115f2308000e56431/packages/plugins/c-sharp/common/scalars.ts).

ardatan:
Would you create a PR for the fix?

dotansimha:
Fixed in:
```
@graphql-codegen/c-sharp@2.0.2
@graphql-codegen/c-sharp-operations@1.19.2

```

PR #5890, opened 2021-04-24
fix(c-sharp): Use `double` instead of `float` 

124C41p:
## Description

The graphql  type "Float" corresponds to the C# type "double". But it was incorrectly translated to "float" before (which also is a valid C# type, but not the right one).

Related #5815

## Type of change

## How Has This Been Tested?

I used graphql-codegen cli to generate a .cs-file before and after my changes to verify that graphql  type "Float" is now correctly translated to C# type "double".

### DIFF
--- a/.changeset/serious-seas-judge.md
+++ b/.changeset/serious-seas-judge.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/c-sharp': patch
+'@graphql-codegen/c-sharp-operations': patch
+---
+
+Replace float with double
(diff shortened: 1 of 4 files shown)


dotansimha:
Thank you @124C41p ! 
Can you please elaborate and share more details? seems like both are valid https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types#characteristics-of-the-floating-point-types ?

124C41p:
@dotansimha Both "float" and "double" are valid C# value types. The difference lies in the precision: As your link shows, float values consist of 4 bytes of information, while double values consist of 8 bytes.

GraphQL floating point numbers also have double precision as mentioned [here](https://graphql.org/learn/schema/) (you also can easily check that GraphQL allows you to receive floating point numbers with way more than 9 digits).

The current behavior of c-sharp-operations is to receive 8 bytes of information and then to throw away 4 of 8 bytes by implicitly casting double to float.

PR #5940, opened 2021-05-05
add vscode extension integration doc

capaj:
Issue templete does not have a doc update as an option so I deleted it.


dotansimha:
Awesome! Thank you @capaj :)

PR #5950, opened 2021-05-06
(urql) - code generation for graphCache

JoviDeCroock:

## Description

This will create a generic to be used in a graphCache config, example output can be found [here](https://github.com/JoviDeCroock/graphcache-code-generation)

## Type of change

## Checklist:


packages/plugins/typescript/urql-graphcache/src/index.ts
```diff
@@ -0,0 +1,244 @@
...
+function getResolversConfig(schema: GraphQLSchema) {
+  const resolvers = [];
+
+  const roots = [schema.getMutationType()?.name, schema.getSubscriptionType()?.name].filter(Boolean);
+
+  const typemap = schema.getTypeMap();
+  const isValidType = (type: string) => !roots.includes(type) && !type.startsWith('__') && !baseTypes.includes(type);
+
+  Object.keys(typemap).forEach(function (key) {
+    if (!isValidType(key)) return;
```
JoviDeCroock, line 98: This should probably change to detecting whether it's not a scalar or internal type 😅

packages/plugins/typescript/urql-graphcache/src/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  if (subscriptionsName) {
+    subscriptionUpaters = getSubscriptionUpdatersConfig(schema, subscriptionsName);
+  }
+
+  return {
+    prepend: [imports],
+    content: [
+      `type RequireFields<T, K extends keyof T> = { [X in Exclude<keyof T, K>]?: T[X] } & { [P in K]-?: NonNullable<T[P]> };`,
```
JoviDeCroock, line 212: Can we check somehow whether another plugin already adds this?

packages/plugins/typescript/urql-graphcache/src/index.ts
```diff
@@ -0,0 +1,244 @@
...
+        return nullable ? `Maybe<Scalars['${typeNode.name.value}']}>` : `Scalars['${typeNode.name.value}']`;
+      }
+
+      switch (type.astNode.kind) {
+        case 'EnumTypeDefinition':
+        case 'UnionTypeDefinition':
+        case 'InputObjectTypeDefinition':
+        case 'ObjectTypeDefinition':
+        case 'ScalarTypeDefinition': {
+          const finalType = `RequireFields<${typeNode.name.value}, '__typename'>${allowString ? ' | string' : ''}`;
```
JoviDeCroock, line 40: Opted for using `RequireFields` here so people that don't force __typename to required in the `typescript` plugin don't get errors.

packages/plugins/typescript/urql-graphcache/src/index.ts
dotansimha: At the moment, we can't, we are currently trying to figure out how to allow interaction between plugins, so it will be simpler to check what other plugins are doing. 
If this type could conflict (I think we use it in `typescript-resolvers` only, no? does it makes sense to use both together?), maybe it can be named differently?

packages/plugins/typescript/urql-graphcache/src/index.ts
dotansimha: Cool. This is an alternative for requiring users to set `nonOptionalTypename: true` explicitly?

packages/plugins/typescript/urql-graphcache/src/index.ts
kitten: I think given this we may just want to name it differently? I think it's fine to make a more specialised type for this generator that explicitly only works with `'__typename'`?

ardatan:
@JoviDeCroock I did some fixes but there are still some missing parts. It still doesn't respect naming conventions of `typescript` plugin. 
@dotansimha Could you take a look? Would it be better to use Visitor pattern like we do in `resolvers` plugin?

ardatan (approved):

Issue #4684, opened 2020-09-02
Duplicated nested fragments on code generation


zaguiini:
**Describe the bug**

When the `TypedDocumentNode` is generated by the tool, it splats (`...`) all the fragment definitions, and then the nested fragments follow the same pattern, recursively, resulting in duplicated fragments being sent over the wire.

A lot of GraphQL APIs have limits regarding how many fragments are allowed per query, so this is also a problem.

**To Reproduce**

[Here's a reproducible example](https://github.com/zaguiini/typed-document-node-issue/blob/master/src/generated/graphql.ts) of what's happening. Clone it, install and run `yarn show-generated-content` and it will output the definitions included in the typed node.

**Expected behavior**
Nested fragments are not duplicated on documents.

**Environment:**

Specified on the reproducible repo.

**Additional context**

`graphql-tag` workarounds it by removing the duplicated definitions. I don't think it's straightforward to achieve the same behavior on this plugin since it relies on static code generation instead of dynamic, but I do believe it's possible to achieve the same result.

In the meantime, I'm using a palliative solution that patches this plugin. Here's the diff:

```diff
diff --git a/node_modules/@graphql-codegen/typed-document-node/index.cjs.js b/node_modules/@graphql-codegen/typed-document-node/index.cjs.js
index b305f91..eeabaa8 100644
--- a/node_modules/@graphql-codegen/typed-document-node/index.cjs.js
+++ b/node_modules/@graphql-codegen/typed-document-node/index.cjs.js
@@ -8,6 +8,47 @@ const graphql = require('graphql');
 const path = require('path');
 const visitorPluginCommon = require('@graphql-codegen/visitor-plugin-common');
 const autoBind = _interopDefault(require('auto-bind'));
+const ts = require('typescript');
+
+const includeFragmentDeduplicator = operation => {
+    const tree = ts.createSourceFile(
+        'file',
+        operation,
+        undefined,
+        true
+    );
+
+    const transformer = context => root => {
+        const visit = (node) => {
+            node = ts.visitEachChild(node, visit, context);
+
+            if(node.kind === ts.SyntaxKind.ObjectLiteralExpression) {
+                const kindPropertyIndex = node.properties.findIndex(property => {
+                    return property.name && property.name.text === 'kind'
+                });
+
+                const hasKindProperty = kindPropertyIndex !== -1;
+                const isDocumentKind = () => node.properties[kindPropertyIndex].initializer.text === 'Document';
+
[...]

PR #6018, opened 2021-05-18
Introduce `dedupeFragments` config flag for dealing with duplicate fragments 

gilgardosh:
Related #4684

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
@@ -188,7 +188,11 @@ export interface RawConfig {
    */
   globalNamespace?: boolean;
   /**
-   * @ignore
+   * @description  Removes fragment duplicants for reducing data transfer.
+   * It is done by removing sub-fragments imports from fragment definition
+   * Instead - import all of them are imported to the Operation node.
+   * @type boolean
+   * @default false
    */
   dedupeFragments?: boolean;
 }
--- a/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
@@ -30,6 +30,7 @@ export interface ParsedConfig {
   fragmentImports: ImportDeclaration<FragmentImport>[];
   immutableTypes: boolean;
   useTypeImports: boolean;
+  dedupeFragments: boolean;
 }
 
 export interface RawConfig {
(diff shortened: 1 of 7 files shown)


dotansimha:
Thoughts on that:
1. I think while generating fragments `DocuemntNode`, we can flatten and remove ALL other fragments' definitions. So each fragment will be standalone, with just spreads for other fragments.
2. Then, in each query `DocumentNode` that we generate, we can add all fragments in the query, and deduplicate it, based on the map of fragments.

### Today

```ts
const fragmentB = `fragment B on B { ... }`;
const fragmentA = `fragment A on A { field { ...B } } ${fragmentB}`;
const query = `query { ...B, ...A } ${fragmentA} ${fragmentB}`
```

What I suggest, is to do:

```ts
const fragmentB = `fragment B on B { ... }`;
const fragmentA = `fragment A on A { field { ...B } }`; // No template literals here
const query = `query { ...B, ...A } ${fragmentA} ${fragmentB}` // Only here
```

packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
```diff
@@ -186,6 +187,10 @@ export interface RawConfig {
    * @ignore
    */
   globalNamespace?: boolean;
+  /**
```
dotansimha, line 190: Why `ignore`?

packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
```diff
@@ -186,6 +187,10 @@ export interface RawConfig {
...
+   * @ignore
```
dotansimha, line 191: Why `ignore`? I think we need to document it :)

dotansimha: You can define it here, and then you don't need to define it in the config of typed-document-node. This is relevant for all of the plugins that inherits.

packages/plugins/typescript/typed-document-node/tests/typed-document-node.spec.ts
```diff
@@ -7,4 +8,54 @@ describe('TypeDocumentNode', () => {
...
+
+      type Job {
+        id: ID!
+        recruiterName: String!
+        title: String!
+      }
+    `);
+
+    const ast = parse(/* GraphQL */ `
+      query GetJobs {
```
dotansimha, line 30: Please add an additional test where the fragment is loaded also by the query (like this one, but we `...JobSimpleRecruiterData` as well)

dotansimha (approved):

packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
gilgardosh: It's documented [here](https://github.com/dotansimha/graphql-code-generator/blob/c47bcba43904ca6abd1da4209e07011ddd7c08d2/packages/plugins/typescript/typed-document-node/src/config.ts#L26), do you think it is relevant to activate it from the base visitor as well?

dotansimha (approved):

Issue #6085, opened 2021-06-04
Would a RTK-Query plugin be welcome?


phryneas:
Hi there,
I'm a maintainer of Redux Toolkit. 

Over the past half year, we've been working on a "server-side API cache" implementation called "RTK Query" that we are integrating into our package. 

Right now we've [just released](https://github.com/reduxjs/redux-toolkit/releases/tag/v1.6.0) (you can find our [documentation over here](https://redux-toolkit.js.org/rtk-query/overview)) after many alphas, betas and release candidates.

The api itself has been tested over the past half year, is already being used in production by a few teams and has been acknowledged by other package authors (for example, we are already on [React Query's "comparison page"](https://react-query.tanstack.com/comparison) .

We already have [code generation for OpenAPI specifications](https://github.com/rtk-incubator/rtk-query-codegen) and I've set off to also create a plugin for `graphql-codegen`. Honestly, implementing this has been a breeze and it's [pretty much finished by now](https://github.com/phryneas/graphql-code-generator/pull/1).

Would you be interested in getting that into the `graphql-codegen` repo? We could also publish it on our own, but honestly, I'm pretty impressed with the documentation, integration with the demo page and everything, so I'd love to see it in here "officially" :)

dotansimha:
Hi @phryneas !
Thank you for creating this issue. We are always welcome any kind of open-source contributions. RTK is definitely something that we can support, and we would love to see a PR for creating a plugin for it. 
I assume you can use `react-query` plugin as reference since it might be similar in terms of wrapping GraphQL operations. Feel free to reach out in case of questions!

phryneas:
@dotansimha it's already almost done, I'll just need to go through it again and adjust the version number to the release. I'll probably open a PR in the next hour or so :)

dotansimha:
`@graphql-codegen/typescript-rtk-query@1.0.0` is now available. Thanks @phryneas !

PR #6101, opened 2021-06-08
NEW PLUGIN: `typescript-rtk-query`

phryneas:
## Description

As already mentioned in #6085, this would create a plugin for rtk-query.

## Type of change

https://githubbox.com/reduxjs/redux-toolkit/tree/example/graphql-codegen/examples/query/react/graphql-codegen

## How Has This Been Tested?

Local code generation so far, also see added CodeSandbox with generated code.

## Checklist:

  I *think* so? There are many resources :sweat_smile: 

## Further comments

I already explain it in the doc, but here a short breakdown of the workflow:

* user creates a RTK-Query API in their codebase 
* graphql-codegen generates a file that imports this api, calls `api.injectEndpoints` to inject endpoints for all queries & mutations and re-exports this new API (and hooks, if applicable), which has now new extended types
* user configures the `baseApi` with their redux store
* user imports the hooks from the generated files. 

This is generally the way RTK-Query achieves code splitting and it does work fine with `preset-near-operation-file` (see the condesandbox).
If the user wants to further adjust the generated API, they can import the generated file, use `api.enhanceEndpoints` to add manual modifications and *then* re-export and use that file. That way the generated files can always be overwritten without risk.


dotansimha:
Thank you @phryneas ! ❤️

PR #6102, opened 2021-06-08
ESM support

PabloSzx:
## Description

ESM Support

## Type of change

## How Has This Been Tested?

**Test Environment**:
- OS: Windows & macOS
- `@graphql-codegen/...`: latest 
- NodeJS: v14.17.0 & v16.3.0

## Checklist:


PabloSzx:
I found an issue in CLI, and it's the usage of "require.resolve", which is not supported in ESM https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/config.ts#L183

PabloSzx:
Another issue found in @graphql-codegen/plugin-helpers https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-helpers/src/resolve-external-module-and-fn.ts#L7

That will most definitely break in ESM

dotansimha:
> I found an issue in CLI, and it's the usage of "require.resolve", which is not supported in ESM [https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/config.ts#L183](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/config.ts?rgh-link-date=2021-06-08T19%3A52%3A38Z#L183)

Yeah, seems to be an issue. And we do use this in codegen core to load plugins, configurations, and loaders.  I guess we need a wrapper for that to make sure it's still supported in both setups. 



> Another issue found in @graphql-codegen/plugin-helpers [https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-helpers/src/resolve-external-module-and-fn.ts#L7](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-helpers/src/resolve-external-module-and-fn.ts?rgh-link-date=2021-06-09T16%3A59%3A41Z#L7)
> 
> That will most definitely break in ESM

How do you think we should address this one? 

cc @ardatan

PabloSzx:
> How do you think we should address this one?

If that question is for me (maybe it wasn't), I feel like that is completely out of my scope since I'm not completely familiar with all the codegen internals, but adding support for esm I'd say is pretty much required if for example GraphQL.js starts adding support for ESM with the exports field, that will trigger duplicated graphql module versions, since if a commonjs module requires graphql, it will use the cjs version, and an esm module will use the esm version

dotansimha:
> > How do you think we should address this one?
> 
> If that question is for me (maybe it wasn't), I feel like that is completely out of my scope since I'm not completely familiar with all the codegen internals, but adding support for esm I'd say is pretty much required for programmatic usage if for example GraphQL.js starts adding support for ESM with the exports field, that will trigger duplicated graphql module versions error, since if a commonjs module requires graphql, it will use the cjs version, and an esm module will use the esm version (this applies for every GraphQL library btw)

Yeah, the question was for you :D 
I see what you mean, so I think maybe we can wait a bit longer on this PR, and plan this for the next major version.

PabloSzx:
image

image

@dotansimha @ardatan 

While working in a example using graphql-codegen programmatically I realized that the plugin-helpers .mjs file is bundling `requires` that come from bundling `change-case-all`, in ESM world bundling dependencies most of the time is forbidden, because it will bundle the requires in the esm files, which of course breaks.

Of course the solution is to add "change-case-all" as a dependency in this specific package, but this applies to every package that intends to support esm

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -180,7 +180,7 @@ export function parseArgv(argv = process.argv): YamlCliFlags {
 
 export async function createContext(cliFlags: YamlCliFlags = parseArgv(process.argv)): Promise<CodegenContext> {
   if (cliFlags.require && cliFlags.require.length > 0) {
-    await Promise.all(cliFlags.require.map(mod => import(require.resolve(mod, { paths: [process.cwd()] }))));
+    await Promise.all(cliFlags.require.map(mod => import(mod)));
```
merceyz, line 183: This doesn't behave the same, it will resolve `mod` from the CLI instead of the `cwd`, you can use `createRequire` instead of `require.resolve`.

Ref https://github.com/dotansimha/graphql-code-generator/pull/5032 and https://github.com/dotansimha/graphql-code-generator/pull/4983

PabloSzx: @merceyz thanks 👍 https://github.com/dotansimha/graphql-code-generator/pull/6102/commits/9b55a9210b170d2438d034b6cd45f93d51a5fefa

PabloSzx:
After adding the test-esm.mjs script (and locally fixing the pending to be merged fix of graphql-tools https://github.com/ardatan/graphql-tools/pull/3289) these errors appear, this script is only able to catch errors in the imports, not the logic inside of the scripts

https://user-images.githubusercontent.com/8672915/127908337-ac08b4be-7db2-45e9-aaf0-4b7c6642394d.mp4

I will take a look into fixing these issues

PabloSzx:
After https://github.com/ardatan/graphql-tools/pull/3300 is merged and publish, this PR should be ready @dotansimha

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -180,7 +181,17 @@ export function parseArgv(argv = process.argv): YamlCliFlags {
 
 export async function createContext(cliFlags: YamlCliFlags = parseArgv(process.argv)): Promise<CodegenContext> {
   if (cliFlags.require && cliFlags.require.length > 0) {
-    await Promise.all(cliFlags.require.map(mod => import(require.resolve(mod, { paths: [process.cwd()] }))));
+    const relativeRequire = (createRequire || createRequireFromPath)(process.cwd());
+    await Promise.all(
+      cliFlags.require.map(
+        mod =>
+          import(
+            relativeRequire.resolve(mod, {
```
merceyz, line 189: This should just be `require.resolve`

PabloSzx:
image

createRequireFromPath is already removed in Node.js v16 (it was already deprecated in earlier versions), I will have to remove it and only use "createRequire"

PabloSzx:
@dotansimha It doesn't allow me to request a review in the PR for some reason, but this PR should be ready

packages/graphql-codegen-cli/src/generate-and-save.ts
```diff
@@ -3,7 +3,7 @@ import { Types } from '@graphql-codegen/plugin-helpers';
 import { executeCodegen } from './codegen';
 import { createWatcher } from './utils/watcher';
 import { fileExists, readSync, writeSync, unlinkFile } from './utils/file-system';
-import { sync as mkdirpSync } from 'mkdirp';
+import mkdirp from 'mkdirp';
```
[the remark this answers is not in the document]
PabloSzx: not relevant for this PR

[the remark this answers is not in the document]
ardatan: @merceyz We'd love to accept a PR for this change!

.changeset/sharp-nails-clean.md
```diff
@@ -0,0 +1,52 @@
+---
+'@graphql-cli/codegen': major
```
ardatan, line 2: Do we really need a major release?

PabloSzx: the changesets can be changed, I think this change is for minor, it was set as major since at first there was some changes that I thought were breaking, now everything should be fine

PabloSzx: @ardatan done 👍

.changeset/unlucky-dots-judge.md
```diff
@@ -0,0 +1,60 @@
...
+'@graphql-codegen/urql-svelte-operations-store': minor
+'@graphql-codegen/typescript-vue-apollo': minor
+'@graphql-codegen/typescript-vue-apollo-smart-ops': minor
+'@graphql-codegen/typescript-vue-urql': minor
+'@graphql-codegen/gql-tag-operations-preset': minor
+'@graphql-codegen/graphql-modules-preset': minor
+'@graphql-codegen/import-types-preset': minor
+'@graphql-codegen/near-operation-file-preset': minor
+'@graphql-codegen/config-schema': minor
+'@graphql-codegen/testing': minor
```
ardatan, line 55: We don't publish testing and website to the NPM so those would changeset fail at NPM release. Could you remove these two?

PabloSzx: @ardatan done 👍 this issue in changesets has been open since so much time... https://github.com/atlassian/changesets/issues/436

PR #6140, opened 2021-06-17
Implemented first generation skip in watch mode.

cbmd:
## Description

Implemented possibility to skip first generation whenever some `pre-script` previously can already generate files which result onto unnecessary re-generation on `cold start` of scripts. With this change it's no longer necessary.

Related # (issue)

## Type of change

## How Has This Been Tested?

- [x] Changed over config values and then manually trigger watcher
- [x] Added `watchSkipFirst` CLI flag and then manually trigger watcher

## Checklist:

### DIFF
--- a/packages/graphql-codegen-cli/src/utils/watcher.ts
+++ b/packages/graphql-codegen-cli/src/utils/watcher.ts
@@ -142,7 +142,7 @@ export const createWatcher = (
 
   // the promise never resolves to keep process running
   return new Promise<void>((resolve, reject) => {
-    executeCodegen(initalContext)
+    (config.watchConfig?.skipFirst ? Promise.resolve([]) : executeCodegen(initalContext))
       .then(onNext, () => Promise.resolve())
       .then(runWatcher)
       .catch(err => {
--- a/.changeset/odd-colts-grow.md
+++ b/.changeset/odd-colts-grow.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': minor
+---
+
+First generation in watch mode can now be skipped.
--- a/packages/graphql-codegen-cli/src/config.ts
+++ b/packages/graphql-codegen-cli/src/config.ts
@@ -12,6 +12,7 @@ import yaml from 'yaml';
 export type YamlCliFlags = {
   config: string;
   watch: boolean | string | string[];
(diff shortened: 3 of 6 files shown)


dotansimha:
Thank you for creating this PR, @cbmd ! 
Can you please explain the background and use-case for this? I'm not sure that I understood it.

cbmd:
Sure @dotansimha, use case if following:
You wanted to run watch more of codegen AND webpack build simultaneously, whiles it's possible when you have some files already generated it's impossible on "fresh install" (due to both processes start at a same time). So what I'm suggesting is to skip mandatory first compilation and have granular control over it. E.g.
```json
{
  "scripts": {
    "prestart": "graphql-codegen",
    "start": "graphql-codegen --watch --skipFirst & webpack"
  }
}
```
And in this case it would still success to compile with webpack (Cuz all files already be in place while it would be boostraped) and will have re-compilation (cuz codegen would re-create files and then would trigger watcher of webpack, or any other bundler in place) of app.

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -142,7 +142,7 @@ export const createWatcher = (
 
   // the promise never resolves to keep process running
   return new Promise<void>((resolve, reject) => {
-    executeCodegen(initalContext)
+    (config.watchConfig?.skipFirst ? Promise.resolve([]) : executeCodegen(initalContext))
```
cbmd, line 145: It might be incorrect to return an array in here, but since by default initial context resulted with empty array - it was pretty clear to me that I should use the same, although i'm not happy with it. Maybe we can unify it with (let's say) some default via function or some other way around?

dotansimha:
I see, thank you @cbmd for explaining! 
I can merge that but may be worth checking with @piglovesyou since he's doing some cool stuff with codegen and Webpack in https://github.com/piglovesyou/graphql-let , so maybe its something that can be beneficial as well? 

Also, in most implementations of codegen and Webpack, the programmatic API is used, and then you can control when and how to call codegen.

binary64 (approved):

coreylight:
fwiw I would really enjoy this feature. We currently commit our generated files so it is only necessary to regen when graphql documents or schema changes.

PR #6142, opened 2021-06-17
Incremental rebuilds

kamilkisiela:
Related: ardatan/graphql-tools#2816

incremental re-builds in codegen. It skips running any computations related to outputs when their documents, schemas didn’t change. Later on we could cache plugin outputs as well and run only those relevant. Even more interesting, we could calculate checksums and avoid generating unnecessary file even outside of watch mode

4s vs 0.4s

https://user-images.githubusercontent.com/8167190/122384173-b9c3cf80-cf6b-11eb-8cd9-e6632e8a1e9d.mov


tgriesser:
@kamilkisiela is this still being actively worked on? Would love to see this sort of improvement and wanted to see if there's anything I could do to assist.

kamilkisiela:
I would have to rebase the PR and adjust it a bit, because of the change in GraphQL tools Loaders.
The Loaders changed quite a bit :(

kamilkisiela:
The idea here was to also persist some kind of metadata file to build/skip things between executions of graphql-codegen command.

tgriesser:
> The idea here was to also persist some kind of metadata file to build/skip things between executions of graphql-codegen command.

Ah nice, yeah that would be pretty useful - was starting to look into different options for speeding up the codegen by potentially building a smarter in-process watcher, which it looks like this is trying to do. We're using graphql-codegen (and several other graphql-tools) in the new v10 branch of Cypress: https://github.com/cypress-io/cypress/blob/10.0-release/graphql-codegen.yml and the codegen is a bit of a bottleneck when doing local development

charlypoly:
We will keep this PR as an inspiration for the upcoming work on CLI improvements.

PR #6154, opened 2021-06-20
fix: reference `@rtk-query/graphql-request-base-query`, remove from codegen package

phryneas:
## Description

This is a bug I unfortunately missed before as I copied the file just over to test it locally and that only surfaced now with the real package. I'm terribly sorry!

The user was supposed to 
```ts
import { graphqlRequestBaseQuery } from '@graphql-codegen/typescript-rtk-query';
```
in their runtime code. 
Unfortunately, the way the graphql-codegen packages are built, those are not side-effect-free, don't tree-shake and would try to execute node code in the browser - which crashes at runtime. Totally my bad, I should have noticed that one.

I've moved the export in question into it's own package and published it as `@rtk-query/graphql-request-base-query` - so this PR removes it from this repository and updates the docs accordingly.

## Type of change

## How Has This Been Tested?

Updated the example at https://github.com/reduxjs/redux-toolkit/pull/1152 to use all the correct & published packages (no more copying and voodoo with `yalc` any more) and run it locally.
I'll do a little more polishing and then we'll also add that example to our docs.

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas (this only removes code and updates an example)
- [x] I have added tests that prove my fix is effective or that my feature works (see https://github.com/reduxjs/redux-toolkit/pull/1152)


phryneas:
(changeset added)

dotansimha:
Thank you @phryneas ! I'll release that in a few minutes and update the docs.

dotansimha:
Available in `@graphql-codegen/typescript-rtk-query@1.1.0`.

Issue #6149, opened 2021-06-19
Fragments do not combine arrays correctly


chrbala:
**Describe the bug**
Including multiple fragments that contain the same arrays results in the types getting combined at the array level. TS doesn't know how to merge these correctly. That is ``Array<TypeA> & Array<TypeB>`` is not equivalent to ``Array<TypeA & TypeB>`` because TS doesn't merge arrays.

**To Reproduce**
This repro has the bug fully reproducible.

https://github.com/chrbala/apollo-bug-repro/tree/codegen-merge-arrays

1. My GraphQL schema:

```graphql
type Item {
  string: String!
  number: Int!
}

type Query {
  items: [Item!]!
}
```

2. My GraphQL operations:

```graphql
query Page {
  ...String
  ...Number
}

fragment String on Query {
  items {
    string
  }
}

fragment Number on Query {
  items {
    number
  }
}
```

3. My `codegen.yml` config file:

```yml
overwrite: true
schema: "src/schema.js"
documents: "src/**/*.tsx"
generates:
  src/generated/graphql.tsx:
    plugins:
      - "typescript"
      - "typescript-operations"
      - "typescript-react-apollo"
```

**Expected behavior**
You should be able to access both the ``string`` and ``number`` properties on ``items``, but you can only access string. If you switch the order of ``...String`` and ``...Number``, then you can only access number. 

The generated fragment is
```ts
export type PageQuery = (
  { __typename?: 'Query' }
  & NumberFragment
  & StringFragment
);
```

But for it to work correctly in TS, it would need to be more like
```
export type PageQuery = (
  { __typename?: 'Query' }
  & {
    items: Array<NumberFragment['items'][0] & StringFragment['items'][0]>
  }
);
```

**Environment:**

- OS: OSX 10.14.6
- Codegen
    "@graphql-codegen/cli": "1.21.5",
    "@graphql-codegen/introspection": "1.18.2",
    "@graphql-codegen/typescript": "1.22.1",
    "@graphql-codegen/typescript-operations": "1.18.0",
    "@graphql-codegen/typescript-react-apollo": "2.2.5",
- NodeJS: v12.13.1

dotansimha:
@n1ru4l what do you think?

n1ru4l:
@dotansimha This is a bit tricky, as we generate types per fragment, but maybe there is some kind of merge TypeScript utility that can take two deep types and merge them?

```tsx
export type PageQuery = (
[...]

PR #6184, opened 2021-06-24
fix: inline all operation types + handle edge cases where generated t…

n1ru4l:
Closes https://github.com/dotansimha/graphql-code-generator/issues/6149

This pr adds the option `inlineFragmentTypes` which is set to `true` per default.

When `inlineFragmentTypes` is set to `true` the generated operation types will deep inline the fragment selection set types instead of combining the fragment type.


packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -470,28 +525,30 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
...
+          if (relevant.length === 0) {
+            return null;
+          } else if (relevant.length === 1) {
+            return relevant[0];
+          } else {
+            return `( ${relevant.join(' & ')} )`;
+          }
+        })
+        .filter(Boolean)
+        .join(' | ') + this.getEmptyObjectTypeString(mustAddEmptyObject)
```
n1ru4l, line 542: Need to add a test to check whether this is correctly omitted if __typename is selected.

.changeset/many-socks-warn.md
```diff
@@ -0,0 +1,9 @@
+---
+'@graphql-codegen/flow-operations': patch
+'@graphql-codegen/visitor-plugin-common': patch
+'@graphql-codegen/typescript-operations': patch
+---
+
+Add option `inlineFragmentTypes` for deep inlining fragment types within operation types. This `inlineFragmentTypes` is set to `true` by default.
```
n1ru4l, line 7: @dotansimha Thoughts on the option? I think if we would instead make it a string: `fragmentSpreadTypeMode` with values (`inline`, `fragment`), we can later add `mask` without introducing a new conflicting option

dev-test/githunt/__generated__/comment.query.stencil-component.tsx
```diff
@@ -15,7 +15,13 @@ declare global {
     entry?: Types.Maybe<
       { __typename?: 'Entry' } & Pick<Types.Entry, 'id' | 'createdAt' | 'commentCount'> & {
           postedBy: { __typename?: 'User' } & Pick<Types.User, 'login' | 'html_url'>;
-          comments: Array<Types.Maybe<{ __typename?: 'Comment' } & CommentsPageCommentFragment>>;
+          comments: Array<
+            Types.Maybe<
```
dotansimha, line 19: So we still have the fragment `CommentsPageCommentFragment` available as type, and it matches to the type here, right?

.changeset/many-socks-warn.md
dotansimha: I think `inlineFragmentTypes` or `inlineFragments` should be fine

dotansimha (approved):
Is there any concern for a breaking change here? I know the types are more accurate, but we kept the fragment type the same, so it's still matches on places where we inlined it?

dev-test/githunt/__generated__/comment.query.stencil-component.tsx
n1ru4l: yes! the fragment types are still generated, and the types match!

PR #6196, opened 2021-06-27
add `overrideExisting` option for typescript-rtk-query plugin

phryneas:
## Description

This adds the `overrideExisting` option to the `typescript-rtk-query` plugin. This enables use cases such as webpack's hot module reloading when running CRA and graphql-codegen both in `watch` mode (without this, every change to a query/mutation would require a page reload, this allows endpoints to be replaced to be switched out).

## Type of change

[Sandbox for of the behaviour without this feature](https://codesandbox.io/s/github/reduxjs/redux-toolkit/tree/117b18c3e697c18b757baa703d80da3564029643/examples/query/react/graphql-codegen?file=/src/features/posts/GetPosts.graphql)

On changing a query, the code is generated, but instead of replacing the endpoint, RTK-Query warns 
> called `injectEndpoints` to override already-existing endpointName GetPosts without specifying `overrideExisting: true`

Which we have in place to prevent multiple files to accidentally override the same endpoint. In this case it's obviously not helping :slightly_smiling_face: 

So a reload is required to see an effect, which is not ideal.

With this PR, `overrideExisting` can be specified.

## How Has This Been Tested?

Build locally, linked in to the example above locally using `yalc`, edited files. Result was directly usable.

A test case for the option has been added, it has been integrated into the generated examples.

## Checklist:

## Further comments

Nothing, really. This is pretty straightforward on the codegen side :)

Oh: the last commit contains a bunch of unrelated changes triggered by starting the docs locally. I assumed it would best to also commit those so that it wouldn't confuse the next person :)


dotansimha:
This is awesome, thank you @phryneas !

Issue #6212, opened 2021-06-30
Suffix not present on hooks anymore


lucasbordeau:
**Describe the bug**
Since I upgraded to 2.2.6 and above, the suffix is not added on hooks anymore with plugin `typescript-react-apollo` and option `typesSuffix: MySuffix`.

**To Reproduce**

- Install `typescript-react-apollo@2.2.5`
- Generate hooks with option `typesSuffix: MySuffix`
- Create some custom hooks calling `useFindUserQueryMySuffix()`
- Upgrade to `typescript-react-apollo@2.2.6` or above.
- Make the exact same generation
- The apollo hooks are generated without the `MySuffix` and this breaks all of the generated hooks imports

1. My GraphQL schema:

```graphql
scalar Date

schema {
  query: Query
}

type Query {
  me: User!
  user(id: ID!): User
  allUsers: [User]
  search(term: String!): [SearchResult!]!
  myChats: [Chat!]!
}

enum Role {
  USER,
  ADMIN,
}

interface Node {
  id: ID!
}

union SearchResult = User | Chat | ChatMessage

type User implements Node {
  id: ID!
  username: String!
  email: String!
  role: Role!
}

type Chat implements Node {
  id: ID!
  users: [User!]!
  messages: [ChatMessage!]!
}

type ChatMessage implements Node {
  id: ID!
  content: String!
  time: Date!
  user: User!
}

```

2. My GraphQL operations:

```graphql
query findUser($userId: ID!) {
  user(id: $userId) {
    ...UserFields
  }
}

fragment UserFields on User {
  id
  username
  role
}
```

3. My `codegen.yml` config file:

```yml
generates:
  types-and-hooks.tsx:
    plugins:
      - typescript
      - typescript-operations
      - typescript-react-apollo
    config: 
      typesSuffix: MySuffix
```

**Expected behavior**
Could we have a way to workaround this new behavior ? There's nothing I can do except renaming all my imports or downgrading to `2.2.5`.

**Environment:**

- OS: `OSX 11.4`
- packages : 
`"@graphql-codegen/cli": "^1.21.6",`
`"@graphql-codegen/introspection": "^1.18.2",`
`"@graphql-codegen/typescript": "^1.22.4",`
`"@graphql-codegen/typescript-operations": "^1.18.3",`
`"@graphql-codegen/typescript-react-apollo": "2.2.6",`
- NodeJS: `v14.17.0`

**Additional context**

[...]

PR #6213, opened 2021-06-30
feat(react-apollo): add hooksSuffix option

ardatan:
Related #6212

### DIFF
--- a/packages/plugins/typescript/react-apollo/tests/react-apollo.spec.ts
+++ b/packages/plugins/typescript/react-apollo/tests/react-apollo.spec.ts
@@ -394,7 +394,7 @@ describe('React Apollo', () => {
       expect(content.content).toEqual(expect.stringMatching(/(?=.useUserQuery)(?!.useUserQueryType)(.+)/));
     });
 
-    it('#6001 - Should use TypesSuffix on function names if hooksSuffix provided', async () => {
+    it('#6212 - Should use TypesSuffix on function names if hooksSuffix provided', async () => {
       const docs = [
         {
           location: '',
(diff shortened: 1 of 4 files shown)


lucasbordeau:
Thanks for your reactivity ! 

Alright I just tested with this alpha version and it works again !

Except now the last letter gets lowercased : 

instead of `useGetEntityQueryGraphQL` we get `useGetEntityQueryGraphQl`

lucasbordeau:
Also do we need to change something to add autocompletion for this new property on `codegen.yml` ?

ardatan:
@lucasbordeau Autocompletion will be available when it's released :)

packages/plugins/typescript/react-apollo/tests/react-apollo.spec.ts
```diff
@@ -393,6 +393,29 @@ describe('React Apollo', () => {
...
+                id
+                username
+                email
+              }
+            }
+          `),
+        },
+      ];
+
+      const content = await plugin(schema, docs, { typesSuffix: 'Type', hooksSuffix: 'Hook' });
```
lucasbordeau, line 413: Use `hooksSuffix: 'HookXYZ'` and `(?=.useUserQueryHookXYZ)` because the last capital letters can get lowercased.

lucasbordeau (changes requested):

packages/plugins/typescript/react-apollo/tests/react-apollo.spec.ts
ardatan: Is it ok now?

lucasbordeau: Yes !!! Perfect

PR #6231, opened 2021-07-04
Fixes relative custom imports & empty plugins

arcanis:
## Description

Given that the changes are fairly small I didn't open an issue; I believe seeing the proposed changes may be easier to understand. There are currently two small fixes I encountered while developing a third-party tool using this project. I've put a comment under each one if you wish to discuss them separately.

## Type of change

## How Has This Been Tested?

Made the changes in the graphql-codegen code within my project, and things worked instead of crashing.

## Checklist:

- n/a I have commented my code, particularly in hard-to-understand areas
- n/a I have made corresponding changes to the documentation
- n/a Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/graphql-codegen-cli/src/plugins.ts
+++ b/packages/graphql-codegen-cli/src/plugins.ts
@@ -22,12 +22,12 @@ export async function getPluginByName(
     try {
       return await pluginLoader(moduleName);
     } catch (err) {
-      if (err.code !== 'MODULE_NOT_FOUND' || !err.message.includes(moduleName)) {
+      if (err.code !== 'MODULE_NOT_FOUND') {
         throw new DetailedError(
           `Unable to load template plugin matching ${name}`,
           `
               Unable to load template plugin matching '${name}'.
-              Reason: 
+              Reason:
                 ${err.message}
             `
         );
@@ -48,7 +48,7 @@ export async function getPluginByName(
     `
         Unable to find template plugin matching '${name}'
         Install one of the following packages:
-        
+
         ${possibleNamesMsg}
       `
   );
--- a/packages/graphql-codegen-cli/src/presets.ts
+++ b/packages/graphql-codegen-cli/src/presets.ts
@@ -18,12 +18,12 @@ export async function getPresetByName(
 
       return loaded as Types.OutputPreset;
     } catch (err) {
-      if (err.code !== 'MODULE_NOT_FOUND' || !err.message.includes(moduleName)) {
+      if (err.code !== 'MODULE_NOT_FOUND') {
         throw new DetailedError(
           `Unable to load preset matching ${name}`,
           `
               Unable to load preset matching '${name}'.
-              Reason: 
+              Reason:
                 ${err.message}
             `
         );
@@ -44,7 +44,7 @@ export async function getPresetByName(
     `
         Unable to find preset matching '${name}'
         Install one of the following packages:
-        
+
         ${possibleNamesMsg}
       `
   );


packages/utils/plugins-helpers/src/helpers.ts
```diff
@@ -25,7 +25,7 @@ export function isOutputConfigArray(type: any): type is Types.OutputConfig[] {
 }
 
 export function isConfiguredOutput(type: any): type is Types.ConfiguredOutput {
-  return typeof type === 'object' && type.plugins;
+  return typeof type === 'object' && (type.preset || type.plugins);
```
arcanis, line 28: Passing a plugin is currently mandatory; this is a bit overkill when it comes to preset, because presets are then expected to provide their own plugins anyway. In my own case I've built a preset that doesn't require any plugin, since it includes the relevant ones from the get-go. The change is to make it so that having a plugin isn't required if a preset is configured.

packages/graphql-codegen-cli/src/plugins.ts
```diff
@@ -22,12 +22,12 @@ export async function getPluginByName(
     try {
       return await pluginLoader(moduleName);
     } catch (err) {
-      if (err.code !== 'MODULE_NOT_FOUND' || !err.message.includes(moduleName)) {
+      if (err.code !== 'MODULE_NOT_FOUND') {
```
arcanis, line 25: The current plugin/preset loader tries to detect whether the plugin exists or not by requiring it, and catching potential resolution errors that may be thrown. One additional check is made to look for the full require path existing in the message; I believe this was intended to not catch transitive resolution errors occurring within the plugin/preset code itself. However, this isn't compatible with all environments - under PnP for example, we turn the require path into an absolute one in the error message (thus breaking the `indexOf` match).

The current change simply removes this guard; however if you wish to keep it I'm open to implement an alternative (separating the call to `require.resolve` from its `require`, thus being able to only catch errors that come from the plugin/preset resolution itself). It's slightly more involved so I didn't do it yet, let me know what you prefer.

dotansimha:
Hi @arcanis ! Thank you for this PR. 
@n1ru4l was doing the same thing (in order to support the gql-tag) plugin, so I think we can close it :) See the conflicts, it's actually the same change.
This will be available in upcoming release.

n1ru4l:
@arcanis I did not see this before, sorry! We now added this preset: https://github.com/dotansimha/graphql-code-generator/tree/master/packages/presets/gql-tag-operations

I know on Twitter you mentioned wanting to have global unique operations and we would appreciate it if you could contribute that by porting it over from your preset/plugin as an opt-in feature (I guess most people will still prefer global unique operation/fragment names)! We fixed some upstream issues with gql-tag-loader that make the API easier to use! See https://github.com/dotansimha/graphql-code-generator/tree/master/dev-test/gql-tag-operations for a code example and https://github.com/dotansimha/graphql-code-generator/blob/3f98977499f4ab014a89deae3b19e669d0e0e103/dev-test/codegen.yml#L391-L394 on the config.

arcanis:
No worry, it's absolutely the best case scenario for me! Thanks for being super-reactive and tackling that! I'll look at contributing extra optional features to the new preset, although I'm a little bit worried that some features I've implemented are perhaps a bit too specialized to my use case (for instance I made it so [the `id` field is always fetched, even if missing from the query](https://github.com/arcanis/graphql-typescript-integration/commit/4facefe7bf9b6fe9021ceabe26b99f9fcea35951)). I don't want to bother you if you don't think these would be the right direction for the project!

Issue #6240, opened 2021-07-05
add sdk generation to react-apollo


cyberhck:
Hello there,

following up on my discussion at https://github.com/dotansimha/graphql-code-generator/discussions/6238

What I figured is if I were to use code-generator outside of a react component, I will either have to use graphql-request plugin which doesn't use apollo client, OR I'll have to use generic sdk where I'll have to implement requester for where subscription might be a little tricky.

Since we already have operations, variable descriptions, adding sdk shouldn't be a big deal. With one config to turn that on, and off by default, it won't even be a breaking change.

I'd like to implement this feature.

Since this will be my first PR, I'll need someone to babysit me :slightly_frowning_face: 

Here's the approach I'm planning to take,

simply modify react-apollo client and basically create a function at last which returns everything using a apollo client based on operation type (query, mutation or subscription),

Currently the only tricky part I'm thinking of is that, react-apollo client goes through each operation and generates the whole hook itself,

but we'll probably need to return that in a object format, which means it's not a simple map from operation to a hook like thing.

Here's two option I'm thinking of:
- For each visited operations, create a private array of those function body and store key name then have a method group everything into one function that returns an object.
- From outside the visitor have another method which gets everything like generic-sdk or graphql-request

Is this something you'd merge if I were to send a PR?

PR #6242, opened 2021-07-05
feat(react-apollo): add withSdk option

cyberhck:

## Description
While I didn't wait for anyone's confirmation, I'm happy to make any changes you request. I didn't want to wait because I've a project for which I want to use this.

Basically instead of only creating hooks, I'm planning to also generate Sdk, that way we can use same generated code without having to generate using multiple plugins.

The change this does is basically allow withSdk field on config, and if it's on, it'll generate getSdk method which expects an apollo client. I think we can also write a useSdk hook which uses the useApolloClient and returns the sdk, let me know if this is something we want to add, it should be very straightforward.

Related #6240 

## Type of change

## How Has This Been Tested?

- [x] I wrote a unit test expecting to see getSdk method, which I did see
- [ ] I want someone's help to write better unit tests, for now I've no idea what else to test, very new to this project (my first PR)
- [ ] I'd also like someone's help to try this plugin out before merging on my project to see if it's actually working.

**Test Environment**:
- OS: Linux <username> 5.8.0-59-generic #66~20.04.1-Ubuntu SMP Thu Jun 17 11:14:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
- `@graphql-codegen/...`: 
- NodeJS: 14.6.1

## Checklist:

## Further comments

### DIFF
--- a/packages/plugins/typescript/react-apollo/src/visitor.ts
+++ b/packages/plugins/typescript/react-apollo/src/visitor.ts
@@ -548,6 +548,10 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
     });
     return `export const getSdk = (client: Apollo.ApolloClient<any>) => {
       ${sdkOperations.join(',\n')}
-    };`;
+    };
+    export const useSdk = () => {
+      const client = Apollo.useApolloClient();
+      return getSdk(client);
+    }`;
   }
 }
--- a/packages/plugins/typescript/react-apollo/src/visitor.ts
+++ b/packages/plugins/typescript/react-apollo/src/visitor.ts
@@ -546,7 +546,7 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
           return client.query<${x.operationResultType}>({variables, query: ${documentVariableName})
       }`;
     });
-    return `export const getSdk = (client: ApolloClient<any>) => {
+    return `export const getSdk = (client: Apollo.ApolloClient<any>) => {
       ${sdkOperations.join(',\n')}
     };`;
   }
--- a/packages/plugins/typescript/react-apollo/src/visitor.ts
+++ b/packages/plugins/typescript/react-apollo/src/visitor.ts
@@ -22,6 +22,7 @@ export interface ReactApolloPluginConfig extends ClientSideBasePluginConfig {
   withHooks: boolean;
   withMutationFn: boolean;
   withRefetchFn: boolean;
+  withSdk: boolean;
   apolloReactCommonImportFrom: string;
   apolloReactComponentsImportFrom: string;
   apolloReactHocImportFrom: string;
@@ -38,6 +39,13 @@ export interface ReactApolloPluginConfig extends ClientSideBasePluginConfig {
 export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPluginConfig, ReactApolloPluginConfig> {
   private _externalImportPrefix: string;
   private imports = new Set<string>();
+  private _operationsToInclude: {
+    node: OperationDefinitionNode;
+    documentVariableName: string;
+    operationType: string;
+    operationResultType: string;
+    operationVariablesTypes: string;
+  }[] = [];
 
   constructor(
     schema: GraphQLSchema,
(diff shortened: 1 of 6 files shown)


packages/plugins/typescript/react-apollo/tests/__snapshots__/react-apollo.spec.ts.snap
```diff
@@ -1,5 +1,48 @@
...
+      },
+SubmitRepositoryMutation(variables?: SubmitRepositoryMutationVariables) {
+          return client.query<SubmitRepositoryMutation>({variables, query: SubmitRepositoryDocument)
+      }
+    };",
+  "prepend": Array [
+    "import { gql } from '@apollo/client';",
+    "import * as Apollo from '@apollo/client';",
+  ],
+}
```
cyberhck, line 43: because the prepend has only Apollo, I think I'll need to change the type of client from: `ApolloClient<any>` to `Apollo.ApolloClient<any>`

packages/plugins/typescript/react-apollo/src/visitor.ts
```diff
@@ -513,4 +529,25 @@ export class ReactApolloVisitor extends ClientSideBaseVisitor<ReactApolloRawPlug
...
+      const documentVariableName = this.getDocumentNodeVariable(x.node, x.documentVariableName);
+
+      const operationName = x.node.name.value;
+      return `${operationName}${x.operationType}(variables${optionalVariables ? '?' : ''}: ${
+        x.operationVariablesTypes
+      }) {
+          return client.query<${x.operationResultType}>({variables, query: ${documentVariableName})
+      }`;
+    });
+    return `export const getSdk = (client: ApolloClient<any>) => {
```
cyberhck, line 549:
```suggestion
    return `export const getSdk = (client: Apollo.ApolloClient<any>) => {
```

packages/plugins/typescript/react-apollo/tests/__snapshots__/react-apollo.spec.ts.snap
```diff
@@ -1,5 +1,48 @@
...
+export const SubmitRepositoryDocument = gql\`
+    mutation SubmitRepository($name: String) {
+  submitRepository(repoFullName: $name) {
+    id
+  }
+}
+    \`;
+export type SubmitRepositoryMutationResult = Apollo.MutationResult<SubmitRepositoryMutation>;
+export type SubmitRepositoryMutationOptions = Apollo.BaseMutationOptions<SubmitRepositoryMutation, SubmitRepositoryMutationVariables>;
+export const getSdk = (client: ApolloClient<any>) => {
```
cyberhck, line 31:
```suggestion
export const getSdk = (client: Apollo.ApolloClient<any>) => {
```

packages/plugins/typescript/react-apollo/src/visitor.ts
dotansimha: Yeah, you have helpers for getting the correct identifier. Using ApolloClient might break in case of a missing import.

dotansimha (changes requested):
Thank you @cyberhck ! please see me comment regarding the hardcoded Apollo usage.

packages/plugins/typescript/react-apollo/src/visitor.ts
cyberhck: Hi, I haven't actually seen this, tried looking into generic-sdk, where can I find an example?

cyberhck: I've added a commit which does add `Apollo`,

I couldn't figure out the helper method you mentioned, let me know where I can look into that, and I will update.

cyberhck:
turns out it's failing because it adds new lines to sdk, technically it's a side effect of introducing the feature, I have two options:
- update snapshot for that plugin (easy, straight forward)
- write in a way that there's no new empty line (potential array.push call, but no tests getting updated)

cyberhck:
On the discussion which is mentioned in issue, I found out that Apollo Client Sdk as a separate plugin might be a good idea.

BUT I don't know how to do that, my only question is, how will I define dependency on another plugin,
also, will I have to emit only the `getSdk` part? Or will I have to emit variable types, documents etc? how will that work?

If that's really the better way of doing things, given a little help, I'll be more than willing to send another PR :slightly_smiling_face: 

also how to test locally to see if my things are working, while snapshot testing is good, I don't know if it'll automatically generate when plugin dependency is defined, or I have to do something.

cyberhck:
closed in favor of #6246

PR #6246, opened 2021-07-06
Graphql apollo sdk

### LINKED ISSUE #6240: add sdk generation to react-apollo

cyberhck:

## Description
The change introduces a new plugin which generates standalone SDK with apollo, This doesn't have an opinion on frameworks you would be using.

Related #6240 
closes #6242 
closes #6240 

## Type of change

## How Has This Been Tested?

- [x] I wrote a unit test expecting to see getSdk method, which I did see
- [x] I also wrote integration tests to verify the generated sdk actually works.

**Test Environment**:
- OS: Linux <username> 5.8.0-59-generic # 66~20.04.1-Ubuntu SMP Thu Jun 17 11:14:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
- `@graphql-codegen/...`: 
- NodeJS: 14.6.1

## Checklist:

## Further comments


packages/plugins/typescript/graphql-apollo/jest.config.js
```diff
@@ -0,0 +1 @@
+module.exports = require('../../../../jest.project')({	dirname: __dirname });
```
ardatan, line 1: Why do we need that?

packages/plugins/typescript/graphql-apollo/package-lock.json
```diff
@@ -0,0 +1,1942 @@
+{
+  "name": "@graphql-codegen/typescript-graphql-apollo",
```
ardatan, line 2: I think you ran npm install on this directory. We use yarn workspaces so we call `yarn test` and `yarn build` on the root.

packages/plugins/typescript/graphql-apollo/tests/graphql-apollo.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    it('Should support useTypeImports', async () => {
+      const config = { useTypeImports: true };
+      const docs = [{ location: '', document: basicDoc }];
+      const result = (await plugin(schema, docs, config, {
+        outputFile: 'graphql.ts',
+      })) as Types.ComplexPluginOutput;
+
+      const usage = `
+async function test() {
+  const Client = require('graphql-request').GraphQLClient;
```
ardatan, line 72: Should we update those?

packages/plugins/typescript/graphql-apollo/jest.config.js
cyberhck: not sure, every typescript I checked had this.

ardatan: Ok let's keep it then.

cyberhck:
@ardatan let me know if any changes are required :slightly_smiling_face:

cyberhck:
not sure why this is failing :slightly_frowning_face:

cyberhck:
They pass locally, and when I comment out removal of generated file in test, it actually generates file correctly over there. But on CI for some reason it's failing

cyberhck:
turns out the project I copied from, `generic-sdk` also commits the generated file (probably because we could see the generated file directly on github or IDE without having to fiddling with tests). I've pushed a commit which does the same.

cyberhck:
hello @dotansimha , please let me know if you'd like any changes :slightly_smiling_face:

cyberhck:
hi @dotansimha I changed the package version to 0.0.0 and start with major version like you suggested on another new plugin PR, can you review this and see if it can be merged? tests were passing, I'm hoping to use this without having to publish a temporary package myself :slightly_smiling_face:

cyberhck:
can we please review this :slightly_smiling_face: I kinda need this :)

packages/plugins/typescript/graphql-apollo/src/visitor.ts
```diff
@@ -0,0 +1,92 @@
...
+        x.node.variableDefinitions.length === 0 ||
+        x.node.variableDefinitions.every(v => v.type.kind !== Kind.NON_NULL_TYPE || v.defaultValue);
+
+      const documentVariableName = x.documentVariableName;
+
+      const operationName = x.node.name.value;
+      return `${operationName}${x.operationType}(variables${optionalVariables ? '?' : ''}: ${
+        x.operationVariablesTypes
+      }) {
+          return client.query<${x.operationResultType}>({variables, query: ${documentVariableName}})
```
MattSchiller, line 83: Should this `.query` not be `.mutation` for mutations?

```suggestion
          return client.${operationType  === 'Mutation' ? 'mutation' : 'query'}<${x.operationResultType}>({variables, query: ${documentVariableName}})
```

MattSchiller (commented):
Hey Nishchal!

Thank you for putting this code together! My team and I were actually running into the same issue where we didn't have syntax parity across our hooks & NextJS SSR/SSG functions (which this solves).

I was wondering if you'd be open to a slightly different API for the package though, so that it looked even more like hooks:

Current:
```ts
const { getSdk } = require('./test-files/graphql-apollo-sdk');
const apolloClient = new ApolloClient({
  cache: new InMemoryCache(),
  link: new HttpLink({ uri: 'http://localhost:4000/graphql', fetch }),
});
const queryResult = await getSdk(apolloClient).AddQuery({ x: 2, y: 3 });
```    

More Hook-like:
```ts
const { makeAddQuery } = require('./test-files/graphql-apollo-sdk'); // generated output
const apolloClient = new ApolloClient({
  cache: new InMemoryCache(),
  link: new HttpLink({ uri: 'http://localhost:4000/graphql', fetch }),
});
const queryResult = await makeAddQuery(apolloClient, { x: 2, y: 3 });
```

I actually already have the code for all of this written up, though I have it embedded within the existing `typescript-react-apollo` package in my local environment (to reuse a bunch of their helpers, though I don't see why they couldn't also live in your module here).

This is the relevant code:
```ts
  private _buildClientMethodsJSDoc(
    node: OperationDefinitionNode,
    operationName: string,
    operationType: string
  ): string {
    const variableString = node.variableDefinitions.reduce((acc, item) => {
      const name = item.variable.name.value;

      return `${acc}\n *      ${name}: // value for '${name}'`;
    }, '');

    const queryDescription = `
 * To run a query from outside a component, call \`apollo${operationName}\` and pass it your Apollo Client and any options that fit your needs.
 * \`make${operationName}\` returns a typed promise from passed Apollo Client that contains loading, error, and data properties;`;

    const queryExample = `
 * const { data, loading, error } = make${operationName}({
 *   variables: {${variableString}
 *   },
 * });`;

    const mutationDescription = `
 * To run a mutation, you first call \`make${operationName}\` and pass it your Apollo Client and any options that fit your needs.
 * \`make${operationName}\` returns a tuple that includes:
 * - A mutate function that you can call at any time to execute the mutation
 * - An object with fields that represent the current status of the mutation's execution`;

    const mutationExample = `
 * const [${camelCase(operationName)}, { data, loading, error }] = make${operationName}({
 *   variables: {${variableString}
 *   },
 * });`;

    return `
/**
 * __make${operationName}__
 *${operationType === 'Mutation' ? mutationDescription : queryDescription}
 *
 * @param baseOptions options that will be passed into the ${operationType.toLowerCase()}, supported options are listed on: https://www.apollographql.com/docs/react/api/react-hooks/#${
      operationType === 'Mutation' ? 'options-2' : 'options'
    };
 *
 * @example${operationType === 'Mutation' ? mutationExample : queryExample}
 */`;
  }

  private _buildClientMethods(
    node: OperationDefinitionNode,
    operationType: string,
    documentVariableName: string,
    operationResultType: string,
    operationVariablesTypes: string,
    hasRequiredVariables: boolean
  ): string {
    const nodeName = node.name?.value ?? '';
    const suffix = this._getHookSuffix(nodeName, operationType);
    const operationName: string =
      this.convertName(nodeName, {
        suffix,
        useTypesPrefix: false,
        useTypesSuffix: false,
      }) + this.config.hooksSuffix;

    const clientMethodName = operationTypeToClientMethod[operationType];

    if (!clientMethodName) {
      return null;
    }

    this.imports.add(this.getApolloReactCommonImport(true));
    this.imports.add(this.getDefaultOptions());

    const apolloPackage = this.getApolloReactCommonIdentifier();

    const methodFns = [
      `export function make${operationName}(apolloClient: ${apolloPackage}.ApolloClient<any>, variables${
        hasRequiredVariables ? '' : '?'
      }: ${operationVariablesTypes}) {
        return apolloClient.${clientMethodName}<${operationResultType}>({
          ${operationType === 'Mutation' ? 'mutation' : 'query'}: ${this.getDocumentNodeVariable(
        node,
        documentVariableName
      )}, variables });
  }`,
    ];

    if (this.config.addDocBlocks) {
      methodFns.unshift(this._buildClientMethodsJSDoc(node, operationName, operationType));
    }

    return methodFns.join('\n');
  }

File: packages/plugins/typescript/graphql-apollo/src/visitor.ts (hunk 4 above)
Comment (cyberhck) [reply packages/plugins/typescript/graphql-apollo/src/visitor.ts:83]: right, I've to handle query, mutations OR subscriptions as well, will update.

**Comment by cyberhck** [comment]
hi @MattSchiller regarding your hooks questions, this project already has hooks plugin. The objective of this PR is to provide a SDK which you'd use when you don't have hooks option like vanillajs environment or in libraries other than react OR in react where you can't use hooks (schema validation on forms with yup)

File: packages/plugins/typescript/graphql-apollo/src/visitor.ts (hunk 4 above)
Comment (cyberhck) [reply packages/plugins/typescript/graphql-apollo/src/visitor.ts]: can you check this? I didn't exactly apply the suggestion you mentioned, I created a private static method instead which will handle `Subscription`, `Mutation` and `Query`

**Comment by cyberhck** [comment]
Can I get a CI/CD check? cc @ardatan

**Comment by MattSchiller** [comment]
Hey @cyberhck!

> May I ask why you can't use the hooks generated by typescript-react-apollo plugin? 

We can't only use the hooks because our use-case is to sometimes use the hooks (inside react components) and sometimes to fetch outside of react components in the static-generated portions of NextJS. But we want to still use Apollo, since Apollo caches/dehydrates the queries in a way that let us send it with the initial page load so that we can keep the server-fetched GQL cache with it, so the client can have that too.

And since the autogenerated hooks are using the Apollo hooks, which are looking to find the `ApolloProvider` context, it won't work in NextJS methods, outside of the React tree.

We could end up using the code generated from your plugin here, but I'd like to have the dev-experience for my junior devs be as consistent as possible when they're learning to use GQL. Having two methods `useXYZQuery` and `makeXYZQuery`, one for hooks, and one for NextJS methods just seems like such a clean experience when we're using both.

> If I were to change to this type of API, it'd be more difficult to write tests where you're not using jest for example. Which is why I want to add a SdkType as well, that's because I'd like to be able to say I need a SDK in a method for example, with TypeScript it'd be better developer experience if I return an object 🙂

For writing tests, I don't know exactly what you meant, though think it'd be about the same, no? Like, importing specific `make*` methods to test is about the same as importing `getSdk` and then calling specific methods on it.

But I don't want to twist your arm on it and I want to respect your vision for this module. Can we at least get a better name for the main import than `getSdk`? perhaps `makeApolloRequest`? Also, it seems quite strange to have the methods pascalCase instead of camelCase

Instead of:
```ts
getSdk(apolloClient).AddQuery({ x: 1, y: 2 });
```
Maybe:
```ts
makeApolloRequest(apolloClient).addQuery({ x: 1, y: 2 })
```

packages/plugins/typescript/graphql-apollo/src/visitor.ts
```diff
@@ -0,0 +1,106 @@
...
+        x.node.variableDefinitions.length === 0 ||
+        x.node.variableDefinitions.every(v => v.type.kind !== Kind.NON_NULL_TYPE || v.defaultValue);
+
+      const documentVariableName = x.documentVariableName;
+
+      const operationName = x.node.name.value;
+      return `${operationName}${x.operationType}(variables${optionalVariables ? '?' : ''}: ${
+        x.operationVariablesTypes
+      }) {
+          return client.${GraphQLApolloVisitor.getApolloOperation(x.operationType)}<${
```
MattSchiller, line 83: Here would be the pascal case'ing I'm requesting. Can be done with:

```ts
import { camelCase } from 'change-case-all';

...
return client.${camelCase(GraphQLApolloVisitor.getApolloOperation(x.operationType))}<${
```

packages/plugins/typescript/graphql-apollo/src/visitor.ts
```diff
@@ -0,0 +1,106 @@
...
+    const sdkOperations = this._operationsToInclude.map(x => {
+      const optionalVariables =
+        !x.node.variableDefinitions ||
+        x.node.variableDefinitions.length === 0 ||
+        x.node.variableDefinitions.every(v => v.type.kind !== Kind.NON_NULL_TYPE || v.defaultValue);
+
+      const documentVariableName = x.documentVariableName;
+
+      const operationName = x.node.name.value;
+      return `${operationName}${x.operationType}(variables${optionalVariables ? '?' : ''}: ${
```
MattSchiller, line 80: Here would be the pascal case'ing I'm requesting. Can be done with:

```ts
import { camelCase } from 'change-case-all';

...
return `${camelCase(operationName})${x.operationType})(...
```

packages/plugins/typescript/graphql-apollo/src/visitor.ts
```diff
@@ -0,0 +1,106 @@
...
+      case 'Subscription':
+        return 'subscribe';
+      case 'Mutation':
+        return 'mutate';
+      case 'Query':
+        return 'query';
+      default:
+        throw new Error('unknown type: ' + operationType);
+    }
+  }
```
MattSchiller, line 105: This solution looks great 👍

cyberhck: this actually comes from gql type, it doesn't really have anything to do with apollo though, but unknown operation type would be much much better thanks, I'll change that :slightly_smiling_face:

packages/plugins/typescript/graphql-apollo/src/visitor.ts
cyberhck: You're right, I'll also update this :slightly_smiling_face: thanks.

cyberhck:
@MattSchiller is it okay if I use the name: `buildApolloSdk`?

MattSchiller:
> @MattSchiller ~is it okay if I use the name: `buildApolloSdk`?~
> Actually `generic-sdk` also uses getSdk and so does `graphql-request`, I'll keep it `getSdk` as well,
> 
> how about you create a makeGraphqlRequest function which just does a `getSdk` and returns, that should solve your requirement 🙂

Sounds good to me. I'm always down to make an adapter layer locally =)

And thank you for making the other changes! Now, let's get this one merged/deployed 🚀

cyberhck:
hi @dotansimha, cc @ardatan 

What are my next steps?

cyberhck:
I've pushed another commit which should fix the build issue, can we run the workflow again :slightly_smiling_face:

cyberhck:
hi @dotansimha cc @ardatan I'd like for you to run workflow again. :slightly_smiling_face:

cyberhck:
@dotansimha cc @ardatan what are my next steps to get this merged?

packages/plugins/typescript/graphql-apollo/src/visitor.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      const documentVariableName = x.documentVariableName;
+
+      const operationName = x.node.name.value;
+      return `${camelCase(operationName)}${x.operationType}(variables${optionalVariables ? '?' : ''}: ${
+        x.operationVariablesTypes
+      }) {
+          return client.${GraphQLApolloVisitor.getApolloOperation(x.operationType)}<${
+        x.operationResultType
+      }>({variables, query: ${documentVariableName}})
```
MattSchiller, line 86: Hey there!

I'm currently building out a bit more robust of an apollo client call, and I'm finding that variables is insufficient, and that I'm using [context](https://www.apollographql.com/docs/react/api/core/ApolloClient/#ApolloClient.query).

Would it be possible for the API for the method be updated from:
```ts
client.query(variables: MyQueryVariables)
```
to 
```ts
client.query({ variables }:  QueryOptions<MyQueryVariables>) // or MutationOptions / SubscriptionOptions
```

This would not only solve my current need, but keep this plugin a bit more future-resilient

cyberhck: I actually love this idea :slightly_smiling_face: Do you mind sending a PR on top of this? I'm swamped currently at work, I'll try to give this a try over the weekend if I happen to get any free time, but would love if you could help me out, either way this is going to happen :slightly_smiling_face:

cyberhck: I think this also helps when we write middlewares, you can access the context you pass :slightly_smiling_face: if I understand correctly, forgive me if I'm wrong, haven't done client side apollo, I'm all backend graphql for now :slightly_smiling_face:

cyberhck: however it doesn't seem like we're ever going to get a review from maintainers though, I'm starting to think to publish this from my account, but I don't think it's such a good idea though.

MattSchiller: > I think this also helps when we write middlewares, you can access the context you pass 🙂 if I understand correctly, forgive me if I'm wrong, haven't done client side apollo, I'm all backend graphql for now 🙂

You're absolutely correct about this. I'm using apollo in all 3, SSG/SSR/CSR and context is accessible in all of them. That's actually why I want this whole plugin to begin with, so that I have consistent and clean types/syntactic sugar. You're the real hero here.

I've got a crazy-busy week here too 😭 😎 .

Let me work up at least the base of this branch this week and if you have the bandwidth this weekend, you could maybe update the tests or something?

> however it doesn't seem like we're ever going to get a review from maintainers though, I'm starting to think to publish this from my account, but I don't think it's such a good idea though.

i'm optimistic that they'll review this soon =) It's not a complicated PR, so... fingers crossed.

And if in another 2 weeks, we still don't hear anything, I wonder what the way to self-publish this would be such that it would still work with the core codegen lib?

cyberhck: I'll give it a stab over the coming weekend (just set a reminder on slack), :slightly_smiling_face: but if you happen to push something I can take care of tests and making these pass.

dotansimha (approved):
Looks good to me. 
I do want @ardatan to take another look before merging.

cyberhck:
@dotansimha Thanks for the review,
 actually @MattSchiller and I decided we'll add one more commit, please give me this weekend, and I'll have everything pushed, I'd actually love it if you could keep an eye out for this PR and periodically run CI so I know I'm on right path :slightly_smiling_face: it shouldn't be a huge change though.

cyberhck:
@dotansimha I'd like to check if CI will pass.

@MattSchiller I've pushed two new commits which should address your usecase as well. I've also published this for temporary usage at [npm](https://www.npmjs.com/package/@cyberhck/typescript-graphql-apollo-temp) please use the latest version. But please be advised that if you were to use that, there's no guarantee that it'll keep getting updates. Also there's a chance API for this still might change in case there's some reviewers asking for change. :slightly_smiling_face: I've published that because I really need it available soon, but as soon as this PR is merged, I'll switch to this one as well.

The reason this feels so dirty is because typings in apollo has one weird quirk :smile: https://github.com/apollographql/apollo-client/issues/8537 they've flipped the type parameters only on Mutation.

I'm sure we can clean this up, please suggest and I'll take care of it.

packages/plugins/typescript/graphql-apollo/src/visitor.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  public get sdkContent(): string {
+    const sdkOperations = this._operationsToInclude.map(x => {
+      const documentVariableName = x.documentVariableName;
+      const optionType = GraphQLApolloVisitor.getApolloOperationOptionType(x.operationType);
+      const generics = [x.operationResultType, x.operationVariablesTypes];
+
+      const operationName = x.node.name.value;
+      const genericParameter = x.operationType !== 'Mutation' ? [...generics].reverse() : [...generics];
```
cyberhck, line 80: maybe it's a good idea to comment why this is needed?

MattSchiller:
> I'm sure we can clean this up, please suggest and I'll take care of it.

You mean to the underlying package and types? I think we're fine as-is, right? Like, presumeably other packages are also built on top of this inconsistent ordering of elements, which isn't the end of the world.

I'm happy with how this looks now =)

Just wondering, and apologies if this is a lazy question: Does this config have the same [skipTypename](https://www.graphql-code-generator.com/docs/plugins/typescript-operations) option as the base plugins? It'll still work with all of those, right? My project uses this (among others) to avoid having some really gnarly `Extract` utility types.

Otherwise, :shipit:

cyberhck:
> Just wondering, and apologies if this is a lazy question: Does this config have the same skipTypename option as the base plugins? It'll still work with all of those, right? My project uses this (among others) to avoid having some really gnarly Extract utility types.

I have absolutely no idea how that part works, if it works with graphql-request or generic-sdk then it should work with this one. One way to find out would be to try it,

MattSchiller:
@dotansimha Hey hey! What's still needed for this package to be fully deployed? Thank you!

cyberhck:
hi, @dotansimha , it's been quite a while I really want this to be merged so I can consume it correctly on my applications, I still have tech debt where I'm using a temporarily published client, it's a mess.

If you need any changes, please let me know and I will do them, is there anything I can do to get this merged?

cyberhck:
bump @dotansimha in case you missed the above notification

cyberhck:
hey @dotansimha I'm pinging you again to see if there's anything I can do to get this reviewed quicker, :slightly_smiling_face:

n1ru4l (approved):
Looks good to me!

PR #6267, opened 2021-07-07
graphql-tools-update | gql-tag preset/plugin | drop node 10 support

n1ru4l:
gql-tag plugin/preset inspired by https://github.com/arcanis/graphql-typescript-integration (https://github.com/n1ru4l/graphql-typescript-integration/pull/1)

NOTE: updating graphql-tools is a major breaking change as it drops Node 10 support.


.changeset/shaggy-crews-stare.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphql-codegen/gql-tag': minor
```
dotansimha, line 2: Maybe `gql-tag-operations`?

packages/graphql-codegen-cli/src/load.ts
```diff
@@ -77,11 +77,29 @@ export async function loadDocuments(
...
+      },
+    }),
+    new GitLoader(),
+    new GithubLoader(),
+    new GraphQLFileLoader(),
+  ];
+
+  const ignore: Array<string> = [];
+  for (const generatePath of Object.keys(config.generates)) {
+    if (extname(generatePath) === '') {
```
dotansimha, line 93: What about directories? Shouldn't we add that as well?

website/docs/presets/gql-tag.md
```diff
@@ -0,0 +1,104 @@
...
+In order to use this preset, you need to add the following configuration to your `codegen.yml`:
+
+```yml
+schema: src/path/to/your/schema.graphql
+documents:
+  - 'src/**/*.ts'
+  - '!src/gql/**/*'
+generates:
+  ./src/gql/:
+    preset: gql-tag-preset
```
dotansimha, line 44: @n1ru4l the goal of the preset is to generate a `d.ts` near the operation file? 
I think maybe we can generate a single `d.ts` as well? without a preset? and use `declare module` with wildcards to add types to a specific identifier in a file?

dotansimha (approved):

website/docs/presets/gql-tag.md
n1ru4l: The file does not generate any d.ts files, only a graphql.ts and gql.ts file: https://github.com/dotansimha/graphql-code-generator/pull/6267/files#diff-a133eb90520d99b44b280c99cc17eee7369c9df5be1f731c22671d12adec06f8

No files are generated near the operation files, the goal is to not clutter the repository with codegen artifact files

packages/graphql-codegen-cli/src/load.ts
n1ru4l: extname will always return an empty string if a path has no extension, theoretically a path with extension could also be a folder, in order to be 100% sure we would have to access the file-system here

website/docs/presets/gql-tag-operations.md
```diff
@@ -0,0 +1,102 @@
+---
+id: gql-tag-operations
+title: gql-tag-operations
+---
+
+This preset generates typings for your inline `gql` function usages, without having to manually specify import statements for the documents. All you need to do is import your `gql` function and run codegen in watch mode.
+
+Huge thanks to [Maël Nison](https://github.com/arcanis), who conceptualized the foundation for this preset [over here](https://github.com/arcanis/graphql-typescript-integration).
```
merceyz, line 8: This is nice and all but the MIT license requires you to include a full copy of it in the source code

n1ru4l: can u send a pr?

PR #6291, opened 2021-07-11
Added new config for `typescript-resolvers`: `externalMappersFrom`

dotansimha:
This PR adds a new config flag for `typescript-resolvers`. If you wish to leverage TypeScript type-system and TypeScript type-inference (and you are using TS > 4), you can use `externalMappersFrom` config flag for defining your GraphQL resolvers mappers in a TypeScript file, instead in the codegen YAML config file.

You can point to an external TS `type` identifier as a general mapper:

```yaml
schema: schema.graphql
generates:
  ./resolvers-types.ts:
    config:
      externalMappersFrom: "./my-mappers#TypeMapping"
    plugins:
      - typescript
      - typescript-resolvers
```

And then in `my-mappers.ts` file, export an identifier called `TypeMapping` with a custom type mapping:

```ts
import { UserModel, UserProfile } from "./models";

export type TypeMapping = {
  User: UserModel,
  Profile: UserProfile,
}
```

### DIFF
--- a/.changeset/heavy-windows-remain.md
+++ b/.changeset/heavy-windows-remain.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+---
+
+Added support for externalMappersFrom flag
--- a/.changeset/twelve-poems-stare.md
+++ b/.changeset/twelve-poems-stare.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-resolvers': minor
+---
+
+Added support for using `mappers` with an external TypeScript type, and use type inference for type-checking. (`externalMappersFrom`)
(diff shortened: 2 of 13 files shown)


dotansimha:
This PR is currently blocked, until we'll figure out how to apply this to child-resolvers (where only partial resolution is being mapped)

n1ru4l:
I am not sure whether this is an improvement as we still would have one central file that we need to touch every time we introduce a new GraphQL type whose business logic model is not 1:1 mappable to the GraphQL type. I know that it is not possible to work around having to define the mapping anywhere, but prefer to declare it locally next to the typedefs instead of at a central place?

dotansimha:
> I am not sure whether this is an improvement as we still would have one central file that we need to touch every time we introduce a new GraphQL type whose business logic model is not 1:1 mappable to the GraphQL type. I know that it is not possible to work around having to define the mapping anywhere, but prefer to declare it locally next to the typedefs instead of at a central place?

With a TS interface, we can also do module augmentation, so you can declare the TS interface in one place, and point codegen there. and you can do module augmentation in each module/typedefs and extend the type

PR #6330, opened 2021-07-17
map directive on FIELD_DEFINITION to custom field resolver type

n1ru4l:
This can be handy for automatically having the correct type when using directives for authorization as it allows providing a custom context, without having to manually type it.

### DIFF
--- a/.changeset/weak-plums-reflect.md
+++ b/.changeset/weak-plums-reflect.md
@@ -38,7 +38,7 @@ export type UnauthenticatedResolver<TResult, TParent, _TContext, TArgs> = (
   info: GraphQLResolveInfo
 ) => Promise<TResult> | TResult;
 
-export type AUthenticatedResolver<TResult, TParent, _TContext, TArgs> = (
+export type AuthenticatedResolver<TResult, TParent, _TContext, TArgs> = (
   parent: TParent,
   args: TArgs,
   context: AuthenticatedContext,
--- a/.changeset/weak-plums-reflect.md
+++ b/.changeset/weak-plums-reflect.md
@@ -9,6 +9,8 @@ Allow overwriting the resolver type signature based on directive usages.
 
 For actually ensuring that a type is correct at runtime you will have to use schema transforms (e.g. with [@graphql-tools/utils mapSchema](https://www.graphql-tools.com/docs/schema-directives)) that apply those rules! Otherwise, you might end up with a runtime type mismatch which could cause unnoticed bugs or runtime errors.
 
+Example configuration:
+
 ```yml
 config:
   # This was possible before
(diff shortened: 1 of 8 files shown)


packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -30,21 +32,59 @@ export const plugin: PluginFunction<TypeScriptResolversPluginConfig, Types.Compl
...
+  const visitor = new TypeScriptResolversVisitor(
+    { ...config, customDirectiveToResolverTypeNameMappings },
+    transformedSchema
+  );
   const namespacedImportPrefix = visitor.config.namespacedImportName ? `${visitor.config.namespacedImportName}.` : '';
 
-  const printedSchema = config.federation
-    ? printSchemaWithDirectives(transformedSchema)
-    : printSchema(transformedSchema);
+  const printedSchema = printSchemaWithDirectives(transformedSchema);
```
n1ru4l, line 81: `printSchema` does not care about directives...

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -971,6 +979,13 @@ export type IDirectiveResolvers${contextType} = ${name}<ContextType>;`
       });
       const mappedTypeKey = isSubscriptionType ? `${mappedType}, "${node.name}"` : mappedType;
 
+      const directiveMappings =
+        node.directives
+          ?.map(directive => this._customDirectiveToResolverTypeNameMappings.get(directive.name as any))
+          .filter(Boolean) ?? [];
+
+      const resolverType = isSubscriptionType ? 'SubscriptionResolver' : directiveMappings[0] ?? 'Resolver';
```
n1ru4l, line 987: We take the first, I don't think this should be over-used. My ideal use-case is having ``@unauthenticated` for only marking those fields that should be available with auth and thus making the viewer within that resolver optional 🤔

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -971,6 +979,13 @@ export type IDirectiveResolvers${contextType} = ${name}<ContextType>;`
       });
       const mappedTypeKey = isSubscriptionType ? `${mappedType}, "${node.name}"` : mappedType;
 
+      const directiveMappings =
+        node.directives
+          ?.map(directive => this._customDirectiveToResolverTypeNameMappings.get(directive.name as any))
```
n1ru4l, line 984: Something is wrong with the typings of `node`. The type of directive indicates accessing via `directive.name.value`, but actually the `directive.name` is already the string at runtime.

.changeset/weak-plums-reflect.md
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export type UnauthenticatedResolver<TResult, TParent, _TContext, TArgs> = (
+  parent: TParent,
+  args: TArgs,
+  context: UnauthenticatedContext,
+  info: GraphQLResolveInfo
+) => Promise<TResult> | TResult;
+
+export type AUthenticatedResolver<TResult, TParent, _TContext, TArgs> = (
```
n1ru4l, line 41:
```suggestion
export type AuthenticatedResolver<TResult, TParent, _TContext, TArgs> = (
```

Issue #6414, opened 2021-08-04
typescript-react-query: `exposeFetcher` fetcher missing dataSource argument


dan-lee:
**Describe the bug**
Generating types with `@graphql-codegen/typescript-react-query` and `exposeFetcher: true` exposes the fetcher but with missing `dataSource` argument (unlike the `useXXX` equivalent)

![image](https://user-images.githubusercontent.com/571589/128164041-2b35aba1-3020-478c-8020-3f2434f95f13.png)

**To Reproduce**

Set `exposeFetcher: true` for `@graphql-codegen/typescript-react-query`

Check CodeSandbox: https://codesandbox.io/s/busy-dhawan-y03d0?file=/types.ts

**Expected behavior**

It should be possible to pass in `dataSource`

**Environment:**

    "@graphql-codegen/cli": "2.0.1",
    "@graphql-codegen/typescript": "2.0.0",
    "@graphql-codegen/typescript-operations": "2.0.1",
    "@graphql-codegen/typescript-react-query": "2.0.0",

paolotiu:
Hey, @dan-lee! I missed this while creating the `exposeFetcher` feature. It  _should_ be a quick fix 😄

PR #6416, opened 2021-08-04
[typescript-react-query] Pass `dataSource` as argument to `fetcher`

dan-lee:
## Description

This passes in the missing `dataSource` argument to `fetcher`, when `exposeFetcher: true` is set.

I am not sure about the order of arguments, for the hook it's the first argument, here it's second.
But changing the order of arguments would be a breaking change, wouldn't it?

Reported issue #6414 

## Type of change

## How Has This Been Tested?

- [x] Original test has been extended

## Checklist:


nenadfilipovic:
Yup, same problem, with provided endpoint and fetchParams inside codegen.yml it is fixed, but to use fetcher standalone this fix is needed.

paolotiu:
Hey, @dan-lee I was about to make my own pull request, but then I saw yours. I think the datasource should be passed as the first argument since the original hook accepts it as the first argument. Although it would technically be a breaking change, current users won't be able to use the exposed fetcher since the datasource isn't an argument.

dan-lee:
Hey @paolotiu, sorry I have been on vacation and couldn't make the changes sooner. It makes sense to change the order of arguments, as you mentioned. I just pushed it and it would be ready to merge if it's not already fixed. :)

paolotiu:
Thanks for the fix @dan-lee! No problem at all, hope you enjoyed your vacation :)

nenadfilipovic:
Is this ready?

paolotiu:
It's just waiting for it to be merged.

dan-lee:
I'd be more than happy if it got merged :)

n1ru4l:
Hey there, there is another potential breaking change and I would like to bundle those together into a single release if possible. You could help me tremendously by adding a changeset to this pull request that showcases how this will affect end-users (ideally with a migration path) and furthermore give feedback on https://github.com/dotansimha/graphql-code-generator/pull/6611

Then I will release those two PRs together next week :)

n1ru4l (changes requested):
Please add a changset as I mentioned in my comment

paolotiu:
Hey @n1ru4l! Thanks for looking at this. Although this pr wouldn't necessarily break anything for users (since it's already broken in the first place), I agree that it would be good practice to bundle this with the other change you suggest regarding the processing of the .env variables.

dan-lee:
Thanks for providing the changeset and merge ✌️

PR #6492, opened 2021-08-17
feat(gql-tag-operations): support module augmentation + babel plugin

n1ru4l:
See https://github.com/dotansimha/graphql-code-generator/pull/6492/files#diff-1a39726074b13ad1586a426c290e1c2213e3d691224eff1de2e3ddc92748ebc0

### DIFF
--- a/packages/presets/gql-tag-operations/tests/gql-tag-operations.spec.ts
+++ b/packages/presets/gql-tag-operations/tests/gql-tag-operations.spec.ts
@@ -24,27 +24,27 @@ describe('gql-tag-operations-preset', () => {
     });
 
     expect(result[0].content).toMatchInlineSnapshot(`
-"/* eslint-disable */
-import * as graphql from './graphql';
-import { TypedDocumentNode as DocumentNode } from '@graphql-typed-document-node/core';
+      "/* eslint-disable */
+      import * as graphql from './graphql';
+      import { TypedDocumentNode as DocumentNode } from '@graphql-typed-document-node/core';
 
-const documents = {
-    \\"\\\\n  query A {\\\\n    a\\\\n  }\\\\n\\": graphql.ADocument,
-    \\"\\\\n  query B {\\\\n    b\\\\n  }\\\\n\\": graphql.BDocument,
-    \\"\\\\n  fragment C on Query {\\\\n    c\\\\n  }\\\\n\\": graphql.CFragmentDoc,
-};
+      const documents = {
+          \\"\\\\n  query A {\\\\n    a\\\\n  }\\\\n\\": graphql.ADocument,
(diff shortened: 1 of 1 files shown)


website/docs/presets/gql-tag-operations.md
```diff
@@ -145,3 +145,140 @@ const Tweets = () => {
   );
 };
 ```
+
+## Improving bundle-size with the gql-tag-operations babel plugin
+
+Because the generated code output is a single `gql` function that looks similar to thw following code, code-splitting and tree-shaking is not easily possible.
```
dotansimha, line 151: small type, `*thw`

dotansimha (approved):

website/docs/presets/gql-tag-operations.md
```diff
@@ -145,3 +145,140 @@ const Tweets = () => {
...
+
+declare module '@urql/core' {
+  export function gql(
+    source: '\n  query Bar {\n    Tweets {\n      ...Lel\n    }\n  }\n'
+  ): typeof import('./graphql').BarDocument;
+  export function gql(source: string): unknown;
+}
+```
+
+Now you can simply use your `gql` function imported from `@url/core` in your application code for having fully typed document nodes.
```
renatorib, line 257:
```suggestion
Now you can simply use your `gql` function imported from `@urql/core` in your application code for having fully typed document nodes.
```

website/docs/presets/gql-tag-operations.md
```diff
@@ -145,3 +145,140 @@ const Tweets = () => {
...
+
+```ts
+import { gql } from 'urql';
```
renatorib, line 260: the import shouldn't be from `@urql/core` as well?

n1ru4l: urql re-exports `gql` from `@urql/core`

website/docs/presets/gql-tag-operations.md
```diff
@@ -145,3 +145,140 @@ const Tweets = () => {
...
+const FooQuery = gql(/* GraphQL */ `
+  query Foo {
+    Tweets {
+      id
+    }
+  }
+`);
+```
+
+**NOTE**: In case you are using fragments you MUST use the gql-tag-operations plugin as otherwise the `gql` calls that reference global fragments will cause runtime errors, as the `gql` operation cannot find the global fragment.
```
n1ru4l, line 271:
```suggestion
**NOTE**: In case you are using fragments you MUST use the gql-tag-operations babel plugin as otherwise, the `gql` calls that reference global fragments will cause runtime errors, as the `gql` operation cannot find the global fragment.
```

renatorib (approved):

PR #6497, opened 2021-08-18
[typescript-react-query] Remove undefined variables from query key

### LINKED ISSUE #6496: [typescript-react-query] Query keys are too specific
**Describe the bug**
Query keys are too strict by default, because they always have a second parameter.

```tsx
const { data } = useGetUsersQuery({ role: "User" });
const {mutate: deleteUser} = useDeleteUserMutation({
	onSuccess: () => queryClient.invalidateQueries(useGetUsersQuery.getKey())
});
```

In the above example the getUsers query internally uses a queryKey of `['getUsers', { role: "User" }]`, which is what I would expect.

After deleting a user, I want to invalidate all getUsers queries, regardless of their variables. That means I only care for the first part of the query key, which I would expect to get with `useGetUsersQuery.getKey()`. Running the mutation won’t invalidate the query though.

**Reason**  

If we look at the generated code, we find out why:
```tsx
export const useGetUsersQuery = <
      TData = GetUsersQuery,
      TError = unknown
    >(
      variables?: GetUsersQueryVariables, 
      options?: GetUseQueryOptions<GetUsersQuery, TError, TData>
    ) => 
    useQuery<GetUsersQuery, TError, TData>(
      ['getUsers', variables],
      fetcher<GetUsersQuery, GetUsersQueryVariables>(GetUsersDocument, variables),
      options
    );
useUsersQuery.getKey = (variables?: GetUsersQueryVariables) => ['getUsers', variables];
```

The query key always has a length of `2`, because `variables` is included even if it has not been defined. In our example above we try to invalidate `['getUsers', undefined]` instead of `['getUsers']`, which does not match `['getUsers', { role: "User" }]` and therefore does not invalidate the getUsers query.

hpohlmeyer:
## Description

Related #6496 by modifying `generateQueryKey` to not include `undefined` variables.

## Type of change

This can be a breaking change if you rely on the undefined entry in the query key. I would consider the previous behavior a bug though.

## How Has This Been Tested?

I have tested it in a private project, where I applied the listed changes via patch-package.

## Checklist:

### DIFF
--- a/.changeset/ninety-insects-kick.md
+++ b/.changeset/ninety-insects-kick.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-react-query': major
+---
+
+Queries without variables will no longer have an undefined entry in their query key
--- a/dev-test/githunt/types.react-query.ts
+++ b/dev-test/githunt/types.react-query.ts
@@ -408,7 +408,7 @@ export const useCurrentUserForProfileQuery = <TData = CurrentUserForProfileQuery
   options?: UseQueryOptions<CurrentUserForProfileQuery, TError, TData>
 ) =>
   useQuery<CurrentUserForProfileQuery, TError, TData>(
-    ['CurrentUserForProfile', variables],
+    variables === undefined ? ['CurrentUserForProfile'] : ['CurrentUserForProfile', variables],
     fetcher<CurrentUserForProfileQuery, CurrentUserForProfileQueryVariables>(
       dataSource.endpoint,
       dataSource.fetchParams || {},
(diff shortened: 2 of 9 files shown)


.changeset/ninety-insects-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-codegen/typescript-react-query': major
+---
+
+Queries without variables will no longer have an undefined entry in their query key
```
n1ru4l, line 5: Wouldn't this be a bug fix instead of a major change?

hpohlmeyer: Sorry, I was on vacation last week.

It depends on how you look at it. If you rely on the `undefined` entry, this might invalidate your query after this change, which would make it a breaking change. I would sill say it is a bug, but was not sure if anyone relies on the current behavior. I can change it to bugfix and make it a minor change if you prefer it though.

Issue #6505, opened 2021-08-19
[jsdocs] Schema & directives cause `[object Object]` to be written to the file


phildenhoff:
**Describe the bug**

If your Schema has directives or includes a `schema` header, the JSDocs plugin dumps an `[object Object]` to the output file. It's invalid Javascript, which is problematic.

**To Reproduce**

https://codesandbox.io/s/nifty-euclid-o963o?file=/types.js

You can see that there are two `[object Object]` lines at the top of the output file. If you use `@deprecated`, as below, you will see the directive be applied, but not removed from the top of the file. If you delete the directive, one `[object Object]` will go away.

```diff
type RootQueryType {
-  allTags: [String]
+  allTags: [String] @deprecated(reason: "I dunno why")
}
```

<!--
#############
#############
🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 

PLEASE READ THE FOLLOWING: 
You can use this template to create a live sandbox of the issue: 
https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template
Make sure to fork this template, and run `yarn generate` in the terminal. 
Please make sure the Codegen and plugins version under `package.json` matches yours. 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~ ISSUES WITH A LIVE REPRODUCTION WILL BE ANSWERED QUICKLY ~~~~~~~
~~~~~~ ISSUES WITHOUT REPRODUCTION MIGHT GET CLOSED WITHOUT RESPONSE ~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#############
#############
-->

[...]

PR #6506, opened 2021-08-19
[jsdocs] Handle directive & schema definitions

phildenhoff:
## Description

See my existing issue for what's up. In short: schema and directives cause the JSDocs plugin to output `[object Object]`, which is kind of annoying. Adding a `DirectiveDefinition` and `SchemaDefinition` which returns `null` solves this problem since the expected behaviour is to not write anything out.

This _does not_ affect directives that are used. If a field is marked `@deprecated(reason: ...)` then that reason will be output as per normal. It just avoids trying to write the definition of the directive to the file.

Related #6505 

## Type of change

Code sandbox is available here: https://codesandbox.io/s/nifty-euclid-o963o?file=/schema.graphql

## How Has This Been Tested?

1. Include an unused directive in the schema. Nothing extra is printed out, the file looks like it should.
2. Include a deprecated field, marked with the directive. The field is marked as deprecated in the JSDocs as expected.

**Test Environment**:
- OS: macOS
- `@graphql-codegen/jsdoc`: 2.0.0
- NodeJS: v16.1.0

## Checklist:

### DIFF
--- a/packages/plugins/other/jsdoc/src/index.ts
+++ b/packages/plugins/other/jsdoc/src/index.ts
@@ -130,6 +130,12 @@ export const plugin: PluginFunction<RawDocumentsConfig> = (schema, documents) =>
         return ` - DEPRECATED: ${reason.value.value}`;
       },
     },
+    DirectiveDefinition: {
+      enter() {
+          /** This plugin currently does not support unused Directives. */
+          return null;
+      }
+    },
     FieldDefinition: {
       enter(node) {
         if (node.type.kind === 'NonNullType') {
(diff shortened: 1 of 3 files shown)


packages/plugins/other/jsdoc/src/index.ts
```diff
@@ -181,6 +187,12 @@ export const plugin: PluginFunction<RawDocumentsConfig> = (schema, documents) =>
         return createDocBlock([createDescriptionBlock(node), `@typedef {*} ${node.name}`]);
       },
     },
+    SchemaDefinition: {
```
ardatan, line 190: I think this is extra

ardatan:
Could you take a look at my comments and add a changeset so we can merge and release?
Thanks for the PR :)

packages/plugins/other/jsdoc/src/index.ts
phildenhoff: Ah you're right! I updated the existing definition to return `null` instead; I found that returning an empty string causes a new line to be generated unnecessarily.

Also: not sure if I created a changeset correctly as I've never used it before. Let me know if I need to change anything!

phildenhoff:
@n1ru4l Tests added! I confirmed they fail without my changes — including avoiding any extra lines generated by `schema {...}`

PR #6541, opened 2021-08-23
update installation command

japrozs:
whenever I install `@graphql-codegen/cli`, it always prompts me to install `@graphql-codegen/visitor-plugin-common`, so I thought I'd add it to the installation command.


ardatan:
Thank you for the PR but we have removed that dependency already :)
So it should be fixed with this PR;
https://github.com/dotansimha/graphql-code-generator/pull/6555

japrozs:
Sounds good

PR #6588, opened 2021-08-30
docs(codegen-config): remove extra backtick

thanoskrg:
This is a small PR removing an extra backtick character from the [codegen.yml](https://www.graphql-code-generator.com/docs/getting-started/codegen-config) documentation page.


ardatan:
Thank you @thanoskrg :)

Issue #6624, opened 2021-09-06
Output schema not sorted


DronBrigadir:
**Describe the bug**
I add sort: true in graphql-codegen.yml and output schema not sorted

**To Reproduce**
run graphql-codegen --config ./graphql-codegen.yml and output schema won't be sorted

1. My `codegen.yml` config file:

```yml
schema: https://kp-graphql-api.unstable.kp.yandex.net/graphql/
hooks:
    afterOneFileWrite:
        - prettier --write
generates:
    ./src/index.ts:
        plugins:
            - add:
                  content:
                      - '/* eslint-disable import/newline-after-import */'
                      - '/* eslint-disable camelcase */'
                      - '// @generated'
                      - '// This file was automatically generated and should not be edited.'
            - add:
                  placement: 'append'
                  content: export * from './type-names';
            - typescript
            - fragment-matcher
        config:
            avoidOptionals: true
            apolloClientVersion: 3
            sort: true
```

**Expected behavior**
Output schema will be lexicographic sorted

- OS: macOS Catalina version: 0.15.7
- `@graphql-codegen/typescript`: "2.2.0"

Hrubas:
`yarn upgrade @graphql-codegen/cli@2.1.1` solved the issue for me

ardatan:
Available in https://github.com/dotansimha/graphql-code-generator/pull/6622 !
Thanks for the PR :)

coyoteecd:
@ardatan @DronBrigadir  so how do I disable this new sorting? I upgraded to the most recent release and see a long list of changes when re-generating the types over an existing schema. I tried adding 
```
config:
  sort: false
```
and it doesn't seem to have any effect.

coyoteecd:
Actually nevermind, I think the config must be added as a global option (in the root config node, not next to plugins node). If I put it there, it does disable the sorting, but the order becomes unpredictable for my case (server-side schema generation) whereas in the previous version it was stable. So it seems I have to keep it enabled anyway.

PR #6625, opened 2021-09-06
Bump GraphQL Tools for Schema Sorting and the fix for the schema definition handling

DronBrigadir:
## Description

Use graphql.lexicographicSortSchema in transformSchemaAST depending on the parameter - sort
Bumps GraphQL Tools package that handles schema definitions correctly

Related #6624 

## Type of change
- Bug fix (non-breaking change which fixes an issue)

**Test Environment**:
- OS: macOS Catalina version: 0.15.7
- `@graphql-codegen/typescript`: "2.2.0"

### DIFF
--- a/.changeset/heavy-tables-float.md
+++ b/.changeset/heavy-tables-float.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': patch
+---
+
+fix: handle convertExtensions properly with schema definitions
--- a/.changeset/quick-candles-suffer.md
+++ b/.changeset/quick-candles-suffer.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/cli': minor
+'@graphql-codegen/schema-ast': minor
+---
+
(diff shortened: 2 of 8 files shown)


ardatan:
Thanks!
Could you add some unit tests to prevent future regressions?

Sytten:
I think this might the cause of the changes I saw in my output. Example
```
export type Balance = {
  __typename?: 'Balance';
- payouts: Array<Payout>;
  liabilities: Array<Liability>;
+ payouts: Array<Payout>;
  total: Scalars['Int'];
};
```
I did not enable sorting so I feel this is a regression @ardatan @DronBrigadir

ardatan:
It is not a regression. It is a fix actually. Sorting is enabled by default by codegen but schema wasn't properly sorted so this was causing an inconsistency in output. You can do "sort: false".

Sytten:
Yeah had the discussion on slack, my bad.

PR #6631, opened 2021-09-07
Test if base module is imported in fragments-only files

kamilkisiela:


n1ru4l:
Gonna close this as it was actually caused by https://github.com/dotansimha/graphql-code-generator/pull/6642

PR #6642, opened 2021-09-09
fix: inline Maybe usages instead of referencing the imported one when…

n1ru4l:
I think this one succeeds https://github.com/dotansimha/graphql-code-generator/pull/6631

### DIFF
--- a/.changeset/little-bags-reflect.md
+++ b/.changeset/little-bags-reflect.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-operations': patch
+---
+
+correctly inline Maybe types if the `preresolveTypes` config option is set to `true`
--- a/dev-test/githunt/__generated__/comment-added.subscription.stencil-component.tsx
+++ b/dev-test/githunt/__generated__/comment-added.subscription.stencil-component.tsx
@@ -9,13 +9,16 @@ declare global {
 
   export type OnCommentAddedSubscription = {
     __typename?: 'Subscription';
-    commentAdded?: Types.Maybe<{
-      __typename?: 'Comment';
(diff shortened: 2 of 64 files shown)


kamilkisiela:
Can we merge it? It works in Outreach

dleavitt:
Agree - the `preResolveTypes` setting now _changes the types_ generated by the operations plugin (but not those generated by the typescript plugin). @n1ru4l this seems like a bug and is at minimum a breaking change.

The issue is that the "maybe" types are not pre-resolved if `preResolveTypes` is true. Instead a hardcoded value is used: 

https://github.com/dotansimha/graphql-code-generator/blob/6afbbf67c66787b9d7316689438a6519c26c1b7c/packages/plugins/typescript/operations/src/visitor.ts#L52


### Example

```yaml
schema.ts:
  plugins:
    - "typescript"
    - "typescript-operations"
  config:
    maybeValue: T | undefined
    preResolveTypes: # see below
```

```graphql
type User {
  name: string
}
```

```typescript
// preResolveTypes: false
export type User = {
  name?: string | undefined
}

// preResolveTypes: true
// not compatible with the above!
export type User = {
  name?: string | null | undefined
}
```

### Recommendations

Could either a. revert this change or b. try to pre-resolve the maybe type instead of just using a hardcoded default. Something like:

```typescript
if (config.maybeType) {
  return maybeType.replace("T", type);
} else {
  return `${type} | null | undefined`; 
}
```

LMK if I should make this into a ticket.

Urigo:
could you please open a new ticket for this? It would be easier for us to track and discuss
thank you!

dleavitt:
Yep!

PR #6720, opened 2021-09-22
Add filter, fixes #6709

### LINKED ISSUE #6709: ResolversParentTypes reference non-existing fields of themselves
**Describe the bug**
 
Sometimes it can happen with the `typescript-resolvers` plugin that some fields in `ResolversParentTypes` contain references to other fields in `ResolversParentTypes` that don't exist. This only happens with enums that are mapped using the "mappers" functionality.
E.g. look at this from the reproduction sandbox:
```ts
  Query: {};
  User: Omit<User, 'gender'> & { gender?: Maybe<ResolversParentTypes['Gender']> };
  Boolean: Scalars['Boolean'];
  String: Scalars['String'];
};
```
`ResolversParentTypes['Gender']` doesn't exist.

Note that this **only happens when mappers are used**.

**To Reproduce**
Repro sandbox: https://codesandbox.io/s/crazy-fast-rcgcb?file=/types.ts:3977-4172

<!--
#############
#############
🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 

PLEASE READ THE FOLLOWING: 
You can use this template to create a live sandbox of the issue: 
https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template
Make sure to fork this template, and run `yarn generate` in the terminal. 
Please make sure the Codegen and plugins version under `package.json` matches yours. 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~ ISSUES WITH A LIVE REPRODUCTION WILL BE ANSWERED QUICKLY ~~~~~~~
~~~~~~ ISSUES WITHOUT REPRODUCTION MIGHT GET CLOSED WITHOUT RESPONSE ~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#############
#############
-->

1. My GraphQL schema:

```graphql
extend type Query {
    gender: User!
}

type User {
    gender: Gender
}

enum Gender {
    MALE
    FEMALE
}

```

2. My GraphQL operations:

```graphql
# None
```

```yml
schema: schema.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      mappers:
        Gender: ./gender#OtherGender
```

**Expected behavior**
<!-- A clear and concise description of what you expected to happen. -->
I think there's two things that would make sense:
1. Include enums in `ResolversParentTypes`
2. The syntax where it omits a field from a type and then concatenates it together again (types.ts line 112 in sandbox) makes no sense in this scenario I think.


**Environment:**

- OS: 
- `@graphql-codegen/cli`: "1.20.1",
- `@graphql-codegen/typescript`: "1.21.0",
- `@graphql-codegen/typescript-resolvers`: "^2.2.1",
- NodeJS: 14.17.6

**Additional context**
<!-- Add any other context about the problem here. -->

Once again: Note that this only happens when mappers are used on enums (as far as I know)

Also, it appears to me that this only happens when the types are nested. I tried adding a field to the `Query` type that returns a `Gender` directly and the result was fine.

I have been digging into the code and [this](https://github.com/dotansimha/graphql-code-generator/blob/0b090e31ad668d993ccdb145d2f0edc8518d740a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts#L889-L893) part seems to have something to do with the problem - although I can't find the logic yet concerning the mappers...

puchm:
## Description

I added a filter to remove fields that are not included (files where shouldInclude is false). I think the implementation should be find but I am not sure if this is the right way to solve #6709.

Related #6709 

## Type of change

- [*] Bug fix (non-breaking change which fixes an issue)

This sandbox demonstrates what the issue was and contains a reproduction: https://codesandbox.io/s/crazy-fast-rcgcb?file=/types.ts:3977-4172

## How Has This Been Tested?

I ran `npm test` which failed on some tests. However, these are unrelated and also fail without any changes at the current state of the repo. This may be because I am using npm instead of the normal tools?

The fix works on the case in the reproduction repo.

**Test Environment**:
- OS: Windows 10
- npm 7.13.0
- Node 16.2.0

## Checklist:

- [*] I have followed the [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project (**I have not created a failing test though, maybe someone can tell me how to do that?**)
- [*] I have performed a self-review of my own code
- [*] I have commented my code, particularly in hard-to-understand areas
- [*] I have made corresponding changes to the documentation
- [*] My changes generate no new warnings
- [*] New and existing unit tests pass locally with my changes (**As mentioned, some tests are failing without any changes, no additional failing tests with my changes though**)

## Further comments

As mentioned I think the code is fine but I am not sure if this is the right way to solve the problem in the first place.

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -564,6 +564,16 @@ export class BaseResolversVisitor<
           replaceWithType: string;
         }[] = this._federation
           .filterFieldNames(Object.keys(fields))
+          .filter(fieldName => {
+            const field = fields[fieldName];
+            const baseType = getBaseType(field.type);
+
+            // Filter out fields of types that are not included
+            if (shouldInclude && !shouldInclude(baseType)) {
+              return false;
+            }
+            return true;
+          })
           .map(fieldName => {
             const field = fields[fieldName];
             const baseType = getBaseType(field.type);


ardatan:
Thanks for the PR :)
Could you add some unit tests to prevent future regressions?

puchm:
Sure, will do that later today. What do you think is the reason why some of the tests are failing for me, even without my changes? Is it because I was using npm as a package manager?

ardatan:
@puchm That's the reason I guess. I'd recommend to use `yarn` for this project.

puchm:
I added a unit test. What do you think? Is this the right way to test it?
Also, is my solution to the issue even the right thing to do?

puchm:
What's next for this now? Is there anything else I need to do? And is there a recommended way I can already use this before it gets merged? Normally I can simply do that by installing it from my forked repo but I'm not sure how that works with this big monorepo...

dotansimha:
As far as I can see, if we are using some GraphQL `enums` that shouldn't be used at all, then using `shouldInclude` is the right way to go (since it's there for ignoring specific types). 

This seems like the correction solution because the fields are being collected before generated.

@kamilkisiela @n1ru4l @ardatan can you please review?

kamilkisiela (approved):

Issue #6748, opened 2021-09-28
skip/include directives do not `avoidOptionals` if field type is a list


brainkim:
**Describe the bug**

The `@skip` and `@include` directives seem to be generating optional types for `GraphQLList` fields, even when `avoidOptionals` is true.

**To Reproduce**
https://codesandbox.io/s/practical-swanson-b93wg

1. My GraphQL schema:

```graphql
type Query {
  me: User!
}

type User {
  messages: [Message!]!
}

type Message {
  content: String!
}
```

2. My GraphQL operations:

```graphql
query MyQuery($include: Boolean!) {
  me {
    messages @include(if: $include) {
      content
    }
  }
}
```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-operations
config:
  avoidOptionals: true
  nonOptionalTypename: true
```

**Expected behavior**

I expect TypeScript optionals not to be included in the generated output.

**Actual behavior**
```ts
export type MyQueryQuery = (
  { __typename: 'Query' }
  & { me: (
    { __typename: 'User' }
    & { messages?: Maybe<(
      { __typename: 'Message' }
      & Pick<Message, 'content'>
    )> }
  ) }
);
```

`messages` is optional.

**Environment:**

- OS: MacOS Big Sur (11.6)
- "@graphql-codegen/add": "2.0.2",
- "@graphql-codegen/cli": "1.20.1",
- "@graphql-codegen/typescript": "1.21.0",
- "@graphql-codegen/typescript-operations": "1.17.14",
- "graphql": "15.5.0"

**Additional context**

Probably related to this PR: 
https://github.com/dotansimha/graphql-code-generator/pull/5104

gilgardosh:
Hi @brainkim ,
could you test it with this alphas, and confirm it fixes your issue?
https://github.com/dotansimha/graphql-code-generator/pull/6752#issuecomment-930122239

brainkim:
I tried upgrading typescript-operations to 2.1.5, but am now running into new errors related to the `preResolveTypes` option erasing `Maybe` types in an unexpected way. Looking into it.

Hello from ApolloGraphQL💜

gilgardosh:
Rethinking this case.
If conditional directive is used, the expected outcome must be optional, even if `avoidOptinals` is set.
case include is `false`, `messages` property shouldn't be in the response.
so:

[...]

PR #6752, opened 2021-09-29
avoidOptionals with skip/include directives fix

gilgardosh:
Related #6748

### DIFF
--- a/packages/plugins/typescript/operations/tests/ts-documents.spec.ts
+++ b/packages/plugins/typescript/operations/tests/ts-documents.spec.ts
@@ -5496,7 +5496,6 @@ export type KittyQuery = { __typename?: 'Query', animals: Array<{ __typename?: '
         showName: Scalars['Boolean'];
       }>;
 
-
       export type UserQuery = (
         { __typename?: 'Query' }
         & { user: (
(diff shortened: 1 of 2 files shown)


packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -5560,6 +5559,58 @@ export type KittyQuery = { __typename?: 'Query', animals: Array<{ __typename?: '
...
+
+      expect(content).toBeSimilarStringTo(`
+        export type MyQueryQuery = (
+          { __typename: 'Query' }
+          & { me: (
+            { __typename: 'User' }
+            & { messages: Maybe<Array<(
+              { __typename: 'Message' }
+              & Pick<Message, 'content'>
+            )>> }
```
n1ru4l, line 5609: This should actually generate the following:

```suggestion
            & { messages?: Array<(
              { __typename: 'Message' }
              & Pick<Message, 'content'>
            )> }
```

which is what the actual server response would be

gilgardosh: discussed [here](https://github.com/dotansimha/graphql-code-generator/issues/6748#issuecomment-931123054)

Issue #6757, opened 2021-09-29
`Maybe` is inlined to include `undefined` in `typescript-operations` version 2.1.5


brainkim:
**Describe the bug**
 
When using the `preResolveTypes` option, `Maybe` types now include `undefined` in their union starting in version 2.1.5, regardless of any `avoidOptionals` options.
 
**To Reproduce**
https://codesandbox.io/s/amazing-montalcini-kuf9p?file=/codegen.yml
Code Sandbox isn’t inlining Maybes for some reason right now. Maybe the dependencies aren’t updating correctly.

1. My GraphQL schema:

```graphql
type Query {
    user(id: ID!): User!
}

type User {
    id: ID!
    username: String!
    email: String
}
```

2. My GraphQL operations:

```graphql
# Put your operations here
query user {
    user(id: 1) {
        id
        username
        email
    }
}
```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-operations
    config:
      avoidOptionals: true
      preResolveTypes: true
```

**Expected behavior**

The codesandbox isn’t inlining maybes for some reason.
An example change for an operation. `Maybe` is being inlined as `T | null | undefined` even though `avoidOptionals` is true.
```diff
-export type CypressCreateOrgMutation = { newAccount: Maybe<{ __typename: 'Account', id: string }> };
+export type CypressCreateOrgMutation = { newAccount: { __typename: 'Account', id: string } | null | undefined };
````
**Environment:**

```
    "@graphql-codegen/add": "^2.0.2",
    "@graphql-codegen/cli": "2.2.0",
    "@graphql-codegen/plugin-helpers": "^2.1.1",
    "@graphql-codegen/typescript": "^2.2.2",
    "@graphql-codegen/typescript-operations": "^2.1.5",
    "@graphql-codegen/typescript-resolvers": "^1.19.1",
    "@graphql-codegen/visitor-plugin-common": "^2.2.1",
```

**Additional context**
Probably related to this commit: 
https://github.com/dotansimha/graphql-code-generator/commit/25cd11d01ec55d5c6a2a83ebd501abdd1ed2f248

gilgardosh:
Hey again @brainkim ,
Could you run it with this and see if fixed?

@graphql-codegen/typescript-operations@2.1.6-alpha-a376e6942.0
[...]

PR #6763, opened 2021-09-30
fix avoidOptional handle of Maybe types

gilgardosh:
fix for #6757

### DIFF
--- a/packages/plugins/typescript/operations/tests/ts-documents.spec.ts
+++ b/packages/plugins/typescript/operations/tests/ts-documents.spec.ts
@@ -5560,6 +5560,37 @@ export type KittyQuery = { __typename?: 'Query', animals: Array<{ __typename?: '
         );
       `);
     });
+
+    it('Should handle "preResolveTypes" and "avoiudOptionals" together', async () => {
+      const schema = buildSchema(/* GraphQL */ `
+        type Query {
+          user(id: ID!): User!
+        }
+
+        type User {
+          id: ID!
+          username: String!
+          email: String
+        }
+      `);
+      const operations = parse(/* GraphQL */ `
+        query user {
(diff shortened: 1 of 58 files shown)


packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -5560,6 +5560,37 @@ export type KittyQuery = { __typename?: 'Query', animals: Array<{ __typename?: '
         );
       `);
     });
+
+    it('Should handle "preResolveTypes" and "avoiudOptionals" together', async () => {
```
n1ru4l, line 5564:
```suggestion
    it('Should handle "preResolveTypes" and "avoidOptionals" together', async () => {
```

small typo

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -5560,6 +5560,37 @@ export type KittyQuery = { __typename?: 'Query', animals: Array<{ __typename?: '
...
+          }
+        }
+      `);
+      const config = { avoidOptionals: true, preResolveTypes: true };
+      const { content } = await plugin(schema, [{ location: '', document: operations }], config, {
+        outputFile: 'graphql.ts',
+      });
+
+      expect(content).toBeSimilarStringTo(
+        `export type UserQuery = { __typename?: 'Query', user: { __typename?: 'User', id: string, username: string, email: Maybe<string> } }`
```
n1ru4l, line 5591: as discussed on slack this should be the following instead:

```suggestion
        `export type UserQuery = { __typename?: 'Query', user: { __typename?: 'User', id: string, username: string, email: string | null } }`
```

n1ru4l (approved):

PR #6799, opened 2021-10-07
Allow ResolverTypeWrapper to return null or Promise<null>

TylerBrock:
In the linked blog post for the typescript-resolvers plugin
there is a TS error (TS2322) because the users query could
find no matching user and therefore return undefined:

```
src/resolvers.ts:16:5 - error TS2322: Type '(parent: {}, args: RequireFields<QueryUserArgs, "id">) => { id: string; email: string; profile: { age: number; name: string; }; } | undefined' is not assignable to type 'Resolver<ResolverTypeWrapper<User>, {}, any, RequireFields<QueryUserArgs, "id">> | undefined'.
  Type '(parent: {}, args: RequireFields<QueryUserArgs, "id">) => { id: string; email: string; profile: { age: number; name: string; }; } | undefined' is not assignable to type 'ResolverFn<ResolverTypeWrapper<User>, {}, any, RequireFields<QueryUserArgs, "id">>'.
    Type '{ id: string; email: string; profile: { age: number; name: string; }; } | undefined' is not assignable to type 'ResolverTypeWrapper<User> | Promise<ResolverTypeWrapper<User>>'.
      Type 'undefined' is not assignable to type 'ResolverTypeWrapper<User> | Promise<ResolverTypeWrapper<User>>'.

16     user: (parent, args) => {
       ~~~~

Found 1 error.
```

I think there is a legitimate case where this would fail
and we'd want to return null or potentially Promise<null>
with `Promise<T | null> | T | null`.

Blog post is here: https://the-guild.dev/blog/better-type-safety-for-resolvers-with-graphql-codegen

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -270,7 +270,7 @@ export interface RawResolversConfig extends RawConfig {
    */
   enumValues?: EnumValuesMap;
   /**
-   * @default Promise<T> | T
+   * @default Promise<T> | T | null
    * @description Allow you to override `resolverTypeWrapper` definition.
    */
   resolverTypeWrapperSignature?: string;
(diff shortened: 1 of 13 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -373,7 +373,7 @@ export class BaseResolversVisitor<
       optionalResolveType: getConfigValue(rawConfig.optionalResolveType, false),
       enumPrefix: getConfigValue(rawConfig.enumPrefix, true),
       federation: getConfigValue(rawConfig.federation, false),
-      resolverTypeWrapperSignature: getConfigValue(rawConfig.resolverTypeWrapperSignature, 'Promise<T> | T'),
+      resolverTypeWrapperSignature: getConfigValue(rawConfig.resolverTypeWrapperSignature, 'Promise<T> | T | null | null'),
```
n1ru4l, line 376: Why `| null | null`

n1ru4l:
@TylerBrock It is a bit hard for me to grasp what is wrong here and what you are trying to address. Can you please give a detailed explanation of what is actually wrong and how we might fix it?

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
TylerBrock: that was probably a mistake

PR #6815, opened 2021-10-11
Introduce `inputMaybeValue` configuration for `typescript`-related plugins

giladd123:
I have added the option for InputMaybe, to create separation between input Maybe and output Maybe, by changing all of the Inputs that are of type Maybe to be of type InputMaybe which could be configured differently than Maybe.

### DIFF
--- a/.changeset/thirty-dodos-sell.md
+++ b/.changeset/thirty-dodos-sell.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript': minor
+---
+
+Added InputMaybe, a different type of Maybe type for input/arguments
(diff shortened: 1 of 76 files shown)


saihaj:
> This is also useful for Prisma users - today, `typescript-resolvers` conflicts on `maybeValue` because Prisma returns `| null` from `find`, while not allowing `null` in `where` parameters, so it always needs to be casted into `undefined`.
> 
> With this change, it should be possible to do and it should be simpler to use with Prisma ORM:
> 
> ```
> config:
>   maybeValue: T | null | undefined # affects resolvers signature only 
>   inputMaybeValue: T | undefined # affects resolvers `args`
> ```

yeah I think we need a way to customize `ResolverTypeWrapper`

dotansimha (approved):

dotansimha:
I pushed a few fixes for this PR. Including:
- Rebase to latest `master`
- Instead of doing `replace('Maybe', 'InputMaybe')`, it's now checking if it's an input value correctly, and apply it based on the actual need.
- Fixed issues with `input` types definition - now this applies to inline arguments fields, and also for fields defined in `input` types.
- Updated all tests, docs and examples to match. 

@giladd123 @n1ru4l @kamilkisiela please review again :)

PR #6822, opened 2021-10-12
[bug] fix typename is undefined

CHAOWEICHIU:
[Reference](https://github.com/dotansimha/graphql-code-generator/issues/6821)


n1ru4l:
@CHAOWEICHIU Can you please add tests for this big fix?

charlypoly:
This bug could not be reproduced: https://github.com/dotansimha/graphql-code-generator/issues/6821

Issue #5869, opened 2021-04-19
TypeScript output shouldn't treat interface as a union of implementing types


travigd:
**Is your feature request related to a problem? Please describe.**

The current treatment of interfaces treats an interface as a union of its types. This means it's actually not safe for backwards compatible usage (i.e., the server **should** be able to add new interface implementations in a backwards compatible way, but graphql-code-generator breaks that contract). It's not the end of the world, but I didn't see anything else about this.

**Describe the solution you'd like**

It's not possible to doo something like

```
interface IAnimal { __typename: string; name: string; }
interface Dog extends IAnimal { __typename: "Dog"; favoriteDogTreatType: string; }
interface Cat extends IAnimal { __typename: "Cat"; meowVolume: number; }
```

since the resulting codegen would be something like this:
```
type Animal = IAnimal | Dog | Cat;
``

and this would be broken:
```
const animal: IAnimal = ...;
if (animal.__typename === "Dog") {
  // This is an error since this type guard only narrows to (IAnimal | Dog)
  // since it's technically valid for IAnimal to have __typename "Dog" in
  // TypeScript's world -- even though that's not possible in GraphQL world
  console.log(animal.favoriteDogTreatType);
}
```

One solution to this could be to use a sentinel value:

```ts
type Animal = (IAnimal & __typename: "_interface_sentinel_do_not_check_for_me_or_you_will_be_fired_") | Dog | Cat;
```

**Describe alternatives you've considered**
Status quo isn't the end of the world! :)

**Additional context**

The actual issues this can cause is essentially forgetting to write fallback code which then breaks when new interface implementations are added on the server side.

```ts
// This type-checks in the current version
function greetAnimal(animal: Animal): string {
  switch (animal.__typename) {
  case "Dog":
    return `Hello, puppy!`;
  case "Cat":
    return `Meow meow.`;
  }
}
```

This breaks when the server adds an `Iguana` implementation because there's no default case.

travigd:
Also, as an aside, the current codegen seems to generate every case for every single interface even when they're not given. I'm not sure that this is desirable behavior.

Ostensibly it's meant to handle the case where a field on the interface is actually a more specific type (e.g., `Animal has a field `owner: Owner`, and `Dog` has a field `owner: DogOwner` which is an implementation of the `Owner` type), but it also means that when we do

```
animal {
  name
}
```

the resulting codegen looks like
```ts
[...]

PR #6826, opened 2021-10-12
feat: Add option to squash exactly similar fragment types

asmundg:
## Description

When selecting from an interface, the default code path will declare the
fragment type as a union of all implementing types. This happens
regardless of whether we actually selecting something from the
implementing types.

E.g.
```
type FooFragment =
  | {
    __typename: "foo";
    shared: string;
  }
  | {
    __typename: "bar";
    shared: string;
  }
```

This behavior can be noisy and redundant. In combination with heavy use of
interfaces, especially in nested schemas, this can also potentially
trigger TypeScript's "union type is too complex to represent".

In order to reduce the size of the generated fragments, we introduce the
`mergeFragmentTypes` configuration option. When enabled, this will
collapse all types with the same selection set to a single union member
of the fragment type

E.g.
```
type FooFragment = {
  __typename: "foo" | "bar";
  shared: string;
}
```

Related #5869

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [X] Added unit tests
- [X] Tested in big proprietary repo

**Test Environment**:
- OS: MacOS 11.6
- NodeJS: 16.7.0

## Checklist:

- [X] Any dependent changes have been merged and published in downstream modules


n1ru4l:
@asmundg This is an interesting feature! Just out of curiosity I tried to add a test in order to see how this option behaves combined with the `skipTypename` options and it raised an error:

```ts
    it.only('mergeFragmentTypes works combined with the skipTypename option', async () => {
      const ast = parse(/* GraphQL */ `
        fragment N on Query {
          notifications {
            ... on TextNotification {
              text
            }
          }
        }
      `);
      const config = { preResolveTypes: true, mergeFragmentTypes: true, skipTypename: true };
      const { content } = await plugin(schema, [{ location: 'test-file.ts', document: ast }], config, {
        outputFile: '',
      });

      expect(content).toBeSimilarStringTo(`
        export type NFragment = { __typename?: 'Query', notifications: Array<{ __typename?: 'TextNotification', text: string } | { __typename?: 'ImageNotification' }> };
      `);
      await validate(content, config);
    });
```

```
● TypeScript Operations Plugin › Selection Set › mergeFragmentTypes works combined with the skipTypename option
  TypeError: Cannot read property 'type' of null
    370 |       // string literal union of typenames.
    371 |       const compacted = Object.keys(grouped).reduce<Record<string, string[]>>((acc, key) => {
  > 372 |         const typeNames = grouped[key].types.map(t => t.type);
        |                                                         ^
    373 |         const typenameUnion = this._processor.transformTypenameField(typeNames.join(' | '), grouped[key].type[0].name);
    374 |         const transformedSet = this.selectionSetStringFromFields([...typenameUnion, ...grouped[key].fields]);
    375 |
    at map (packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts:372:57)
        at Array.map (<anonymous>)
    at reduce (packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts:372:46)
        at Array.reduce (<anonymous>)
    at SelectionSetToObject._buildGroupedSelections (packages/plugins/other/visitor-plugin-common/srcselection-set-to-object.ts:371:46)
    at SelectionSetToObject.transformSelectionSet (packages/plugins/other/visitor-plugin-common/srcselection-set-to-object.ts:587:50)
    at SelectionSetToObject.buildSelectionSet (packages/plugins/other/visitor-plugin-common/src/selection-set-to-objectts:509:24)
    at reduce (packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts:358:43)
        at Array.reduce (<anonymous>)
    at SelectionSetToObject._buildGroupedSelections (packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts:346:37)
```

This should definitely not happen. Any ideas how we can resolve this?

asmundg:
Thanks @n1ru4l, I'll have a look.

asmundg:
All right, it should work nicely with skipTypenames now.

asmundg:
@charlypoly, @n1ru4l how do we feel about this change? We're currently using patch-package to apply it to an internal project where vanilla graphql-code-generator generates types that are too complex for TypeScript to handle (the schema is a bit convoluted).

charlypoly:
> @charlypoly, @n1ru4l how do we feel about this change? We're currently using patch-package to apply it to an internal project where vanilla graphql-code-generator generates types that are too complex for TypeScript to handle (the schema is a bit convoluted).

Hi @asmundg,

Your PR is in my TODO list for months, this part of the code is beyond my knowledge but I'll make a double effort this week to do a final review!

Thank you for your patience!

charlypoly (changes requested):
@asmundg 

I found a regression with a new feature introduced in the meantime: `inlineFragmentTypes: "mask"` ([Fragment masking](https://www.graphql-code-generator.com/plugins/gql-tag-operations-preset#fragment-masking)). (cc @n1ru4l 👀)

When both are configured (`mergeFragmentTypes: false, inlineFragmentTypes: 'mask'`), the generated code is missing the fragment types definition:


**output with `mergeFragmentTypes: false, inlineFragmentTypes: 'mask'`**

```
export type TestQueryVariables = Exact<{ [key: string]: never; }>;
    
    
    export type TestQuery = { __typename?: 'Query', notifications: Array<(
        { __typename?: 'TextNotification' }
        & { ' $fragmentRefs': { 'N_TextNotification_Fragment': N_TextNotification_Fragment } }
      ) | (
        { __typename?: 'ImageNotification' }
        & { ' $fragmentRefs': { 'N_ImageNotification_Fragment': N_ImageNotification_Fragment } }
      )> };
    
    type N_TextNotification_Fragment = { __typename?: 'TextNotification', id: string };
    
    type N_ImageNotification_Fragment = { __typename?: 'ImageNotification', id: string };
    
    export type NFragment = N_TextNotification_Fragment | N_ImageNotification_Fragment;
```

**output with `mergeFragmentTypes: true, inlineFragmentTypes: 'mask'`**

```
   export type TestQueryVariables = Exact<{ [key: string]: never; }>;
    
    
    export type TestQuery = (
      { notifications: Array<(
        { ' $fragmentRefs': { 'N_TextNotification_Fragment': N_TextNotification_Fragment } }
        & { __typename?: 'TextNotification' }
      ) | (
        { ' $fragmentRefs': { 'N_ImageNotification_Fragment': N_ImageNotification_Fragment } }
        & { __typename?: 'ImageNotification' }
      )> }
      & { __typename?: 'Query' }
    );
    
    export type NFragment = (
      { id: string }
      & { __typename?: 'TextNotification' | 'ImageNotification' }
    ) & { ' $fragmentName': 'NFragment' };
```

The issue comes from `SelectionSetToObject#transformFragmentSelectionSetToTypes()` (called from `BaseDocumentsVisitor#FragmentDefinition()`) that does not get the proper `grouped` fragments types:


**`grouped` value `mergeFragmentTypes: false, inlineFragmentTypes: 'mask'`**

```
{
      TextNotification: [ "{ __typename?: 'TextNotification', id: string }" ],
      ImageNotification: [ "{ __typename?: 'ImageNotification', id: string }" ]
    }
```

**`grouped` value with `mergeFragmentTypes: true, inlineFragmentTypes: 'mask'`**

```
{
      "'TextNotification'_'ImageNotification'": [
        '(\n' +
          '  { id: string }\n' +
          "  & { __typename?: 'TextNotification' | 'ImageNotification' }\n" +
          ')'
      ]
    }
```



I have a failing test locally, if you are interested I can push it to your branch.

asmundg:
@charlypoly thanks, I'll take a look.

charlypoly:
> @charlypoly thanks, I'll take a look. Please feel free to push a test.

@asmundg Actually, I'll need your help to rebase/sync your branch with the updated `master` 👀

charlypoly:
@asmundg Also, after thinking about it, I actually think that it does not make sense to support `mergeFragmentTypes: true` when `inlineFragmentTypes: 'mask'` is set.

Let me know what you think, but if you agree then the fix would be straightforward and would just require a bit of added documentation about the compatibility.

asmundg:
@charlypoly interesting stuff. While I think these should be possible to get working at the same time, the current mechanism for collapsing similar types doesn't play well with the need for having a fragment reference as part of the type. We would need to check whether we can collapse the type _before_ we added the fragment reference, make the fragment reference dependent on whether the types were merged.

I'd be happy to punt a bit on making that work though.

charlypoly (approved):

charlypoly:
Thank you for this great contribution @asmundg! 🚀

simPod:
Did anyone manage to make typeguards work? I used

```ts
export const isType = <T extends string>(
  typename: T,
) => <N extends { __typename: string }>(node: N): node is Extract<N, { __typename: T }> => node.__typename === typename;
```

that was used as `isType('TypeName')(value)` which doesn't work for unions now.

asmundg:
@simPod yes, `Extract` doesn't work there, as `N.__typename` can still be a union, which clearly cannot be assigned to `T`. I didn't figure out a pretty way around this for cases when I needed an unambiguous type, so I settled for a bit of assertion along with normal narrowing.

I haven't looked very closely at this use case though, so I might have missed something.

```typescript
  type ExactFragment<
    T extends { __typename: string },
    K extends string
  > = T & {
    __typename: K;
  };

  if (value.__typename === "TypeNameA") {
    // v now has the shared fragment properties and __typename = "TypeNameA"
    const v = value as ExactFragment<
      typeof value,
      typeof value.__typename
    >;
  }
```

simPod:
I see. I guess I'll open new issue so it's easier to find it and collaborate ideas. This option is very nice but had to revert it since I don't want to use `as`.

Issue #6724, opened 2021-09-23
[TypeScript React-Query] exposeFetcher for mutation and exposeMutationKeys


judastheo:
**Is your feature request related to a problem? Please describe.**

We are migrating more and more of our API calls to the queries and mutations generated. And sometimes, we want to replace old mutation calls by new ones that are used outside of functional component, hence we cannot use hooks.

**Describe the solution you'd like**

For queries, you already enabled `exposeFetcher` and `exposeQueryKeys` that matches our needs! It would be great to also have  `getKey` and `fetcher` for mutations.

The config of `typescript-react-query` could have two more boolean options:
- `exposeMutationKeys`
- `exposeMutationFetcher`

The generated code could look exactly the same as queries generated with `exposeFetcher` and `exposeQueryKeys`:

```tsx
useMyMutation.getKey = (
  variables: MyMutationVariables
) => ['my', variables];

useMyMutation.fetcher = (
  variables: MyMutationVariables
) =>
  fetch<MyMutation, MyMutationVariables>(
    MyDocument,
    variables
  );
```

**Failing test proposal**

Not very used to it, but here is what I think could be added inside [plugins/typescript/react-query/tests/react-query.spec.ts](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/react-query/tests/react-query.spec.ts)

```ts
    it("Should generate fetcher field when exposeMutationFetcher is true and the fetcher isn't a react hook", async () => {
      const config = {
        fetcher: {
          func: './my-file#customFetcher',
        },
        exposeMutationFetcher: true,
      };

      const out = (await plugin(schema, docs, config)) as Types.ComplexPluginOutput;
      expect(out.content).toBeSimilarStringTo(
        `useTestMutation.fetcher = (variables?: TTestMutationVariables) => customFetcher<TTestMutation, TTestMutationVariables>(TestDocument, variables);`
      );
    });

    it('Should NOT generate fetcher field when exposeMutationFetcher is true and the fetcher IS a react hook', async () => {
      const config = {
        fetcher: {
          func: './my-file#useCustomFetcher',
          isReactHook: true,
        },
        exposeMutationFetcher: true,
      };

      const out = (await plugin(schema, docs, config)) as Types.ComplexPluginOutput;
      expect(out.content).not.toBeSimilarStringTo(`useTestMutation.fetcher`);
    });

  describe('exposeMutationKeys: true', () => {
[...]

PR #6863, opened 2021-10-19
feat(react-query): Adding 'exposeMutationKeys' config option to react…

alex-misch:
react-query plugin

## Description

We need to use keys and fetcher of generated mutations. Also, right now `useMutation` is generated without key and every call of `useMutation` creates new subscriber. This PR adds key for each mutation that provide their uniqually so we can subscribe to mutation in case of passed variables.

Related #6724 

## Type of change

## How Has This Been Tested?

The tests is similar to useQuery `exposeFetcher` and `exposeQueryKeys` configuration

- [x] Should generate mutation fetcher field when exposeFetcher is true and the fetcher isn't a react hook
- [x] Should NOT generate mutation fetcher field when exposeFetcher is true and the fetcher IS a react hook
- [x] Should generate getKey for each mutation if exposeMutationKeys is true

**Test Environment**:
- OS: macOS Big Sur 11.2.3 (20D91)
- NodeJS: v12.22.6

## Checklist:

### DIFF
--- a/dev-test/githunt/types.react-query.ts
+++ b/dev-test/githunt/types.react-query.ts
@@ -534,6 +534,7 @@ export const useSubmitRepositoryMutation = <TError = unknown, TContext = unknown
   options?: UseMutationOptions<SubmitRepositoryMutation, TError, SubmitRepositoryMutationVariables, TContext>
 ) =>
   useMutation<SubmitRepositoryMutation, TError, SubmitRepositoryMutationVariables, TContext>(
+    'submitRepository',
     (variables?: SubmitRepositoryMutationVariables) =>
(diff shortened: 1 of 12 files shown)


alex-misch:
Hi! Sure, I've just updated the PR. And fixes little bug: useMutation's key can't be based on variables because it returns `mutate(variables)` function and we don't know about variables in hook. So it must provide string key only. The implementation on [this kind of ts overload](https://github.com/tannerlinsley/react-query/blob/19861272f57da5e587bce5bff449d0c3020243e0/src/react/useMutation.ts#L42) in react-query

alex-misch:
@dotansimha I'm finished, can you please approve workflows runs?

n1ru4l (approved):
Legit! I rebased and will merge it now. Sorry for the delay!

Issue #5212, opened 2020-12-07
React Query infinite pagination 


galkatz373:
I would love to use the react query plugin, but unfortunately it doesn't have a support for infinite scrolling and for me itvs really important, I would really appreciate it if you add it.

lynxtaa:
Since my PR (https://github.com/dotansimha/graphql-code-generator/pull/5863) is merged and released in `typescript-react-query@1.3.0` now there is an option `exposeDocument` for exposing document from the hook. So I've written a custom hook for infinite queries:

```ts
interface UseQueryFn<
  TData extends Record<string, any>,
  TVariables extends Record<string, any>
> {
  (variables: TVariables, options?: UseQueryOptions<TData>): unknown
  document: string
  getKey: (variables: TVariables) => unknown[]
}

export function useInfiniteGraphQLQuery<
  TData extends Record<string, any>,
  TVariables extends Record<string, any>
>(
  useQuery: UseQueryFn<TData, TVariables>,
  getVariables: ({ pageParam }: { pageParam?: number }) => TVariables,
  options?: UseInfiniteQueryOptions<TData, Error>,
): UseInfiniteQueryResult<TData, Error> {
  return useInfiniteQuery<TData, Error>(
    useQuery.getKey(getVariables({})),
    ({ pageParam }) => gqlRequest(useQuery.document, getVariables({ pageParam })),
    options,
  )
}
```

And use it like this:
```ts
// All types, key and document are infered from the passed hook
const { fetchNextPage, hasNextPage, data, isLoading } = useInfiniteGraphQLQuery(
  useGetFolderDocumentsQuery, // this is a hook generated by typescript-react-query
  ({ pageParam = 0 }) => ({ // return variables
    folderId,
    limit: 100,
    offset: 100 * pageParam,
   }),
  // getNextPageParam and other options
)
```

In codegen config I'm setting  `exposeQueryKeys: true` and `exposeDocument: true`. I hope you find it helpful

Urigo:
anyone who tracks this issue, do you think @lynxtaa solution (thank again for the PR!) is enough, or something else is needs?
(that might include just a receipt or a mention in the docs)
Thanks!

lynxtaa:
I think the best solution is to add an option `useInfiniteQuery` which takes a list of operations names similarly to `useSRWInfinite` option in `plugin-typescript-swr`:  https://www.npmjs.com/package/graphql-codegen-plugin-typescript-swr#useswrinfinite

[...]

PR #6961, opened 2021-11-03
Add useInfiniteQuery to react-query plugin

jrwpatterson:

## Description

Related # (https://github.com/dotansimha/graphql-code-generator/issues/5212)

## Type of change

## How Has This Been Tested?

- I have added the tests to the branch. 
- I have also run the built results for several weeks in our production system as a yarn patch to ensure the API both generates the correct code and that the code generated isn't buggy. 

**Test Environment**:
- OS: linux/macos
- "@graphql-codegen/typescript-operations": "2.2.0"
- "@graphql-codegen/cli": "2.2.2"
- "@graphql-codegen/plugin-helpers": "^2.3.0"
- "@graphql-codegen/typescript": "2.3.0"
- "@graphql-codegen/typescript-document-nodes": "2.2.0"
- "@graphql-codegen/typescript-graphql-files-modules": "2.1.0"
- "@graphql-codegen/visitor-plugin-common": "2.5.0"
- NodeJS: 14

## Checklist:

## Further comments

I originally passed the whole options object down with the intention of allowing the fetcher to work it out. 

By passing the key down this gives the dev the ability to change the variable that he wants the pageParam to update easily as you can't change the fetcher like you can with a standard infinitequery, but also nothing in the options was useful for the fetcher anyway. 

the pageparam key is typesafe off the variables as to my knowledge you need one of the variables to update. 

This allows the fetcher to be universal as it injects the pageparam for all of them and also allows the fetcher to continue to be universal between mutation, query, and infinitequery


n1ru4l:
Hey @jrwpatterson, I just mered some older PRs that affected the same files, can you please rebase this PR?

Subalee:
Future idea: Skip the `addInfiniteQuery` option, which would solve the issue of generating infinite queries where they are not needed.

The infinite query would be based on whether the schema contains a Connection Type based on [Connection Specification](https://graphql.org/learn/pagination/).

But this is just throwing it out there as a very opinionated idea.

jrwpatterson:
@n1ru4l done! 

@Subalee interesting was wondering how that would work...

cabljac:
> should I be looking at the query in the graphql or is there something obvious I missed transversing the types when I tried to find it?

 You could maybe look for the pageInfo field, that should indicate it's a connection?

Subalee:
The easiest thing would to look at the  `__typename` of the Object type returned by the query which should end in `Connection` as that's what a paginated object should end in as per the recommended specification by GQL.

However that seems a little "brittle" as it  could cause issues with naming in some projects so I'd use this AND/OR 
the structure of such query which according to the spec should look something like this

```gql
{
  cats(first: $first, after: $after) {
    edges: {  # contains an array of Cat nodes
      node: {}  # Cat 
      cursor: {} # cursor of the cat
    }
   pageInfo: {} # meta data such as hasNextPage, endCursor of the currently retrieved edges (the last cat in the array)
  }
}
```

- where `cats` is a `__typename: 'CatConnection'`
- where `edges` is a `__typename: 'CatEdge[]'`
- where `node` is a `__typename: 'Cat'`
- where pageInfo is a `__typename: 'PageInfo'`

So i'd suggest at looking at several of these attributes to determine whether the query should be infinite or not, although I'd still keep generating the standard query.

btw, I really appreciate your effor on this, I probably wouldn't be able to dive that deep into this library.

jrwpatterson:
> @jrwpatterson The easiest thing would to look at the `__typename` of the Object type returned by the query which should end in `Connection` as that's what a paginated object should end in as per the recommended specification by GQL.
> 
> However that seems a little "brittle" as it could cause issues with naming in some projects so I'd use this AND/OR the structure of such query which according to the spec should look something like this (basically what @cabljac suggested)
> 
> ```graphql
> {
>   cats(first: $first, after: $after) {
>     edges: {  # contains an array of Cat nodes
>       node: {}  # Cat 
>       cursor: {} # cursor of the cat
>     }
>    pageInfo: {} # meta data such as hasNextPage, endCursor of the currently retrieved edges (the last cat in the array)
>   }
> }
> ```
> 
> * where `cats` is a `__typename: 'CatConnection'`
> * where `edges` is a `__typename: 'CatEdge[]'`
> * where `node` is a `__typename: 'Cat'`
> * where pageInfo is a `__typename: 'PageInfo'`
> 
> So i'd suggest at looking at several of these attributes to determine whether the query should be infinite or not, although I'd still keep generating the standard query.
> 
> btw, I really appreciate your effor on this, I probably wouldn't be able to dive that deep into this library.

that's awesome the only question would be if for example we are using it with a range and start point 

eg if I want 41-50 I as for {range: 10, starting: 41}

Let me have a look at this, If we get the first pr in as an initial step then I'll see what I can think of for this range of different ways! :)

cabljac:
@jrwpatterson 

Out of interest, what happens in this feature when you have an operation with two fields to be paginated?

e.g if i had something like:
```
query Query($catsRange: Int, $catsStarting: Int, $dogsRange: Int, $dogsStarting: Int) {
  cats(range: $catsRange, starting: $catsStarting) {
      ...
      }
  dogs(range: $dogsRange, starting: $dogsStarting) {
    ...
    }
}
```

P.S +1 to appreciation on this feature, it sounds really cool

jrwpatterson:
> @jrwpatterson
> 
> Out of interest, what happens in this feature when you have an operation with multiple fields to be paginated?
> 
> e.g if i had something like:
> 
> ```
> query Query($catsRange: Int, $catsStarting: Int, $dogsRange: Int, $dogsStarting: Int) {
>   cats(range: $catsRange, starting: $catsStarting) {
>       ...
>       }
>   dogs(range: $dogsRange, starting: $dogsStarting) {
>     ...
>     }
> }
> ```
> 
> P.S +1 to appreciation on this feature, it sounds really cool

right now you might need to have them as two separate queries... its a limitation, I know but I don't even know if its possible in react-query... however the way that both graph and react-query work together, if you set up two infinite queries then while the data would be called twice they would be called concurrently so from a UX perspective you would be fine there might be a slight performance lag in the server depending on the way the queries are setup. But as with all these things I think its one step at a time, simple, complex, hard! eventually we could always bring in something like an object on there :)

jrwpatterson:
@cabljac I've updated the pr so now it you return an object in the function so for your script here is what you need 


```graphql
query AnimalsQuery($catsRange: Int, $catsStarting: Int, $dogsRange: Int, $dogsStarting: Int) {
  cats(range: $catsRange, starting: $catsStarting) {
      ...
      }
  dogs(range: $dogsRange, starting: $dogsStarting) {
    ...
    }
}
```

```tsx
import { useInfiniteMyQuery } from './generated';

export const MyComponent = () => {
  const { status, data, error, isFetching } = useInfiniteAnimalsQuery(
    {
      catsRange: 5,
      catsStarting: 0,
      dogsRange: 10,
      dogsStarting: 0,
    },
    {
      getNextPageParam: (lastPage, allPages) => {
        const totalLocal = (allPages.length ?? 0) * (queryParams.limit ?? 1);
        const totalDogs = lastPage.dogs.items?.length ?? 0;
        if (totalLocal < totalDogs) {
          return {
            catsStarting: totalLocal * 5,
            dogsStarting: totalLocal * 10,
          };
        }
      },
    }
  );
};
```

cabljac:
@jrwpatterson  looks good!

cabljac:
@jrwpatterson is there anything special i need to do to get your fork running locally? Keen to see it in action

jrwpatterson:
@cabljac nothing really I have a patch I've been using if you are using yarn berry in your project... you can use the following patch [typescript-react-query.patch.txt](https://github.com/dotansimha/graphql-code-generator/files/7541976/typescript-react-query.patch.txt) remove the `.txt` at the end (if you want) 

I have put that in a folder called `patches` at base level and then in my packages I have `"@graphql-codegen/typescript-react-query": "patch:@graphql-codegen/typescript-react-query@3.1.1#patches/typescript-react-query.patch",`

if you are on yarn 1 or npm I think you can use https://www.npmjs.com/package/patch-package which will do the same thing in which case you need to download the fork, and at base level run `yarn` then `yarn build` navigate to `packages/plugins/typescript/react-query/dist/index.js` and in your project patch `@graphql-codegen/typescript-react-query` and overwrite the index.js file with the contents from there... once done you should have it and all the other users if you wanted it will have it. just remember to add 

```
    config:
      addInfiniteQuery: true
```
Which is in the doco's then run the codegen and you should be golden. Once this is in I'll look at @Subalee's suggestion to make it so it limits it to types that return a list with options rather than every query so its a little cleaner. (will probably make it so that addInfiniteQuery will accept true for all queries or something other word for the queries it thinks it needs. :)

dotansimha (approved):

dotansimha:
Thanks!! Sorry for the delay on reviewing this.

cabljac:
awesome!!

Issue #6993, opened 2021-11-07
Documentation for default of wrapFieldDefinitions is incorrect


dchenk:
**Describe the bug**

On https://www.graphql-code-generator.com/docs/plugins/typescript-resolvers the `wrapFieldDefinitions` option is documented a `true` by default but it's actually `false`.

dotansimha:
Merged, thanks!

PR #6994, opened 2021-11-07
Fix default of wrapFieldDefinitions

dchenk:
See issue: https://github.com/dotansimha/graphql-code-generator/issues/6993


dotansimha:
Thanks @dchenk !

PR #7004, opened 2021-11-09
fix(typescript-resolvers): Fix optional resolver types

noahseger:
## Description

Fix https://github.com/dotansimha/graphql-code-generator/issues/7005.

Possibly broken in #5112.

## Type of change

Failing test diff:

```diff
-   users?: Resolver<Array<ResolversTypes['User']>, ParentType, ContextType, RequireFields<QueryUsersArgs, 'filter'>>;
-   ping?: Resolver<ResolversTypes['String'], ParentType, ContextType>;
+   users: Resolver<Array<ResolversTypes['User']>, ParentType, ContextType, RequireFields<QueryUsersArgs, 'filter'>>;
+   ping: Resolver<ResolversTypes['String'], ParentType, ContextType>;
+ };
```

## How Has This Been Tested?

- [x] `yarn test packages/plugins/typescript/resolvers/tests/ts-resolvers.spec.ts`

**Test Environment**:
- OS: 
- `@graphql-codegen/...`: 
- NodeJS: 

## Checklist:

## Further comments

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -987,7 +987,7 @@ export class BaseResolversVisitor<
 
       const resolverType = isSubscriptionType ? 'SubscriptionResolver' : directiveMappings[0] ?? 'Resolver';
 
-      const avoidOptionals = this.config.avoidOptionals?.object ?? this.config.avoidOptionals === true;
+      const avoidOptionals = this.config.avoidOptionals?.resolvers ?? this.config.avoidOptionals === true;
       const signature: {
         name: string;
         modifier: string;
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -987,6 +987,7 @@ export class BaseResolversVisitor<
 
       const resolverType = isSubscriptionType ? 'SubscriptionResolver' : directiveMappings[0] ?? 'Resolver';
 
+      const avoidOptionals = this.config.avoidOptionals?.object ?? this.config.avoidOptionals === true;
       const signature: {
         name: string;
         modifier: string;
@@ -994,7 +995,7 @@ export class BaseResolversVisitor<
         genericTypes: string[];
       } = {
         name: node.name as any,
-        modifier: this.config.avoidOptionals ? '' : '?',
+        modifier: avoidOptionals ? '' : '?',
         type: resolverType,
         genericTypes: [
           mappedTypeKey,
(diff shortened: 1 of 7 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -972,14 +972,15 @@ export class BaseResolversVisitor<
 
       const resolverType = isSubscriptionType ? 'SubscriptionResolver' : directiveMappings[0] ?? 'Resolver';
 
+      const avoidOptionals = this.config.avoidOptionals?.object ?? this.config.avoidOptionals === true;
```
noahseger, line 975: Something is wrong with the types here, because it thinks it can only be `AvoidOptionalsConfig` but it can apparently also still be a `boolean` per `ts-resolvers.spec.ts`.

noahseger:
@n1ru4l sorry to bug you — do you have any concerns about this fix that I can address?

n1ru4l:
I am actually not sure whether optional resolver functions should be affected by `avoidOptionals.object`. Maybe instead we should add either a `avoidOptionals.resolvers` config object property or a complete configuration option SPECIFIC to typescript-resolvers with the name `optionalResolvers` (that only supports `true` or `false`).

What do you think @gilgardosh and @dotansimha Since we talked about removing the `avoidOptionals` completely in the next major release, we should not introduce new configuration options for it?

`avoidOptionals` (in my understanding) was originally only meant for the `typescript` plugin.

gilgardosh:
I agree with @n1ru4l , IMO adding a `resolvers` property would be the best solution.

@noahseger could you update this PR accordingly?
note that [this](https://github.com/dotansimha/graphql-code-generator/blob/49242c20cd985addba5f74442d435a527bdbe163/packages/plugins/other/visitor-plugin-common/src/avoid-optionals.ts) should also be updated, and `normalizeAvoidOptionals` can be used to convert boolean AvoidOptionals

noahseger:
@n1ru4l I totally understand, thanks for getting me some feedback!

@n1ru4l @gilgardosh it's a great suggestion, I think it's the best way to avoid having to follow up for more broken behavior in case anybody is now expecting the `avoidOptionals` to behave the same way as the latest release.

My only concern is, since the individual options don't behave as expected, should one or more of them be deprecated? I'd like to document them all more clearly, but I'm not sure who is able to describe the behavior of each in detail.

No problem adding the new option though, I'll get that ready before the weekend.

noahseger:
@n1ru4l @gilgardosh done and open to any other suggests 🙏

noahseger:
If there are no other suggestions, do you think we can get this into a release? Really looking forward to upgrading our tools to the latest and greatest :D

charlypoly (approved):

charlypoly:
Hi @noahseger,

Sorry for the delay and thank you very much for your contribution! 🚀 

It will be soon released, I'll keep you updated.

noahseger:
Thank you very much :D

PR #7015, opened 2021-11-12
loosen resolvers subscribe return type

n1ru4l:


Urigo:
maybe worth adding a test case with the IxJS types?

dotansimha:
> maybe worth adding a test case with the IxJS types?

It can also be with dev-test examples

dotansimha (approved):

Cellule:
This change seems incompatible with https://github.com/apollographql/graphql-subscriptions which uses `AsyncIterator` not `AsyncIterable` and the 2 types are not compatible.
I'm wary of changing any runtime code since what we have currently is working in apollo

dotansimha:
> 

`graphql-subscriptions` considered legacy implementation for subscriptions. It doesn't have any added value with today's tools. 
@n1ru4l from what I understand, the `AsyncIterable` is the correct type here, so I guess `graphql-subscriptions` is just using the incorrect type?

n1ru4l:
graphql-subscriptions is using an incorrect type. I created a PR for fixing it a long time ago. No ody from apollo seems to care. We shouldnt generate wrong typings because of some unmaintained library. AsynIterable (which has a Symbol.asyncIterator property) is the correct return type here. You can use patch-package or `as any` if you use graphql-subscriptions.

____

relevant links

https://github.com/apollographql/graphql-subscriptions/pull/232
https://github.com/apollographql/graphql-subscriptions/issues/240

Cellule:
Thanks for the explanation. Since we're using Apollo, I don't want to diverge from their implementation.
I did use patch-package to change it back in our codebase so all good on my end, just wanted to understand the reasoning behind the change :D

n1ru4l:
@Cellule You can also patch graphql-subscriptions, this is what I am doing in a project where we are using redis:

```diff
diff --git a/node_modules/graphql-subscriptions/dist/pubsub-engine.d.ts b/node_modules/graphql-subscriptions/dist/pubsub-engine.d.ts
index 58fa1b1..c8168e5 100644
--- a/node_modules/graphql-subscriptions/dist/pubsub-engine.d.ts
+++ b/node_modules/graphql-subscriptions/dist/pubsub-engine.d.ts
@@ -2,5 +2,5 @@ export declare abstract class PubSubEngine {
     abstract publish(triggerName: string, payload: any): Promise<void>;
     abstract subscribe(triggerName: string, onMessage: Function, options: Object): Promise<number>;
     abstract unsubscribe(subId: number): any;
-    asyncIterator<T>(triggers: string | string[]): AsyncIterator<T>;
+    asyncIterator<T>(triggers: string | string[]): AsyncIterable <T>;
 }

```

```diff
diff --git a/node_modules/graphql-redis-subscriptions/dist/redis-pubsub.d.ts b/node_modules/graphql-redis-subscriptions/dist/redis-pubsub.d.ts
index af38928..c4784f3 100644
--- a/node_modules/graphql-redis-subscriptions/dist/redis-pubsub.d.ts
+++ b/node_modules/graphql-redis-subscriptions/dist/redis-pubsub.d.ts
@@ -20,7 +20,7 @@ export declare class RedisPubSub implements PubSubEngine {
     publish<T>(trigger: string, payload: T): Promise<void>;
     subscribe<T = any>(trigger: string, onMessage: OnMessage<T>, options?: unknown): Promise<number>;
     unsubscribe(subId: number): void;
-    asyncIterator<T>(triggers: string | string[], options?: unknown): AsyncIterator<T>;
+    asyncIterator<T>(triggers: string | string[], options?: unknown): AsyncIterable<T>;
     getSubscriber(): RedisClient;
     getPublisher(): RedisClient;
     close(): Promise<Ok[]>;

```

In other projects where I only have a single server instance I use the following PubSub instead:

```ts
// Node.js events module
import { EventEmitter, on } from "events";

export const createChannelPubSub = <
  TPubSubPublishArgsByKey extends PubSubPublishArgsByKey
>() => {
  const emitter = new EventEmitter();

  return {
    publish: <TKey extends Extract<keyof TPubSubPublishArgsByKey, string>>(
      routingKey: TKey,
      ...args: TPubSubPublishArgsByKey[TKey]
    ) => {
      if (args[1] !== undefined) {
        emitter.emit(`${routingKey}:${args[0] as number}`, args[1]);
      }

      emitter.emit(routingKey, args[0]);
    },
    subscribe: async function* <
      TKey extends Extract<keyof TPubSubPublishArgsByKey, string>
    >(
      ...[routingKey, id]: TPubSubPublishArgsByKey[TKey][1] extends undefined
        ? [TKey]
        : [TKey, TPubSubPublishArgsByKey[TKey][0]]
    ): AsyncGenerator<
      TPubSubPublishArgsByKey[TKey][1] extends undefined
        ? TPubSubPublishArgsByKey[TKey][0]
        : TPubSubPublishArgsByKey[TKey][1]
    > {
      const asyncIterator = on(
        emitter,
        id === undefined ? routingKey : `${routingKey}:${id as number}`
      );
      for await (const [value] of asyncIterator) {
        yield value as any;
      }
    },
  };
};
```

This is pretty cool as it allows typing the event payloads.

```ts
const pubSub = createChannelPubSub<{
  eventName: { payload: string }
}>()

pubSub.publish("eventName", {payload: "ayy" })
```

Cellule:
Thanks for tip! 
We've already greatly improved typings of our subscriptions thanks to graphql-code-generator (I couldn't work with graphql without you guys ♥)
The payload is one the last portion using manual typings so I'll definitely look into improving this in the future

PR #7017, opened 2021-11-12
export generateSearchPlaces

capaj:
sorry about deleting the template, but it seemed superflous for this trivial change


dotansimha:
Thanks @capaj !

capaj:
Thanks @dotansimha. I see I forgot to write why I needed this. It will be helpful in the vscode extension that I maintain. I have copied the function in there for now, but I will import this function when it gets to npm.

Issue #6794, opened 2021-10-06
[react-query] Fetcher headers get generated as ...([object Object])


ovidiuwin:
**Describe the bug**
When passing headers to the fetcher config, those get generated as ...([object Object]). Maybe I'm missing something here.

**To Reproduce**
https://codesandbox.io/s/stoic-swanson-qhck5?file=/generated/index.ts Check the codegen.yml and how the headers get generated in this file.

**Expected behavior**
The headers to get generated as expected.

achen0115:
It works for me without the trailing back-tick symbol. Thanks a lot. 👍

hpohlmeyer:
For anyone using a JS [GraphQL Config](https://graphql-config.com/introduction/), you can stringify the object yourself until it is fixed:

```ts
fetcher: {
  endpoint: "http://localhost:3001/graphql",
  fetchParams: JSON.stringify({
    headers: {
      "Content-Type": "application/json"
    }
  })
}
```

royderks:
I have the same error..

Urigo:
Hi and thank you all for the report

Sorry but I'm not adding a lot here but just labeling it according to our new Contribution Guide and [issue flow](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow).

It seems already got into `stage 1` thanks to your reproduction! Thank you for that!

Now in order to advance to `stage 2` we'll need a failing test, would be great if someone could help progress the issues through the [stages](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow).

Thank you and sorry that this comment is not a complete solution (yet).

zirkelc:
I've added a test case via PR https://github.com/dotansimha/graphql-code-generator/pull/7024. It also adapts the `react-query/src/config.ts` type definition for `HardcodedFetch`.

Please let me know if everything is correct :-)

Urigo:
Wow thank you so much @zirkelc !!

It seems already got into `stage 2` thanks to your reproductions and tests, that is a huge push for us!

The next `stage 3` phase is about finding a solution for the issue locally. 
If anyone wants to help, check out the instructions [here](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow) on how to clone the repo and gradually become aware of the internal codebase.

We would love to help with any questions you might have!

[...]

PR #7024, opened 2021-11-15
issue 6794: add stage 2 failing test case

zirkelc:
## Description

This PR adds a test case `React-Query › fetcher: hardcoded-fetch › Should generate query correctly with fetch config and fetchParams object` into spec `packages/plugins/typescript/react-query/tests/react-query.spec.ts` for the config parameter `fetchParams` as requested in issue https://github.com/dotansimha/graphql-code-generator/issues/6794#issuecomment-968057226. 

In case the `fetchParams` parameter was set via a YAML config, it does contain an Object instead of a String. However, the Typescript type definition in [config.ts](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/react-query/src/config.ts#L3) expects a String and doesn't handle the case if an Object was passed via YAML. It is valid to define the parameters as Object according to the documentation: https://www.graphql-code-generator.com/docs/plugins/typescript-react-query#using-fetch-with-codegen-configuration

The newly added test case does currently fail with the following output:

image

Related # (issue)
https://github.com/dotansimha/graphql-code-generator/issues/6794#issuecomment-968057226

## Type of change

### DIFF
--- a/.changeset/pretty-frogs-leave.md
+++ b/.changeset/pretty-frogs-leave.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-react-query': patch
+---
+
+stringify fetchParams if an object was provided via YAML config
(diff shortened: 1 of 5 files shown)


zirkelc:
I've added a 2nd commit to fix the issue `fetchParams`. If `fetchParams` is an Object, the content will be stringified before it's added to the output content. The newly added test case succeeds now:

image

Please let me know if I missed something :-)

Issue #6641, opened 2021-09-09
inlineFragmentTypes: combine generates import for variables in same file


tw0517tw:
**Describe the bug**
 

When using `inlineFragmentTypes: combine` config, codegen will generate `import` statements to import variables in the same file.

**To Reproduce**

see reproducing repo:
https://github.com/tw0517tw/codegen
https://github.com/tw0517tw/codegen/runs/3555409058

Line 3 of generated file is a unnecessary import
https://github.com/tw0517tw/codegen/blob/main/client/foo.fragments.generated.ts#L1-L4

1. My GraphQL schema:

```graphql
type Query {
   greeting:String
   students:[Student]
}

type Student {
   id:ID!
   firstName:String
   lastName:String
   password:String
   collegeId:String
}
```

2. My GraphQL operations:

```ts
import gql from 'graphql-tag';

const bar = gql`
  fragment bar on Student {
    id
    lastName
  }
`;

export default {
  foo: gql`fragment foo on Query {
    greeting

    students{
      id
      firstName

      ...bar
    }

    ${bar}
  }`,

  bar,
};

```

3. My `codegen.yml` config file:

```yml
overwrite: true
generates:
  ./client/graphql.ts:
    schema: "./schema.graphql"
    plugins:
      - typescript

  ./client:
    schema: "./schema.graphql"
    preset: near-operation-file
    presetConfig:
      extension: .generated.ts
      baseTypesPath: graphql.ts
    documents:
      - "./client/**/*.fragments.ts"
    plugins:
      - typescript-operations
      - typescript-react-apollo
    config:
      documentMode: external
      importDocumentNodeExternallyFrom: near-operation-file
      inlineFragmentTypes: combine

```

**Expected behavior**

typescript not reporting error.

**Environment:**

- OS: Windows/Linux
- "@graphql-codegen/cli": "^2.2.0",
- "@graphql-codegen/near-operation-file-preset": "^2.1.3",
- "@graphql-codegen/typescript": "^2.2.1",
- "@graphql-codegen/typescript-operations": "^2.1.3",
- "@graphql-codegen/typescript-react-apollo": "^3.1.3",
- "typescript": "^4.4.2"
- NodeJS: 14

**Additional context**

PR #7122, opened 2021-11-30
Preset: near-operation-file. Don't emit imports to same location.

FritsvanCampen:
When using the preset `near-operation-file` types for gql documents in a source file will be combined to `[sourcefile].generated.ts`. Additionally when using `inlineFragmentTypes: combine` fragments will be re-used using imports (instead of redeclared inline). When implementing this, a practical case was missed: fragments that are defined in the same source file don't need to be imported. You would get something like: `import { myFragment } from "[sourcefile].generated.ts";` and then tsc warns about a nameclash because `myFragment` is already defined in this file.

For more details the see related issue.

Related issue: https://github.com/dotansimha/graphql-code-generator/issues/6641

This is a bug fix. The emitted output changes, but not in a breaking way. The previous output was syntactically incorrect and tsc would give an error (TS2440).

I've add a test "should not add imports for fragments in the same location" which tests for this behavior.
Additionally running the codegen with this fix on my codebase gives the output I expect.

There are many ways to implement this behavior. I thought it would make the most sense to add the path checking logic as close to the place where the import is generated. In theory the pruning of 'bad imports' (imports from itself) can be done at any stage.

It might be worthwhile to verify that the logic for other properties (such as `externalFragments`) is also correct. I'm not sure what it's used for so I haven't touched it.

### DIFF
--- a/packages/presets/near-operation-file/src/fragment-resolver.ts
+++ b/packages/presets/near-operation-file/src/fragment-resolver.ts
@@ -152,10 +152,12 @@ export default function buildFragmentResolver<T>(
           (dedupeFragments &&
             ['OperationDefinition', 'FragmentDefinition'].includes(documentFileContent.definitions[0].kind))
         ) {
-          if (fragmentFileImports[fragmentDetails.filePath] === undefined) {
-            fragmentFileImports[fragmentDetails.filePath] = fragmentDetails.imports;
-          } else {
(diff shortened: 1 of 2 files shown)


FritsvanCampen:
@charlypoly what is the status of this PR? Are there plans to merge it?

PR #7287, opened 2021-12-29
Cache schema and document loading result

kamilkisiela:


charlypoly:
Outdated since https://github.com/dotansimha/graphql-code-generator/pull/7344

PR #7359, opened 2022-01-12
[ReactQuery]Add infinite query options

andrei-datcu:
## Description
- Allow `addInfiniteQuery` to be a configuration object (which will imply the queries are generated)
- Add `exposeQueryKey` for infinite queries as well. If omitted defaults to the outer `exposeQueryKey`
- Add `markAsInfinite` which will generate a `useInfiniteQueryName.isInfinite = true` for the infinite queries

## Usage:
- I am using both properties (`getKey` and `isInfinite`) to update the query cache.

## Type of change

## How Has This Been Tested?

- Added unit tests

## Checklist:

### DIFF
--- a/.changeset/neat-cougars-confess.md
+++ b/.changeset/neat-cougars-confess.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-react-query': minor
+---
+
+Added extra options for infinite queries generation
--- a/packages/plugins/typescript/react-query/src/config.ts
+++ b/packages/plugins/typescript/react-query/src/config.ts
@@ -2,6 +2,7 @@ import { RawClientSideBasePluginConfig } from '@graphql-codegen/visitor-plugin-c
 
 export type HardcodedFetch = { endpoint: string; fetchParams?: string | Record<string, any> };
 export type CustomFetch = { func: string; isReactHook?: boolean } | string;
+export type InfiniteQueryOptions = { exposeQueryKeys?: boolean; markAsInfinite?: boolean };
 
 /**
  * @description This plugin generates `React-Query` Hooks with TypeScript typings.
@@ -91,5 +92,5 @@ export interface ReactQueryRawPluginConfig
    * @default false
    * @description Adds an Infinite Query along side the standard one
    */
-  addInfiniteQuery?: boolean;
+  addInfiniteQuery?: boolean | InfiniteQueryOptions;
 }
--- a/packages/plugins/typescript/react-query/src/fetcher-custom-mapper.ts
+++ b/packages/plugins/typescript/react-query/src/fetcher-custom-mapper.ts
@@ -1,5 +1,10 @@
 import { ParsedMapper, buildMapperImport, parseMapper } from '@graphql-codegen/visitor-plugin-common';
-import { generateInfiniteQueryKey, generateMutationKey, generateQueryKey } from './variables-generator';
+import {
+  generateInfiniteQueryKey,
+  generateMutationKey,
+  generateQueryKey,
+  generateQueryVariablesSignature,
+} from './variables-generator';
 
 import { CustomFetch } from './config';
 import { FetcherRenderer } from './fetcher';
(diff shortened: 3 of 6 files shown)


charlypoly:
Hi @andrei-datcu, thank you for your contribution!

Could you please update your branch with `master`?

Is seems that many of your changes already exists in the actual codebase 👀

andrei-datcu:
@charlypoly Done! Some things have been added in the mean time so this only adds:
* the possibility to only generate `getKey()` just for the normal queries or just for the infinite queries
* the option to generate a constant `isInfinite` property for all the infinite queries. Personally, I'm using this when editing the cache, to automatically wrap a query data as an `InfiniteData`.

charlypoly:
@andrei-datcu 

Regaring the infiniteQueries `getKey()`, I would suggest the following:

- add a new `exposeInfiniteQueryKeys: boolean` option

OR 

- add a new `exposeKeys: { query?: boolean, mutation?: boolean, infiniteQuery?: boolean } | boolean` option and depreciate - while keeping support for - existing `exposeQueryKeys` and `exposeMutationKeys`


-----

> * the option to generate a constant `isInfinite` property for all the infinite queries. Personally, I'm using this when editing the cache, to automatically wrap a query data as an `InfiniteData`.

could you please provide a snippet of your use-case? Thank you!

charlypoly:
Hi @andrei-datcu!

Just a heads-up on the previous comment 👀

charlypoly:
Hi @andrei-datcu 👋🏼 
I am still waiting for your answer on the following question:

> > * the option to generate a constant `isInfinite` property for all the infinite queries. Personally, I'm using this when editing the cache, to automatically wrap a query data as an `InfiniteData`.
> 
> could you please provide a snippet of your use-case? Thank you!

Best!

andrei-datcu:
> @andrei-datcu
> 
> Regaring the infiniteQueries `getKey()`, I would suggest the following:
> 
> * add a new `exposeInfiniteQueryKeys: boolean` option
> 
> OR
> 
> * add a new `exposeKeys: { query?: boolean, mutation?: boolean, infiniteQuery?: boolean } | boolean` option and depreciate - while keeping support for - existing `exposeQueryKeys` and `exposeMutationKeys`
> 

Both would work, but will allow config inconsistencies. Namely:
one could set `exposeInfiniteQueryKeys: true` in config, yet that setting will have no influence on the final output unless you also set `addInifniteQuery: true`. So you now have a key in the config which you can set to either `true` or `false` without changing the final result at all. Ditto for your other option.

That being said, I do see little value of having a per-type-of-query possibility to generate the key. I merely implemented it this way because at the time of writing this patch `exposeQueryKeys` would not generate keys for the infinite queries. This was just my way of solving that problem which has now already been solved. We could drop it all together as far as I'm concerned.

> > * the option to generate a constant `isInfinite` property for all the infinite queries. Personally, I'm using this when editing the cache, to automatically wrap a query data as an `InfiniteData`.
> 
> could you please provide a snippet of your use-case? Thank you!

```
function matchQueries(dataKey: string) {
    const result = kGlobalInitialData.filter((hq) => typeof hq.key === "string" ?
        hq.key === dataKey : hq.key[0] === dataKey);
    if (result.length === 0) {
        throw new Error(`Was not able to find hydration data for ${dataKey}`);
    }
    return result;
}


    const hydrationQueryies = [useMyInfiniteQuery, useMyNormalQuery];
    for (const hq of hydrationQueries) {
            const k = hq.getKey();
            for (const {key, data} of matchQueries(k[0] as string)) {
                client.setQueryData(key, hq.isInfinite ? {
                    pages: [data],
                    pageParams: [undefined]
                } : data);
            }
    }
```
Granted, I could just check if the key ends in `.infinite` and hope for the best (provided nobody ever changes the way the key is computed), so I'm guessing this is not that critical as well? It's nice to have tough and it has minimal costs to maintain.

charlypoly:
@andrei-datcu let's keep `exposeInfiniteQueryKeys` as it is, however, let's remove the `. infinite` which is targeting a too specific use-case

Thanks!

andrei-datcu:
Not going to add a `exposeInfiniteQueryKeys` because it brings no value for me and because it can lead to to bad configs as stated above.

Issue #7362, opened 2022-01-12
[gql-tag-operations-preset] Generated index.ts should contain \n regardless of whether source contains LF or CRLF


vyacheslav-pushkin:
First of all, I would like to thank the creators of `gql-tag-operations-preset`, it's a really clever solution to have type safety while avoiding explicit imports.

### TL;DR

Generated `index.ts` will cause TypeScript compilation to fail when source files contain CRLF (e.g. when using Windows).

### Long description

When source file contains **LF**, generated `index.ts` will contain **\n** and TypeScript will compile successfully.

When source file contains **CRLF**, generated `index.ts` will contain **\r\n** and TypeScript **will not compile**.

When source file contains **CRLF**, and generated `index.ts` is manually changed to contain **\n**, TypeScript **will compile successfully**.

It seems that generated `index.ts` should contain `\n` regardless of whether the source contains `LF` or `CRLF`.

## To Reproduce

**NOTE: I have created a [minimal reproduction example](https://github.com/vyacheslav-pushkin/crlf-bug-repro). It contains  schema, `codegen.yml` and operations.**

Steps to reproduce the problem (using Linux):

#### 1. Modify `src/app/app.tsx` by replacing `LF` with `CRLF`:

```
$ unix2dos src/app/app.tsx 
unix2dos: converting file src/app/app.tsx to DOS format...
```

New lines are now `CRLF`:

```
$ cat -A src/app/app.tsx
import {gql} from "@my/gql";^M$
import {useQuery} from "@apollo/client";^M$
import React from "react";^M$
^M$
const PET_LIST = gql(`^M$
  query Get_Pet_List {^M$
    petList {^M$
      id^M$
      identificationNumber^M$
    }^M$
  }^M$
`);^M$
^M$
export const MyComponent = () => {^M$
  const {data} = useQuery(PET_LIST);^M$
^M$
  return <>{data?.petList?.[0]?.id}</>;^M$
};^M$
```

#### 2. Run `graphql-codegen` by executing `npm run generate`:

```
$ npm run generate

> crlf-bug@1.0.0 generate
> graphql-codegen

(node:9963) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
  ✔ Parse configuration
  ✔ Generate outputs
```

The generated `index.ts` file will now contain `\r\n`:

```
$ cat src/gql/index.ts
/* eslint-disable */
[...]

PR #7369, opened 2022-01-17
fix: gql-tag-operations generates invalid types on Windows #7362

vyacheslav-pushkin:
## Description

Related #7362 

#### Problem summary

`gql-tag-operations` generates invalid types on Windows, causing compile-time error when attempting to launch the application.

#### Root cause

Source file is read by `@graphql/tools` using `fs.promises.readFile`, which means that the linebreaks are read as-is and the result will be different depending on the OS: it will contain LF (`\n`) on Linux/MacOS and CRLF (`\r\n`) on Windows.

In most scenarios that would be OK. However, `gql-tag-operations` is using the resulting string as a TypeScript type. Which means that the string will be compared against a template literal, for example:

```typescript
`
query a {
   a
 }
` === '\n query a {\n    a\n  }\n '
```

According to [clause 12.8.6.2 of ECMAScript Language Specification](https://tc39.es/ecma262/#sec-static-semantics-trv), when comparing strings, JavaScript doesn't care which linebreaks does the source file contain, any linebreak (CR, LF or CRLF) is LF from JavaScript standpoint (otherwise the result of the above comparison would be OS-dependent, which doesn't make sense).

The reason `gql-tag-operations` breaks on Windows is that it generates `\r\n query a {\r\n    a\r\n  }\r\n `, which is NOT equal to

```
`
query a {
   a
 }
`
```

#### Solution

`@graphql/tools` is doing its job correctly. Also it probably doesn't make sense to make changes on `typescript` plugin level, as `gql-tag-operations` is probably the only place where it matters. Therefore the solution is to replace `\r\n` with `\n` during the preset-specific processing.

## Type of change

## How Has This Been Tested?

- [x] This PR adds a unit test replicating the issue. In order to reproduce the issue, revert the `fix` commit (leaving only `test` commit) and run the tests.
- [x] In addition I have tested a local build of `gql-tag-operations` within [this application](https://github.com/Amplicode/amplicode-frontend/tree/master/example-app) manually on a Windows machine.

**Test Environment**:
- OS: Windows
- `@graphql-codegen/typescript: ^2.2.2`: 
- NodeJS: 14

## Checklist:

### DIFF
--- a/.changeset/orange-keys-guess.md
+++ b/.changeset/orange-keys-guess.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/gql-tag-operations-preset': patch
+---
+
+fix: gql-tag-operations generates invalid types on Windows #7362
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# This fixture must contain CRLF as line breaks. See https://github.com/dotansimha/graphql-code-generator/issues/7362
+packages/presets/gql-tag-operations/tests/fixtures/crlf-operation.ts eol=crlf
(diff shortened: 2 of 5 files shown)


n1ru4l (approved):

n1ru4l (approved):
@vyacheslav-pushkin Thank you for this very detailed investigation and fix for this issue. ❤️

PR #7430, opened 2022-01-28
fix: RequireFields generic making all other fields optional

GreenGremlin:
## Description

The generated `RequireFields` type generic has the unintended side-effect of making all other properties optional. I haven't seen this cause any issues in generated types, but it can cause issues if it is used outside of the generated types. I ran into this issue by wrapping a resolver result type with `RequireFields`, something like `RequireFields<ResolvedThing>`. Wrapping the type with `RequireFields` did make the noted fields required, but it had the unexpected side effect of making fields that were already required optional. This is happening because the `RequireFields` generic explicitly makes all fields that are not passed in the union optional. The fix is to use the build-in `Omit` generic instead.

Since I don't know of a bug this causes with the generated code, I decided to skip creating an issue. Let me know if that's a problem and I'll see what I can come up with.

## Type of change

### DIFF
--- a/dev-test/modules/types.ts
+++ b/dev-test/modules/types.ts
@@ -5,7 +5,7 @@ export type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K]
 export type MakeOptional<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]?: Maybe<T[SubKey]> };
 export type MakeMaybe<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]: Maybe<T[SubKey]> };
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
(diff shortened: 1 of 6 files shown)


charlypoly (changes requested):
Hi @GreenGremlin,

Thank you for your fix!

Looks good to me.
However, the updates to the `dev-test/` files need to be reformatted since its making test fail for now ➡️ https://github.com/dotansimha/graphql-code-generator/runs/4975168321?check_suite_focus=true

charlypoly (approved):

charlypoly (changes requested):
@GreenGremlin we did a deeper analysis of the issue and found many places of the code impacted by this change.

I'll take care of the changes and keep you updated!

Thanks! ⚡

charlypoly (approved):

charlypoly:
@GreenGremlin thank you for your contribution!

Please update your `@graphql-codegen/*` package to get the updated code! 📦

GreenGremlin:
Awesome! Thanks for the quick merge / publish!

PR #7452, opened 2022-02-01
Performance work: resolvers plugins, documents loading

charlypoly:
- [x] skip the type mapping (`shouldMapType()` calls in `createResolversFields()`) if not `mappers` config is provided
- [x] improve documents loading cache

------

On a 4sec total build project:

**Before**
image

**After**
image

### DIFF
--- a/.changeset/witty-carrots-clap.md
+++ b/.changeset/witty-carrots-clap.md
@@ -1,0 +1,6 @@
+---
+"@graphql-codegen/cli": patch
+"@graphql-codegen/visitor-plugin-common": minor
+---
+
+Performance work: resolvers plugins, documents loading
--- a/.changeset/witty-carrots-clap.md
+++ b/.changeset/witty-carrots-clap.md
@@ -1,6 +1,6 @@
 ---
 "@graphql-codegen/cli": patch
-"@graphql-codegen/visitor-plugin-common": minor
+"@graphql-codegen/visitor-plugin-common": patch
 ---
 
 Performance work: resolvers plugins, documents loading
--- a/packages/graphql-codegen-cli/src/codegen.ts
+++ b/packages/graphql-codegen-cli/src/codegen.ts
@@ -257,11 +257,17 @@ export async function executeCodegen(input: CodegenContext | Types.Config): Prom
                     task: wrapTask(async () => {
                       debugLog(`[CLI] Loading Documents`);
 
-                      const allDocuments = [...rootDocuments, ...outputSpecificDocuments];
+                      // get different cache for shared docs and output specific docs
+                      const results = await Promise.all(
+                        [rootDocuments, outputSpecificDocuments].map(docs => {
+                          const hash = JSON.stringify(docs);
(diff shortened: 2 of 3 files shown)


.changeset/witty-carrots-clap.md
```diff
@@ -0,0 +1,6 @@
+---
+"@graphql-codegen/cli": patch
+"@graphql-codegen/visitor-plugin-common": minor
```
dotansimha, line 3: Does this require a minor? why?

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -510,10 +510,13 @@ export class BaseResolversVisitor<
     const nestedMapping: { [typeName: string]: boolean } = {};
     const typeNames = this._federation.filterTypeNames(Object.keys(allSchemaTypes));
 
-    typeNames.forEach(typeName => {
-      const schemaType = allSchemaTypes[typeName];
-      nestedMapping[typeName] = this.shouldMapType(schemaType, nestedMapping);
-    });
+    // avoid checking all types recursively if we have no `mappers` defined
```
dotansimha, line 513: nice! :) 
in cases where `mappers` is defined, do you think we can optimize something within `shouldMapType`? 
I guess the result of `shouldMapType` can be cached for a given type? so we can only check it once? 🤔

charlypoly: `shouldMapType()` has already a cache system that relies on `checkedBefore` parameter
However, we could try to have `nestedMapping` (in `createResolversFields()`) moved as an instance variable, however, I'm not sure of the side effect...

I'll make the changes and test!

charlypoly: @dotansimha tested on a real project, no regression, and slightly better perf when `mappers` is defined.

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -257,11 +257,17 @@ export async function executeCodegen(input: CodegenContext | Types.Config): Prom
...
+                          return documentsLoadingCache.load(hash);
+                        })
+                      );
+
+                      const documents: Types.DocumentFile[] = [];
 
-                      const hash = JSON.stringify(allDocuments);
-                      const result = await documentsLoadingCache.load(hash);
-                      const documents: Types.DocumentFile[] = result.documents;
+                      results.forEach(source => documents.push(...source.documents));
```
n1ru4l, line 270: Won't this break if you have identical operations with near-operation-file preset?

charlypoly: Discussed offine 👍🏼

n1ru4l (approved):

tvvignesh (approved):

kamilkisiela (approved):

PR #7481, opened 2022-02-07
Loading API URL from TypeScript file example

patryk-smc:
This PR adds an example of loading an API URL from a TypeScript file. 

Thanks to @n1ru4l for pointing me in the right direction.


n1ru4l (approved):
thank you 🥳

PR #6969, opened 2021-11-03
Allow customization for extensions type

dodomorandi:
## Description

This introduces a new configuration for the `graphql-request` plugin for
typescript. It allows to use `unknown` (or a custom type) instead of
`any` for the `extensions` parameter without breaking changes.

Related #6966 

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

No jest tests has been added for this. The change is trivial, but if you want me to add a test, just say it.

## Checklist:

- [x] I have followed the [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes -- **Only if I only run tests related to the plugin, otherwise I have lots tests that don't pass, like in master**

PR #7592, opened 2022-02-28
fix: Remove explicit return types from sdk methods

P4sca1:
## Description

Previously, the return type of all SDK methods was declared manually. This can
cause the return type to become outdated when graphql-request gets updated.
This already happened, as graphql-request removed the errors attribute from the
return type (see https://github.com/prisma-labs/graphql-request/issues/174).
Also, `data` was marked as possibly undefined which is wrong. `data` always exists
on a successful response (failed requests throw a `ClientError`).

A sideeffect of this change is that the `extensionsType` config became obsolete
and has been removed. Because of this, I marked this change as a breaking change.

## Type of change

I removed tests for the now removed `extensionsType` config and ran `yarn test -u`.  All tests are still passing. I also tested the change in a private project and did not experience any issues with the change.

### DIFF
--- a/.changeset/shiny-dogs-press.md
+++ b/.changeset/shiny-dogs-press.md
@@ -1,0 +1,16 @@
+---
+'@graphql-codegen/typescript-graphql-request': major
+'@graphql-codegen/website': patch
+---
+
+Do not specify return types of sdk methods explicitly.
+
+Previously, the return type of all SDK methods was declared manually. This can
+cause the return type to become outdated when graphql-request gets updated.
+This already happened, as graphql-request removed the errors attribute from the
+return type (see https://github.com/prisma-labs/graphql-request/issues/174).
+Also, data was marked as possibly undefined which is wrong. data always exists
+on a successful response (failed requests throw a ClientError).
+
+A sideeffect of this change is that the extensionsType config became obsolete
+and has been removed.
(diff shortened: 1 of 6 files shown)


charlypoly:
Hi @P4sca1,

The `extensionTypes` config has been [added quite recently](https://github.com/dotansimha/graphql-code-generator/pull/6969), it seems a bit premature to simply drop it because of concern around the future typing changes of `graphql-request`.
Typing `extensions` when using `rawRequest` is a legit use-case.
In order to completely remove the `extensionTypes` option without dropping this use-case, a good approach would be to [patch `graphql-request` to provide `extensions` generic type parameter](https://github.com/prisma-labs/graphql-request/blob/24ca53bafb4c7f777ab124604345cb64ad3cf301/src/index.ts#L201).

For now, we should especially fix the `data` as a non-undefined property and make the typings evolve in time.

charlypoly (changes requested):
cf: https://github.com/dotansimha/graphql-code-generator/pull/7592#issuecomment-1066571735

P4sca1:
I agree, we should wait for `graphql-request` to provide the ability to change the `extensions` type, before making this change.
In the mean time, I will make a seperate PR to fix the current typings.

charlypoly:
> In the mean time, I will make a seperate PR to fix the current typings.

Thank you @P4sca1!

Issue #7562, opened 2022-02-18
enumPrefix = false applying to non-enum types


bmarker:
**Describe the bug**
The enumPrefix = false config setting is stripping the prefix from non-enum types in `IResolversTypes` and `IResolversParentTypes`.

**To Reproduce**
https://codesandbox.io/s/quirky-proskuriakova-vmpz0v?file=/types.ts

1. My GraphQL schema:

```graphql
type Query {
    user: User!
}

type User {
    name: String!
}
```

2. My GraphQL operations:

None

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      typesPrefix: I
      enumPrefix: false
```

**Expected behavior**
Only skip the type prefix for enum types. The type for `User` should match the generated type `IUser`.
```
export type IResolversParentTypes = {
  Boolean: Scalars['Boolean'];
  Query: {};
  String: Scalars['String'];
  User: IUser;
};
```

Instead of:
```
export type IResolversParentTypes = {
  Boolean: Scalars['Boolean'];
  Query: {};
  String: Scalars['String'];
  User: User;
};
```

**Environment:**
https://codesandbox.io/s/quirky-proskuriakova-vmpz0v?file=/types.ts

```
    "@graphql-codegen/add": "^3.1.1",
    "@graphql-codegen/cli": "^2.4.0",
    "@graphql-codegen/typescript": "2.4.5",
    "@graphql-codegen/typescript-resolvers": "2.5.2",
    "graphql": "^16.2.0"
```

**Additional context**
Also happens with `typesSuffix` set and using `{T}` in `defaultMapper`.

charlypoly:
Hi @bmarker,

Thank you for your report!


Could you please provide a reproduction on Codesandbox by using our template?
➡️ https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template/tree/master/

Thank you!

bmarker:
There's already a Codesandbox link. Are you asking for a different format?

charlypoly:
Hi @bmarker, you're right, I missed it.

I created a fix PR, could you try again with the following alpha versions?

```
@graphql-codegen/typescript-resolvers@2.5.3-alpha-643ff6c1d.0
@graphql-codegen/typescript@2.4.6-alpha-643ff6c1d.0
```

Thank you!

bmarker:
[...]

PR #7626, opened 2022-03-07
fix(resolvers): fix conflict between `typesPrefix: true` and `enumPrefix: false`

charlypoly:
Fixes https://github.com/dotansimha/graphql-code-generator/issues/7562

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -563,9 +563,11 @@ export class BaseResolversVisitor<
           .getTypes()
           .map(type => getTypeToUse(type.name))
           .join(' | ');
+      } else if (isEnumType(schemaType)) {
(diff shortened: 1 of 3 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -563,9 +563,11 @@ export class BaseResolversVisitor<
           .getTypes()
           .map(type => getTypeToUse(type.name))
           .join(' | ');
+      } else if (isEnumType(schemaType)) {
+        prev[typeName] = this.convertName(typeName, { useTypesPrefix: this.config.enumPrefix }, true);
       } else {
         shouldApplyOmit = true;
-        prev[typeName] = this.convertName(typeName, { useTypesPrefix: this.config.enumPrefix }, true);
```
charlypoly, line 568: This branch was catching object + enum types 👀

n1ru4l (approved):

PR #7679, opened 2022-03-22
flow-resolvers: export type as same convension as typescript-resolvers

cometkim:
## Description

This only changes the type definition. For users who use `@graphql-codegen/core` with `@graphql-codegen/flow-resolvers`.

Importing the type definition is optional and I think gatsby-plugin-typegen is the only project that wants to reference it.

https://github.com/cometkim/gatsby-plugin-typegen/blob/e009537/plugin/src/services/codegen.ts#L4-L9

As you can see, the flow-resolvers is the only plugin that doesn't export type definition from `./config`


charlypoly:
@cometkim 

> As you can see, the flow-resolvers is the only plugin that doesn't export type definition from ./config

Many other plugins export the config type from `index.ts` (for example, the `schema-ast` plugin, [see the list here](https://github.com/dotansimha/graphql-code-generator/blob/423a93b9baf2071fe3d947aa0eb8c9ab0ec41939/website/src/lib/plugins-docs.ts#L194-L207))

I'm still merging your PR, but it is not necessary to move all plugins config type definition to a `config.ts` file.

PR #7710, opened 2022-03-30
Add Prettier Check action

B2o5T:


website/package.json
```diff
@@ -7,8 +7,6 @@
     "build": "yarn generate-json-config && next build",
     "dev": "concurrently -r 'next-remote-watch ./docs ./src/pages/_app.tsx' 'wait-on -s 1 http://localhost:3000'",
     "lint": "eslint --ignore-path .gitignore --ext js,jsx,cjs,mjs,ts,tsx,cts,mts .",
-    "prettier:check": "prettier --ignore-path .gitignore --check .",
```
B2o5T, line 10: no need as we have prettier scripts in the root package.json

charlypoly (approved):
That might conflict with all existing PRs but let's do it! 😅

PR #7715, opened 2022-03-31
feat(cli): bump code-file-loader to support svelte files

ardatan:
Now GraphQL Codegen can extract GraphQL SDL from `.svelte` files.
Thanks @jycouet for his work in;
https://github.com/ardatan/graphql-tools/pull/4348

`documents: **/*.svelte`

### DIFF
--- a/.changeset/smart-trees-argue.md
+++ b/.changeset/smart-trees-argue.md
@@ -1,0 +1,9 @@
+---
+'@graphql-codegen/cli': minor
+---
+
+feat(cli): bump code-file-loader to support svelte files
+
+Now we can use graphql-tag-pluck to extract GraphQL SDL from `.svelte` files.
+Thanks @jycouet for his work in;
+https://github.com/ardatan/graphql-tools/pull/4348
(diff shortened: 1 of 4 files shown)


charlypoly:
We need to update the guide: https://www.graphql-code-generator.com/docs/guides/svelte

@jycouet, would you have time to do it? 👀  (While keeping both Svelte Apollo and KitQL sections)

jycouet:
I think that we should not update the doc as the full story is not done yet.
To have full support of inline graphql in .svelte files, I want to have: 
- [x] codegen (done)
- [x] syntax highlighting in vscode
- [ ] autocompletion in vscode (I will work on this soon)

I fine releasing the version, but would wait all these points before updating the doc.

charlypoly (approved):
@ardatan good to merge ✅

charlypoly:
This PR is no longer since `@graphql-tools/code-file-loader` has been bumped by Renovate since then.

jycouet:
Got it, it's good that renovate was faster than us 😅
I guess that the only drawback is that in the changeset it's `Thanks @renovate for...`, nevermind ^^

Good cleanup 🎉

Issue #7720, opened 2022-04-02
Out of memory when generating typescript operations for gatsby wordpress schema


rburgst:
### Issue workflow progress

_Progress of the issue based on the [Contributor Workflow](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow)_

- [x] 1. The issue provides a reproduction available on [Github](https://github.com/dotansimha/graphql-code-generator-issue-sandbox-template), [Stackblitz](https://stackblitz.com/github/dotansimha/graphql-code-generator-issue-sandbox-template) or [CodeSandbox](https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template)

_Make sure to fork this template and run `yarn generate` in the terminal._

_Please make sure the Codegen and plugins version under `package.json` matches yours._

- [ ] 2. A failing test has been provided

- [ ] 3. A local solution has been provided

- [ ] 4. A pull request is pending review

---

**Describe the bug**

When using a gatsby wordpress schema together with typescript-operations I get an out of memory error.

**To Reproduce**

https://github.com/rburgst/graphql-code-generator-issue-ts-ops

run 
```
yarn install
yarn generate
```

**Expected behavior**

The codegen should build through without errors.

**Environment:**

- OS: Mac OSX 12.3
- ```    "@graphql-codegen/add": "^3.1.1",
    "@graphql-codegen/cli": "^2.4.0",
    "@graphql-codegen/typescript": "2.4.8",
    "@graphql-codegen/typescript-operations": "2.3.5",
`???: 
- NodeJS: 16.3.1

**Additional context**

The problem only appears once I add the `fragments.ts` to the `documents` config.

When running  `yarn generate` I get an OOM

```
yarn generate   
yarn run v1.22.17
$ yarn codegen
$ graphql-codegen --config ./codegen.yml
(node:66908) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
  ✔ Parse configuration
  ❯ Generate outputs
    ❯ Generate types.ts
      ✔ Load GraphQL schemas
      ⠙ Load GraphQL documents
        Generate

<--- Last few GCs --->

[...]

PR #7723, opened 2022-04-02
add failing test #7720

B2o5T:
related https://github.com/dotansimha/graphql-code-generator/issues/7720

@charlypoly @dotansimha @n1ru4l
For anyone who wanna works on this I created a failing test

There is an infinite recursion in `packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts` file in `_buildGroupedSelections()` method


n1ru4l (changes requested):
Please also fix the issue 😁

PR #7728, opened 2022-04-04
enable `no-lonely-if` rule

B2o5T:


packages/plugins/java/apollo-android/src/input-type-visitor.ts
```diff
@@ -107,15 +107,13 @@ export class InputTypeVisitor extends BaseJavaVisitor<VisitorConfig> {
...
-        return typeof wrapWith === 'function' ? `${wrapWith(typeToUse)} ${name}` : `${wrapWith}<${typeToUse}> ${name}`;
-      } else {
-        if (applyNullable) {
-          this._imports.add(Imports.Nullable);
-        }
-        return `${applyNullable ? '@Nullable ' : ''}${typeToUse} ${name}`;
+      if (applyNullable) {
+        this._imports.add(Imports.Nullable);
       }
+      return `${applyNullable ? '@Nullable ' : ''}${typeToUse} ${name}`;
```
charlypoly, line 116: why this is not translated to:

```suggestion
      else if (applyNullable) {
        this._imports.add(Imports.Nullable);
      } else {
         return `${applyNullable ? '@Nullable ' : ''}${typeToUse} ${name}`;
```

(with the removing of the parent `else`)

B2o5T: @charlypoly sorry for long respond, missed it 😐.
Answer - because the original behaviour of code expects to execute
```js
return `${applyNullable ? '@Nullable ' : ''}${typeToUse} ${name}`;
```
line no matter `applyNullable` value, in your case you changed the original behaviour of code, execution of line with return statement happens only on falsy `applyNullable` value.

charlypoly: Okay, thanks! 🤓

charlypoly (approved):

PR #7731, opened 2022-04-04
enable `no-duplicate-imports` rule

B2o5T:


packages/utils/graphql-codegen-testing/src/typings.d.ts
```diff
@@ -1 +0,0 @@
-declare module 'lz-string';
```
B2o5T, line 1: I removed this file and installed `@types/lz-string`

charlypoly (approved):

B2o5T:
@charlypoly we didn't decide to enable this rule yet (we'll compare it with `imports/no-duplicates` rule), but I merge current fixes of duplicated imports

PR #7732, opened 2022-04-04
add `no-unreachable-loop` rule

B2o5T:


packages/plugins/java/apollo-android/src/operation-visitor.ts
```diff
@@ -143,7 +143,7 @@ export class OperationVisitor extends BaseJavaVisitor<VisitorConfig> {
   private createUniqueClassName(inUse: string[], name: string, count = 0): string {
     const possibleNewName = count === 0 ? name : `${name}${count}`;
 
-    while (inUse.includes(possibleNewName)) {
```
charlypoly, line 146: weird code 🤔

charlypoly (approved):

PR #7734, opened 2022-04-04
enable `prefer-destructuring` rule

B2o5T:


charlypoly (approved):

charlypoly:
@B2o5T general question to all ESLint PRs: should we add changesets? 👀

B2o5T:
@charlypoly we didn't decide to enable this rule yet, but I merge fixes of this rule

Issue #6873, opened 2021-10-20
[typescript-operations] operation type not restricted to inline fragment


OscarBarrett:
**Describe the bug**
typescript-operations does not appear to take into account the specific inline fragments used, instead generating a type that could include anything from the returned interface.

**To Reproduce**
Have two or more types that share an interface.

Define a query that returns the interface, but only include one of the types in an inline fragment.

Codesandbox: https://codesandbox.io/s/long-http-1wibc

1. My GraphQL schema:

```graphql
type Query {
    entries(type: [String]): EntryInterface
}

interface EntryInterface {
    title: String
}

type first_Entry implements EntryInterface {
    title: String
    first: String
}

type second_Entry implements EntryInterface {
    title: String
    second: String
}
```

2. My GraphQL operations:

```graphql
query first {
    entries(type: "first") {
        ...on first_Entry {
            __typename
            title
            first
        }
    }
}

```

3. My `codegen.yml` config file:

```yml
schema: schema.graphql
documents: document.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-operations
```

**Expected behavior**

`FirstQuery` is generated as follows:

```ts
export type FirstQuery = {
  __typename ? : 'Query',
  entries ? : {
    __typename: 'first_Entry',
    title ? : string | null | undefined,
    first ? : string | null | undefined
  } | {
    __typename ? : 'second_Entry'
  } | null | undefined
};
```

The expected type would be:

```ts
export type FirstQuery = {
  __typename ? : 'Query',
  entries ? : {
    __typename: 'first_Entry',
    title ? : string | null | undefined,
    first ? : string | null | undefined
  } | null | undefined
};
```

(i.e. no `{ __typename?: 'second_Entry' }` in the union).

**Environment:**

- OS: MacOS
- `@graphql-codegen/typescript`: 2.2.4
- `@graphql-codegen/typescript-operations`:  2.1.8
- NodeJS: 14

**Additional context**
In typescript-operations v1, when using `skipTypename: true` objects only containing `__typename?` would be removed, so this could be avoided. In v2 the union will include `{}` instead.

gbettencourt:
Hitting this same issue, which is preventing us from updating.  We'd have to make type assertions to force the correct type which is not ideal.

peternedap:
I'm running into the same problem. This is also preventing an upgrade to graphql 16 in our codebase. 
[...]

PR #7782, opened 2022-04-21
Add config option to omit export of empty fragment subtypes

peternedap:
## Description

Adds an extra config field for omitting empty objects when using inline fragment spreads in combination with `skipTypename`.
This makes an upgrade from `@graphql-codegen/typescript-operations` version 1.18.0 to version 2 possible for people who run into problems with this. This also allows the upgrade from graphql 15 to 16, since `typescript-operations` 1.18 is incompatible with version 16.

Related # [(issue)](https://github.com/dotansimha/graphql-code-generator/issues/6873)

## Type of change

## How Has This Been Tested?

Added unit tests:

- [x] Config Should not generate empty fragments when omitEmptyFragments is set to true
- [x] Config Should generate empty fragments when omitEmptyFragments is set to false

**Test Environment**:

- OS: Ubuntu
- `@graphql-codegen/...`: typescript-operations
- NodeJS: v17.8.0

## Checklist:

## Further comments
n/a

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-documents-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-documents-visitor.ts
@@ -36,6 +36,7 @@ export interface ParsedDocumentsConfig extends ParsedTypesConfig {
   omitOperationSuffix: boolean;
   namespacedImportName: string | null;
   exportFragmentSpreadSubTypes: boolean;
+  omitEmptyFragments: boolean;
   skipTypeNameForRoot: boolean;
   experimentalFragmentVariables: boolean;
   mergeFragmentTypes: boolean;
@@ -101,6 +102,11 @@ export interface RawDocumentsConfig extends RawTypesConfig {
   exportFragmentSpreadSubTypes?: boolean;
   /**
    * @default false
+   * @description If set to true, it will not export empty fragment spread sub-types
+   */
+  omitEmptyFragments?: boolean;
+  /**
+   * @default false
    * @description If set to true, it will enable support for parsing variables on fragments.
    */
   experimentalFragmentVariables?: boolean;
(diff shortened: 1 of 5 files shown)


peternedap:
As a first-time contributor I'm not allowed to run the other workflows

n1ru4l (changes requested):
Could you explain why you would want to introduce a flag for generating wrong types? Omitting the types of empty fragments can cause runtime errors as the TypeScript type would no longer mirror the actual GraphQL result type.

peternedap:
We use global unique IDs in our code base, and every queryable type implements 

```gql
""" Node, to make a type queryable, it needs to implement this interface. """
interface Node {
    """ A node requires an ID. """
    id: ID!
}
```

Furthermore, every queryable type is queried via a spread:

```gql
""" The query root. """
type Query {
    """ Retrieves a Node through its ID. """
    node(id: ID!): Node
}

type Person implements Node {
    """ The ID of the person. """
    id: ID!

    """ Self explanatory. """
    name: String
}

```
queried by 

```ts
export const PERSON_DATA_QUERY = gql`
    query person($id: ID!) {
        node(id: $id ) {
            ... on Person {
                id
                name
            }
        }
    }
`;
```

This is a flow that was well supported in old versions, but with new versions we run into the problem that the resulting generated type cannot be easily narrowed to only the person data. We'd either have to include the typename and add an extra check to every query result, or find a way to remove `{}` from the resulting query types. 
The resulting type can indeed be mismatched runtime when a different ID than that of a person is given as a query variable, but that would also result in different errors since there'd be no Person with that ID (because of the global unique IDs). 

This MR thus introduces an optional way to reintroduce support for the previously existing behaviour where empty spreads were not added to the type.

PR #7855, opened 2022-05-18
fixes for `unicorn/no-lonely-if` rule

B2o5T:


packages/presets/near-operation-file/src/utils.ts
```diff
@@ -36,21 +36,19 @@ export function extractExternalFragmentsInUse(
...
-              extractExternalFragmentsInUse(
-                fragmentNameToFile[node.name.value].node,
-                fragmentNameToFile,
-                result,
-                level + 1
-              );
-            }
+        if (
+          !ignoreList.has(node.name.value) &&
+          (result[node.name.value] === undefined || level < result[node.name.value])
```
B2o5T, line 41: I replaced 
```ts
if (
  result[node.name.value] === undefined ||
  (result[node.name.value] !== undefined && level < result[node.name.value])
)
```

by

```ts
if (
  result[node.name.value] === undefined ||
  level < result[node.name.value]
)
```

since `result[node.name.value] !== undefined` condition is redundant as is always will be `true`

PR #7872, opened 2022-05-22
Fix typo in docs

lorensr:
I also wanted to fix this, which errors (there is no `graphql-codegen` script—you have to do `npx graphql-codegen init`), but I don't see where `<PackageRun>` is defined?

image


dotansimha (approved):

dotansimha:
thanks @lorensr !

B2o5T:
@lorensr Hi Loren, thank you! `<PackageRun/>` is defined in this place https://github.com/the-guild-org/the-guild-docs/blob/main/packages/client/src/components/Package.tsx#L178-L234 if you could fix this inconsistency with `npx graphql-codegen init` this will be so great!

PR #7886, opened 2022-05-25
feat: support generating correct types for `@oneOf` input types

n1ru4l:
Support for the `@oneOf` directive on input object types.

Things to consider:

- Should we make this available behind a flag?
  
  `@onOf` is still in the RFC stage and the behavior could change slightly. Should we, therefore, make this opt-in?

____

References:
- https://github.com/graphql/graphql-js/pull/3513#issuecomment-1137703851
- https://github.com/graphql/graphql-spec/pull/825

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -40,6 +40,7 @@ import {
   wrapWithSingleQuotes,
   getConfigValue,
   buildScalarsFromConfig,
+  isOneOfInputObjectType,
 } from './utils';
 import { OperationVariablesToObject } from './variables-to-object';
 import { parseEnumValues } from './enum-values';
@@ -466,7 +467,7 @@ export class BaseTypesVisitor<
     if (this.config.onlyEnums) return '';
 
     // Why the heck is directive.name a string and not { value: string } at runtime ?!
-    if (node.directives?.some(directive => (directive.name as any) === 'oneOf')) {
+    if (isOneOfInputObjectType(this._schema.getType(node.name as unknown as string))) {
       return this.getInputObjectOneOfDeclarationBlock(node).string;
     }
 
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -466,7 +466,7 @@ export class BaseTypesVisitor<
   InputObjectTypeDefinition(node: InputObjectTypeDefinitionNode): string {
     if (this.config.onlyEnums) return '';
 
-    // Why the heck is directive.name a string and not { value: string } at runtime ?!
+    // Why the heck is node.name a string and not { value: string } at runtime ?!
     if (isOneOfInputObjectType(this._schema.getType(node.name as unknown as string))) {
       return this.getInputObjectOneOfDeclarationBlock(node).string;
     }
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -453,8 +453,22 @@ export class BaseTypesVisitor<
       .withBlock(node.fields.join('\n'));
   }
 
+  getInputObjectOneOfDeclarationBlock(node: InputObjectTypeDefinitionNode): DeclarationBlock {
+    return new DeclarationBlock(this._declarationBlockConfig)
+      .export()
+      .asKind(this._parsedConfig.declarationKind.input)
+      .withName(this.convertName(node))
+      .withComment(node.description as any as string)
+      .withContent(`\n` + node.fields.join('\n  |'));
+  }
+
   InputObjectTypeDefinition(node: InputObjectTypeDefinitionNode): string {
     if (this.config.onlyEnums) return '';
+
+    // Why the heck is directive.name a string and not { value: string } at runtime ?!
+    if (node.directives?.some(directive => (directive.name as any) === 'oneOf')) {
+      return this.getInputObjectOneOfDeclarationBlock(node).string;
+    }
 
     return this.getInputObjectDeclarationBlock(node).string;
   }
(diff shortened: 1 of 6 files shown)


packages/plugins/typescript/typescript/tests/typescript.spec.ts
```diff
@@ -2536,6 +2536,79 @@ describe('TypeScript', () => {
...
+          }
+        `.concat(oneOfDirectiveDefinition)
+        );
+
+        const result = await plugin(schema, [], {}, { outputFile: '' });
+
+        expect(result.content).toBeSimilarStringTo(`
+          export type Input =
+            { int: Scalars['Int']; }
+            | { boolean: Scalars['Boolean']; };
```
benjie, line 2608: This is accurate, but I think it'll be uncomfortable to work with, because you'll have to do something like `if ('int' in foo)` rather than `if (foo.int)`.

Maybe something like:

```
          export type Input =
            | { int: Scalars['Int']; boolean?: never }
            | { boolean: Scalars['Boolean']; int?: never };
```

https://www.typescriptlang.org/play?#code/C4TwDgpgBAygxgQwDYIE4GcoF4oG8BQURUAkgHbABcUZArgLYBGEqA3IcQEID23SECMtUa9+g9gF98+CAA8w3VMCihIpMmFrKsHAD54oASwrV4yNOgDaAcnLBrAXVZQpRfbigi+AobEQoMGx5vQUdnCXYZeUVlVWhyTWAAJmw9A2MqP3NA2wowz1EfAH5qMggANxYXNI8vMV8zAKtrYPr8jJKaCqqI6ThuMnRlBGoErWwaWiQkKARMQRB2fsHlRlGNLRScOmnZ+bJF6UMAMygACgQAOgyASjwOImX0b0ukbgBzC+uKG+coAHp-lBOABBAAi+Ck+BO52sGWsRjIszuBGIUCeLzenyut0kR1OZ0Y32AKIe6IGz34rw+hOJv0h+Nh8MRnlJaIxVKxtNxAKBoIhEiAA

Take this suggestion with a pinch of salt; I don't know what the nicest TypeScript would be - it's a tradeoff either way

n1ru4l: @benjie Yeah I think this is definitely the better solution when dealing with it in your resolver function! So far I only thought of it from a client perspective.

n1ru4l: @benjie https://github.com/dotansimha/graphql-code-generator/pull/7886/commits/fe6369a6ea945b79c273b2f0fc5c72e15ba2ff8a

benjie:
oneOf requires that all fields are nullable in the schema, and promises/requires that exactly one will be non-null. You have tests for non-null fields on a oneOf; that’s expressly forbidden in the spec.

packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -453,9 +454,23 @@ export class BaseTypesVisitor<
...
+      .asKind(this._parsedConfig.declarationKind.input)
+      .withName(this.convertName(node))
+      .withComment(node.description as any as string)
+      .withContent(`\n` + node.fields.join('\n  |'));
+  }
+
   InputObjectTypeDefinition(node: InputObjectTypeDefinitionNode): string {
     if (this.config.onlyEnums) return '';
 
+    // Why the heck is directive.name a string and not { value: string } at runtime ?!
```
n1ru4l, line 469:
```suggestion
    // Why the heck is node.name a string and not { value: string } at runtime ?!
```

packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -495,3 +497,24 @@ function clearOptional(str: string): string {
...
+
+  if (isOneOfType !== undefined) {
+    return isOneOfType;
+  }
+
+  isOneOfType =
+    isInputObjectType(namedType) &&
+    ((namedType as unknown as Record<'isOneOf', boolean | undefined>).isOneOf ||
+      namedType.astNode?.directives?.some(d => d.name.value === 'oneOf'));
+
```
n1ru4l, line 518:
```suggestion
isOneOfTypeCache.set(namedType, isOneOfType);
```

packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -495,3 +497,26 @@ function clearOptional(str: string): string {
...
+  }
+  let isOneOfType = isOneOfTypeCache.get(namedType);
+
+  if (isOneOfType !== undefined) {
+    return isOneOfType;
+  }
+
+  isOneOfType =
+    isInputObjectType(namedType) &&
+    ((namedType as unknown as Record<'isOneOf', boolean | undefined>).isOneOf ||
```
n1ru4l, line 516: In the GraphQL version that will ship stable `oneOf` the input object type will actually have a boolean `isOneOf` boolean value.

packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -495,3 +497,26 @@ function clearOptional(str: string): string {
...
+      namedType.astNode?.directives?.some(d => d.name.value === 'oneOf'));
```
n1ru4l, line 517: as a fallback we are looking for the directive instead

packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -495,3 +497,26 @@ function clearOptional(str: string): string {
 function stripTrailingSpaces(str: string): string {
   return str.replace(/ +\n/g, '\n');
 }
+
+const isOneOfTypeCache = new WeakMap<GraphQLNamedType, boolean>();
```
n1ru4l, line 501: Dumb perf optimization for iterating over directives array

dotansimha (approved):

purkhusid:
@n1ru4l @dotansimha  Can this change be published in a prerelease?

ardatan:
@purkhusid You can use the versions here;
https://github.com/dotansimha/graphql-code-generator/pull/7886#issuecomment-1137805757

purkhusid:
@ardatan Ah, thought some commits were missing from that release but it seems like everything is there. Thanks!

levrik:
Sadly this is breaking for me and blocking me from upgrading in general, see https://github.com/dotansimha/graphql-code-generator/issues/7938

Issue #7898, opened 2022-06-01
Enums with falsy values are generated incorrectly when using enumsAsTypes


georgehb:
### Issue workflow progress

_Progress of the issue based on the [Contributor Workflow](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow)_

- [x] 1. The issue provides a reproduction available on [CodeSandbox](https://codesandbox.io/s/intelligent-dubinsky-5rerq2)
- [x] 2. A failing test has been provided
- [x] 3. A local solution has been provided
- [x] 4. A pull request is pending review (#7900)

---

**Describe the bug**

When generating types programmatically for typescript and enumsAsTypes is set, If an enum type has an item which maps to a falsy value (ie 0) then the generated type uses the enum value name instead of the value 

**To Reproduce**

1. My GraphQL schema:

```javascript
  const stageType = new GraphQLEnumType({
    name: "Stage",
    values: {
      INITIALIZED: {value: 0},
      STARTED: {value: 1},
      FINISHED: {value: 2},
    }
  });

  const queryType = new GraphQLObjectType({
    name: "Query",
    fields: {
      getStage: {
        type: new GraphQLNonNull(stageType),
      }
    }
  });

  const schema = new GraphQLSchema({
    query: queryType
  });
```

2. My GraphQL operations:

Not used

3. Generated programmatically using:

```javascript
  const output = await codegen({
    schemaAst: schema,
    pluginMap: {
      typescript: typescriptPlugin
    },
    plugins: [
      {
        typescript: {
          enumsAsTypes: true,
        },
      },
    ],
  });
```

**Expected behavior**

The generated type for the enum should be:
```typescript
export type Stage =
  | 0
  | 1
  | 2;
```

instead of
```typescript
export type Stage =
  | 'INITIALIZED'
  | 1
  | 2;
```

**Environment:**

- OS: Linux
- NodeJS: v16.15.0
```
    "@graphql-codegen/core": "2.5.1",
    "@graphql-codegen/typescript": "2.4.11",
    "graphql": "16.5.0"
```

saihaj:
released https://github.com/dotansimha/graphql-code-generator/releases/tag/%40graphql-codegen%2Fvisitor-plugin-common%402.10.0

PR #7900, opened 2022-06-01
fix: ensure falsy enum values are still mapped

georgehb:
## Description

Enums with falsy values were getting generated incorrectly when using enumsAsTypes. The cause was truthyness check on the value before creating the value map.

I've removed the truthy check so instead it only checks if `value !== name`.

I considered adding a `typeof value !== "undefined"` but as far as I'm aware, value is never undefined. For enums without a value map, then the value is the same as the name.

Related #7898

## Type of change

[CodeSandbox](https://codesandbox.io/s/intelligent-dubinsky-5rerq2) demonstrating problem

## How Has This Been Tested?

A test has been added in this PR

**Test Environment**:

- OS: Linux
- NodeJS: v16.15.0
```
    "@graphql-codegen/core": "2.5.1",
    "@graphql-codegen/typescript": "2.4.11",
    "graphql": "16.5.0"
```

## Checklist:

### DIFF
--- a/.changeset/spotty-waves-whisper.md
+++ b/.changeset/spotty-waves-whisper.md
@@ -1,0 +1,6 @@
+---
+"@graphql-codegen/typescript": patch
+"@graphql-codegen/visitor-plugin-common": patch
+---
+
+Ensure falsy enum values are still mapped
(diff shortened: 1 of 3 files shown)


saihaj (approved):
Thanks!

PR #7974, opened 2022-06-26
New issue template

Urigo:
Moving to a new issue creation form


saihaj (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,87 @@
...
+  - type: input
+    id: link
+    attributes:
+      label: Your Example Website or App
+      description: |
+        Which website or app were you using when the bug happened?
+        Note:
+        - Your bug will may get fixed much faster if we can run your code and it doesn't have dependencies other than the `@graphql-codegen/*` npm related packages / dependencies.
+        - To create a shareable code example you can use [GitHub](https://github.com/dotansimha/graphql-code-generator-issue-sandbox-template), [Stackblitz](https://stackblitz.com/github/dotansimha/graphql-code-generator-issue-sandbox-template) or [CodeSandbox](https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template). Please no localhost URLs.
+        - Please read these tips for providing a minimal example: https://stackoverflow.com/help/mcve.
```
dotansimha, line 37: `https://stackoverflow.com/help/mcve` link is broken

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,87 @@
...
+        to the following link: https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/attaching-files
+      placeholder: |
+        You can drag your video or image files inside of this editor ↓
+  - type: textarea
+    id: platform
+    attributes:
+      label: Platform
+      value: |
+        - OS: [e.g. macOS, Windows, Linux]
+        - NodeJS: [e.g. 18.5.0]
```
dotansimha, line 79: in codegen it also worth mentioning the `graphql` version they are using

dotansimha, line 79: Also worth mentioning here the codegen config file, probably this should be a separate input for the YAML config file? (it can be optional i guess)

Urigo: Added, I've made it mandatory as I thought we keep asking people that all the time :)

Urigo: I can change it back if you prefer

dotansimha: I think optional is better in this case, because sometimes people don't have a config file and they have issues with packages/releases/cli and things around the config file.

Urigo: done

PR #7978, opened 2022-06-27
feat: generate TypeScript modules compatible exports map

n1ru4l:
- Upgrades to the latest bob which supports generating a correct `exports` map for TypeScript with `types`.
- Change all relative imports to `.js` (required by ESM and TypeScript esm)
- Add import integrity checks using `bob check` (in favor of the old `test-esm.js` script)
- Introduces the `@graphql-codegen/c-sharp-common` for shared c-sharp-x code (the way it worked before causes issues after the code is no longer bundled into a single file)

_____

After https://github.com/kamilkisiela/bob/pull/87 is merged we can upgrade to bob `3.0.0` "stable" and merge this.


saihaj (approved):
do we need to add a `patch` changeset for all packages so everything gets a release which is TS module compatible?

Issue #7990, opened 2022-06-28
Security vulnerability with `latest-version` in `graphql-codegen-cli`


kevcube:
### Describe the bug

<img width="918" alt="image" src="https://user-images.githubusercontent.com/716806/176322242-69f59bdc-101b-4470-8d90-bab40517803a.png">

verify with `yarn audit`

latest-version requires package-json which requires got.

This PR upgrades latest-version to 6 because 6 still supports node 12.

If node 12 support can be dropped, then upgrade to latest-version 7.

### Steps to Reproduce the Bug or Issue

```sh
git checkout main
yarn audit
```

### Platform

- `@graphql-codegen/cli` version: 2.6.2

saihaj:
should be fixed in https://github.com/dotansimha/graphql-code-generator/releases/tag/%40graphql-codegen/cli%402.6.3

dotansimha:
Fixed in `@graphql-codegen/cli@2.6.4`. Keeping open until we'll figure out a more stable way to upgrade and make sure we do not have regressions.

saihaj:
> Fixed in `@graphql-codegen/cli@2.6.4`. Keeping open until we'll figure out a more stable way to upgrade and make sure we do not have regressions.

https://github.com/dotansimha/graphql-code-generator/issues/8008

PR #7991, opened 2022-06-28
Upgrade latest version

kevcube:
## Description

fix security alert regarding package `got`

related: #7990 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

ran `yarn test`

**Test Environment**:

- OS: macOS
- NodeJS: 18 (should retain compatibility back to node 12, however)

## Checklist:

- [x] I have followed the [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #7998, opened 2022-06-30
Revert recent `latest-version` upgrade due to ESM issues

charlypoly:
Reverts dotansimha/graphql-code-generator#7991

Upgrading `latest-version` to `6.0.0` broke the CLI, let's find another way to fix https://github.com/dotansimha/graphql-code-generator/issues/7990

### DIFF
--- a/.changeset/afraid-clocks-ring.md
+++ b/.changeset/afraid-clocks-ring.md
@@ -1,0 +1,5 @@
+---
+"@graphql-codegen/cli": patch
+---
+
+Revert "Upgrade latest version"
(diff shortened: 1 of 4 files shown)


emilio-martinez:
@charlypoly can you share more details as to what broke? Happy to look into contributing, if nobody is doing so already, but I'd like to know what specifically broke so I can cover it – my understanding is that node 12.20 supports ESM?

charlypoly:
@emilio-martinez 

> @charlypoly can you share more details as to what broke? Happy to look into contributing, if nobody is doing so already, but I'd like to know what specifically broke so I can cover it – my understanding is that node 12.20 supports ESM?


This PR is a hotfix for the upgrade of `latest-version` that broke our CLI (due to an ESM compatibility issue)
The explanation of the decision to remove it from our dependencies is detailed here: https://github.com/dotansimha/graphql-code-generator/issues/8008

PR #8013, opened 2022-07-01
ci: run on 12,14,16,18 node versions and run tests once linting is success

saihaj:
Just like [`graphql-yoga`](https://github.com/dotansimha/graphql-yoga) we should probably ensure this on all the Node versions we support. 

To avoid running too many CI tasks ensure that linting and prettier passes. Since if they fail we wouldn't merge PR anyways so no point running tests since user will have to push again.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,6 +108,11 @@ jobs:
   test:
     name: Unit Test on Node ${{matrix.node_version}} (${{matrix.os}}) and GraphQL v${{matrix.graphql_version}}
     runs-on: ubuntu-latest
+    needs:
+      - lint
+      - prettier-check
+      - dev-tests
+      - esm
     strategy:
       matrix:
         os: [ubuntu-latest] # remove windows to speed up the tests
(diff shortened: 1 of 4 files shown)


B2o5T (approved):

Urigo:
could/should we share parts of `.github/workflows/main.yml` between our libraries?

saihaj:
> could/should we share parts of `.github/workflows/main.yml` between our libraries?

Yeah we should create composite actions I think with new bob we will be able to simplify more scripts cc @n1ru4l

ardatan (approved):

.github/workflows/main.yml
```diff
@@ -108,10 +111,15 @@ jobs:
...
+    needs:
+      - lint
+      - prettier-check
+      - dev-tests
+      - esm
     strategy:
       matrix:
         os: [ubuntu-latest] # remove windows to speed up the tests
-        node_version: [12, 17]
+        node_version: [12, 14, 16, 18]
```
B2o5T, line 122: I guess you can already remove 12
```suggestion
        node_version: [14, 16, 18]
```

B2o5T (approved):

.github/workflows/main.yml
saihaj: we still support `v12` I think?

Issue #8008, opened 2022-07-01
remove latest-version dependency


n1ru4l:
Unfortunatly, `latest-version` is an ESM only hipster library and the current version has a security vulnerability. The solution is to replace this library and the dependencies (lol https://npm.anvaka.com/#/view/2d/latest-version) with this one-liner 😅

```ts
const getLatestVersion = () => fetch("https://unpkg.com/latest-version/package.json").then(res => res.json()).then(pkg => pkg.version)
```

PR #8035, opened 2022-07-05
feat: remove latest-version dependency

n1ru4l:
Closes https://github.com/dotansimha/graphql-code-generator/issues/8008
Closes https://github.com/dotansimha/graphql-code-generator/issues/7990


saihaj (approved):

ardatan (approved):

packages/graphql-codegen-cli/src/utils/get-latest-version.ts
```diff
@@ -0,0 +1,12 @@
+import { fetch } from 'cross-undici-fetch';
+
+/**
+ * Fetches the version directly from the registry instead of depending on
+ * an ESM only module as latest-version does.
+ * @param packageName
+ */
+export async function getLatestVersion(packageName: string): Promise<string> {
+  return fetch(`https://unpkg.com/${packageName}/package.json`)
```
ardatan, line 9: Do we need an error handling here?
You know sometimes `fetch` gives random errors

n1ru4l: good idea I guess in case it errors we should log something and then skip it. We can do it in a followup PR, there wasnt error handling before. I will create an issue

ardatan: Makes sense. I didn't think it is a blocker for this PR already. It just came to my mind to make Codegen throw random errors :D

Issue #8075, opened 2022-07-11
`@graphql-codegen/cli` is missing `tslib` dependency


mhassan1:
### Describe the bug

`@graphql-codegen/cli` is missing an explicit `tslib` runtime dependency. Running `yarn graphql-codegen ...` in a strict environment like Yarn Plug'n'Play leads to:
```txt
Error: @graphql-codegen/cli tried to access tslib, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```
This is an issue in `@graphql-codegen/cli@2.8.0`, but not `@graphql-codegen/cli@2.7.0`.

### Your Example Website or App

n/a

### Steps to Reproduce the Bug or Issue

1. `yarn set version berry`
2. `yarn add --dev @graphql-codegen/cli@2.8.0`
3. `yarn graphql-codegen ...`
4. `Error: @graphql-codegen/cli tried to access tslib, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.`

### Expected behavior

Successful command execution

### Platform

- OS: MacOS
- NodeJS: 16.14.2
- `graphql` version: 16.5.0
- `@graphql-codegen/*` version(s): 2.8.0

### Codegen Config File

```yml
schema: "./schema.gql"
generates:
  "./types.ts":
    plugins: ["typescript"]
    config:
      enumsAsTypes: true
```

### Additional context

I am using the following workaround:

```yml
# .yarnrc.yml

packageExtensions:
  "@graphql-codegen/cli@^2":
    dependencies:
      tslib: ~2.4.0
```

saihaj:
hey @mhassan1 you can try [alpha release](https://github.com/dotansimha/graphql-code-generator/pull/8078#issuecomment-1180744376) and provide feedback

mhassan1:
Works for me!

saihaj:
https://github.com/dotansimha/graphql-code-generator/releases/tag/%40graphql-codegen%2Fcli%402.8.1

PR #8078, opened 2022-07-11
chore: add tslib as dep

saihaj:
## Description

`tslib` is used in generated code. In the monorepo it doesn't crash since it uses the one we have in `monorepo` but we should provide this part of package.

Related #8075 

## Further comments

Or should we just put this as a peer-dep 🤔


n1ru4l (changes requested):
We should not pin dependencies

n1ru4l (approved):

PR #8077, opened 2022-07-11
feat: add `emitLegacyCommonJSImports` flag

saihaj:
## Description

See changeset

Related #8067 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Screenshots/Sandbox (if appropriate/relevant):

Adding links to sandbox or providing screenshots can help us understand more about this PR and take action on it as appropriate

## How Has This Been Tested?

Updated `dev-tests`

## Checklist:

- [X] I have followed the [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

## Further comments

Couldn't figure out how to make it work for `oclif` plugin

PR #8094, opened 2022-07-14
Fix .js extension when agumentedModuleName is empty

kamilkisiela:
It was missing in 525ad580b


n1ru4l (changes requested):
@kamil You need to regenerate the examples using `generate:examples` and commit the diff.

n1ru4l:
@saihaj I think we can close this, as it was solved via https://github.com/dotansimha/graphql-code-generator/pull/8077?

PR #8125, opened 2022-07-21
allow mappers config overriding

cometkim:
## Description

When workspaces include multiple GraphQL projects, and graphs share the same core model. In this case, I would like to write the core model mappers only once in the root config and additional mappers per plugin, but that is not currently allowed.

## Type of change

## How Has This Been Tested?

I skipped it because there are no existing tests for config. Please tell me if it requires

## Checklist:

### DIFF
--- a/.changeset/angry-apples-protect.md
+++ b/.changeset/angry-apples-protect.md
@@ -7,8 +7,8 @@ Introduce `mergeMappers` config flags will allows to reuse mappers across genera
 ```yml
 schema: "schema.graphql"
 documents: src/*.ts
+mergeMappers: true
 config:
-    mergeMappers: true
     mappers:
       ID: IDType
 generates:
--- a/.changeset/angry-apples-protect.md
+++ b/.changeset/angry-apples-protect.md
@@ -1,0 +1,29 @@
+---
+'@graphql-codegen/cli': minor
+---
+
+Introduce `mergeMappers` config flags will allows to reuse mappers across generated files:
+
+```yml
+schema: "schema.graphql"
+documents: src/*.ts
+config:
+    mergeMappers: true
+    mappers:
+      ID: IDType
+generates:
+  resolvers-types-1.ts:
+    plugins:
+      - typescript
+      - typescript-resolvers
+    config:
+      mappers:
+        String: StringType
+  resolvers-types-2.ts:
+    plugins:
+      - typescript
+      - typescript-resolvers
+    config:
+      mappers:
+        String: StringType
+```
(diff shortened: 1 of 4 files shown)


charlypoly (changes requested):
Hi @cometkim!

Thank you for your proposal; it definitely would help codegen work better on large-scale or monorepo projects; we are definitely looking forward to having this new behavior merged! ⚡ 

Your proposed changes won't allow set `mappers` at the root level of the config since you're getting the current `mappers` values from the generates section.
To get `mappers` to be defined and fetched from the root of the config, you'll need to introduce changes related to the [`Config`](https://github.com/dotansimha/graphql-code-generator/blob/2cbcbb37101708705ab94cec747997c8063f698d/packages/utils/plugins-helpers/src/types.ts#L345) type.

Let me know if we can help!

cometkim:
@charlypoly I expect that the JSONSchema for editor support is also need to be changed. Is it automatically generated? Or is it managed elsewhere?

charlypoly:
Hi @cometkim,

What do you mean by "JSONSchema for editor support"?

cometkim:
I meant this https://www.graphql-code-generator.com/config.schema.json

charlypoly:
> I meant this https://www.graphql-code-generator.com/config.schema.json

This file is automatically generated

charlypoly:
@cometkim, I've fixed your PR to support the following configuration:

```yml
schema: "schema.graphql"
documents: src/*.ts
mergeMappers: true
config:
    mappers:
      ID: IDType
generates:
  resolvers-types-1.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      mappers:
        String: StringType
  resolvers-types-2.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      mappers:
        String: StringType
```

Example setup is available here: https://github.com/charlypoly/codegen-repros/tree/master/merged-config

You can see that I added a `mergeMappers` flag (default: `false`) since introducing this change will be breaking for most configurations.

Let me know what you think 👀

charlypoly:
@cometkim, actually, this new options is not necessary since you can achieve same result using YAML references: 
[codegen.yml](https://github.com/charlypoly/codegen-repros/blob/master/reusable-mappers/codegen.yml)

```yaml
schema: "schema.graphql"
documents: src/*.ts

shared_mappers: &shared_mappers
  ID: IDType

generates:
  resolvers-types-1.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      mappers:
        <<: *shared_mappers
        String: StringType
  resolvers-types-2.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      mappers:
        <<: *shared_mappers
        String: StringType
```


I'm closing this PR and will update the `typescript-resolvers` documentation with the above example

charlypoly:
https://github.com/dotansimha/graphql-code-generator/pull/8177

Issue #6488, opened 2021-08-16
hooks hide logs of their tasks


jantimon:
Hooks are really powerful - unfortunately right now all logs are hidden:

https://github.com/dotansimha/graphql-code-generator/blob/71b9a8925d9df494adf8496d16f660308499461a/packages/graphql-codegen-cli/src/hooks.ts#L47-L84

Would it be possible to log them to console? For example in VERBOSE mode?
Without those logs its really hard to find out if those hooks are setup correctly.

charlypoly:
@jantimon, hooks scripts logs are now forwarded to console when codegen is run with `DEBUG=1`, from `@graphql-codegen/cli@2.10.0` 📦

PR #8142, opened 2022-07-26
feat(hooks): forward hooks logs to debug logs

charlypoly:
https://github.com/dotansimha/graphql-code-generator/issues/6488

@saihaj, I also restored the debug logs that were no longer printed with `DEBUG=1`


saihaj (changes requested):
let's add a changeset

Issue #4725, opened 2020-09-09
Add a --dry-run option


justinwaite:
**Is your feature request related to a problem? Please describe.**

The problem is that a developer could forget to run the code generator, or forget to keep it in watch mode, and then commit operations to master that don't have associated types with them. Then another developer comes along and wonders why their PR has a bunch of types generated that are unrelated to the work they were doing. 

**Describe the solution you'd like**

This _could_ be handled using lint-staged if there was a way to possibly do a dry-run and a non-zero exit if there are types that need to be generated. Then in a pre-commit hook, the developer would be warned that there are types that must be generated. This could also be used in a CI/CD environment and fail if necessary.

**Describe alternatives you've considered**

I tried using lint-staged and just automatically add the generated files with git, but that could lead to unforeseen issues, and quite frankly I couldn't get it to work.

**Additional context**

dotansimha:
Thank you for this suggestion @justinwaite . 
In most setups, I think it's better to avoid committing generated files to git. But regardless of that, I think adding this kind of flag could help.

nandorojo:
I'd love to `--dry-run` so from my CI, just to confirm that the app's schema is valid, without generating any actual files.

charlypoly:
Duplicate of https://github.com/dotansimha/graphql-code-generator/issues/2872

Merging the issues together

charlypoly:
The dry-run mode is now available via the `--check` CLI flag, from `@graphql-codegen/cli@2.11.0` 📦

PR #8149, opened 2022-07-27
feat(cli): add a dry-run mode with `--check` cli flag

### LINKED ISSUE #2872: Add "--check" flag to detect stale code
I need some way to check whether the generated code is up-to-date with the source code, so that I can check before every commit with husky and during CI.

There could be a `--check` flag in the CLI that would generate all the code in memory, compare to what's in my working directory and exit with an error if they are different.

None.

**Additional context**
It would be similar to how `prettier --check` or `prettier --list-different` works.

charlypoly:
Closes #2872 #4725

### DIFF
--- a/.changeset/real-toes-dance.md
+++ b/.changeset/real-toes-dance.md
@@ -1,0 +1,6 @@
+---
+"@graphql-codegen/cli": patch
+"@graphql-codegen/plugin-helpers": patch
+---
+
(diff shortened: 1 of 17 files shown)


charlypoly:
@saihaj tested with a local repro project, here is the output before and after changes to the schema:

image

saihaj (approved):

Issue #6174, opened 2021-06-23
& in path causes .graphql files not to be found


WillSquire:
**Describe the bug**
If an `&` character is the a parent directory that codegen is running in it will produce the error `Unable to find any GraphQL type definitions for the following pointers using schema file`.

**To Reproduce**
Put the whole project in a folder containing `&` character. I.e. `something & something` directory.

<!--
#############
#############
🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 🛑 STOP! 

PLEASE READ THE FOLLOWING: 
You can use this template to create a live sandbox of the issue: 
https://codesandbox.io/s/github/dotansimha/graphql-code-generator-issue-sandbox-template
Make sure to fork this template, and run `yarn generate` in the terminal. 
Please make sure the Codegen and plugins version under `package.json` matches yours. 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~ ISSUES WITH A LIVE REPRODUCTION WILL BE ANSWERED QUICKLY ~~~~~~~
~~~~~~ ISSUES WITHOUT REPRODUCTION MIGHT GET CLOSED WITHOUT RESPONSE ~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#############
#############
-->

My `codegen.yml` config file:

```yml
overwrite: true
schema: "../schema/schema.graphql"
documents: "src/**/*.graphql"
generates:
  src/middleware/graphql.tsx:
    plugins:
      - "typescript"
      - "typescript-operations"
      - "typescript-react-apollo"
```

**Expected behavior**
Escape the `&` character properly in whatever part of the script is causing the issue...

**Environment:**

[...]

PR #8156, opened 2022-07-28
chore(deps): update `@graphql-tools/graphql-file-loader` to `7.5.0`

charlypoly:
Fixes https://github.com/dotansimha/graphql-code-generator/issues/6174


charlypoly:
@saihaj, thanks for the `graphql-tools` fix!

I've tested locally with and without the `@graphql-tools/graphql-file-loader` bump and it's now working as expected ✅

PR #8204, opened 2022-08-09
feat(generic-sdk): respect live queries & use AsyncIterable if possible

ardatan:
- Respect GraphQL Live Queries like Subscriptions and use the stream return types (`AsyncIterable` or `Observable`).
- Previously if there was no `usingObservableFrom` set in the configuration, the plugin was using `Promise` as subscriptions' return type, and this is wrong. Now it uses `AsyncIterable` in this case.

### DIFF
--- a/.changeset/twelve-insects-invite.md
+++ b/.changeset/twelve-insects-invite.md
@@ -1,5 +1,5 @@
 ---
-'@graphql-codegen/typescript-generic-sdk': minor
+'@graphql-codegen/typescript-generic-sdk': major
 ---
 
 - Respect GraphQL Live Queries like Subscriptions and use the stream return types (`AsyncIterable` or `Observable`).
--- a/.changeset/twelve-insects-invite.md
+++ b/.changeset/twelve-insects-invite.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/typescript-generic-sdk': minor
+---
+
+- Respect GraphQL Live Queries like Subscriptions and use the stream return types (`AsyncIterable` or `Observable`).
+- Previously if there was no `usingObservableFrom` set in the configuration, the plugin was using `Promise` as subscriptions' return type, and this is wrong. Now it uses `AsyncIterable` in this case.
--- a/packages/plugins/typescript/generic-sdk/src/visitor.ts
+++ b/packages/plugins/typescript/generic-sdk/src/visitor.ts
@@ -15,6 +15,19 @@ export interface GenericSdkPluginConfig extends ClientSideBasePluginConfig {
   rawRequest: boolean;
 }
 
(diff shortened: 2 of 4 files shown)


packages/plugins/typescript/generic-sdk/src/visitor.ts
```diff
@@ -91,9 +104,7 @@ export class GenericSdkVisitor extends ClientSideBaseVisitor<RawGenericSdkPlugin
 
     const documentNodeType = this.config.documentMode === DocumentMode.string ? 'string' : 'DocumentNode';
     const resultData = this.config.rawRequest ? 'ExecutionResult<R, E>' : 'R';
-    const returnType = usingObservable
-      ? `Promise<${resultData}> & Observable<${resultData}>`
-      : `Promise<${resultData}>`;
+    const returnType = `Promise<${resultData}> | ${usingObservable ? 'Observable' : 'AsyncIterable'}<${resultData}>`;
```
charlypoly, line 107: why changing `Promise & ...` to `Promise | ...` ?

packages/plugins/typescript/generic-sdk/src/visitor.ts
```diff
@@ -15,6 +15,19 @@ export interface GenericSdkPluginConfig extends ClientSideBasePluginConfig {
   rawRequest: boolean;
 }
 
+function isStreamOperation(operationAST: OperationDefinitionNode) {
+  if (operationAST.operation === 'subscription') {
+    return true;
+  }
+  if (
+    operationAST.operation === 'query' &&
+    operationAST.directives?.some(directiveNode => directiveNode.name.value === 'live')
```
charlypoly, line 24: it should also take `@defer` into account, right?

packages/plugins/typescript/generic-sdk/src/visitor.ts
ardatan: I think `&` is wrong because result can be `Promise` or stream. `&` makes it to have both signatures which is wrong actually, right?

packages/plugins/typescript/generic-sdk/src/visitor.ts
charlypoly: Discussed offline.

.changeset/twelve-insects-invite.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphql-codegen/typescript-generic-sdk': minor
```
charlypoly, line 2:
```suggestion
'@graphql-codegen/typescript-generic-sdk': major
```

As discussed offline, let's make if a major 📦

charlypoly (approved):
@ardatan, feel free to merge!

Issue #8022, opened 2022-07-04
Change plugins to attach `.js` extension to emitted code


n1ru4l:
We need to attach the `.js` extension to all relative imports within the code files emitted by GraphQL Code Generator  for TypeScript module/ESM compliance.

Issue #8065, opened 2022-07-08
near-operation-file + typescript-react-query = Failed to resolve .js import


pix2D:
### Describe the bug

https://github.com/dotansimha/graphql-code-generator/pull/8034 seems to have changed the way imports work by adding .js everywhere.

I use preset `near-operation-file` together with `typescript-operations` and `typescript-react-query` which now results in TS files with fragment imports like this:

```
import { BlaFragmentDoc } from './Bla.js'
```

In my Next.js app this now fails to compile with:

```
Module not found: Can't resolve './Bla.js'
```

Sorry I don't have time to make a repro repo. I provided my codegen.yml which should result in the same issue.

### Your Example Website or App

Sorry I don't have time to make a repro repo.

### Steps to Reproduce the Bug or Issue

1. Generate code using provided config
2. Import into Next.js app
3. Run dev server

### Expected behavior

There should be no .js in the fragment filename.

### Platform

- OS: macOS
- NodeJS: v16.15.0

    "graphql": "^16.5.0",
   "@graphql-codegen/add": "^3.2.0",
    "@graphql-codegen/cli": "^2.8.0",
    "@graphql-codegen/near-operation-file-preset": "^2.3.0",
    "@graphql-codegen/typescript": "^2.7.0",
    "@graphql-codegen/typescript-operations": "^2.5.0",
    "@graphql-codegen/typescript-react-query": "^3.6.0",

### Codegen Config File

```
overwrite: true
schema: 'schema.graphql'
hooks:
  afterAllFileWrite:
    - prettier --loglevel silent --write "**/*.{ts,tsx}"
documents:
  - 'src/**/*.graphql'
generates:
  src/types.ts:
    config:
      onlyOperationTypes: true
      useTypeImports: true
    plugins:
      - add:
          content: >
            /* eslint-disable */

            /**
             * WARNING: THIS FILE IS AUTO-GENERATED, DO NOT EDIT IT DIRECTLY!
             */

            export type Nullable<T> = {
              [P in keyof T]: T[P] | null
            }
      - typescript
  src:
    config:
      dedupeFragments: true
      exposeFetcher: true
      exposeQueryKeys: true
      fetcher: graphql-request
      preResolveTypes: true
      pureMagicComment: true
      useTypeImports: true
    preset: near-operation-file
    presetConfig:
      baseTypesPath: types.ts
      extension: .ts
    plugins:
      - add:
          content: >
            /* eslint-disable */

            /**
[...]

PR #8034, opened 2022-07-05
feat: attach .js extension to all relative/absolute file imports/exports

n1ru4l:
Closes https://github.com/dotansimha/graphql-code-generator/issues/8022
Closes https://github.com/dotansimha/graphql-code-generator/pull/8034

PR #8235, opened 2022-08-14
 graphql-request: Specify explicit extension when importing dom types

just-boris:
This is needed to support "moduleResolution": "nodenext" in typescript

## Description

Related #8022

It looks like the related issue attempted to add `.js` extensions to all imports, but this one got missing

## Type of change

n/a

## How Has This Been Tested?

- [x] Modified generated code locally, it passes typescript compilation

**Test Environment**:

n/a

## Checklist:

## Further comments

n/a

### DIFF
--- a/packages/plugins/typescript/react-query/src/fetcher-graphql-request.ts
+++ b/packages/plugins/typescript/react-query/src/fetcher-graphql-request.ts
@@ -66,8 +66,9 @@ function fetcher<TData, TVariables>(client: GraphQLClient, query: string, variab
     const variables = generateQueryVariablesSignature(hasRequiredVariables, operationVariablesTypes);
 
     const typeImport = this.visitor.config.useTypeImports ? 'import type' : 'import';
+    const fileExtension = this.visitor.config.emitLegacyCommonJSImports ? '' : '.js';
     this.visitor.imports.add(`${typeImport} { GraphQLClient } from 'graphql-request';`);
-    this.visitor.imports.add(`${typeImport} { RequestInit } from 'graphql-request/dist/types.dom';`);
+    this.visitor.imports.add(`${typeImport} { RequestInit } from 'graphql-request/dist/types.dom${fileExtension}';`);
 
     const hookConfig = this.visitor.queryMethodMap;
     this.visitor.reactQueryHookIdentifiersInUse.add(hookConfig.query.hook);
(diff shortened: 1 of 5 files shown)


n1ru4l (changes requested):
Hey there! The file extension should be added conditionally via the `emitLegacyCommonJSImports` flag, which we introduced here: https://github.com/dotansimha/graphql-code-generator/pull/8077

The reason for this is that some tools (such as Next.js) require additional configuration for being able to process the `.js` extension, thus we introduced it behind a flag for now.

See the following issues for a brief history:
- https://github.com/dotansimha/graphql-code-generator/pull/8034
- https://github.com/dotansimha/graphql-code-generator/issues/8065
- https://github.com/dotansimha/graphql-code-generator/issues/8065
- https://github.com/dotansimha/graphql-code-generator/pull/8077

packages/plugins/typescript/react-query/src/fetcher-graphql-request.ts
```diff
@@ -67,7 +67,7 @@ function fetcher<TData, TVariables>(client: GraphQLClient, query: string, variab
 
     const typeImport = this.visitor.config.useTypeImports ? 'import type' : 'import';
     this.visitor.imports.add(`${typeImport} { GraphQLClient } from 'graphql-request';`);
-    this.visitor.imports.add(`${typeImport} { RequestInit } from 'graphql-request/dist/types.dom';`);
+    this.visitor.imports.add(`${typeImport} { RequestInit } from 'graphql-request/dist/types.dom.js';`);
```
charlypoly, line 70: The same applies here, we need to rely on `emitLegacyCommonJSImports` config

packages/plugins/typescript/react-query/src/fetcher-graphql-request.ts
```diff
@@ -134,7 +134,7 @@ function fetcher<TData, TVariables>(client: GraphQLClient, query: string, variab
   ): string {
     const variables = generateQueryVariablesSignature(hasRequiredVariables, operationVariablesTypes);
     const typeImport = this.visitor.config.useTypeImports ? 'import type' : 'import';
-    this.visitor.imports.add(`${typeImport} { RequestInit } from 'graphql-request/dist/types.dom';`);
+    this.visitor.imports.add(`${typeImport} { RequestInit } from 'graphql-request/dist/types.dom.js';`);
```
charlypoly, line 137: The same applies here, we need to rely on `emitLegacyCommonJSImports` config

charlypoly (changes requested):

charlypoly (changes requested):
Please update the plugin's test to pass `emitLegacyCommonJSImports: true` in `config` since it is the expected default behavior when `emitLegacyCommonJSImports` is missing at the `codegen.yml` level (the CLI normalizes `emitLegacyCommonJSImports` to `true` when missing, then pass its value to the plugins).
It will also prevent all those changes on the snapshots.

Thanks!

jcarrus:
Hi everyone. I just ran into this issue also and would like to help move this along. So excited to see a fix already in progress!

> Please update the plugin's test to pass `emitLegacyCommonJSImports: true` in `config` since it is the expected default behavior when `emitLegacyCommonJSImports` is missing at the `codegen.yml` level (the CLI normalizes `emitLegacyCommonJSImports` to `true` when missing, then pass its value to the plugins). It will also prevent all those changes on the snapshots.

@charlypoly, are you recommending to explicitly pass `emitLegacyCommonJSImports: true` to just be clear about what the default value is? Or is there some other reason?

It seems to me that the existing tests should stay the same and a new test should be added with `emitLegacyCommonJSImports: true` to verify that the extension is attached.

I've mocked these proposed changes here: #8334 if that's easier to see.

Let me know how we can move this forward!

charlypoly:
Close in favor of https://github.com/dotansimha/graphql-code-generator/pull/8334

PR #7701, opened 2022-03-28
Feat: Hasura allow list global fragments support

BenoitRanque:
## Description

Added support for globally defined fragments for the hasura-allow-list plugin.
Previously it was required that fragments be defined in the same document as the operations that use them.
Now, with the `global_fragments` config set, fragments will be pulled from all documents in the project.

Because this configuration is off by default, this feature does not break previous behavior and is thus not a breaking change.

Related #7700

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Screenshots/Sandbox (if appropriate/relevant):

Adding links to sandbox or providing screenshots can help us understand more about this PR and take action on it as appropriate

## How Has This Been Tested?

Added tests to test suite for the plugin, testing success case and two error cases.
Also built the plugin and tested on a local project to see the desired output.

**Test Environment**:
- OS: windows
- `@graphql-codegen/cli`: 2.6.0 
- NodeJS: 14.17.6

## Checklist:

- [x] I have followed the [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

Issue #7700, opened 2022-03-28
Hasura Allow list plugin: support fragments defined globally


BenoitRanque:
**Is your feature request related to a problem? Please describe.**

Currently the hasura allow list plugin requires that fragments be defined in the same document as the operations that use them.

**Describe the solution you'd like**

It would be better if instead, fragments were pulled from all documents. This works as long as fragment names are unique throughout the project.

shinnoki:
+1
We use hasura-allow-list plugin but patch it by ourself for this issue.

Note that Hasura's allow lists strictly respect the order of documents, fragments and fields.
I think it would be better if adopt Visitor Pattern, because this plugin is usually used together with other typescript plugins such as typed-document-node, and it's better to be guaranteed mechanically to behave the same way as them.

franciscolourenco:
> This works as long as fragment names are unique throughout the project.

@BenoitRanque to clarify, are you saying that the plugin currently support fragments defined in separate files, as long as each fragment has a unique name?

BenoitRanque:
> > This works as long as fragment names are unique throughout the project.
> 
> @BenoitRanque to clarify, are you saying that the plugin currently support fragments defined in separate files, as long as each fragment has a unique name?
> 

Yes, if the global fragments option is enabled, the plug-in will recursively append fragments found in processed files (usually a target directory) to the generated allow list entry.

This requires that that fragment name be unique across your project as that is how required fragments are identified.

PR #8242, opened 2022-08-16
Feature/hasura allow list global fragments

BenoitRanque:
This is a replacement PR for #7701 where I messed up the git history

## Description

Added support for globally defined fragments for the hasura-allow-list plugin.
Previously it was required that fragments be defined in the same document as the operations that use them.
Now, with the global_fragments config set, fragments will be pulled from all documents in the project.

Because this configuration is off by default, this feature does not break previous behavior and is thus not a breaking change.
Related #7700 

## Type of change

## How Has This Been Tested?

Added tests to test suite for the plugin, testing success case and two error cases.
Also built the plugin and tested on a local project to see the desired output.

**Test Environment**:

- OS: Windows
- `@graphql-codegen/cli`: 2.6.0
- NodeJS: 14.17.6

## Checklist:

### DIFF
--- a/packages/plugins/other/hasura-allow-list/src/config.ts
+++ b/packages/plugins/other/hasura-allow-list/src/config.ts
@@ -11,4 +11,9 @@ export interface HasuraAllowListPluginConfig {
    * The default value _will change_ in the future if/when newer config versions are released.
    */
   config_version?: 2 | 3;
+  /**
+   * @default false
+   * @description Whether to source fragments per-document, or globally. If set, will enforce fragment name uniqueness
+   */
+  global_fragments?: boolean;
 }
--- a/packages/plugins/other/hasura-allow-list/src/config.ts
+++ b/packages/plugins/other/hasura-allow-list/src/config.ts
@@ -15,5 +15,5 @@ export interface HasuraAllowListPluginConfig {
    * @default false
    * @description Whether to source fragments per-document, or globally. If set, will enforce fragment name uniqueness
    */
-  global_fragments?: boolean;
+  globalFragments?: boolean;
 }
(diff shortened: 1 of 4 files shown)


packages/plugins/other/hasura-allow-list/src/config.ts
```diff
@@ -11,4 +11,9 @@ export interface HasuraAllowListPluginConfig {
    * The default value _will change_ in the future if/when newer config versions are released.
    */
   config_version?: 2 | 3;
+  /**
+   * @default false
+   * @description Whether to source fragments per-document, or globally. If set, will enforce fragment name uniqueness
+   */
+  global_fragments?: boolean;
```
charlypoly, line 18: I missed it in the initial PR, but can you use camelCase config properties like all plugins? 🙏🏼 👀 

```suggestion
  globalFragments?: boolean;
```

charlypoly (changes requested):

packages/plugins/other/hasura-allow-list/src/config.ts
BenoitRanque: Absolutely!

BenoitRanque:
Accepted the config change, then realized I'd also need to update the rest of the code to reflect that change. Will commit that in a moment

BenoitRanque:
Will go ahead and change all config properties to camel case. This is a breaking change.

Will add error handling code to detect older config versions and throw a descriptive error to help users make the change.

BenoitRanque:
Checked the website documentation and I don't believe changes are required there, assuming the config reference is pulled from the type definitions which seems to be the case :)

BenoitRanque:
Unsure where to add a changeset. Changeset text:

```
added support for global fragments
changed config options to camelCase for consistency
```

charlypoly (approved):

Issue #8206, opened 2022-08-09
`gql-tag-operations-preset` `babelPlugin` doesn't work with `vite`


PatrykWalach:
### Describe the bug

Using `babelPlugin` from `gql-tag-operations-preset` inside `vite` project throws `_babel_template.default is not a function`

### Your Example Website or App

https://stackblitz.com/edit/vitejs-vite-toyias

### Steps to Reproduce the Bug or Issue

Use the link above or:
1. Create new vite project with react
2. `npm install graphql @graphql-codegen/gql-tag-operations-preset`
3. Add babel plugin
```ts
// vite.config.js
import { defineConfig } from 'vite';
import react from '@vitejs/plugin-react';
import { babelPlugin } from '@graphql-codegen/gql-tag-operations-preset';

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [react({ babel: { plugins: [[babelPlugin, {}]] } })],
});

```
4. Define a gql operation
```ts
// src/App.jsx
const Query = gql(`query AppQuery { hello }`);
```

### Expected behavior

Works out of the box

### Screenshots or Videos

![](https://imgur.com/fcspSUH.jpg)

### Platform

- `graphql` version: 16.5.0
- `@graphql-codegen/gql-tag-operations-preset` version: 1.5.4
- `vite` version: 3.0.4

### Additional context

Possible fix is to use `template.smart` instead of default export in [gql-tag-operations/src/babel.ts#L56](https://github.com/dotansimha/graphql-code-generator/blob/2e727e5b53718d829f6b3f6fd52357c63b897d35/packages/presets/gql-tag-operations/src/babel.ts#L56):

```diff
- const importDeclaration = template(`
+ const importDeclaration = template.smart(`
import { %%importName%% } from %%importPath%%
`);
```

Default export from `@babel/template` is `smart.bind(undefined)` so using `tempalte.smart` shouldn't be a breaking change.

saihaj:
hey @PatrykWalach can you please try this alpha release and give feedback https://github.com/dotansimha/graphql-code-generator/pull/8260#issuecomment-1220876702

PatrykWalach:
It works as expected with the alpha release

PR #8260, opened 2022-08-19
fix: babel plugin for vite

saihaj:
Related #8206 

## Type of change
Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
Tested this change in the repro provided in the issue.

## Further comments

we should figure out integration test strategy for different bundlers 🤔


charlypoly (approved):
⚡

PR #8301, opened 2022-08-30
GraphQL Code Generator v3: `@graphql-codegen/cli` changes

charlypoly:
Introduces support for TypeScript config file and a new preset lifecycle (required for `client-preset`)


n1ru4l (approved):

packages/graphql-codegen-cli/package.json
```diff
@@ -57,6 +57,7 @@
     "chalk": "^4.1.0",
     "chokidar": "^3.5.2",
     "cosmiconfig": "^7.0.0",
+    "cosmiconfig-typescript-loader": "4.0.0",
```
n1ru4l, line 60: I don't think we need to pin it and can sue a range

packages/utils/plugins-helpers/src/types.ts
```diff
@@ -218,6 +218,14 @@ export namespace Types {
   export type NamedPreset = string;
   export type OutputConfig = NamedPlugin | ConfiguredPlugin;
 
+  export type PresetNamesBase =
```
charlypoly, line 221: We will improve this in the future by inferring the preset options from the preset function's typings (or another build-based approach)

n1ru4l:
Should we also do the doc updates for recommending a typescript config as part of this PR?

packages/graphql-codegen-cli/package.json
dotansimha: Agree with @n1ru4l , we can do `^`

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -202,21 +202,29 @@ export async function executeCodegen(input: CodegenContext | Types.Config): Prom
...
+                let outputSpecificDocuments = normalizeInstanceOrArray<Types.OperationDocument>(outputConfig.documents);
+
+                const preset: Types.OutputPreset | null = hasPreset
+                  ? typeof outputConfig.preset === 'string'
+                    ? await getPresetByName(outputConfig.preset, makeDefaultLoader(context.cwd))
+                    : outputConfig.preset
+                  : null;
+
+                if (preset) {
+                  if (preset.prepareDocuments) {
```
dotansimha, line 224: That's the new feature we discussed yesterday, right? where the plugin can modify the `document` section

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -124,6 +131,7 @@ export async function loadCodegenConfig({
       '.yaml': customLoader('yaml'),
       '.yml': customLoader('yaml'),
       '.js': customLoader('js'),
+      '.ts': customLoader('ts'),
```
dotansimha, line 134: should we consider supporting `mts`?

packages/graphql-codegen-cli/src/presets.ts
```diff
@@ -31,9 +31,9 @@ export async function getPresetByName(
         /** ESM Error code */
         err.code !== 'ERR_MODULE_NOT_FOUND'
       ) {
-        throw new DetailedError(
-          `Unable to load preset matching ${name}`,
-          `
+        throw new Error(
```
dotansimha, line 34: Does this change play well with listr?

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -79,7 +79,7 @@ export const createWatcher = (
       .map(filename => ({ filename, config: normalizeOutputParam(config.generates[filename]) }))
       .forEach(entry => {
         if (entry.config.preset) {
-          const extension = entry.config.presetConfig && entry.config.presetConfig.extension;
+          const extension = entry.config.presetConfig && (entry.config.presetConfig as any).extension;
```
dotansimha, line 82: can we avoid `any`?

packages/utils/plugins-helpers/src/types.ts
dotansimha: Ideally, we should aim to infer that options/config and flags based on the preset/plugins name (in TS files). 
But that's for later

packages/graphql-codegen-cli/src/codegen.ts
charlypoly: yes!

packages/graphql-codegen-cli/src/config.ts
charlypoly: Good point, but then, should we support `mjs` files too?

packages/graphql-codegen-cli/src/presets.ts
charlypoly: Yes, I made this change for Listr support actually.

Listr does not support `DetailledError`, leading to missing information in the console output.

packages/graphql-codegen-cli/src/config.ts
charlypoly: Discussed with @dotansimha and @n1ru4l, let's see later if someone is asking for it
cf: https://github.com/dotansimha/graphql-code-generator/issues/8303

charlypoly:
> Should we also do the doc updates for recommending a typescript config as part of this PR?

@n1ru4l, done in b42d628b4

dotansimha (approved):

website/src/pages/docs/getting-started/development-workflow.mdx
```diff
@@ -63,14 +72,23 @@ If you are using a monorepo structure, with tools such as [Yarn Workspaces](http
...
-  server/src/models.ts:
-    - typescript
-    - typescript-resolvers
+```ts {7,11}
+import { CodegenConfig } from '@graphql-codegen/cli'
+
+const config: CodegenConfig = {
+  schema: 'server/src/**/*.graphql',
+  documents: 'client/src/**/*.graphql'
+  generates {
```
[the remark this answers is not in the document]
roottool: I think so. Because `:` exisits after `generates` in the other file.
https://github.com/dotansimha/graphql-code-generator/blob/a87a2f53fee866d3e444743f81f3104b139553ea/website/src/pages/docs/getting-started/esm-typescript-usage.mdx?plain=1#L34

PR #8304, opened 2022-08-30
GraphQL Code Generator v3: front-end examples

charlypoly:
#### TODO

- [x] better READMEs


examples/front-end/angular/src/app/films.service.ts
```diff
@@ -0,0 +1,37 @@
...
+  votes: number;
+  author: {
+    id: string;
+    firstName: string;
+    lastName: string;
+  };
+}
+export interface Response {
+  posts: Post[];
+}
```
kamilkisiela, line 17: I think we can remove it

Issue #5616, opened 2021-02-25
Collapse whitespaces in query strings


AradAral:
There might be a reason why this isn't already implemented, and if there is tell me, but:
It seems that it would be better if the whitespaces inside the generated strings that contain operations and fragments were collapsed.

For example, this is what currently gets generated:
Given that you have a .graphql file that contains the following query:
```graphql
query getCurrentUserInfo {
  currentUser {
    fullName
    cartItemsCount
    purchasedCoursesCount
    specialRole
  }
}
```

This is its string version that gets generated by GraphQL Code Generator:
```javascript
export const GetCurrentUserInfoDocument = `
    query getCurrentUserInfo {
  currentUser {
    fullName
    cartItemsCount
    purchasedCoursesCount
    specialRole
  }
}
    `;
```
However, all those double whitespaces, tabs, and of course line breaks are basically redundant.
It could instead generate:
```javascript
export const GetCurrentUserInfoDocument = "query getCurrentUserInfo { currentUser { fullName cartItemsCount purchasedCoursesCount specialRole } }";
```
The 2 queries above have absolutely no functional difference, but if the line breaks, tabs, and double whitespaces are preserved, it would completely unnecessarily add to the bundle size of the app (if you're building a client-side web app). Not only that, it's also just extra unnecessary bits sent over the network to the server every time.

So, why not simply do a `.replace(/\s+/g, ' ').trim();` on these strings to get rid of all the unnecessary whitespace characters?

ardatan:
Would you create a PR for this?

I think `documentMode: documentNode` would be better option for client-side app because `gql` tag runs everytime app is loaded in the client.

AradAral:
> I think documentMode: documentNode would be better option for client-side app because gql tag runs everytime app is loaded in the client.

You're right, but 2 points:
1. You could be using a GraphQL client that just uses plain strings instead of the AST that the `gql` tag generates, and you've set `documentMode` to `string`.
For instance, [graphql-request](https://github.com/prisma-labs/graphql-request) is an example of such a client.
2. Even if for any reason you've decided to use `documentMode: graphQLTag`, which many people actually do, you again don't need any of the unnecessary characters in the strings that get generated.

> Would you create a PR for this?
I'll give it a shot.

Thanks for your quick response.

dotansimha:
I'm all it for that, feel free to send a PR! :)

Aaron2321d:
[...]

PR #8322, opened 2022-09-03
feat: Add `stripIgnoredCharacters` option (visitor-plugin-common)

aradalvand:
Related #5616

## Description

Adds a new `stripIgnoredCharacters: boolean` option to `client-side-base-visitor` that removes redundant characters (like line-breaks, whitespaces, etc.) from the query strings.

In the linked issue, I originally proposed doing `.replace(/\s+/g, ' ').trim()` on the strings, but that turned out to be an unreliable approach as it would fail to cover a number of cases:
- It doesn't remove extraneous whitespaces that are not consecutive, so like `query ( $var : Int! )` (should become `query ($var: Int!)`) or more commonly `query { foo bar }` (should become `query{foo bar}`).
- If you have a string literal inside the query, that contains more than one consecutive space or line breaks, it removes those but it obviously shouldn't.

This all means a **GraphQL-specific** minifier is actually needed to do this reliably,
Happily I found out there's a built-in function in [`graphql-js`](https://github.com/graphql/graphql-js) for this very task! `stripIgnoredCharacters` — see https://github.com/graphql/graphql-js/issues/1523, which does precisely this.

## Type of change

## How Has This Been Tested?

I actually didn't add tests for this because I didn't know where to add the tests, `client-side-base-visitor` itself didn't include any tests, and I don't really know what tests I should add and more importantly where. I would appreciate if you point me in the right direction.

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -7,6 +7,7 @@ import {
   FragmentSpreadNode,
   GraphQLSchema,
   Kind,
+  stripIgnoredCharacters,
 } from 'graphql';
 import { DepGraph } from 'dependency-graph';
 import gqlTag from 'graphql-tag';
@@ -160,6 +161,11 @@ export interface RawClientSideBasePluginConfig extends RawConfig {
    * @description If set to true, it will enable support for parsing variables on fragments.
    */
   experimentalFragmentVariables?: boolean;
+  /**
+   * @default false
+   * @description If set to true, any extraneous whitespace characters, etc. in the strings will be removed — using `graphql-js`'s `stripIgnoredCharacters` function. Has no effect if `documentMode` is set to `documentNode`.
+   */
+  stripIgnoredCharacters?: boolean;
 }
 
 export interface ClientSideBasePluginConfig extends ParsedConfig {
@@ -180,6 +186,7 @@ export interface ClientSideBasePluginConfig extends ParsedConfig {
   pureMagicComment?: boolean;
   optimizeDocumentNode: boolean;
   experimentalFragmentVariables?: boolean;
+  stripIgnoredCharacters?: boolean;
 }
 
 export class ClientSideBaseVisitor<
@@ -333,12 +340,14 @@ export class ClientSideBaseVisitor<
       return JSON.stringify(gqlObj);
     }
     if (this.config.documentMode === DocumentMode.string) {
-      return '`' + doc + '`';
+      return '`' + this.config.stripIgnoredCharacters ? stripIgnoredCharacters(doc) : doc + '`';
     }
 
     const gqlImport = this._parseImport(this.config.gqlImport || 'graphql-tag');
 
-    return (gqlImport.propName || 'gql') + '`' + doc + '`';
+    return (gqlImport.propName || 'gql') + '`' + this.config.stripIgnoredCharacters
+      ? stripIgnoredCharacters(doc)
+      : doc + '`';
   }
 
   protected _generateFragment(fragmentDocument: FragmentDefinitionNode): string | void {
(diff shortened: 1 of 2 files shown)


aradalvand:
Could you guys help me out with the failing checks? I'm new to all of this, not sure what's being done wrong.

aradalvand:
@ardatan @dotansimha Hi guys could you check this out please? Thank you.

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -333,12 +340,14 @@ export class ClientSideBaseVisitor<
...
     }
     if (this.config.documentMode === DocumentMode.string) {
-      return '`' + doc + '`';
+      return '`' + this.config.stripIgnoredCharacters ? stripIgnoredCharacters(doc) : doc + '`';
     }
 
     const gqlImport = this._parseImport(this.config.gqlImport || 'graphql-tag');
 
-    return (gqlImport.propName || 'gql') + '`' + doc + '`';
+    return (gqlImport.propName || 'gql') + '`' + this.config.stripIgnoredCharacters
```
charlypoly, line 348: This change impact all configurations of `documentMode`, while this new feature should only apply to `documentMode: "string"`

charlypoly (changes requested):
Hi @aradalvand, 

Thank you for your contribution! ⚡ 

I left a comment on potentially unwanted behavior.
Also, could you please:
- a new test case for this new configuration (with `stripIgnoredCharacters` enabled and disabled)
- ensure to run `yarn lint` locally and push the changes to your PR


Thank you!

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
aradalvand: @charlypoly Well, this is expected though. This feature should apply to any `documentMode` that spits out strings, which includes both `string` and `graphQLTag` — `graphQLTag` is basically a string with a `gql` tab in front of it, so we should remove redundant characters for that string too.

And so this is what's happening here.

aradalvand:
Hi @charlypoly, thank you for the response!
Sure, I would certainly be willing to add a test case for this, I just don't know where I'm supposed to add the test. As I said in my original comment:
> I actually didn't add tests for this because I didn't know where to add the tests, client-side-base-visitor itself didn't include any tests, and I don't really know what tests I should add and more importantly where.

Could you please guide me on this? Thank you!

charlypoly:
> Hi @charlypoly, thank you for the response! Sure, I would certainly want to add a test case for this, I just don't know where I'm supposed to add the test. As I said in my original comment:
> 
> > I actually didn't add tests for this because I didn't know where to add the tests, client-side-base-visitor itself didn't include any tests, and I don't really know what tests I should add and more importantly where.
> 
> Could you please guide me on this? Thank you!

Hi @aradalvand,

A good place to write functional tests would be in `packages/plugins/typescript/typescript/tests/typescript.spec.ts` which runs complete schema to types tests using snapshots.

In the meantime, I'll do some manual E2E tests to ensure that this new flag is not breaking some plugins (especially the upcoming codegen v3 one).

JanStevens:
Hi,

This is awesome as it greatly reduces the amount of lines in our generated types. I did notice that it does not work in combination with the following plugins 🤔 

```yaml
      - 'typescript'
      - 'typescript-operations'
      - 'typescript-generic-sdk'
```

aradalvand:
I'm from Iran where the government has recently been limiting Internet access serverely and most VPNs don't work, I've found that I can't actually push to GitHub even though I can visit the website in the browser.

I would appreciate it if someone else take over this PR and continues this (e.g. writing the tests), it should be pretty simple. I can't do that at the moment and I'm not sure when/if this will change.

charlypoly:
The discussion continues on https://github.com/dotansimha/graphql-code-generator/pull/8489

Issue #8338, opened 2022-09-08
Typescript Resolvers Plugin: Ability to provide a custom visitor


dac09:
In Redwood, we need the ability to slightly modify how the `TypeScriptResolversVisitor` works, while keeping all of the other functionality within the typescript-resolvers plugin as is. 

**Describe the solution you'd like**
Ability to pass a customResolver in the config via an import string

**Describe alternatives you've considered**
- passing visitor when we instantiate the plugin (but this does not match existing patterns of configuring codegen plugins

**Additional context**
See draft PR for discussion

Tobbe:
**Describe alternatives you've considered**

We've also considered calling the original typescript-resolver plugin (with the original visitor) as it is, and then calling our own visitor. And then trying to surgically replace specific parts of the output string from the original plugin with the output from our custom visitor. But it feels inefficient and prone to breaking when the output from the original plugin or visitor changes.

PR #8339, opened 2022-09-08
feat(ts-resolvers): Add ability to configure a custom resolver

dac09:
Draft PR (WIP) - one way to implement the feature described in #8338

## Description
In Redwood, we need the ability to slightly modify how the TypeScriptResolversVisitor works, while keeping all of the other functionality within the typescript-resolvers plugin as is.

I've tried to describe what we're trying to achieve in this recording: https://s.tape.sh/xK6JyeYz?s=1.25

Related #8338

## Type of change

## How Has This Been Tested?
This is a draft PR for discussion, I will add tests as advised!

**Test Environment**:

- OS:
- `@graphql-codegen/...`:
- NodeJS:

## Checklist:

### DIFF
--- a/packages/plugins/typescript/resolvers/src/index.ts
+++ b/packages/plugins/typescript/resolvers/src/index.ts
@@ -76,7 +76,18 @@ export type Resolver${capitalizedDirectiveName}WithResolve<TResult, TParent, TCo
   }
 
   const transformedSchema = config.federation ? addFederationReferencesToSchema(schema) : schema;
-  const visitor = new TypeScriptResolversVisitor({ ...config, directiveResolverMappings }, transformedSchema);
+
+  let visitor;
+
+  if (config.customVisitor) {
+    // @TODO: figure out how to use custom-visitor#VisirClass syntax
+    // instead of using default
+    const CustomVistor = require(config.customVisitor).default;
+    visitor = new CustomVistor({ ...config, directiveResolverMappings }, transformedSchema);
+  } else {
+    visitor = new TypeScriptResolversVisitor({ ...config, directiveResolverMappings }, transformedSchema);
+  }
+
   const namespacedImportPrefix = visitor.config.namespacedImportName ? `${visitor.config.namespacedImportName}.` : '';
 
   const astNode = getCachedDocumentNodeFromSchema(transformedSchema);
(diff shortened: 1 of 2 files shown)


packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -76,7 +76,18 @@ export type Resolver${capitalizedDirectiveName}WithResolve<TResult, TParent, TCo
...
 
   const transformedSchema = config.federation ? addFederationReferencesToSchema(schema) : schema;
-  const visitor = new TypeScriptResolversVisitor({ ...config, directiveResolverMappings }, transformedSchema);
+
+  let visitor;
+
+  if (config.customVisitor) {
+    // @TODO: figure out how to use custom-visitor#VisirClass syntax
+    // instead of using default
+    const CustomVistor = require(config.customVisitor).default;
```
dac09, line 85: I need a little guidance on how I can load the supplied string, in the same way mappers or resolverFn loads external imports!

n1ru4l: For loading modules correctly on both ESM and CJS the following is recommended: https://github.com/dotansimha/graphql-code-generator/blob/25c76c110999fe985b59bcb8e353edcec49ac0b9/packages/graphql-codegen-cli/src/codegen.ts#L33-L52

dac09: I was looking for the function that loads strings like this:

`./my-custom-visitor#VisitorClass`

dac09: @n1ru4l, what I'm trying is using `parseMapper` so that the syntax for configuring an import stays the same.

It'll look a little like this:

```ts
import { ExternalParsedMapper, parseMapper } from '@graphql-codegen/visitor-plugin-common';

    const parsedVisitorString = parseMapper(config.customVisitor) as ExternalParsedMapper;
```

This way you can specify the name of the class, rather than have to specifically use the default export.

It also looks like the `makeDefaultLoader` is in the codegen-cli package, should I duplicate it in this one? I'll need to spend a bit more time on this, but hopefully async imports are possible in a visitor!

Issue #8382, opened 2022-09-26
Args passed to commands in CLI hooks (e.g. paths for Prettier) are not escaped


vhenzl:
### Describe the bug

Related to #7088 which is about spaces in paths. This issue is about escaping arguments in general.

When using `near-operation-file-preset` with [Remix's dynamic routes](https://remix.run/docs/en/v1/api/conventions#dynamic-route-parameters), Prettier fails to format the generated files because the paths passed to `prettier --write` are not escaped and therefore their parts are interpreted as shell variables. 

For example

```
prettier --write /project/app/routes/blog/$postId/comments.generated.ts
```

results in

```
prettier --write /project/app/routes/blog//comments.generated.ts
```

Arguments needs to be qouted with single quotes or escaped; either `'/project/app/routes/blog/$postId/comments.generated.ts'` or `/project/app/routes/blog/\\$postId/comments.generated.ts`.

---

BTW, the current version of Codegen (2.12.1) ignores the underlying error silently whereas older versions (for example @graphql-codegen/cli 2.6.2) print a useful error message:

```
Command failed: prettier --write /project/app/routes/blog/$postId/comments.generated.ts
[error] No files matching the pattern were found: "/project/app/routes/blog//comments.generated.ts".
```

### Your Example Website or App

none

### Steps to Reproduce the Bug or Issue

1. Setup `@graphql-codegen` with `near-operation-file-preset` and a hook for Prettier
2. Run codegen and check that generated code is properly formatted
3. Rename a directory with generated code to include the `$` symbol or a space character
4. Run codegen – generated files in the dir are no longer formatted

### Expected behavior

Arguments passed to CLI hook scripts/commands are correctly escaped/quoted to support spaces and special characters in paths.

### Platform

- OS: [WSL]
- NodeJS: 16
- `graphql` version: irrelevant
- `@graphql-codegen/cli` version(s): any (e.g. 2.6.2 or 2.12.1)

PR #8383, opened 2022-09-26
Fix `executeHooks` to quote args for shell command

vhenzl:
## Description

Related #8382

Arguments for commands in CLI hooks are escaped properly using `shell-quote`.

Arguments are escaped only for `execShellCommand`; those passed to a custom `script` function are preserved in their original form.

## Type of change

Possibly breaking for `onError` hook

## How Has This Been Tested?

Locally in our project.

## Checklist:


packages/graphql-codegen-cli/src/hooks.ts
```diff
@@ -74,10 +75,11 @@ async function executeHooks(
...
 
+  const quotedArgs = quote(args);
   for (const script of scripts) {
     if (typeof script === 'string') {
-      debugLog(`Running lifecycle hook "${hookName}" script: ${script} with args: ${args.join(' ')}...`);
-      await execShellCommand(`${script} ${args.join(' ')}`);
+      debugLog(`Running lifecycle hook "${hookName}" script: ${script} with args: ${quotedArgs}...`);
+      await execShellCommand(`${script} ${quotedArgs}`);
     } else {
       debugLog(`Running lifecycle hook "${hookName}" script: ${script.name} with args: ${args.join(' ')}...`);
```
vhenzl, line 84: `quotedArgs` is not used here on purpose as the text in the log message would contain different values than `script` receives. It could be confusing to have escaped arguments in the log but not in the function.

packages/graphql-codegen-cli/src/hooks.ts
```diff
@@ -92,7 +94,7 @@ export const lifecycleHooks = (_hooks: Partial<Types.LifecycleHooksDefinition> =
     afterStart: async (): Promise<void> => executeHooks('afterStart', hooks.afterStart),
     onWatchTriggered: async (event: string, path: string): Promise<void> =>
       executeHooks('onWatchTriggered', hooks.onWatchTriggered, [event, path]),
-    onError: async (error: string): Promise<void> => executeHooks('onError', hooks.onError, [`"${error}"`]),
+    onError: async (error: string): Promise<void> => executeHooks('onError', hooks.onError, [error]),
```
vhenzl, line 97: This change is possibly breaking for `onError` hooks with a custom script as they won't any longer receive quoted text, however, I reckon it's more correct this way, and the impact of this change should be minimal.

charlypoly (changes requested):
@vhenzl, I did some local QA and found out that the `script` should be also `quote()`.
With the following `hooks` configuration:

```yaml
hooks:
  afterAllFileWrite:
    - prettier --write /project/app/routes/blog/$postId/comments.generated.ts
```

The `prettier --write /project/app/routes/blog/$postId/comments.generated.ts` command is not properly escaped with your current implementation.
It seems that quoting the `script` variables (in `executeHooks()`) could do the work.

vhenzl:
Rebased to resolve conflict in `yarn.lock`.

vhenzl:
@charlypoly, I dissagre. 

Whatever is specified in a hook should be a valid command on its own. You should be able to copy&paste it to a terminal, and it should run. You are in control of what's in the config, so you should escape it if needed:

```yaml
hooks:
  afterAllFileWrite:
    - prettier --write '/project/app/routes/blog/$postId/comments.generated.ts'
```

But if you really want to use variables in your hooks, you should be able to do that:

```yaml
hooks:
  afterAllFileWrite:
    - prettier --write $MY_PROJECT/frontend
```

If we escape everything automatically, this wouldn't be possible.

dotansimha (approved):

Dodobibi:
This doesn't work on Windows...

```yaml
hooks:
  afterAllFileWrite:
    - prettier --write
```
is breaking because the quoted path is wrong when absolute.
C:/PATH/TO/FILE is transformed as C\\:/PATH/TO/FILE instead of C\:/PATH/TO/FILE

charlypoly:
> This doesn't work on Windows...
> 
> ```yaml
> hooks:
>   afterAllFileWrite:
>     - prettier --write
> ```
> 
> is breaking because the quoted path is wrong when absolute. C:/PATH/TO/FILE is transformed as C\:/PATH/TO/FILE instead of C:/PATH/TO/FILE

Hi @Dodobibi,

Could you open an issue? 🙏🏼 

Thank you!

Issue #8370, opened 2022-09-20
Option importDocumentNodeExternallyFrom for near-operation-file only imports JS files, but should support TS


bohdanbirdie:
**Is your feature request related to a problem? Please describe.**

When using `importDocumentNodeExternallyFrom: near-operation-file` with `documentMode: external` the imported operation is expected to be a `*.js` file

See:
https://github.com/dotansimha/graphql-code-generator/blob/47d0a57e27dd0d2334670bfc6c81c45e00ff4e74/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts#L487

**Describe the solution you'd like**

Add another config property to change the import file format or inherit it from the operation.

**Describe alternatives you've considered**

Not using `importDocumentNodeExternallyFrom`

**Additional context**

<img width="1011" alt="image" src="https://user-images.githubusercontent.com/23266928/191267254-282d6d20-7f2b-47b6-a06c-5310bfab8c2a.png">

bohdanbirdie:
Actually, it probably makes sense not to have the extension at all for the imports

charlypoly:
Hi @bohdanbirdie,

Could you update all your `@graphql-codegen/*` packages to the last version?
A path has been released that should fix your issue.

bohdanbirdie:
Thanks, @charlypoly 
I tried to do it quickly, but looks like the config file format has been changed from yaml to json, so I will take me a while to edit that

But I can see that the related PR is merged and it looks right to me, so I'll close this issue, and if that won't work for me later I'll re-open it

Thanks again!

Issue #8331, opened 2022-09-06
near-operation-file + typescript-react-apollo = Failed to resolve .js import


tpinne:
### Describe the bug

I still have the same issues as in #8065 even with `emitLegacyCommonJSImports: false`.

### Your Example Website or App

Currently I have no simple repo to provide. If you can't get anywhere without let me know. I try to find time to provide a stripped down repo.

### Steps to Reproduce the Bug or Issue

1. Generate code using provided config
2. Import into Next.js app
3. Run dev server

### Expected behavior

There should be no .js in the fragment filename.

### Platform

- OS: macOS, Linux
- NodeJS: 16.15.1, 16.17.0
- `graphql` version: 16.6.0
- `@graphql-codegen/*` version(s): 

```
├─ @graphql-codegen/add@3.2.1
├─ @graphql-codegen/cli@2.12.0
│  └─ @graphql-codegen/plugin-helpers@2.7.0
├─ @graphql-codegen/core@2.6.2
├─ @graphql-codegen/fragment-matcher@3.3.1
├─ @graphql-codegen/introspection@2.2.1
├─ @graphql-codegen/near-operation-file-preset@2.4.1
├─ @graphql-codegen/plugin-helpers@2.6.2
├─ @graphql-codegen/schema-ast@2.5.1
├─ @graphql-codegen/typescript-apollo-client-helpers@2.2.3
├─ @graphql-codegen/typescript-operations@2.5.3
├─ @graphql-codegen/typescript-react-apollo@3.3.3
├─ @graphql-codegen/typescript@2.7.3
└─ @graphql-codegen/visitor-plugin-common@2.12.1
```

### Codegen Config File

```yaml
overwrite: true
schema: "src/graphql/Schema/*.graphqls"
documents: "src/graphql/**/*.graphql"
emitLegacyCommonJSImports: false
hooks:
  afterAllFileWrite:
    - prettier --write
config:
    avoidOptionals: true
    withComponent: false
    withHOC: false
    withHooks: true
    nonOptionalTypename: true
    preResolveTypes: true
    documentMode: documentNodeImportFragments
    pureMagicComment: true
    dedupeFragments: true
    scalars:
        Cursor: string
        Time: Date
generates:
    src/graphql/types.generated.ts:
        plugins:
            - add:
                  content: "/* eslint-disable */"
            - "typescript"
    src/utilities/Apollo/FragmentTypes.generated.ts:
        plugins:
            - add:
[...]

PR #8432, opened 2022-10-04
Remove extension from operations file import

mvestergaard:
## Description

#7978 added `.js` extensions all over the place. But this one seems wrong. In a typescript project, it will for example generate
`import * as Operations from './UserSelection.gql.js';` resulting in a bunch of errors because the files aren't there (they are `.ts` files)

#8370 #8331

## Type of change

### DIFF
--- a/.changeset/beige-buckets-sleep.md
+++ b/.changeset/beige-buckets-sleep.md
@@ -1,0 +1,5 @@
+---
+"@graphql-codegen/visitor-plugin-common": patch
+---
+
+Remove extension from operations file import
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -483,9 +483,12 @@ export class ClientSideBaseVisitor<
       case DocumentMode.external: {
         if (this._collectedOperations.length > 0) {
           if (this.config.importDocumentNodeExternallyFrom === 'near-operation-file' && this._documents.length === 1) {
-            this._imports.add(
-              `import * as Operations from './${this.clearExtension(basename(this._documents[0].location))}';`
-            );
+            let documentPath = `./${this.clearExtension(basename(this._documents[0].location))}`;
+            if (this.config.emitLegacyCommonJSImports) {
+              documentPath += '.js';
+            }
+
+            this._imports.add(`import * as Operations from '${documentPath}';`);
           } else {
             if (!this.config.importDocumentNodeExternallyFrom) {
               // eslint-disable-next-line no-console
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -484,7 +484,7 @@ export class ClientSideBaseVisitor<
         if (this._collectedOperations.length > 0) {
           if (this.config.importDocumentNodeExternallyFrom === 'near-operation-file' && this._documents.length === 1) {
             this._imports.add(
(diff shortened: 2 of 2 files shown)


ardatan:
This will break the behavior described in here;
https://www.the-guild.dev/graphql/codegen/docs/getting-started/esm-typescript-usage#codegen-configuration

If this flag doesn't work, that flag should be fixed instead of removing `.js` extension.

mvestergaard:
Then someone who knows anything about that ESM stuff should probably pick this up.
It seems to just add the extension always as things are now.

I changed our config to not use external operations to get around the problem for now.

charlypoly (approved):

charlypoly:
Hi @mvestergaard,

Thank you for your contribution! ⚡ 
Your patch has been released in all impacted packages: https://github.com/dotansimha/graphql-code-generator/pull/8469

mvestergaard:
@charlypoly Seems like there's still issues. The `emitLegacyCommonJSImports` setting is not read from config files. It's always true no matter how you try to set it in configuration.

mvestergaard:
@charlypoly Adding ` --emit-legacy-common-js-imports=false` to the command line is the only way to actually set the value it seems

mvestergaard:
And now the types import gets the .js extension instead, sigh...

charlypoly:
> @charlypoly Adding ` --emit-legacy-common-js-imports=false` to the command line is the only way to actually set the value it seems

@mvestergaard could you share your configuration please?

mvestergaard:
```ts

  schema: './src/graphql/schema/schema.graphql',
  generates: {
    './src/graphql/schema/index.ts': {
      plugins: [{ add: { content: '/* eslint-disable */' } }, 'typescript'],
    },
    './src/graphql/schema/introspection.ts': {
      plugins: [{ add: { content: '/* eslint-disable */' } }, 'fragment-matcher'],
    },
    './src/graphql/schema/apollo-helpers.ts': {
      plugins: [{ add: { content: '/* eslint-disable */' } }, 'typescript-apollo-client-helpers'],
    },
    './src': {
      documents: './src/**/*.gql.ts',
      preset: 'near-operation-file',
      presetConfig: {
        extension: '.generated.tsx',
        baseTypesPath: './graphql/schema',
      },
      plugins: ['typescript-operations', 'typescript-react-apollo'],
      config: {
        apolloReactHooksImportFrom: '~/graphql/hooks',
        documentMode: 'external',
        importDocumentNodeExternallyFrom: 'near-operation-file',
      },
    },
  },
  config: {
    enumsAsConst: true,
    strictScalars: true,
    defaultScalarType: 'unknown',
    useTypeImports: true,
    emitLegacyCommonJSImports: false,
    scalars: {
      Date: 'Date',
      DateTime: 'Date',
      LocalDateTime: 'Date',
      LocalTime: 'Date',
      UUID: 'string',
      Decimal: 'number',
      Float: 'number',
      Long: 'number',
      Any: 'unknown',
    },
  },
};

export default config;
```

charlypoly:
@mvestergaard can you move the `emitLegacyCommonJSImports: false` at the root of the config?

mvestergaard:
@charlypoly Ok, that does work with toggling it off. But there's an issue with the types import being generated still. With it set to false, it generates 
`import type * as Types from '../../../../graphql/schema.js';`
should be
`import type * as Types from '../../../../graphql/schema';`

It works fine with `emitLegacyCommonJSImports: true` so the logic is probably the wrong way around somewhere, haven't been able to find where.

charlypoly:
> 

You are right, we need to patch the conflict between `useTypeImports` and `emitLegacyCommonJSImports: false`
I've created an issue: https://github.com/dotansimha/graphql-code-generator/issues/8490

PR #8436, opened 2022-10-04
fix missing config in @graphql-codegen/client-preset

### LINKED ISSUE #8435: `client-preset` doesn't take into account `namingConvention`
### Describe the bug

When using the `client-preset`, if I add `namingConvention` config it won't be taken into account in the "types". It is however applied to the "documents".

### Your Example Website or App

Not public

### Steps to Reproduce the Bug or Issue

1. Use the following config:
Config used to reproduce
```

  overwrite: true,
  schema: 'http://localhost:8080/api/graphql',
  documents: ['src/**/*.{ts,tsx}', '!src/gql/**/*'],
  generates: {
    'src/gql': {
      preset: 'client',
      plugins: [],
    },
  },
  config: { namingConvention: 'change-case-all#upperCase' },
};

export default config;

```
2. Run the codegen
3. Types case doesn't match the config in `src/gql/graphql`

### Expected behavior

`namingConvention` should be taken into account

### Screenshots or Videos

_No response_

### Platform

- OS: all
- NodeJS: v14
- `graphql` version: 16.6.0
- "@graphql-codegen/cli": "2.13.2",
- "@graphql-codegen/client-preset": "1.0.2",

### Codegen Config File

```
{
  overwrite: true,
  schema: 'http://localhost:8080/api/graphql',
  documents: ['src/**/*.{ts,tsx}', '!src/gql/**/*'],
  generates: {
    'src/gql': {
      preset: 'client',
      plugins: [],
    },
  },
  config: { namingConvention: 'change-case-all#upperCase' },
}
```

### Additional context

It seems to be because `options.config` isn't passed down here: https://github.com/dotansimha/graphql-code-generator/blob/f3c5b7a0044a0618898676ee6ee06f2494ac0b7c/packages/presets/client/src/index.ts#L177

Hugodby:
## Description

Pass down config in client-preset config.

Related #8435

## Type of change

## How Has This Been Tested?

- Confirmed it fixes #8435 in local

**Test Environment**:

Same as in #8435

### DIFF
--- a/packages/presets/client/src/index.ts
+++ b/packages/presets/client/src/index.ts
@@ -175,6 +175,7 @@ export const preset: Types.OutputPreset<ClientPresetConfig> = {
         pluginMap,
         schema: options.schema,
         config: {
+          ...options.config,
           inlineFragmentTypes: isMaskingFragments ? 'mask' : options.config['inlineFragmentTypes'],
           useTypeImports: options.config.useTypeImports,
         },


charlypoly:
Hi @Hugodby,

Thank you for raising and proposing a fix!
The fact that the `preset: client` does not forwards configuration is intentional (we want to limit the available options), and I was planning to restore some of them (`scalars`, `namingConvention`, `skipTypename`, and more).
However, we don't want to re-enable all the options.

I'll work on a proposal today and keep you updated here!

charlypoly:
@Hugodby, I'll fix it tomorrow.
I've been blocked on other major issues today.

Issue #8437, opened 2022-10-05
`@graphql-codegen/cli` v2.13.3 fails and only logs `{}`


ryparker:
### Describe the bug

Updating  `@graphql-codegen/cli` to v2.13.3 causes `graphql-codegen` command to fail:

```sh
❯ yarn graphql-codegen --verbose --debug --config ./codegen.ts
{}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Your Example Website or App

I don't think this is necessary. Let me know if you can't reproduce this and i'll put something together.

### Steps to Reproduce the Bug or Issue

1. Install v2.13.3 of `@graphql-codegen/cli`
2. Run `graphql-codegen`

### Expected behavior

- CLI not to fail.
- Would be nice if verbose/debug logging could print more info on errors like this.

### Platform

- OS: macOS
- NodeJS: 16.17.1
- Yarn: 1.22.19
- `graphql` version: 16.6.0
- `@graphql-codegen/cli` version: 2.13.3
- `@graphql-codegen/near-operation-file-preset`: 2.4.1
- `@graphql-codegen/typed-document-node`: 2.3.3
- `@graphql-codegen/typescript`: 2.7.3
- `@graphql-codegen/typescript-operations`: 2.5.3
- `@graphql-typed-document-node/core`: 3.1.1

### Codegen Config File

```ts
import { CodegenConfig } from '@graphql-codegen/cli';
import * as dotenv from 'dotenv';

dotenv.config({ path: '.env.local' });
const GRAPHQL_ENDPOINT = String(process.env.GRAPHQL_ENDPOINT);
const GRAPHQL_SECRET = String(process.env.GRAPHQL_SECRET);

const config: CodegenConfig = {
  schema: [
    {
      [GRAPHQL_ENDPOINT]: {
        headers: {
          'x-api-secret': GRAPHQL_SECRET,
        },
      },
    },
  ],
  documents: ['./src/graphql/*.graphql'],
  generates: {
    './src/graphql-operations.ts': {
      plugins: ['typescript', 'typescript-operations', 'typed-document-node'],
      config: {
        namingConvention: {
          typeNames: 'change-case-all#pascalCase',
          transformUnderscore: true,
        },
      },
    },
  },
};

export default config;
```

### Additional context

My workaround is to downgrade back to v2.13.1 by fixing the package version.

e.g. in package.json:

```json
{
	...
[...]

PR #8439, opened 2022-10-05
Fix issue when loading TS config with imports

charlypoly:
Closes https://github.com/dotansimha/graphql-code-generator/issues/8437


charlypoly:
Aborted in favor of #8441

PR #8455, opened 2022-10-06
feat(preset/client): allow the use of essential `config` options

charlypoly:
https://github.com/dotansimha/graphql-code-generator/issues/8435

### DIFF
--- a/packages/presets/client/src/index.ts
+++ b/packages/presets/client/src/index.ts
@@ -69,6 +69,17 @@ export const preset: Types.OutputPreset<ClientPresetConfig> = {
   buildGeneratesSection: options => {
     const reexports: Array<string> = [];
 
+    // the `client` preset is restricting the config options inherited from `typescript`, `typescript-operations` and others.
+    const forwardedConfig = {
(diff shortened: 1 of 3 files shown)


packages/presets/client/src/index.ts
```diff
@@ -69,6 +69,17 @@ export const preset: Types.OutputPreset<ClientPresetConfig> = {
   buildGeneratesSection: options => {
     const reexports: Array<string> = [];
 
+    // the `client` preset is restricting the config options inherited from `typescript`, `typescript-operations` and others.
+    const forwardedConfig = {
+      scalars: options.config.scalars,
+      defaultScalarType: options.config.defaultScalarType,
+      strictScalars: options.config.strictScalars,
+      namingConvention: options.config.scalars,
```
Hugodby, line 77: It should be mapped to `options.config.namingConvention`

charlypoly: Good catch, fixing: https://github.com/dotansimha/graphql-code-generator/pull/8457

PR #8487, opened 2022-10-15
doc: update docs

yassh:


website/src/pages/docs/getting-started/index.mdx
```diff
@@ -268,7 +268,10 @@ const config: CodegenConfig = {
    schema: 'https://localhost:4000/graphql',
    documents: ['src/**/*.tsx'],
    generates: {
-      './src/gql/': { preset: 'client' }
+      './src/gql/': {
+        preset: 'client',
+        plugins: []
+      }
```
saihaj, line 274: I don’t think this change is necessary

yassh: This change is necessary because `plugins` is not an optional property now.

https://github.com/dotansimha/graphql-code-generator/blob/4976f512b862b3c50a1e0eb800b02813e7a00e5c/packages/utils/plugins-helpers/src/types.ts#L249

charlypoly (approved):

Issue #7549, opened 2022-02-16
[typescript react-query] unused `pageParamKey` when calling `useInfiniteQuery`.


ndthinh8796:
**Describe the bug**
Unused `pageParamKey` in file https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/react-query/src/fetcher-graphql-request.ts

I think this:
```typescript
    return `export const useInfinite${operationName} = <
      TData = ${operationResultType},
      TError = ${this.visitor.config.errorType}
    >(
      pageParamKey: keyof ${operationVariablesTypes},
      client: GraphQLClient,
      ${variables},
      ${options},
      headers?: RequestInit['headers']
    ) =>
    ${hookConfig.infiniteQuery.hook}<${operationResultType}, TError, TData>(
      ${generateInfiniteQueryKey(node, hasRequiredVariables)},
      (metaData) => fetcher<${operationResultType}, ${operationVariablesTypes}>(client, ${documentVariableName}, {...variables, ...(metaData.pageParam ?? {})}, headers)(),
      options
    );`;

```
should be change to:
```typescript
    return `export const useInfinite${operationName} = <
      TData = ${operationResultType},
      TError = ${this.visitor.config.errorType}
    >(
      pageParamKey: keyof ${operationVariablesTypes},
      client: GraphQLClient,
      ${variables},
      ${options},
      headers?: RequestInit['headers']
    ) =>
    ${hookConfig.infiniteQuery.hook}<${operationResultType}, TError, TData>(
      ${generateInfiniteQueryKey(node, hasRequiredVariables)},
      (metaData) => fetcher<${operationResultType}, ${operationVariablesTypes}>(client, ${documentVariableName}, {...variables, [pageParamKey]: metaData.pageParam }, headers)(),
      options
    );`;

```

EandrewJones:
Any intention to fix this or should I create a PR? 

The infinite query does not work correctly if it isn't passing the page param to the correct variable. Plus, even if you write a temporary solution, the generator overwrites the fix anytime you update your update and need to regenerate the hooks.

saihaj:
@EandrewJones go for it! Would love to review and land it in :)

EandrewJones:
Great -- will look to push something in the coming week or two.

Best

Evan Jones
Website: www.ea-jones.com

EandrewJones:
@saihaj Will you please assign me the issue?

PR #8497, opened 2022-10-17
fixes issue #7549

EandrewJones:
## Description

In the typescript-react-query plugin, there are `useInfinite${operationName}Query` hooks. These require a `pageParamKey` to be specified in order for the hook to fetch the next page in paginated queries. Previously, this param was defined as a private param with a preceding '_' and was left unused in the call to the fetcher.

I dropped the private underscore prefix `pageParamKey` and updated the fetcher call so that the `metaData.pageParam` is passed to the `pageParamKey` variable. This allows pagination function properly.

Related #7549 

## Type of change

N/A

## How Has This Been Tested?

- [x] Updated the typescript-react-query plugin test suite  (`packages/plugins/typescript/react-query/tests/react-query.spec.ts`) to match the new outputs 

Can be tested with:
```
$ yarn build
$ yarn test
```

**Test Environment**:

- OS: Ubuntu 18.04
- NodeJS: v14.20.1

## Checklist:

## Further comments

N/A

### DIFF
--- a/.changeset/sweet-nails-tie.md
+++ b/.changeset/sweet-nails-tie.md
@@ -1,0 +1,5 @@
+---
+"@graphql-codegen/typescript-react-query": patch
+---
+
+fixes issue #7549
--- a/dev-test/githunt/types.react-query.ts
+++ b/dev-test/githunt/types.react-query.ts
@@ -466,7 +466,7 @@ export const useCommentQuery = <TData = CommentQuery, TError = unknown>(
   );
 export const useInfiniteCommentQuery = <TData = CommentQuery, TError = unknown>(
   dataSource: { endpoint: string; fetchParams?: RequestInit },
-  _pageParamKey: keyof CommentQueryVariables,
+  pageParamKey: keyof CommentQueryVariables,
   variables: CommentQueryVariables,
   options?: UseInfiniteQueryOptions<CommentQuery, TError, TData>
 ) =>
@@ -475,7 +475,7 @@ export const useInfiniteCommentQuery = <TData = CommentQuery, TError = unknown>(
     metaData =>
       fetcher<CommentQuery, CommentQueryVariables>(dataSource.endpoint, dataSource.fetchParams || {}, CommentDocument, {
         ...variables,
-        ...(metaData.pageParam ?? {}),
+        [pageParamKey]: metaData.pageParam,
       })(),
     options
   );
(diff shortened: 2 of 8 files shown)


saihaj:
@EandrewJones can you fix tests?

EandrewJones:
I will take a look later today. All tests passed locally when I ran yarn
test, so I need to see which ones are failing.

EandrewJones:
@dotansimha I ran:
```
yarn run generate:examples:esm
yarn run generate:examples:cjs
```

I believe this updated the types for the dev-test flow and should've fixed the issue.

charlypoly (approved):

marcmll:
@EandrewJones it seems this change has broken all of my generated `useInfiniteQuery` hook calls.
cc @charlypoly

### Context
One example of a paginated query I am calling is `EmployeeQuery`. The GraphQL schema has some variables which are non-nullable, including the one that I have set as the `pageParamKey` (`paginationInput`):
```ts
export type EmployeesQueryVariables = Exact<{
  paginationInput: WorkPaginationInput;
  filterInput?: InputMaybe<WorkEmployeesFilterInput>;
  sortInput: WorkEmployeeSortInput;
}>;
```

I call it in the following way:
```ts
const {
  data,
  fetchNextPage,
  hasNextPage,
  isFetching,
  isFetchingNextPage,
  isLoading,
  isStale,
} = useInfiniteEmployeesQuery(
  'paginationInput',
  getClient(),
  {
    filterInput,
    paginationInput: { first: PAGE_SIZE },
    sortInput,
  },
  {
    getNextPageParam,
    onError: () => {
      showToast('general', 'error');
    },
  }
);
```

As you can see, `paginationInput` is part of the variables.

### Issue
When the query is called for the first time, `metaData.pageParam` is `undefined`. Previously, `{ ...variables, ...(metaData.pageParam ?? {}) }` would result in all variables being passed during the first call. Now, the new implementation means the `paginationInput` variable is overwritten with `undefined`. By overwriting the `[pageParamKey]` with `undefined` during the first call, GraphQL returns an error as a required variable isn't defined.

### Solution
Instead of making the assumption that `metaData.pageParam` is always defined, only overwrite the `pageParamKey` variable when `metaData.pageParam` has a value.

For example, in [packages/plugins/typescript/react-query/src/fetcher-graphql-request.ts](https://github.com/dotansimha/graphql-code-generator/pull/8497/files#diff-ad57cc93bd7e5269368fb9fd723188868a6c4cd03b9c2f04f812c96cc650ee01):
```ts
(metaData) => fetcher<${operationResultType}, ${operationVariablesTypes}>(client, ${documentVariableName}, {...variables, ...(metaData.pageParam ? { [pageParamKey]: metaData.pageParam } : {}}, headers)(),
```

### Conclusion
Overall... I'm unsure about two things:
1. How could you assume this was a "non-breaking change which fixes an issue", when this change clearly has breaking potential. Something not being used → something being used = something can potentially break
2. You state that "This allows pagination function properly." Did it not function properly before? I have been using the `typescript-react-query` plugin with `react-query` since `react-query` v3 all the way up until the newest version and have never had an issue with the pagination function not working properly. With this change in place... I will no longer be able to update the `typescript-react-query` plugin as this change results in it no longer working for my use case.

EandrewJones:
@marcmll 

Thanks for your comment and I apologize for the breaking change. I will promptly add the ternary operator.

>  How could you assume this was a "non-breaking change which fixes an issue"

I don't think I said that verbatim anywhere, but point well taken.

> Did it not function properly before? 

In my case, no. The page variable is indeed part of the variables and the page param was being updated but not correctly passed back into the variables because I couldn't define the key. I'm curious as to why the types generated for your infiniteQuery hooks correctly allowed for defining the pageParamKey but mine did not...

PR #8527, opened 2022-10-25
fix(typed-document-node): update directory in package.json

zingerj:


charlypoly (approved):

charlypoly:
Good catch! 🦅 

Thank you @zingerj

PR #8535, opened 2022-10-27
feat(cli): upgrade `@graphql-tools/load` to get benefits of debug info on file loading

charlypoly:
`DEBUG=1` now prints some profiling info on files loading:

image

### DIFF
--- a/packages/graphql-codegen-cli/package.json
+++ b/packages/graphql-codegen-cli/package.json
@@ -51,7 +51,7 @@
     "@graphql-tools/github-loader": "^7.3.6",
     "@graphql-tools/graphql-file-loader": "^7.5.0",
     "@graphql-tools/json-file-loader": "^7.4.1",
-    "@graphql-tools/load": "^7.7.1",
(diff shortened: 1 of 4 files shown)


packages/graphql-codegen-cli/package.json
```diff
@@ -51,7 +51,7 @@
     "@graphql-tools/github-loader": "^7.3.6",
     "@graphql-tools/graphql-file-loader": "^7.5.0",
     "@graphql-tools/json-file-loader": "^7.4.1",
-    "@graphql-tools/load": "^7.7.1",
+    "@graphql-tools/load": "7.8.0",
```
SimenB, line 54: why is this locked down?

charlypoly: Because we required to update to this exact version for performance purposes

This will be fixed automatically by Renovate soon

SimenB: 👍

Issue #8566, opened 2022-11-02
PR #8497 typescript-react-query infiniteQuery pageParamKey update introduced breaking change for some users


EandrewJones:
Please see @marcmll's comment below:
https://github.com/dotansimha/graphql-code-generator/pull/8497#issuecomment-1300617915

I will open a new PR to only overwrite the variables with `pageParamKey` when its value is defined:

```
(metaData) => fetcher<${operationResultType}, ${operationVariablesTypes}>(client, ${documentVariableName}, {...variables, ...(metaData.pageParam ? { [pageParamKey]: metaData.pageParam } : {}}, headers)(),
```

saihaj:
released https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1667578711788

saihaj:
https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1667578711788

Issue #8517, opened 2022-10-24
react-query plugin generates invalid infinite query if query has no arguments


iscekic:
### Describe the bug

The latest version introduces an issue where, if an infinite query has no arguments, the `react-query` plugin generates invalid types.

### Your Example Website or App

/

### Steps to Reproduce the Bug or Issue

1. write a query without arguments
2. observe type error in generated code

The error comes from the 3rd argument passed to fetcher:
```
error TS2345: Argument of type '{ [x: string]: any; }' is not assignable to parameter of type 'Exact<{ [key: string]: never; }>'.
  'string' index signatures are incompatible.
    Type 'any' is not assignable to type 'never'.
```

### Expected behavior

No error.

### Platform

- OS: macOS
- NodeJS: 16.7.1
- `graphql` version: 16.6.0

```
    "@graphql-codegen/cli": "^2.13.7",
    "@graphql-codegen/typescript": "^2.8.0",
    "@graphql-codegen/typescript-graphql-request": "^4.5.7",
    "@graphql-codegen/typescript-operations": "^2.5.5",
    "@graphql-codegen/typescript-react-query": "^4.0.3",
```

### Codegen Config File

`scalars` omitted to keep it short

```
overwrite: true
generates:
  src/modules/common/generated/client.graphql.ts:
    documents: "src/**/*.graphql"
    plugins:
      - "typescript"
      - "typescript-operations"
      - "typescript-react-query"
    config:
      dedupeFragments: true
      immutableTypes: true
      skipTypename: true
      maybeValue: T | null | undefined
      namingConvention:
        enumValues: keep
        typeNames: keep
      strictScalars: true
      fetcher: graphql-request
      exposeQueryKeys: true
      addInfiniteQuery: true
```

### Additional context

Only happens in latest version.

wileymc:
This is not only a generated type issue, but infinite queries are now broken.  The work around is to use `exposeDocument: true` and manually build the infinite query using the document.

EandrewJones:
This is on my radar. The last hotfix fixed one subset of issues, but not this one.

I believe the issue stems from trying to unpack metaData.PageParam when it exists into the variables if the infiniteQuery call has no variables. 

[...]

PR #8567, opened 2022-11-02
Hotfix for typescript react query infinite query breaking change

EandrewJones:
## Description

As per @marcmll's comment https://github.com/dotansimha/graphql-code-generator/pull/8497#issuecomment-1300617915, the change introduced in PR #8497 breaks infinite query calls for some users if `pageParam` is undefined which happen on the first call.

Related #8566 

The code now safeguards against this via a ternary operator that check whether pageParam is defined and only overwrites variable value when it's defined.

## Type of change

## How Has This Been Tested?

- [x] Updated the typescript-react-query plugin test suite (packages/plugins/typescript/react-query/tests/react-query.spec.ts) to match the new outputs

**Test Environment**:

- OS: Ubuntu 18.04
- NodeJS: v14.20.1

## Checklist:


marcmll (approved):
Thank you so much for your quick response and for opening a hotfix so quickly!

EandrewJones:
My apologies for breaking your code. There's nothing worse than updating a
package and everything stops working.

Best

Evan Jones
Website: www.ea-jones.com

saihaj (approved):

wileymc:
this is great work! lets get this merged :)

saihaj:
https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1667578711788

wileymc:
Interesting, after updating I'm still getting tsc complaining about: 

```
Argument of type '{ [x: string]: any; }' is not assignable to parameter of type 'Exact<{ [key: string]: never; }>'.
  'string' index signatures are incompatible.
    Type 'any' is not assignable to type 'never'.
```

for infinite queries.  any ideas?

EandrewJones:
Hmm. I'll try to take a look at this tomorrow. Could you provide an example
of the query call you're making and the fetcher you're using?

Best

Evan Jones
Website: www.ea-jones.com

wileymc:
Hi @EandrewJones.  Here is one example of this in a generated query: https://github.com/populist-vote/web/blob/w/upgrade-rq/generated.ts#L3676-L3679

Our codegen.yml: https://github.com/populist-vote/web/blob/w/upgrade-rq/codegen.yml

And the associated .gql: https://github.com/populist-vote/web/blob/w/upgrade-rq/graphql/mutations/auth.graphql#L13-L33

Note, the offending generated hooks are for queries that we'll never need an infinite query for so perhaps there is a way to simply ignore these?

I can get to get a minimal reproducible example if you need! Thanks for your attention on this

EandrewJones:
Thanks for sharing.

Update: I probably won't be able to look deeply into this until Monday. Is
it breaking your type checks on commit/deploy?

Best

Evan Jones
Website: www.ea-jones.com

wileymc:
No worries, its breaking type checks on build but its not blocking or pressing at all.  Thanks for investigating!

EandrewJones:
@wileymc

Are the type warnings appearing in https://github.com/populist-vote/web/blob/w/upgrade-rq/generated.ts itself or in front-end front-end query calls itself?

I'm a bit confused on where you're seeing the issue since you said the offending hooks are for queries that you'll never need an infinite query for, which suggests to me that the warnings must be in the generated.ts file itself.

wileymc:
> @wileymc
> 
> Are the type warnings appearing in https://github.com/populist-vote/web/blob/w/upgrade-rq/generated.ts itself or in front-end front-end query calls itself?
> 
> I'm a bit confused on where you're seeing the issue since you said the offending hooks are for queries that you'll never need an infinite query for, which suggests to me that the warnings must be in the generated.ts file itself.
> 
> Is the entire query underlined by your linter or just a specific part? I'm trying to replicate this but I can't.

I'm not consuming any of the offending generated hooks.  The warnings are indeed in [the generated file:](https://github.com/populist-vote/web/blob/w/upgrade-rq/graphql/mutations/auth.graphql#L4527-L4530) 

image

```
Argument of type '{ [x: string]: any; }' is not assignable to parameter of type 'Exact<{ [key: string]: never; }>'.
  'string' index signatures are incompatible.
    Type 'any' is not assignable to type 'never'.ts(2345)
```
However, I am noticing that one of the infinite query hooks that I am consuming is now failing at runtime when making a call for the next page.  

```
Error: Invalid value for argument "after", expected type "String"`
```

The hooks usage looks [like this](https://github.com/populist-vote/web/blob/w/upgrade-rq/pages/politicians/index.tsx#L100-L133): 
```
  const {
    data,
    isLoading,
    error,
    fetchNextPage,
    hasNextPage,
    isFetchingNextPage,
  } = useInfinitePoliticianIndexQuery(
    "cursor",
    {
      pageSize: PAGE_SIZE,
      filter: {
        query: debouncedSearchQuery || null,
        homeState: state as State,
        politicalScope: scope as PoliticalScope,
        chambers: chamber as Chambers,
      },
    },
    {
      queryKey: [
        "politicianIndex",
        debouncedSearchQuery,
        scope,
        chamber,
        state,
      ],
      getNextPageParam: (lastPage) => {
        if (!lastPage.politicians.pageInfo.hasNextPage) return {};
        return {
          cursor: lastPage.politicians.pageInfo.endCursor,
        };
      },
    }
  );
```

In debugging, i've logged out `lastPage.politicians.pageInfo.endCursor` and it is defined and indeed a string.  

Let me know if you need any other information, thank you

EandrewJones:
@wileymc 
### Type Mismatch
The issue with the type checking is that CurrentUserQuery doesn't accept any variables. So there's a type mismatch between the type definition for `CurrentUserQueryVariables`:
https://github.com/populist-vote/web/blob/37fb984b378497cfc7de572301287bd92ec98df4/generated.ts#L1754

and the unpacking that's happening during pagination:

```
        {
          ...variables,
          ...(metaData.pageParam ? { [pageParamKey]: metaData.pageParam } : {}),
        }
```
What confuses me here is how you're storing currentUser (client-side state) in a database and looking it up without passing any variables such as a userId to the query??

For example, I see similar logic in DeleteAccount: https://github.com/populist-vote/web/blob/37fb984b378497cfc7de572301287bd92ec98df4/generated.ts#L1947

How does the back end know _which_ account to delete if you're not passing a key of some sort?

I'm definitely missing something important here.

### Fix for the broken calls
> However, I am noticing that one of the infinite query hooks that I am consuming is now failing at runtime when making a call for the next page.
> 
> ```
> Error: Invalid value for argument "after", expected type "String"`
> ```
> 
> The hooks usage looks [like this](https://github.com/populist-vote/web/blob/w/upgrade-rq/pages/politicians/index.tsx#L100-L133):
> 
> ```
>   const {
>     data,
>     isLoading,
>     error,
>     fetchNextPage,
>     hasNextPage,
>     isFetchingNextPage,
>   } = useInfinitePoliticianIndexQuery(
>     "cursor",
>     {
>       pageSize: PAGE_SIZE,
>       filter: {
>         query: debouncedSearchQuery || null,
>         homeState: state as State,
>         politicalScope: scope as PoliticalScope,
>         chambers: chamber as Chambers,
>       },
>     },
>     {
>       queryKey: [
>         "politicianIndex",
>         debouncedSearchQuery,
>         scope,
>         chamber,
>         state,
>       ],
>       getNextPageParam: (lastPage) => {
>         if (!lastPage.politicians.pageInfo.hasNextPage) return {};
>         return {
>           cursor: lastPage.politicians.pageInfo.endCursor,
>         };
>       },
>     }
>   );
> ```
> 
> In debugging, i've logged out `lastPage.politicians.pageInfo.endCursor` and it is defined and indeed a string. May be something to do with react-query v4
> 
> Let me know if you need any other information, thank you

Well this one is more apparent: your `getNextPageParam` function isn't just returning `lastPage.poiticians.pageInfo.endCursor`, it's returning an object with a `cursor` field. You only need to return `lastPage.poiticians.pageInfo.endCursor` and the hook should work.

The way the code was changed, the infiniteQuery no longer unpacks the pageParam object `...(metaData.pageParam ?? {})` instead it simply expects the value for the pageParamKey. 

Originally, I thought this was the issue, but now I believe all I needed to fix was the missing pageParamKey argument in the custom-fetcher generation files...

blaenk:
Also experiencing this issue. Is there already an issue to track it? Not sure if it's https://github.com/dotansimha/graphql-code-generator/issues/8517

wileymc:
> @wileymc
> 
> ### Type Mismatch
> The issue with the type checking is that CurrentUserQuery doesn't accept any variables. So there's a type mismatch between the type definition for `CurrentUserQueryVariables`: https://github.com/populist-vote/web/blob/37fb984b378497cfc7de572301287bd92ec98df4/generated.ts#L1754
> 
> and the unpacking that's happening during pagination:
> 
> ```js
>         {
>           ...variables,
>           ...(metaData.pageParam ? { [pageParamKey]: metaData.pageParam } : {}),
>         }
> ```
> 
> What confuses me here is how you're storing currentUser (client-side state) in a database and looking it up without passing any variables such as a userId to the query??
> 
> For example, I see similar logic in DeleteAccount: https://github.com/populist-vote/web/blob/37fb984b378497cfc7de572301287bd92ec98df4/generated.ts#L1947
> 
> How does the back end know _which_ account to delete if you're not passing a key of some sort?
> 
> I'm definitely missing something important here.
> 
> ### Fix for the broken calls
> > However, I am noticing that one of the infinite query hooks that I am consuming is now failing at runtime when making a call for the next page.
> > ```
> > Error: Invalid value for argument "after", expected type "String"`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The hooks usage looks [like this](https://github.com/populist-vote/web/blob/w/upgrade-rq/pages/politicians/index.tsx#L100-L133):
> > ```
> >   const {
> >     data,
> >     isLoading,
> >     error,
> >     fetchNextPage,
> >     hasNextPage,
> >     isFetchingNextPage,
> >   } = useInfinitePoliticianIndexQuery(
> >     "cursor",
> >     {
> >       pageSize: PAGE_SIZE,
> >       filter: {
> >         query: debouncedSearchQuery || null,
> >         homeState: state as State,
> >         politicalScope: scope as PoliticalScope,
> >         chambers: chamber as Chambers,
> >       },
> >     },
> >     {
> >       queryKey: [
> >         "politicianIndex",
> >         debouncedSearchQuery,
> >         scope,
> >         chamber,
> >         state,
> >       ],
> >       getNextPageParam: (lastPage) => {
> >         if (!lastPage.politicians.pageInfo.hasNextPage) return {};
> >         return {
> >           cursor: lastPage.politicians.pageInfo.endCursor,
> >         };
> >       },
> >     }
> >   );
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In debugging, i've logged out `lastPage.politicians.pageInfo.endCursor` and it is defined and indeed a string. May be something to do with react-query v4
> > Let me know if you need any other information, thank you
> 
> Well this one is more apparent: your `getNextPageParam` function isn't just returning `lastPage.poiticians.pageInfo.endCursor`, it's returning an object with a `cursor` field. You only need to return `lastPage.poiticians.pageInfo.endCursor` and the hook should work.
> 
> ```js
>       getNextPageParam: (lastPage) => {
>         if (!lastPage.politicians.pageInfo.hasNextPage) return ''; // Might need to change this to undefined or change the logic so this callback isn't even triggered unless hasNextPage is true
>         return lastPage.politicians.pageInfo.endCursor;
>       },
> ```
> 
> The way the code was changed, the infiniteQuery no longer unpacks the pageParam object `...(metaData.pageParam ?? {})` instead it simply expects the value for the pageParamKey.
> 
> Originally, I thought this was the issue, but now I believe all I needed to fix was the missing pageParamKey argument in the custom-fetcher generation files...

Thanks for the info on the infiniteQuery hook, I will give that a shot later today.  Per my getCurrentUser and deleteAccount queries, they do not require any variables to be be passed because they rely on an http only cookie that is passed with the request which provides the server with all the context it needs to get/delete the user record

EandrewJones:
@wileymc Cookie-based -- got it. Makes sense.

StampixSMO:
> Interesting, after updating I'm still getting tsc complaining about:
> 
> ```
> Argument of type '{ [x: string]: any; }' is not assignable to parameter of type 'Exact<{ [key: string]: never; }>'.
>   'string' index signatures are incompatible.
>     Type 'any' is not assignable to type 'never'.
> ```
> 
> for infinite queries. any ideas?

Also having this error for queries without any variables

EandrewJones:
@wileymc @StampixSMO @marcmll 

I spent some time looking into this. A conditional check against queryVariables that are of type `Exact<{[key: string]: never}>` is out of the question for various reasons.

I think a reversion to the original variable unpacking implementation is the only choice since I don't believe that was the original source of my issue and it also worked fine with your code. Original unpacking implementation:

```ts
{...variables, ...(metaData.pageParam ?? {})}
```
@wileymc This would mean that the fix for broken calls I gave you above would no longer work, but your prior implementation would.

If this sounds good by you, I'll go ahead and implement.

blaenk:
> This would mean that the fix for broken calls I gave you above would no longer work, but your prior implementation would.

Curious what that 'prior implementation' is? I'm interested as someone who didn't make any changes to my graphql-codegen config but can't currently generate code due to this issue. Do you mean that it will now work again without any changes, or is some change necessary and if so what is it? I personally don't mind if any changes are necessary, I'm just curious what it would entail.

Thank you so much for looking into this and working on it @EandrewJones!

EandrewJones:
@blaenk Sorry for the vague language. By "prior implementation", I was referring to the getNextPageParam in @wileymc's code which broke when I made the original change.

If you haven't made any changes to your code to try to address the change introduced in this PR, then reverting _should_ return your existing code to working state. The original way nextPageParam variables were unpacked and passed back into the query was:
```ts
{...variables, ...(metaData.pageParam ?? {})}
```
I thought this was a partial source of the issue I encountered because the pageParamKey wasn't being passed in during unpacking (unless you explicitly returned an object from the getNextPageParam function with the pageParamKey e.g. `return {[pageParamKey]: pageParam}`), but now I believe the problem was actually a missing line of code elsewhere. So, if I revert this line of code, I _think_ everyone should be made whole again.

EandrewJones:
@blaenk That is the intention. Will update this thread when the PR is in. Might be a few days because of holidays.

blaenk:
Looks like that other PR you mentioned is now merged. Nice work!

EandrewJones:
Yes, thanks for all the nudges -- this slipped off my radar as it did the
reviewer's. The codebase should now be almost completely reverted.

Best

Evan Jones
Website: www.ea-jones.com

blaenk:
It does appear to work for me, thanks again for your contribution on this!

EandrewJones:
Thanks for the confirmation.

Best

Evan Jones
Website: www.ea-jones.com

Issue #8588, opened 2022-11-08
Top level TypeScriptLoader(ts-node) registration issue


Diizzayy:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

The [change introduced here](https://github.com/dotansimha/graphql-code-generator/pull/8452/commits/572d3287964ff83dc9a7102cd3d0ac7a0e701c97) by #8452 results in the unnecessarily early registration of `ts-node` which subsequently attempts to load a project's `tsconfig.json` if present.

This isn't an issue for most projects, because the tsconfig tends to be static and independent. 

However though, with [Nuxt 3](https://github.com/nuxt/framework) we've taken the approach of immensely improving upon the developer experience by providing auto imports among with other typescript first conventions to make the workflow as seamless as possible.

Nuxt 3 projects contain a root `tsconfig.json` file which extends from an auto generated tsconfig located at `.nuxt/tsconfig.json`, hence the issue explained above. This has resulted in Nuxt graphql modules which leverage the amazing graphql code generator developed here to crash whenever that `.nuxt/tsconfig.json` hasn't been previously generated.

### Steps to Reproduce the Bug or Issue

1. Clone [`nuxt-graphql-client`](https://github.com/Diizzayy/nuxt-graphql-client)
2. Run `pnpm i` to install dependencies
3. Run `pnpm prepare` to successfully prepare the app

#### Trigger Issue
1. Delete the `playground/.nuxt/tsconfig.json` file
2. Open `package.json` and removed the patched dependency from line 70
3. Run `pnpm i` to apply changes
4. Run `pnpm prepare` to see issue

### Expected behavior

Should run properly without the prerequisites of the project's `tsconfig.json` file being present.

### Platform

- OS: macOS
- NodeJS: v18.11.0
- `graphql` version: 16.6.0
- `@graphql-codegen/cli` version(s): 2.13.11

charlypoly:
@Diizzayy, thanks for the fix!
It has been released in `@graphql-codegen/cli@2.15.0` 📦

PR #8590, opened 2022-11-08
fix: top level ts-node registration bug

Diizzayy:
## Description

This PR simply averts from using a top level variable for the `TypeScriptLoader` to prevent the unnecessary early registration of `ts-node`

Related #8588

## Type of change

### DIFF
--- a/.changeset/silent-singers-cover.md
+++ b/.changeset/silent-singers-cover.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': major
+---
+
+register TypeScriptLoader when needed
--- a/.changeset/silent-singers-cover.md
+++ b/.changeset/silent-singers-cover.md
@@ -1,5 +1,5 @@
 ---
-'@graphql-codegen/cli': major
+'@graphql-codegen/cli': minor
 ---
 
 register TypeScriptLoader when needed
--- a/packages/graphql-codegen-cli/src/config.ts
+++ b/packages/graphql-codegen-cli/src/config.ts
@@ -21,9 +21,6 @@ import { createHash } from 'crypto';
 
 const { lstat } = promises;
 
-// #8437: conflict with `graphql-config` also using TypeScriptLoader(), causing a double `ts-node` register.
-const tsLoader = TypeScriptLoader({ transpileOnly: true });
-
 export type CodegenConfig = Types.Config;
 
 export type YamlCliFlags = {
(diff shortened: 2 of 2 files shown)


packages/graphql-codegen-cli/src/config.ts
```diff
@@ -77,6 +74,8 @@ function customLoader(ext: 'json' | 'yaml' | 'js' | 'ts') {
     }
 
     if (ext === 'ts') {
+      // #8437: conflict with `graphql-config` also using TypeScriptLoader(), causing a double `ts-node` register.
```
charlypoly, line 77: @Diizzayy, doing so will register ts-node multiple times 😣

charlypoly (changes requested):

packages/graphql-codegen-cli/src/config.ts
Diizzayy: @charlypoly Thank you for taking the time to look at this, I must've overlooked something when I tested for the double ts-node registration.

I'm hoping desperately that we'll be able to come up with a workaround

Diizzayy: @charlypoly I must admit that I'm a bit confused here, In my tests thus far, my proposed change doesn't reintroduce the issue raised in #8437 and I can't seem to detect double `ts-node` registrations. Can you have another look into this or possibly direct me as to how to properly check for the ts-node issue

.changeset/silent-singers-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-codegen/cli': major
```
charlypoly, line 2:
```suggestion
'@graphql-codegen/cli': minor
```

charlypoly (approved):
@Diizzayy, I've updated the changeset to minor since we use minor for breaking changes and major for big releases.

Besides that, I [conducted some manual QA](https://github.com/dotansimha/graphql-code-generator/pull/8665) and did not found any regression, testing with the following config:

```ts
import { CodegenConfig } from "@graphql-codegen/cli";
import dotenv from "dotenv";
import { schemaPath } from "./test";

dotenv.config();

  schema: schemaPath,
  documents: [process.env.DOCS!],
  generates: {
    "types.ts": {
      plugins: ["typescript", "typescript-operations"],
    },
  },
};

export default config;
```

Looks good to me, thanks! ⚡

Diizzayy:
> Besides that, I https://github.com/dotansimha/graphql-code-generator/pull/8665 and did not found any regression, testing with the following config

@charlypoly Happy to hear! Thanks again for taking the time to have another look into this

Issue #8570, opened 2022-11-03
@graphql-codegen/cli >= 2.13.10 causes "Unknown fragment" error


rosskevin:
### Describe the bug

```sh
$ yarn codegen
$ graphql-codegen --config ./codegen.ts
✔ Parse Configuration
⚠ Generate outputs
  ❯ Generate to types.ts
    ✔ Load GraphQL schemas
    ✔ Load GraphQL documents
    ✖ GraphQL Document Validation failed with 1 errors;
      Error 0: Unknown fragment "CurrentUser".
      at /sandbox/.graphql:9:8
error Command failed with exit code 1.
```

Broken sandbox:

https://codesandbox.io/s/gql-codegen-fragment-error-reversion-narrow-dep-mtze4x?file=/package.json

```json
  "dependencies": {
    "@graphql-codegen/cli": "2.13.10", // <---------
    "@graphql-codegen/typescript": "2.8.1",
    "@graphql-codegen/typescript-operations": "2.5.6",
    "graphql": "^16.2.0"
  }
```

-------

Working sandbox:

https://codesandbox.io/s/gql-codegen-fragment-error-reversion-working-9j9i6q?file=/package.json

```json
  "dependencies": {
    "@graphql-codegen/cli": "2.13.9",
    "@graphql-codegen/typescript": "2.8.1",
    "@graphql-codegen/typescript-operations": "2.5.6",
    "graphql": "^16.2.0"
  },
```

### Your Example Website or App

https://codesandbox.io/s/gql-codegen-fragment-error-reversion-narrow-dep-mtze4x?file=/package.json

### Steps to Reproduce the Bug or Issue

`yarn generate`

### Expected behavior

I expect `@graphql-codegen/cli >= 2.13.10` to work like previous versions.

### Platform

see codesandbox

### Codegen Config File

default codesandbox file

### Additional context

This has been working fine and I was surprised with a fresh `ncu -u && yarn install` with only patch releases updated that my codegen is broken.  I've narrowed the test sandbox down to @graphql-codegen/cli@2.13.9 being successful, and @graphql-codegen/cli@2.13.10 failing.

cheald:
I've just run into this too. I reverted back to my previous .8 install and things are fine.

MokumJ:
Same same

n33pm:
`2.13.9` works too

rosskevin:
[...]

PR #8606, opened 2022-11-14
fix(core): fix validations on fragments/operations naming conflicts

charlypoly:
Fix https://github.com/dotansimha/graphql-code-generator/issues/8570


saihaj (approved):
can we add a test here for this so when we upgrade tools next time we hopefully know it doesn't break?

charlypoly:
> can we add a test here for this so when we upgrade tools next time we hopefully know it doesn't break?

Agree, we added a integration test in tools 👀

Issue #8632, opened 2022-11-22
Better JSDOC for gql-tag helper


jantimon:
### Which packages are impacted by your issue?

`@graphql-codegen/gql-tag-operations`

It would be nice to add some jsdoc comments to support devs when using the `client-preset`.

<img width="1374" alt="gql" src="https://user-images.githubusercontent.com/4113649/203368721-d198ed0c-38da-4b27-99f9-28c619879f3a.png">

In case the generated types are out of sync we could provide a hint to regenerate them:

<img width="823" alt="unknown" src="https://user-images.githubusercontent.com/4113649/203367847-4346e21a-e216-4d71-9c59-e74d1d515a07.png">

It would also be nice to explain why the babel-plugin should be used for production:

<img width="807" alt="babel" src="https://user-images.githubusercontent.com/4113649/203368148-21de4bd0-50c8-41fd-ade1-fcc6506c7340.png">

Here is a PR: https://github.com/dotansimha/graphql-code-generator/pull/8633

jantimon:
Thanks for merging @charlypoly

PR #8633, opened 2022-11-22
improve autogenerated jsdocs

jantimon:
Related #8632

## Type of change

It adds some jsdoc comments to support devs when using the `client-preset`:

image

In case the generated types are out of sync it provides a hint to regenerate them:

image

It explains why the babel-plugin should be used for production:

image

## How Has This Been Tested?

Unfortunately I couldn't update the snapshots on my machine.  
Jest showed the following error:

`Jest: Couldn't locate all inline snapshots. `

## Checklist:

### DIFF
--- a/.changeset/thick-ravens-try.md
+++ b/.changeset/thick-ravens-try.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/gql-tag-operations': patch
+---
+
+provide jsdoc comments for better IDE support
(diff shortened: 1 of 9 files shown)


charlypoly:
Hi @jantimon,

That's a great idea, thank you!

Could you run the test suite locally with the following cli flags to update the snapshots?

```
yarn test --updateSnashot
```

Also, you will need to run the E2E tests and commit the result:

```
yarn generate:examples
```

_Note: both commands must be run at the project root_

jantimon:
Unfortunately `yarn test --updateSnashot` does not work there is any inline snapshots mismatch 🤷‍♂️

I managed to fix all tests manually :)

Pipeline is now green

PR #8652, opened 2022-11-25
improve life cycle hooks typings

jantimon:
## Description

I would like to prepare a PR for #8643.  
Right now the life cycle hooks are using generics and reducers which kind of blocks an easy implementation.

Therefore I removed the reducer and the generic type.

Related #8643

## Type of change

- [x] Internal Typings

## Checklist:

### DIFF
--- a/.changeset/plenty-bottles-prove.md
+++ b/.changeset/plenty-bottles-prove.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/cli': patch
+'@graphql-codegen/plugin-helpers': patch
+---
+
+improve typings for life cycle hooks
--- a/.changeset/plenty-bottles-prove.md
+++ b/.changeset/plenty-bottles-prove.md
@@ -1,6 +1,6 @@
 ---
 '@graphql-codegen/cli': patch
-'@graphql-codegen/plugin-helpers': patch
+'@graphql-codegen/plugin-helpers': major
 ---
 
 improve typings for life cycle hooks
--- a/packages/graphql-codegen-cli/src/hooks.ts
+++ b/packages/graphql-codegen-cli/src/hooks.ts
@@ -15,6 +15,40 @@ const DEFAULT_HOOKS: Types.LifecycleHooksDefinition = {
   beforeAllFileWrite: [],
 };
 
+function normalizeHooks(_hooks: Partial<Types.LifecycleHooksDefinition>): {
+  [key in keyof Types.LifecycleHooksDefinition]: (string | Types.HookFunction)[];
+} {
+  const keys = Object.keys({
+    ...DEFAULT_HOOKS,
+    ..._hooks,
+  });
+
+  return keys.reduce(
+    (prev: { [key in keyof Types.LifecycleHooksDefinition]: (string | Types.HookFunction)[] }, hookName: string) => {
+      if (typeof _hooks[hookName] === 'string') {
+        return {
+          ...prev,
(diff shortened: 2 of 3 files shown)


packages/utils/plugins-helpers/src/types.ts
```diff
@@ -524,58 +524,60 @@ export namespace Types {
...
   export type HookFunction = (...args: any[]) => void | Promise<void>;
 
+  export type LifeCycleHookValue = string | HookFunction | (string | HookFunction)[];
+
   /**
    * @description All available lifecycle hooks
    * @additionalProperties false
    */
-  export type LifecycleHooksDefinition<T = string | HookFunction | (string | HookFunction)[]> = {
+  export type LifecycleHooksDefinition = {
```
saihaj, line 533: this I guess would be a breaking type change 🤔

jantimon: good point 👍
are those types exported?

should I adjust the changeset?

jantimon: if I read https://unpkg.com/browse/@graphql-codegen/cli@2.14.1/typings/index.d.ts correctly `hooks.ts` is not exported.

So I guess it is not a breaking change - or did I miss sth?

saihaj: Not in Cli but for the `@graphql-codegen/plugin-helpers` which I think many plugins use https://unpkg.com/browse/@graphql-codegen/plugin-helpers@2.7.2/typings/types.d.ts

jantimon: Yes that's absolutely valid 👍 
I believe the risk that a plugin makes use of those generics is very low but it is definitely possible.

Should I change the changeset to be a breaking change?

.changeset/plenty-bottles-prove.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphql-codegen/cli': patch
+'@graphql-codegen/plugin-helpers': patch
```
saihaj, line 3:
```suggestion
'@graphql-codegen/plugin-helpers': major
```

packages/graphql-codegen-cli/src/hooks.ts
```diff
@@ -15,37 +15,6 @@ const DEFAULT_HOOKS: Types.LifecycleHooksDefinition<string[]> = {
...
-    }
-    if (Array.isArray(_hooks[hookName])) {
-      return {
-        ...prev,
-        [hookName]: _hooks[hookName] as string[],
-      };
-    }
-    return prev;
-  }, {} as Types.LifecycleHooksDefinition<(string | Types.HookFunction)[]>);
-}
```
saihaj, line 47: let's refactor in seperate PR and keep this just isolated for type changes

saihaj (changes requested):

packages/graphql-codegen-cli/src/hooks.ts
jantimon: this code contains some type hacks (e.g. casting functions to strings) so for me it felt better to shop both changes at once.

but if it makes the review easier it's fine for me: https://github.com/dotansimha/graphql-code-generator/pull/8652/commits/b1cb681008d1a8ba15f61efed0c3fe639b6a1dda

saihaj (approved):

saihaj:
sorry for the delay here

PR #8723, opened 2022-12-12
transform documents hooks

kazekyo:
## Description
I have a [preset](https://github.com/kazekyo/nau/tree/main/packages/graphql-codegen-preset) to transform user-defined documents. 
I want to use the new `client` preset, so I want to be able to transform documents with a plugin.

Plugin will have the following functions:
```my-plugin.js
module.exports = {
  plugin: () => {
    return 'hello'
  },
  transformDocuments: (_schema, documents) => {
    // Make some changes to the documents    
    return documents;
  },
  validateBeforeTransformDocuments: () => {
    // Raises an error if necessary
  },
};
```
Use it as follows:
```ts
import type { CodegenConfig } from '@graphql-codegen/cli'
 
   schema: 'https://localhost:4000/graphql',
   documents: ['src/**/*.tsx'],
   generates: {
      './src/gql/': {
        preset: 'client',
        documentTransformPlugins: ['./my-plugin.js'],
        plugins: ['./my-plugin.js']
      }
   }
}
export default config
```

One of the main uses of this feature is to add fields. 
For example, assuming that there is a `@useIcon` directive that adds fields related to an icon:
```graphql
query exampleQuery {
  user @useIcon {
    name
    email
  }
}
```
If the plugin transforms the document, we can get the following document:
```graphql
query exampleQuery {
  user {
    name
    email
    iconUrl # Added by plugin
    iconFileName # Added by plugin
  }
}
```

Related: 
- https://github.com/dotansimha/graphql-code-generator/discussions/8593
- https://github.com/dotansimha/graphql-code-generator/issues/8296#issuecomment-1308372643

## Type of change

Please see unit tests for usage.

## How Has This Been Tested?
I added tests to the spec file.

## Checklist:
- [x] I have followed the [CONTRIBUTING](#)(https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project

### About validation
No validation of documents is executed before transforming.
Because I understand that validation before transforming often leads to errors when this feature is needed.
This decision will allow future use of plugins such as relay-operation-optimizer without the `skipDocumentsValidation` flag.

### About relay-operation-optimizer
We can probably remove the special implementation for `relay-operation-optimizer` and replace it with the document transform plugin.
But I did not include that changes in this pull request. If there is no problem, I will do that next.


kazekyo:
@saihaj @n1ru4l Could you please review this?
https://github.com/dotansimha/graphql-code-generator/discussions/8593#discussioncomment-4253778

.changeset/short-toes-relax.md
```diff
@@ -0,0 +1,10 @@
+---
+"@graphql-codegen/cli": minor
+"@graphql-codegen/core": minor
+"@graphql-codegen/plugin-helpers": minor
+"@graphql-codegen/client-preset": minor
+"@graphql-codegen/gql-tag-operations-preset": minor
+"@graphql-codegen/graphql-modules-preset": minor
+---
+
+Add a feature to transform documents
```
n1ru4l, line 10: It would be great if you could add an usage example here and also a section in the codegen documentation. 😇 The documentation change can be a follow-up PR.

kazekyo:
@n1ru4l  I have added a page to the documentation. I'm concerned about the quality of my texts as I don't usually use English.
Please feel free to point it out 🙏

website/src/pages/docs/custom-codegen/document-transform.mdx
```diff
@@ -0,0 +1,117 @@
+import { Callout } from '@theguild/components'
+
+# Document Transform
+
+## Basic Usage
+
+Each plugin can also provide a function `transformDocuments` for transforming `documents`:
+
+```js {5-14}
+const { visit } = require('graphql')
```
saihaj, line 10:
```suggestion
import { visit } from 'graphql'
```

website/src/pages/docs/custom-codegen/document-transform.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+## Example
+
+For example, let's remove a `@loacalOnlyDirective` directive from documents:
+
+```js
+const { visit, print } = require('graphql')
+
+module.exports = {
+  plugin(schema, documents, config) {
```
n1ru4l, line 60: I am not really sure if we need this hook/function. What benefit is it trying to solve?

could this file not simply be the followng?

```ts

module.exports = function transformDocuments (...args) => {
  
}

```

website/src/pages/docs/custom-codegen/document-transform.mdx
```diff
@@ -0,0 +1,117 @@
...
+module.exports = {
+  plugin() {},
+  transformDocuments(schema, documents, config) {
+    return documents
+  },
+  validateBeforeTransformDocuments(schema, documents, config) {
+    if (config.somethingWrong) {
+      throw new Error(`Something wrong!`)
+    }
+  }
```
n1ru4l, line 115: Could the validation not be done within `transformDocuments`?

n1ru4l:
Hey @kazekyo, sorry for letting this PR untouched for so long. I just rebased your PR and I have some questions for you:

With the `client-preset` we are aiming for a relay-compiler-like experience. Features we want to implement there include the following:

- fragment masking (done)
- fragment arguments (pending)
- operation document optimization (pending)

Our goal would be that all new projects using GraphQL Code Generator would pick the `client-preset` over other presets/plugin combinations.

This feature would kind of allow implementation fragment arguments and operation document optimization as you already pointed out. This is great as it would work nicely with this change 😍. [See this `client-preset` artifact as an example](https://github.com/dotansimha/graphql-code-generator/blob/a83786e78e0b5b263e7243641f207f177b072a9c/examples/yoga-tests/src/gql/gql.ts#L32-L45).

Would you be interested in helping out by having fragment arguments and document optimization built into the client preset? Of course not part of this pull request!

______

Here is some more feedback that I noticed while playing around with it.

Right now it is impossible only possible to reference `documentTransforms` via a file.

```ts
// eslint-disable-next-line import/no-extraneous-dependencies
import { type CodegenConfig } from '@graphql-codegen/cli';

  schema: './src/yoga.ts',
  documents: ['src/**/*.ts'],
  generates: {
    './src/gql/': {
      plugins: [],
      preset: 'client-preset',
      documentTransforms: ['./foo.js'],
    },
  },
};

export default config;
```

Should we also allow inline functions?

```ts
// eslint-disable-next-line import/no-extraneous-dependencies
import { type CodegenConfig } from '@graphql-codegen/cli';

  schema: './src/yoga.ts',
  documents: ['src/**/*.ts'],
  generates: {
    './src/gql/': {
      plugins: [],
      preset: 'client-preset',
      documentTransforms: [{
        tarnsformDocuments() {}
      }],
    },
  },
};

export default config;
```


Let me know what you think!

website/src/pages/docs/custom-codegen/document-transform.mdx
kazekyo: Good idea! 🎉 
I'll keep API simple by removing `validateBeforeTransformDocuments`. There may be a possibility of issues arising from the order of validation execution in multiple `transformDocuments` calls, but I cannot give specific examples, and if such a case arises we can always add the function back in.

website/src/pages/docs/custom-codegen/document-transform.mdx
kazekyo: In terms of the current implementation, I'm using [Plugin mechanism](https://github.com/dotansimha/graphql-code-generator/blob/f16fc4c7973a7d51689151539db363bd3d03faba/packages/graphql-codegen-cli/src/codegen.ts#L322) to execute`transformDocuments`. Currently, we get the `transformDocuments` function from Plugin.

Regarding your question, are you suggesting we create a plugin without the `plugin` function?
Currently, the `plugin` function is not optional, but I can change it to be optional. However, I have concerns that making the `plugin` function optional may confuse users and may not be a desirable change to the API. 🤔 

Another option would be not to use the Plugin mechanism at all. Instead, we could introduce a new concept called something like **DocumentTransform** that works independently of Plugin. Just as Preset is a separate concept from Plugin, DocumentTransform is separate from Plugin.
This would be a slightly over-engineered solution, and most implementations would probably be copies of Plugin, but it would make  API cleaner. :sparkles:
If this design is acceptable, I will implement it. 🚀

kazekyo:
@n1ru4l  Thanks for reviewing this pull request!

---

> Would you be interested in helping out by having fragment arguments and document optimization built into the client preset? Of course not part of this pull request!

Of course, I'd be happy to help! It seems that some of our goals overlap. 😃 
Please guide me to the relevant issues, discussions, Discord threads, etc., to discuss these tasks. 

As for this PR, I am unsure whether it can be used as part of your project or should be re-implemented in the `client-preset`. I'll keep developing it if it can be used, but if you have a different opinion, feel free to close it. 🙏 

---

> Should we also allow inline functions?
> 
> ```ts
> // eslint-disable-next-line import/no-extraneous-dependencies
> import { type CodegenConfig } from '@graphql-codegen/cli';
> 
> const config: CodegenConfig = {
>   schema: './src/yoga.ts',
>   documents: ['src/**/*.ts'],
>   generates: {
>     './src/gql/': {
>       plugins: [],
>       preset: 'client-preset',
>       documentTransforms: [{
>         tarnsformDocuments() {}
>       }],
>     },
>   },
> };
> 
> export default config;
> ```

It's great to allow inline functions. It would make it easier for users to experiment.

Just to confirm, Plugin cannot be written as inline functions, correct?
```ts
import { type CodegenConfig } from '@graphql-codegen/cli';

  schema: './src/yoga.ts',
  documents: ['src/**/*.ts'],
  generates: {
    './src/gql/': {
      plugins: [{
        plugin() {}
      }],
      preset: 'client-preset',
    },
  },
};

export default config;
```
Currently, [I use Plugin to execute `transformDocuments`](https://github.com/dotansimha/graphql-code-generator/pull/8723#discussion_r1096513884), so it would be necessary to allow Plugin to write functions inline. I think a separate pull request for this would be appropriate. 🤔 
If we don't use Plugin for `transformDocuments`, I will implement it in this pull request! 💪

website/src/pages/docs/custom-codegen/document-transform.mdx
kazekyo: Fixed code and documentation 👍

n1ru4l: thank you 🥳

n1ru4l:
> Currently, https://github.com/dotansimha/graphql-code-generator/pull/8723#discussion_r1096513884, so it would be necessary to allow Plugin to write functions inline. I think a separate pull request for this would be appropriate. 🤔
If we don't use Plugin for transformDocuments, I will implement it in this pull request! 💪

As discussed here: https://github.com/dotansimha/graphql-code-generator/pull/8723#discussion_r1108655465, let's allow passing a function.

That would then allow you to do something like the following:

```ts
import { type CodegenConfig } from '@graphql-codegen/cli';
import { addTypenameFieldSelections, addIdFieldSelection, inlineFragments } from '@wherever'

  schema: './src/yoga.ts',
  documents: ['src/**/*.ts'],
  generates: {
    './src/gql/': {
      preset: 'client-preset',
      transformDocuments: [addTypenameFieldSelections, addIdFieldSelection, inlineFragments]
    },
  },
};

export default config;
```

n1ru4l (changes requested):
Sorry, for the delayed answer. I added some comments, I am excited to get this close to getting merged!

kazekyo:
@n1ru4l 
I have introduced the concept of Document Transform and made changes to enable document transformation without using plugins. 🎉 
In addition, I implemented the `addToSchema` function to resolve https://github.com/dotansimha/graphql-code-generator/discussions/8724.
Please refer to the documentation for details on how to use it.

Regarding the API designs, please let me know if you have any feedback on the following points:

1. `documentTransforms` is specified with objects and file names. It is also possible to implement a more simplified way of specifying it by directly writing the `transform` function, but I did not implement it as I did not need it. If you think it is necessary to make it more user-friendly, please let me know your feedback.
```ts

  schema: 'https://localhost:4000/graphql',
  documents: ['src/**/*.tsx'],
  generates: {
    './src/gql/': {
      preset: 'client',
      documentTransforms: [
        ({ documents }) => { // Specify only the function
          return documents;
        },
      ],
    },
  },
};
export default config;
```

2. The `transform` function and the `addToSchema` function have completely different required arguments depending on the case, so I am using destructured arguments to extract only the necessary arguments. Therefore, the `addToSchema` function has the same name as the function in the plugin, but the arguments are different. I chose to use the same function name to make it easier to understand how to use it, but if it causes confusion, I can change the function name to something like `extendSchema`.

If you notice anything else, please let me know. I'm confident that the API has improved through our discussions and I welcome further feedback! 😃

website/src/pages/docs/custom-codegen/document-transform.mdx
kazekyo: Fixed https://github.com/dotansimha/graphql-code-generator/pull/8723#issuecomment-1435936804

n1ru4l:
@kazekyo Can you give more context for the use case of `addToSchema`/`extendSchema`?

What is the use-case of coupling the schema transform with a specific document transform?

I think this is redundant with the following:

```ts

  schema: ['https://localhost:4000/graphql', `extend type Query { foo: String }`], // see here
  documents: ['src/**/*.tsx'],
  generates: {
    './src/gql/': {
      preset: 'client',
    },
  },
};
export default config;
```

kazekyo:
@n1ru4l 

> @kazekyo Can you give more context for the use case of `addToSchema`/`extendSchema`?
> 
> What is the use-case of coupling the schema transform with a specific document transform?
> 
> I think this is redundant with the following:
> 
> ```ts
> import type { CodegenConfig } from '@graphql-codegen/cli';
> 
> const config: CodegenConfig = {
>   schema: ['https://localhost:4000/graphql', `extend type Query { foo: String }`], // see here
>   documents: ['src/**/*.tsx'],
>   generates: {
>     './src/gql/': {
>       preset: 'client',
>     },
>   },
> };
> export default config;
> ```

My use case is to implement Relay's [`__id`](https://relay.dev/docs/guided-tour/list-data/updating-connections/#accessing-connections-using-__id). To do that, I need the `addToSchema` function which takes `documents` and `schema` as arguments. This is not possible with the current functionality. You can find more details in https://github.com/dotansimha/graphql-code-generator/discussions/8724.

> Would be happy, if we could immediately merge the already discussed stuff without the `addToSchema` part and then in a follow-up PR/issue discuss the necessity and use-cases for `addToSchema`. I want to ship the `documentTransforms` as soon as possible as we also want to use them. 😇

If after reading the information I just provided, you still have additional discussion or questions regarding `addToSchema`, I will immediately remove `addToSchema`.
Afterward, I will either open another pull request or we can discuss further at https://github.com/dotansimha/graphql-code-generator/discussions/8724 if more discussion is needed!

n1ru4l:
@kazekyo Let's keep `addToSchema` out of scope of this PR!

kazekyo:
@n1ru4l I have removed `addToSchema`! 🚀 

 I don't think the failing test `algolia/algolia-records-check` is a problem with the code but please check it out if you need.

n1ru4l (approved):

n1ru4l:
Thank you @kazekyo 🎉

Issue #8745, opened 2022-12-16
make `enumsAsTypes` export a const of the values as well as the type


pikachu:
### Is your feature request related to a problem? Please describe.

If I want to use `enumsAsTypes` because I want the string union, I have no way to get the list of values of the string union. There are cases in which I want both the string union AND the list of all values in my enum. You _could_ use `enumsAsConsts`, but one could also want to use `enumsAsTypes` (for example, you might not like the weird capitalization things that happen with keys in the generated const).

This change seems pretty harmless - 99% backwards compatible (pending small questions) and adds to the existing feature.

Some questions for you:
- Should we do something better with naming?  `MyEnumValues` or something? need to avoid collisions

### Describe the solution you'd like

When `enumsAsTypes` is true, I want there to be the following lines:

```
export const MyEnum = ['A', 'B'] as const;
export type MyEnum = typeof MyEnum[number];
```

DanielHoffmann:
This change is definitely not completely harmless since it adds a lot of runtime definitions. If you have a really big schema with a lot of enums it could really explode the runtime bundle size. Even if you assume the bundler tree-shakes these arrays out it could still make bundling time significantly longer. All those type inferences using `typeof` could also have a significant type-checking cost.

This feature would be nice as an opt-in, but also allow you to specify which enums you want to have arrays out of (ie opt in by enum name or an "all" option).

eddeee888:
Hello, 
In the next major version, literal string will be the default since it doesn't incur any runtime. Whilst not released yet, the [migration guide](https://the-guild.dev/graphql/codegen/docs/migration/operations-and-client-preset-from-5-0#7-enum-config-options-are-consolidated-and-the-default-value-is-changed) is open for preview and you can try the alpha versions.

This request won't make the cut to the major version, however, we are open for contributions after the release 🙂

PR #8746, opened 2022-12-16
Add enums as types update

pikachu:

## Description
This offers the solution described in the issue. 

Related # (issue)
https://github.com/dotansimha/graphql-code-generator/issues/8745
## Type of change

- [x] This change requires a documentation update (maybe?)

## How Has This Been Tested?

- I modified all of the existing tests for this and they all pass! Waiting to see if I messed up anything more broadly with testing..

## Checklist:

### DIFF
--- a/packages/plugins/typescript/typescript/src/visitor.ts
+++ b/packages/plugins/typescript/typescript/src/visitor.ts
@@ -357,7 +357,7 @@ export class TsVisitor<
     };
 
     const withFutureAddedValue = [
-      this.config.futureProofEnums ? [indent('| ' + wrapWithSingleQuotes('%future added value'))] : [],
+      this.config.futureProofEnums ? [indent(wrapWithSingleQuotes('%future added value'))] : [],
     ];
 
     const enumTypeName = this.convertName(node, {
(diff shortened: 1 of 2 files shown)


saihaj (approved):
change looks good. Can you add changeset with a `minor` update. Learn more [here](https://github.com/dotansimha/graphql-code-generator/pull/8746#issuecomment-1355278811).

saihaj (commented):
tests are failing

pikachu:
@saihaj thanks for the review - i will fix tests. what do you think about the variable naming? i feel like we should do what we do with `enumsAsConsts` and make the type and const with the same name. This way, we avoid the chance of collisions (and are consistent).

dev-test/codegen.ts
```diff
@@ -98,12 +98,6 @@ const config: CodegenConfig = {
       config: { enumsAsTypes: true },
       plugins: ['typescript', 'typescript-operations'],
     },
-    './dev-test/githunt/types.d.ts': {
```
pikachu, line 101: if i'm understanding this file correctly, this is a dupe of the other file (in other words, 
 `./dev-test/githunt/types.enumsAsTypes.ts` and  `./dev-test/githunt/types.d.ts` are the same) - here is the config that we keeping :
```
    './dev-test/githunt/types.enumsAsTypes.ts': {
      schema: './dev-test/githunt/schema.json',
      documents: './dev-test/githunt/**/*.graphql',
      config: { enumsAsTypes: true },
      plugins: ['typescript', 'typescript-operations'],
    },
```

so, I got rid of this one because `enumsAsTypes` no longer outputs `.d.ts` files. if you have consts in .d.ts files, you'd get an error that looks like:
```
TS1254: A 'const' initializer in an ambient context must be a string or numeric literal or literal enum reference
```

@saihaj, will defer to you on the severity of that change.

pikachu:
friendly bump @saihaj :)

saihaj (commented):
thing I am thinking now is should we just add a new options `enumAsConst` as you suggested the name 🤔 rather than updating this one. 

I do prefer this `const` and would recommend others to use this more over TS Enum

dev-test/codegen.ts
saihaj: if we add as a new options https://github.com/dotansimha/graphql-code-generator/pull/8746#pullrequestreview-1248379596 then I guess we can keep this and also add new tests 🤔

pikachu:
@saihaj do you think this change is too major to make?

saihaj:
> @saihaj do you think this change is too major to make?


Even though I agree this is what default really should be moving forward. I just want to be safe and not break anyone. So adding thisa new option seems a safer thing to do.

vinassefranche:
> Hey @pikachu, I am a bit hesitant about merging these changes, can you provide us some real-world use cases and examples where this is handy? This will also change the plugin to emit runnable code instead of only type definitions. It feels like this could instead live inside a dedicated plugin.

Hi! Jumping on the conversation as I might be interested in this option.
I have no real opinion on whether this should be in a dedicated plugin or not but I find the point `This will also change the plugin to emit runnable code instead of only type definitions` flawed: When emitting enums (which is the default, if I'm not mistaken) you are emitting runnable code.

The use case I have for this is when validating a form. I sometimes want to validate that some values are the ones of the union type (with yup, zod or io-ts for example) For this, I need to pass the list of possible values to my form validator and for now, I copy-paste the possible values in a new array. This would be unnecessary if the array was emitted automatically.

Anyway, thanks for the great job on this tool. It's really great!

Issue #8756, opened 2022-12-21
[client-preset] built-in support for persisted documents


n1ru4l:
### Is your feature request related to a problem? Please describe.

Currently, it is hard to implement persisted operations easily.

### Describe the solution you'd like

The client preset should have a functionality that allows doing the following things:

#### Attach a hash to the DocumentNodes

**gql/graphql.ts**

```ts
export const AllFilmsWithVariablesQueryDocument = ({
  ["__hash"]: "<SHA1 HASH OF THE OPERATION>",
  kind: "Document",
  definitions: [
    {
      kind: "OperationDefinition",
      operation: "query",
      name: { kind: "Name", value: "allFilmsWithVariablesQuery" },
      variableDefinitions: [
        {
          kind: "VariableDefinition",
          variable: {
            kind: "Variable",
            name: { kind: "Name", value: "first" }
          },
          type: {
            kind: "NonNullType",
            type: { kind: "NamedType", name: { kind: "Name", value: "Int" } }
          }
        }
      ],
      selectionSet: {
        kind: "SelectionSet",
        selections: [
          {
            kind: "Field",
            name: { kind: "Name", value: "allFilms" },
            arguments: [
              {
                kind: "Argument",
                name: { kind: "Name", value: "first" },
                value: {
                  kind: "Variable",
                  name: { kind: "Name", value: "first" }
                }
              }
            ],
            selectionSet: {
              kind: "SelectionSet",
              selections: [
                {
                  kind: "Field",
                  name: { kind: "Name", value: "edges" },
                  selectionSet: {
                    kind: "SelectionSet",
                    selections: [
                      {
                        kind: "Field",
                        name: { kind: "Name", value: "node" },
                        selectionSet: {
                          kind: "SelectionSet",
                          selections: [
                            {
                              kind: "FragmentSpread",
                              name: { kind: "Name", value: "FilmItem" }
                            }
                          ]
                        }
                      }
                    ]
                  }
                }
              ]
            }
          }
        ]
      }
    },
    ...FilmItemFragmentDoc.definitions
  ]
} as unknown) as DocumentNode<
  AllFilmsWithVariablesQueryQuery,
  AllFilmsWithVariablesQueryQueryVariables
>;
```

[...]

PR #8757, opened 2022-12-21
feat: add support for metadata and persisted operations in client preset

n1ru4l:
Closes https://github.com/dotansimha/graphql-code-generator/issues/8756
Closes https://github.com/dotansimha/graphql-code-generator/issues/8822

This implementation is very ugly IMHO, but I could not see a different way in how things are structured right now. 🥲

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -360,7 +360,15 @@ export class ClientSideBaseVisitor<
     return documentStr;
   }
 
-  private _generateDocumentNodeHash(definitions: ReadonlyArray<DefinitionNode>, fragmentNames: Array<string>): string {
+  private _generateDocumentNodeHash(
+    definitions: ReadonlyArray<DefinitionNode>,
+    fragmentNames: Array<string>
+  ): string | undefined {
+    // If the document does not contain any executable operation, we don't need to hash it
+    if (definitions.every(def => def.kind !== Kind.OPERATION_DEFINITION)) {
+      return undefined;
+    }
+
     const allDefinitions = [...definitions];
 
     for (const fragment of fragmentNames) {
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -1,5 +1,6 @@
 import { basename, extname } from 'path';
 import { oldVisit, Types } from '@graphql-codegen/plugin-helpers';
+import { printExecutableGraphQLDocument } from '@graphql-tools/documents';
 import { optimizeDocumentNode } from '@graphql-tools/optimize';
 import autoBind from 'auto-bind';
 import { pascalCase } from 'change-case-all';
@@ -14,7 +15,6 @@ import {
   print,
   DefinitionNode,
   visit,
-  DocumentNode,
 } from 'graphql';
 import gqlTag from 'graphql-tag';
 import { BaseVisitor, ParsedConfig, RawConfig } from './base-visitor.js';
@@ -33,8 +33,6 @@ export enum DocumentMode {
 }
 
 const EXTENSIONS_TO_REMOVE = ['.ts', '.tsx', '.js', '.jsx'];
-
-const stablePrintDocumentNode = (documentNode: DocumentNode): string => print(documentNode).replace(/\s+/g, ' ');
 
 export interface RawClientSideBasePluginConfig extends RawConfig {
   /**
@@ -373,6 +371,7 @@ export class ClientSideBaseVisitor<
 
     for (const fragment of fragmentNames) {
       const fragmentRecord = this._fragments.find(
+        // TODO: this is very ugly :)
         f => f.name === fragment.substring(0, fragment.length - 'FragmentDoc'.length)
       );
       if (fragmentRecord) {
@@ -402,7 +401,7 @@ export class ClientSideBaseVisitor<
       }
     );
 
-    const operation = stablePrintDocumentNode(sanitizedDocument);
+    const operation = printExecutableGraphQLDocument(sanitizedDocument);
     const shasum = crypto.createHash('sha1');
     shasum.update(operation);
     const hash = shasum.digest('hex');
--- a/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
@@ -230,13 +230,13 @@ export interface ClientSideBasePluginConfig extends ParsedConfig {
   pureMagicComment?: boolean;
   optimizeDocumentNode: boolean;
   experimentalFragmentVariables?: boolean;
-  unstable_onDocumentNode?: Unstable_OnDocumentNode;
+  unstable_onExecutableDocumentNode?: Unstable_OnExecutableDocumentNode;
   unstable_omitDefinitions?: boolean;
 }
 
-type DocumentNodeMeta = Record<string, unknown>;
+type ExecutableDocumentNodeMeta = Record<string, unknown>;
 
-type Unstable_OnDocumentNode = (documentNode: DocumentNode) => void | DocumentNodeMeta;
+type Unstable_OnExecutableDocumentNode = (documentNode: DocumentNode) => void | ExecutableDocumentNodeMeta;
 
 export class ClientSideBaseVisitor<
   TRawConfig extends RawClientSideBasePluginConfig = RawClientSideBasePluginConfig,
@@ -247,7 +247,7 @@ export class ClientSideBaseVisitor<
   protected _additionalImports: string[] = [];
   protected _imports = new Set<string>();
 
-  private _onDocumentNode?: Unstable_OnDocumentNode;
+  private _onExecutableDocumentNode?: Unstable_OnExecutableDocumentNode;
   private _omitDefinitions?: boolean;
   private _fragments: Map<string, LoadedFragment>;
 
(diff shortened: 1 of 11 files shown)


shmax:
@n1ru4l I've spent the last few days trying to figure out how to get persisted queries working with codegen, myself, and I stumbled across this PR. One thing isn't clear to me: does this actually transform the client code for production builds? I don't use Apollo or Urql (I'm stubborn, but weakening), so all I really want to see is for this kind of transformation to happen...

from (dev):
```
// dev code
const skuQuery = graphql('query productDetailsPage($id: ID) {
  sku(id: $id) {
    id
  }'
);
```

to (prod):
```
const skuQuery = {
  hash: `e3423f1c`
};
```

Hope that makes some sense. Please let me know what you think.

n1ru4l:
@shmax Yes, this is possible 🎉 https://github.com/dotansimha/graphql-code-generator/blob/edb9a806c39ddffb906e8bd01e00413632f5f329/packages/presets/client/tests/client-preset.spec.ts#L1230-L1232

shmax:
> Yes, this is possible

Ah, very cool. Okay, so that's the `graphql.ts` file, but what about calls to `graphql` (with query strings) in the source files?

n1ru4l:
@shmax The graphql function returns the document node that has the hash included 😎

shmax:
Sure, I understand that, but in my code I type my queries right into the source, eg:

```
// someclientfile.ts

const ProductDetails: FC<ProductDetailsProps> = (_props) => {
  const { data } = useQuery(["sku"], async () =>
    graphqlAsync(
      graphql(/* GraphQL */ `
        query productDetailsPage($skuId: ID, $loadMod: Boolean) {
          sku(id: $skuId, loadMod: $loadMod) {
            id
            fullName
          }
        }
      `),
      {
        skuId: (9).toString(),
      }
    )
  );

   ...
}

```

That's the whole point of this package, right? So you can generate typings for the response document? Do you have some alternate approach that hasn't occurred to me?

packages/presets/client/src/index.ts
```diff
@@ -218,10 +266,50 @@ export const preset: Types.OutputPreset<ClientPresetConfig> = {
...
+                {
+                  [`persisted-operations`]: {},
+                },
+              ],
+              pluginMap: {
+                [`persisted-operations`]: {
+                  plugin: async () => {
+                    await tdnFinished.promise;
+                    return {
+                      content: JSON.stringify(Object.fromEntries(persistedOperations.entries()), null, 2),
```
saihaj, line 283: we can use fast json here?

n1ru4l: What is fast json?

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -346,6 +358,59 @@ export class ClientSideBaseVisitor<
...
+    if (definitions.every(def => def.kind !== Kind.OPERATION_DEFINITION)) {
+      return undefined;
+    }
+
+    const allDefinitions = [...definitions];
+
+    for (const fragment of fragmentNames) {
+      const fragmentRecord = this._fragments.find(
+        // TODO: this is very ugly :)
+        f => f.name === fragment.substring(0, fragment.length - 'FragmentDoc'.length)
```
n1ru4l, line 375: Anyone having a better idea here?

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -346,6 +358,59 @@ export class ClientSideBaseVisitor<
     return documentStr;
   }
 
+  private _generateDocumentNodeHash(
```
n1ru4l, line 361: To me, it feels like this function shouldn't be here at all. Given the way how different document nodes are constructed (see `_gql` method), I don't see any other way right now...

packages/presets/client/src/index.ts
```diff
@@ -115,20 +134,50 @@ export const preset: Types.OutputPreset<ClientPresetConfig> = {
...
-      [`typed-document-node`]: typedDocumentNodePlugin,
+      [`typed-document-node`]: {
+        ...typedDocumentNodePlugin,
+        plugin: async (...args: Parameters<PluginFunction>) => {
+          try {
+            return await typedDocumentNodePlugin.plugin(...args);
+          } finally {
+            tdnFinished.resolve();
+          }
+        },
```
n1ru4l, line 153: This also feels dirty... But I have no other solution in mind right now.

packages/presets/client/src/index.ts
saihaj: https://www.npmjs.com/package/fast-json

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -228,8 +230,14 @@ export interface ClientSideBasePluginConfig extends ParsedConfig {
   pureMagicComment?: boolean;
   optimizeDocumentNode: boolean;
   experimentalFragmentVariables?: boolean;
+  unstable_onDocumentNode?: Unstable_OnDocumentNode;
+  unstable_omitDefinitions?: boolean;
```
n1ru4l, line 234: Added them as unstable as I don't want anyone outside the client preset to use these features tbh.

dotansimha (approved):

saihaj (approved):
I guess we can wait for `@graphql-tools/documents` to publish then merge this?

saihaj:
@n1ru4l published https://github.com/ardatan/graphql-tools/releases/tag/release-1673979911912

shmax:
@n1ru4l hey there, I've had a chance to tinker with this awesome feature a bit, and I have some feedback for you. 

## Invalid AST?
When `persistedDocuments` is turned on, we wind up with this kind of structure in our graphql.ts file:

```ts
export const UpdateWishlistMutationDocument = {
  __meta__: { hash: "c3cc8d68aaaa31edda041890c6444d412ab72208" },
  kind: "Document",
  definitions: [
```

Interesting, but this AST structure doesn't validate. The library I use (graphql-php) throws an exception [here](https://github.com/webonyx/graphql-php/blob/master/src/Utils/AST.php#L75) ("Node is missing kind").

## Behavior of `replaceDocumentWithHash`
I tried switching on this flag, and the objects in graphql.ts reduce down to _just_ the hash information, which is great, but why isn't this the default behavior? Why do I have to opt-in when the `__meta__` node isn't validating (in at least one popular GraphQL implementation), and when one of the key justifications for using persisted queries is to reduce the amount of information being sent along with back end requests? What use could a person interested in persisted queries have for the entire AST along with the hash?

## Purpose and implementation of the `onExecutableDocumentNode`  feature
It's not clear to me what the purpose of this feature is. I originally thought it was a hook for me to transform a document node to my own taste, but it seems all it does is allow one to add a little additional payload to the `__meta__` node. After examining the samples I'm not sure what the use case is, or what it has to do with persisted queries. I'm also curious about the decision to add a callback to the codegen.ts file. It's TypeScript, so it's valid, but what happens if a user is using a .yml file, instead? Aren't the two meant to be functionally equivalent?

## There are still long graphql queries left in the code.
I don't know if this is, strictly speaking, part of persisted queries, but I think a lot of people are interested in replacing the queries with hashes all throughout the code (and not just in graphql.ts). I was asking about this on the original PR, but you didn't really resolve the question. This results in smaller bundle sizes, and more private code (I don't need to advertise to outside eyes details about my schema, or even the fact that I use GraphQL).

## Why does persisted-documents.json contain query strings instead of ASTs? What about fragments?
You took ASTs out of graphql.ts and replaced them with hash keys, but the values in the generated persisted-documents.json are the original query strings. I guess it's not a huge problem, but what do we do with fragments? They appear in the json file with their own hashes, but they're not cross-referenced anywhere, and it's not clear to me how one would assemble all the pieces on the back end.


So, I guess I'm curious about the background of this feature. Is it inspired by something you saw in a different project? What about the design of the persisted-documents.json file? Is that format compatible with something like Apollo Server, for example? Do fragments work for you?

Maybe I should also mention that I've prepared my own solution that manages to avoid many of the problems described above. I do it all as a custom babel plugin, which I configure in my webpack config file:

```ts
// webpack.prod.ts
...

module: {
    rules: [
      {
        test: /\.ts|tsx?$/,
        exclude: /node_modules/,
        loader: "babel-loader",
        options: {
          plugins: [
            [
              persistentQueries,
              {
                strip: true,
                parser: graphql,
                onlyMatchImportSuffix: true,
                transform: (source, ast) => {
                  const h = hash(source);
                  graphqlAstHashes[h] = ast;
                  return {
                    queryId: h,
                  };
                },
              } as PersistentQueriesConfig,
            ],
          ],
        },
      },
```

When the transformation is complete I write out my own file, `queries.json`. Because I fetch the ASTs with calls to `graphql` (the one from `gql/graphql.ts`), the fragments are all nicely resolved, and I can just write them directly to my json file and the queries run without issue once they're collected on the back end. 

Furthermore, the `transform` callback allows me to transpile the source code from this:

```ts
const { data, isLoading, isFetching } = useQuery(
    ["sku", params.skuId, isPreview],
    async () => {
      return graphqlAsync(
        graphql(/* GraphQL */ `
          query productDetailsPage($skuId: ID, $loadMod: Boolean) {
            loggedIn
            isModerator
            skuSubmission(skuId: $skuId) {
            ...
```

to this:
```
const { data, isLoading, isFetching } = useQuery(
    ["sku", params.skuId, isPreview],
    async () => {
      return graphqlAsync({
        queryId: 'ecf6bfcf1eb9ea5dfa42954795940140072b30748cb59be8274b32c2ef88003a'
       }...
```

While I'm at it, I remove any imports of `graphql`, so to me it doesn't matter what's in it, or how big it gets--it never makes it into the bundle.

I was planning on putting my plugin up on github somewhere, but I wasn't sure if it might be more appropriate to add it to this repo, or if it would conflict with the work you've done here. I'd be grateful for your thoughts, and those of the other contributors.

Sorry for the long post. Thanks for reading, and there's no rush on responding.

Issue #7885, opened 2022-05-25
Support @defer and @stream in operations plugin


n1ru4l:
defer and stream are close to land and should be correctly supported within the code output.

glasser:
Is there a theory of roughly what sorts of types should be generated for incremental delivery operations?

marcesengel:
@hwillson 😍 any way to support you on this?

@glasser for `@defer` at least it's pretty basic, the deferred fields need to be made optional and that's all. Per the apollo documentation:
> Our deferred fields will not exist as keys on data yet [...]

https://www.apollographql.com/docs/react/data/defer/

alesso-x:
Is defer supported now? Looking at the [PR](https://github.com/dotansimha/graphql-code-generator/pull/9196) referenced and this example https://github.com/dotansimha/graphql-code-generator/tree/master/examples/react/apollo-client-defer.

n1ru4l:
Yes, we will release it today 😇

jennyk1:
@n1ru4l I am still unable to use the defer directive with the latest version of graphql-codegen.
I see the event streams coming in but I think Apollo is having an issue parsing the data that is returned.

![Screenshot 2024-05-16 2 50 35 PM](https://github.com/dotansimha/graphql-code-generator/assets/90651818/c950a520-1ce5-4268-8eda-38c8f2772f22)
![Screenshot 2024-05-16 2 50 31 PM](https://github.com/dotansimha/graphql-code-generator/assets/90651818/ea75e20d-af0c-4833-81c8-0cd6e4008248)


This is the error:
```
ServerParseError: Network Error: Please try again
    at JSON.parse (<anonymous>)
    at parseJsonBody (webpack-internal:///../../node_modules/@apollo/client/link/http/parseAndCheckHttpResponse.js:134:21)
    at eval (webpack-internal:///../../node_modules/@apollo/client/link/http/parseAndCheckHttpResponse.js:167:48)
```

alessbell:
Hi there @jennyk1 👋 I work on Apollo Client and it looks like your server is using SSE to communicate with the client which isn't supported out of the box. 

Can you please open an issue over on the Apollo Client repository since this issue doesn't have to do with codegen? Thanks!

PR #8785, opened 2023-01-03
Add initial typescript `@defer` support

hwillson:
This PR takes a first pass at introducing `@defer` typescript codegen support. It follows a similar approach as the `@skip` / `@include` functionality introduced in #5017, with a few notable exceptions:

1. Since `@defer` is currently only supported on fragments (whereas `@skip` / `@include` are supported on fragments and fields), some of the field work from #5017 is not needed.

2. While working on this PR I've uncovered a few problems with the current `@skip` / `@include` support, such as skipped / included inline fragments not being supported properly. I'll try to get these details submitted in follow up issue(s) (and maybe PR(s)), but this has lead to taking a bit of a different approach in this PR to better support fragments.

All tests (new and old) are currently passing but I'll keep this in draft until I've vetted this work with a real application. I'll also add inline review comments shortly to help with code review.

Please note that this PR is currently only focused on `@defer`, but I've tried to keep things implemented in a way that should facilitate supporting `@stream` without a lot of additional effort. I'm super focused on `@defer` at the moment, but will aim to circle back on `@stream` in a subsequent PR.

Related issue: https://github.com/dotansimha/graphql-code-generator/issues/7885

Thanks!

### DIFF
--- a/.changeset/calm-oranges-speak.md
+++ b/.changeset/calm-oranges-speak.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/visitor-plugin-common': minor
+'@graphql-codegen/typescript-operations': minor
+---
+
+Add typescript codegen `@defer` support
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -34,12 +34,19 @@ import {
   PrimitiveField,
   ProcessResult,
 } from './selection-set-processor/base.js';
-import { ConvertNameFn, GetFragmentSuffixFn, LoadedFragment, NormalizedScalarsMap } from './types.js';
+import {
+  ConvertNameFn,
+  FragmentDirectives,
+  GetFragmentSuffixFn,
+  LoadedFragment,
+  NormalizedScalarsMap,
+} from './types.js';
 import {
   DeclarationBlock,
   getFieldNodeNameValue,
   getPossibleTypes,
   hasConditionalDirectives,
+  hasIncrementalDeliveryDirectives,
   mergeSelectionSets,
   separateSelectionSet,
 } from './utils.js';
@@ -50,6 +57,8 @@ type FragmentSpreadUsage = {
   onType: string;
   selectionNodes: Array<SelectionNode>;
 };
+
+type CollectedFragmentNode = (SelectionNode | FragmentSpreadUsage | DirectiveNode) & FragmentDirectives;
 
 function isMetadataFieldName(name: string) {
   return ['__schema', '__type'].includes(name);
@@ -99,8 +108,8 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
    */
   _collectInlineFragments(
     parentType: GraphQLNamedType,
-    nodes: InlineFragmentNode[],
-    types: Map<string, Array<SelectionNode | FragmentSpreadUsage | DirectiveNode>>
+    nodes: Array<InlineFragmentNode & FragmentDirectives>,
+    types: Map<string, Array<CollectedFragmentNode>>
   ) {
     if (isListType(parentType) || isNonNullType(parentType)) {
       return this._collectInlineFragments(parentType.ofType as GraphQLNamedType, nodes, types);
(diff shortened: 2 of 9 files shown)


.changeset/calm-oranges-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphql-codegen/visitor-plugin-common': minor
+'@graphql-codegen/typescript-operations': minor
```
hwillson, line 3: Let me know if you'd prefer these to be `patch`.

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -112,8 +121,18 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
         const spreadsUsage = this.buildFragmentSpreadsUsage(spreads);
         const directives = (node.directives as DirectiveNode[]) || undefined;
 
+        // When we collect the selection sets of inline fragments we need to
+        // make sure directives on the inline fragments are stored in a way
+        // that can be associated back to the fields in the fragment, to
+        // support things like making those fields optional when deferring a
+        // fragment (using @defer).
```
hwillson, line 128: This alludes to the biggest change in this PR. When the `buildSelectionSet` function is trigged in this file, we need a way to associate fields in the selection set back to directives that might have been set on the fragment containing those fields. I'm doing this by adding a new `fragmentDirectives` property to the collected fields that holds any fragment directives that were specified, so when we later need to decide if a field should be marked as optional, we can look into its `fragmentDirectives` property to see if something like `@defer` was used. This approach also makes sure only the directives defined on the fragment related to a field are used with that field specifically, which means multiple inline fragments with different/no directives can now be properly handled.

hwillson:
👋 I've validated this appears to be working properly with a real application. This should now be ready for review. Thanks!

saihaj:
> 👋 I've validated this appears to be working properly with a real application. This should now be ready for review. Thanks!

Is there any public project you are aware of that I can try it out on?

n1ru4l (changes requested):
Can you please also add test-cases in which there are multiple fields selected on a `defer` fragment?

```graphql
fragment Foo on Query {
  a
  b
}

query Query {
  ...Foo @defer
  c
}
```

AFAIK this should then result in types similar to the following:

```ts
type Query = {
  a?: never;
  b?: never;
  c: CType
} | {
 a: AType;
 b: BType;
 c: CType;
}
```

Also for the cases where there are multiple deferred fragments 😢 

```graphql
fragment Foo on Query {
  a
  b
}

fragment Bar on Query {
  d
  e
}

query Query {
  ...Foo @defer
  ...Bar @defer
  c
}
```

hwillson:
@saihaj I've been testing with a private app; I'll find/create something public and send it over.

@n1ru4l I've updated the `should mark fields in deferred fragments as optional (preResolveTypes: true)` test to include both of the requested test scenarios. Let me know if you would also like to see this added to the other tests (I opted to just add it in the one place since `preResolveTypes: true` seems to be the preferred / default option).

.changeset/calm-oranges-speak.md
n1ru4l: It is a new feature so `minor` seems adequate.

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -99,8 +108,8 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
    */
   _collectInlineFragments(
     parentType: GraphQLNamedType,
-    nodes: InlineFragmentNode[],
-    types: Map<string, Array<SelectionNode | FragmentSpreadUsage | DirectiveNode>>
+    nodes: (InlineFragmentNode & FragmentDirectives)[],
```
n1ru4l, line 111:
```suggestion
    nodes: Array<InlineFragmentNode & FragmentDirectives>,
```

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -6434,6 +6434,571 @@ function test(q: GetEntityBrandDataQuery): void {
...
+        export type UserQueryVariables = Exact<{ [key: string]: never; }>;
+        export type UserQuery = {
+          __typename?: 'Query',
+          user: {
+            __typename?: 'User',
+            email?: string,
+            clearanceLevel: string,
+            name: string,
+            widgetCount?: number,
+            widgetPreference?: string,
```
n1ru4l, line 6537: If I spread `WidgetFragment`, it feels a bit annoying having to do both a null check for `widgetCount` and `widgetPreference`. If one exists the other should exist as well. 🤔

Thus, it might make more sense to implement this with `& { widgetCount?: never; widgetPreference?: never; } | { widgetCoun: number; widgetPreference: string; }` instead.

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
hwillson: 👍

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
hwillson: Thanks @n1ru4l, I'm looking into this.

danieljvdm:
Hey all, any reason why this has stagnated? Let me know if I can help out. I'd love to be able to use @defer with this library.

sync:
Is there anything I could do to help get this going?

dotansimha:
@hwillson what do you think should be the next step here? how can we help?

beerose:
Hey @hwillson! I'm gonna take over this PR  and add the union thing mentioned here: https://github.com/dotansimha/graphql-code-generator/pull/8785#discussion_r1083740719

hwillson:
👋 all - sorry for the delay; I'm away on vacation with limited availability until March 20th. If anyone is interested in jumping in to help with this PR in the meantime, that's would be amazing! Otherwise I'll be able to dive back in next week. Thanks!

AugustDev:
With `@defer` being established, this PR is critical part of graphql code generator. Any progress on it much appreciated!

beerose:
Closing this PR in favor of https://github.com/dotansimha/graphql-code-generator/pull/9196

Issue #8827, opened 2023-01-17
Add possibility to generate intermediary types in documents generating plugins ( `typescript-operations`)


antoine-malliarakis:
### Is your feature request related to a problem? Please describe.

When trying to migrate away from the apollo cli and its codegen feature there is one feature which seems to be missing and is described here: https://github.com/apollographql/apollo-tooling/issues/2053#issuecomment-767293917

Basically it generated intermediary core types which could then be reused.

For example it would generate code such as below:

```
export type UserFieldsFragment_profile = { __typename?: 'Profile', age?: number | null };


export type UserFieldsFragment = { __typename?: 'User', id: string, username: string, profile?: UserFieldsFragment_profile | null };

```

In our codeline we use those intermediary types a lot and it would really helps us minimizing the cost of migration.

It would also seem pretty useful to me but then I could be biased.

### Describe the solution you'd like

Have the possibility, when using the `typescript-operations` to explicitely generate intermediary types.

### Describe alternatives you've considered

I've tried playing with the option `exportFragmentSpreadSubTypes` even though the description weren't really giving much hopes about it but it seemed like the closest option.

saihaj:
This is not a feature we want codegen to support. This encourages an anti-pattern. Fragments is how _intermediary types_ should works. Fragments are built in language feature to support things like this we do not need to create some types wrapper because this just gets messier to manage.

PR #8828, opened 2023-01-17
Add `generateIntermediaryTypes` parameter

antoine-malliarakis:
## Description

The idea here is to be able, when generating fragment / operation related code, to introduce explicit intermediary types instead of necessarily inlining them.

This is trying to tackle https://github.com/dotansimha/graphql-code-generator/issues/8827

## Type of change

## How Has This Been Tested?

By adding new unit tests to make sure it has the expected impacts.

## Checklist:

### DIFF
--- a/.changeset/fast-poets-appear.md
+++ b/.changeset/fast-poets-appear.md
@@ -1,4 +1,5 @@
 ---
+'@graphql-codegen/visitor-plugin-common': minor
 '@graphql-codegen/typescript-operations': minor
 ---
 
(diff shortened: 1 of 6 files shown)


antoine-malliarakis:
(Investigating a code simplification option)

antoine-malliarakis:
I tried investigating on how to avoid the `_intermediaryDeclarations` field and it turns out that it would involve passing a lot of additional parameters to the functions involved in the code generation part and I'm therefore not sure it would ease the maintenance.

saihaj:
Thanks for the PR but not something we should move forward with. See https://github.com/dotansimha/graphql-code-generator/issues/8827#issuecomment-1385825565

PR #8863, opened 2023-01-30
feat: use the client preset for the esm example + testing guide

n1ru4l:
Update the examples for latest best practices


examples/typescript-esm/codegen.cjs
```diff
@@ -0,0 +1,16 @@
...
+  emitLegacyCommonJSImports: false,
+  generates: {
+    './src/gql/': {
+      plugins: [],
+      preset: 'client-preset',
+    },
+  },
+};
+
+module.exports = config;
```
n1ru4l, line 16: Wanted to use `.ts`, but couldnt due to cosmicconfig using `require` instead of `import`.
Tried using `.mts`, but cosmicconfig complained about an unknown file extension 😬

PR #8881, opened 2023-01-31
docs: add example for persisted documents

n1ru4l:
**Todo:**
- [x] Guide


saihaj (commented):
should we commit the generated `gql/*` folder?

n1ru4l:
@saihaj yeah for the examples it makes sense as it helps people understand the stuff without needing to generate stuff when browsing on github

Issue #8967, opened 2023-02-07
@graphql-codegen/typescript-operations dedupeFragments doesn't work


DominicGBauer:
### Which packages are impacted by your issue?

@graphql-codegen/typescript-operations

### Describe the bug

When adding `dedupFragment: true` to config it doesn't do anything.

`codegen.yml`:
```
  src/:
    preset: near-operation-file
    presetConfig:
      baseTypesPath: services/api/types.generated.ts
    plugins:
      - typescript-operations
          dedupFragment: true
```

### Your Example Website or App

none

### Steps to Reproduce the Bug or Issue

1. Add a graphql query that has duplicated fragments
2. Run the codegen with `dedupFragment: true`

### Expected behavior

As a user I expect fragments to be deduplicated

### Platform

"@graphql-codegen/typescript-operations": "^3.0.0"

### Codegen Config File
```
overwrite: true
schema:
  - "https://api-dev.nomanini.com/graphql":
      headers:
        Authorization: 'Bearer zzz'
        X-Nomanini-Organisation: 'zzz'
documents: 'src/**/*.graphql'
watchConfig:
  usePolling: true
  interval: 1000
generates:
  src/services/redux/api/types.generated.ts:
    plugins:
      - add:
          content: /* eslint-disable */
      - typescript
  src/services/redux/api/mocks.generated.ts:
    plugins:
      - add:
          content: '/* eslint-disable camelcase,no-underscore-dangle,@typescript-eslint/no-use-before-define,@typescript-eslint/no-unused-vars,no-prototype-builtins */'
      - typescript-mock-data:
          typesFile: './types.generated.ts'
          listElementCount: 2
          enumValues: change-case-all#pascalCase
          typenames: keep
  src/:
    preset: near-operation-file
    presetConfig:
      baseTypesPath: services/redux/api/types.generated.ts
    config:
      link:
        name: 'nomanini'
        endpoint: 'https://api-dev.nomanini.com/graphql'
    plugins:
      - add:
          content:  import { getEnv } from '@services/utils/general'
      - typescript-operations
          dedupFragment: true
      - typescript-rtk-query:
          importBaseApiFrom: '@services/redux/api/baseApi'
          importBaseApiAlternateName: 'graphqlApi'
          exportHooks: true
          overrideExisting: getEnv() === 'dev'
      - typescript-msw
  .graphql.schema.json:
    plugins:
      - introspection
```

Issue #8842, opened 2023-01-21
Client preset generates duplicate fragments definitions


Banou26:
### Which packages are impacted by your issue?

`@graphql-codegen/client-preset^1.2.6`

### Describe the bug

Using these definitions

```ts
export const COMMON_FRAGMENT = gql(`
  fragment CommonFragment on Common {
    common
    common2
  }
`)

export const FOO_FRAGMENT = gql(`
  fragment FooFragment on Foo {
    ...CommonFragment
    foo
    foo2
    bar {
      ...BarFragment
    }
  }
`)

export const BAR_FRAGMENT = gql(`
  fragment BarFragment on Bar {
    ...CommonFragment
    bar
    bar2
  }
`)

export const _GET_FOO = gql(`
  query GetFoo {
    foo {
      ...FooFragment
    }
  }
`)
```

the client preset generates a `GetFoo` document that contains the `CommonFragment` twice, which causes most servers to reject the request

### Your Example Website or App

[graphql-code-generator-issue-8842](https://github.com/Banou26/graphql-code-generator-issue-8842)

### Steps to Reproduce the Bug or Issue

1. `npm i && npm run dev`
2. Go to http://localhost:2222
3. open devtools and check out the logs

### Expected behavior

Document should not contain duplicate fragments

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/5209149/213889562-b90fba7f-2d99-4ada-a1e2-77d9b80d15d3.png)

### Platform

- NodeJS: 19.0.1

```
    "graphql": "^16.6.0",
    "@graphql-codegen/cli": "^2.16.4",
    "@graphql-codegen/client-preset": "^1.2.6",
    "@graphql-codegen/typescript-react-apollo": "^3.3.7",
    "@graphql-codegen/typescript-resolvers": "^2.7.12",
```

### Codegen Config File

```ts
const config: CodegenConfig = {
  schema: [typeDefs],
  documents: ['src/**/*.ts', 'src/**/*.tsx'],
  generates: {
    './src/generated/': {
      preset: 'client',
      plugins: ['typescript-resolvers'],
      presetConfig: {
        gqlTagName: 'gql',
        fragmentMasking: false
      }
    }
  },
  ignoreNoDocuments: true
}
```

Julhol-droid:
I have the same Problem, resulting in nextjs outputing the generated Definition files in the Server Console, with a Error that my Fragment is defined twice in the generated Typescript file.

[...]

PR #8971, opened 2023-02-08
fix: always inline and deduplicate fragment definitions

n1ru4l:
The option `dedupeFragments` has been problematic from the beginning and caused a lot of unexpected behavior that lead to either duplicated fragment definitions within a document node or missing fragment definitions within a document node.

***ATTENTION !!!*** This PR basically replaces the existing definitions spread (`...`) with a duplication/inline strategy. While this increases the bundle size of the client applications, it is actually the only possible way of solving this issue once and for all. 

I marked this change as a fix, though it might be debatable that this should instead be a breaking change, which we could also ship IMHO.

```diff
-    ...VoteButtonsFragmentDoc.definitions,
-    ...RepoInfoFragmentDoc.definitions,
+   {
+     kind: 'FragmentDefinition',
+     name: { kind: 'Name', value: 'VoteButtons' },
+     typeCondition: { kind: 'NamedType', name: { kind: 'Name', value: 'Entry' } },
+     directives: [],
+     selectionSet: {
+       kind: 'SelectionSet',
+       selections: [
+         { kind: 'Field', name: { kind: 'Name', value: 'score' }, arguments: [], directives: [] },
+         {
+           kind: 'Field',
+           name: { kind: 'Name', value: 'vote' },
+           arguments: [],
+           directives: [],
+           selectionSet: {
+             kind: 'SelectionSet',
+             selections: [
+               { kind: 'Field', name: { kind: 'Name', value: 'vote_value' }, arguments: [], directives: [] },
+             ],
+           },
+         },
+       ],
+     },
+   },
+   {
+     kind: 'FragmentDefinition',
+     name: { kind: 'Name', value: 'RepoInfo' },
+     typeCondition: { kind: 'NamedType', name: { kind: 'Name', value: 'Entry' } },
+     directives: [],
+     selectionSet: {
+       kind: 'SelectionSet',
+       selections: [
+         { kind: 'Field', name: { kind: 'Name', value: 'createdAt' }, arguments: [], directives: [] },
+         {
+           kind: 'Field',
+           name: { kind: 'Name', value: 'repository' },
+           arguments: [],
+           directives: [],
+           selectionSet: {
+             kind: 'SelectionSet',
+             selections: [
+               { kind: 'Field', name: { kind: 'Name', value: 'description' }, arguments: [], directives: [] },
+               { kind: 'Field', name: { kind: 'Name', value: 'stargazers_count' }, arguments: [], directives: [] },
+               { kind: 'Field', name: { kind: 'Name', value: 'open_issues_count' }, arguments: [], directives: [] },
+             ],
+           },
+         },
+         {
+           kind: 'Field',
+           name: { kind: 'Name', value: 'postedBy' },
+           arguments: [],
+           directives: [],
+           selectionSet: {
+             kind: 'SelectionSet',
+             selections: [
+               { kind: 'Field', name: { kind: 'Name', value: 'html_url' }, arguments: [], directives: [] },
+               { kind: 'Field', name: { kind: 'Name', value: 'login' }, arguments: [], directives: [] },
+             ],
+           },
+         },
+       ],
+     },
+   },
```

___

Closes https://github.com/dotansimha/graphql-code-generator/issues/8967
Closes https://github.com/dotansimha/graphql-code-generator/issues/8842


packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -399,47 +375,50 @@ export class ClientSideBaseVisitor<
...
+      );
 
-        return `{${hashPropertyStr}"kind":"${Kind.DOCUMENT}", "definitions":[${definitions}]}`;
+      for (const fragmentName of fragmentNames) {
+        fragmentDependencyNames.add(fragmentName);
       }
 
-      let meta: ExecutableDocumentNodeMeta | void;
+      const jsonStringify = (json: unknown) =>
+        JSON.stringify(json, (key, value) => (key === 'loc' ? undefined : value));
```
n1ru4l, line 397: loc is only bloat

packages/presets/client/tests/client-preset.spec.ts
```diff
@@ -1304,7 +1304,7 @@ export * from "./gql.js";`);
 
         export type CFragment = { __typename?: 'Query', c?: string | null } & { ' $fragmentName'?: 'CFragment' };
 
-        export const CFragmentDoc = {} as unknown as DocumentNode<CFragment, unknown>;
+        export const CFragmentDoc = {"kind":"Document","definitions":[{"kind":"FragmentDefinition","name":{"kind":"Name","value":"C"},"typeCondition":{"kind":"NamedType","name":{"kind":"Name","value":"Query"}},"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":{"kind":"Name","value":"c"}}]}}]} as unknown as DocumentNode<CFragment, unknown>;
```
n1ru4l, line 1307: Now this can actually be used with readFragment and stuff 😬

dotansimha: amazing!

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
dotansimha: nice

dotansimha (approved):

saihaj (approved):
Let's mark `dedupeFragments` as a deprecated option so we can drop it in the next major.

.changeset/rotten-lizards-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-codegen/client-preset': patch
```
n1ru4l, line 2:
```suggestion
'@graphql-codegen/client-preset': patch
'@graphql-codegen/typed-document-node': patch
```

JanStevens:
This is a breaking change IMO and should only be done in a major version.

Our generated file size just jumped from 600kb to a whopping 3.7MB causing issues on our frontend since there is a lot of fragment reuse going on (150 queries, lots and lots of fragments). Our Lighthouse score totally got trashed since we shipped a gigantic file, please reconsider 🙏  or keep the option in there like it was before.

We are using this in combination of typed-document-node, to be honest we never had issues with duplicate fragments (we are not using client-present atm).

Regards,

n1ru4l:
@JanStevens Please use the SWC or babel plugin. Also, you can use persisted documents and completely drop the AST/string being shipped to the client. https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#persisted-documents

JanStevens:
> @JanStevens Please use the SWC or babel plugin. Also, you can use persisted documents and completely drop the AST/string being shipped to the client. https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#persisted-documents

I know about persisted documents but this is not implemented everywhere. For example when querying headless CMS like Contentful or DatoCMS, they don't offer a persisted query implementation. The only option is to add your own graphql server on top of their existing implementation, which is not really maintainable.

nandorojo:
I upgraded all my graphql-codegen versions, and I added `dedupeFragments: true` to my `client` preset. However, I still am getting an error from my server saying we're using duplicate fragments with the same name. Is there any other step I should take?

This is the error I see:

```tsx
There can be only one fragment named `MyFragment`
```


It happens when I do this:

```graphql
query Test {
  test {
    ...MyFragment
    ...Nested
  }
}

fragment Nested on Test { 
  someField
  ...MyFragment 
}
```

Thanks!

negezor:
Unfortunately, I noticed this Pull Request too late, since `@graphql-codegen/typescript-vue-urql` had long depended on `@graphql-codegen/visitor-plugin-common` version ^2. After upgrading and regenerating types, my project size increased by 3MB, because fragment documents are no longer reused. My project has 148 queries, 216 mutations, and 203 fragments that are reused across each other.

Unfortunately, I cannot switch to `documentMode.graphQLTag` generation, as it significantly impacts TBT in Lighthouse. Fortunately, I can still stay on the latest version of graphql-codegen, but with the following override in `package.json`:

```json
{
    "overrides": {
        "@graphql-codegen/typescript-vue-urql": {
            "@graphql-codegen/visitor-plugin-common": "=3.0.0"
        }
    }
}
```

Perhaps it would be worth reconsidering this decision and bringing the option back?

PR #8982, opened 2023-02-13
chore: embed examples in mono-repo

n1ru4l:
- [x] make examples part of the monorep
- [x] fix the code generation of the examples
- [x] fix the examples so they actually run
- [x] add ci check for re-generating the codegen and checking whether the output has changed
- [x] add ci check for running build of the examples in order to check the thing still works


examples/react/babel-optimized/src/setupTests.ts
```diff
@@ -2,4 +2,4 @@
 // allows you to do things like:
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
-import '@testing-library/jest-dom';
+export {};
```
saihaj, line 5: can we not delete this file?

saihaj (approved):

examples/react/babel-optimized/src/setupTests.ts
n1ru4l: done!

PR #8988, opened 2023-02-14
show that type:module will break codegen

n1ru4l:


beerose:
ESM issues were fixed in https://github.com/dotansimha/graphql-code-generator/pull/9086. I'm closing this PR then.

PR #8991, opened 2023-02-15
example end2end tests

n1ru4l:
This adds integration tests to most example apps, so it is easier to upgrade 🥳


saihaj (commented):
Do we need those commented files?

n1ru4l:
@saihaj the files got auto-generated by cypress. I guess we can ditch most of the contents

saihaj:
> @saihaj the files got auto-generated by cypress. I guess we can ditch most of the contents

Yeah we should if we can

PR #9009, opened 2023-02-18
feat: use @parcel/watcher

### LINKED ISSUE #8963: watch mode should pick up codegen config changes
### Is your feature request related to a problem? Please describe.

It would be great when editing codegen.ts/yaml/toml/jesus would be detected by the watch mode.

### Describe the solution you'd like

It would be great when editing codegen.ts/yaml/toml/jesus would be detected by the watch mode.

### Describe alternatives you've considered

restart codegen lol

### Is your feature request related to a problem? Please describe.

_No response_

saihaj:
As the title says use [`@parcel/watcher`](https://github.com/parcel-bundler/watcher) over `chokidar`. It is way performant and watch mode feels fast! Based on platform it uses has the following watcher backends, listed in priority order:

1. [FSEvents](https://developer.apple.com/documentation/coreservices/file_system_events) on macOS
2. [Watchman](https://facebook.github.io/watchman/) if installed
3. [inotify](http://man7.org/linux/man-pages/man7/inotify.7.html) on Linux
4. [ReadDirectoryChangesW](https://msdn.microsoft.com/en us/library/windows/desktop/aa365465%28v%3Dvs.85%29.aspx) on Windows

One thing to note with this PR is we can no longer support `polling` config which was a feature in `chokidar` so things like https://github.com/dotansimha/graphql-code-generator/discussions/5580 will not really be possible. There is an open issue related to this https://github.com/parcel-bundler/watcher/issues/99 

closes #8963


kamilkisiela (approved):

n1ru4l (approved):

n1ru4l:
@saihaj Can this solve https://github.com/dotansimha/graphql-code-generator/issues/8963?

saihaj:
> @saihaj Can this solve #8963?

Looks like so
https://user-images.githubusercontent.com/44710980/220399499-0dc4f80e-ea87-4375-aaef-41f552081084.mp4

lauraseidler:
Looks like this introduces a dependency on `node-gyp`, and with that on Python, which at least for us, is a breaking change - we are using node js alpine docker images, which do not have python installed by default, so this requires a change to the build process.

ardatan:
@lauraseidler I am not if it can be considered as a breaking change. `node-gyp` is the native part of Node.js ecosystem, and GraphQL Codegen's CLI is already Node.js specific package. Python is one of prerequisites of Node.js as far as I know.

szilarddoro:
I can confirm that this is a breaking change for alpine images. We are 99% sure that our pipeline [started failing](https://github.com/nhost/nhost/actions/runs/4303614192/jobs/7503845696) because of `@parcel/watcher`: 

Our `builder` step didn't have `python` installed, so we had to change our Dockerfile.

mogelbrod:
Looks like the switch to `@parcel/watcher` may have also resulted in negated glob patterns no longer working as expected. Using the following documents list I can see that this line doesn't bail as expected for events that should be ignored (such as `create .git/.watchman-cookie-name-869-12641`), but happily continues:
```yaml
documents:
  - "app/**/!(*.d).{graphql,ts,tsx,js,jsx}"
  - "!app/graphql/**"
```

milesrichardson:
Watch mode broke for us (it simply doesn't pickup changes) when migrating from `1.21.3` to `3.2.2`. Granted, that's a large change, so I've been trying to track down what broke it.

In our case, the breakage is that changes are _not_ picked up. My hypothesis was also around glob pattern matching - has the underlying glob matching library changed?

xli1000:
We also encountered problems building @parcel/watcher when upgrading. node-gyp dependencies seem to be much more troublesome/inconvenient than regular JS dependencies because there are often differences in the environment that cause the build to fail or there are network issues with node-gyp requesting certain files when building behind a corporate proxy.

PR #9039, opened 2023-02-21
rust plugin ci and docs update

saihaj:
Made the rust plugin flow to be triggered manually making and added documentation on release flow.

### DIFF
--- a/packages/presets/swc-plugin/README.md
+++ b/packages/presets/swc-plugin/README.md
@@ -16,7 +16,7 @@ You will need to provide the `artifactDirectory` path that should be the same as
 
 #### `.swrc`
 
-```json
+```json5
 {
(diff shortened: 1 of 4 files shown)


packages/presets/swc-plugin/README.md
```diff
@@ -0,0 +1,53 @@
...
+yarn add -D @graphql-codegen/client-preset-swc-plugin
+```
+
+### Usage
+
+You will need to provide the `artifactDirectory` path that should be the same as the one configured in your `codegen.ts`
+
+#### `.swrc`
+
+```json
```
n1ru4l, line 19:
```suggestion
```json5
```

n1ru4l (approved):

Issue #9045, opened 2023-02-23
@graphql-codegen/cli 3.2.0 watch mode regeneration triggers on all file changes (not just files specified in config documents pattern)


GrantASL19:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

In `@graphql-codegen/cli` 3.2.0 watch mode regeneration is triggered by editing any file in the project. In <=3.1.0 regeneration only triggers when a file in the config’s `documents` pattern is edited (correct me if I’m wrong!).

In our projects this is triggering many regenerations, e.g. seemingly as files are written into the Next.js build folder.

This seems to be a result of https://github.com/dotansimha/graphql-code-generator/pull/9009

Please let me know if I’m missing something! I tried passing a pattern as the `--watch` argument or `watch` config but it seems to make no difference. But IMO even if there’s a configuration workaround this was an unexpected breaking change in a minor release.

### Your Example Website or App

https://github.com/GrantASL19/graphql-codegen-cli-watch-regression-repro

### Steps to Reproduce the Bug or Issue

* `npm i`
* `npm run generate-graphql-watch`
* Edit `src/test.js` and `src/graphQl/staticPage.graphql` and note that the following is output when either file changes:
  ```
  ✔ Parse Configuration
  ✔ Generate outputs
  ℹ Watching for changes...
  ```
* Stop the watch with Ctrl-C
* `npm i @graphql-codegen/cli@3.1.0`
* `npm run generate-graphql-watch`
* Edit `src/test.js` and `src/graphQl/staticPage.graphql` and note that the regeneration is only triggered when `src/graphQl/staticPage.graphql` changes.

### Expected behavior

Regeneration should only trigger when I change a file specified in the config `documents` field.

### Platform

- OS: macOS 13.2.1
- NodeJS: 19.6.1
- `@graphql-codegen/cli` version 3.2.0

### Codegen Config File

```js
/** @type {import('graphql-config').IGraphQLConfig} */
const graphQlConfig = {
  documents: "./src/graphQl/**/*.graphql",
  extensions: {
    /** @type {import("@graphql-codegen/cli").CodegenConfig} */
    codegen: {
      generates: { 
        "./src/generated/graphQl.ts": {
          overwrite: true,
          plugins: [
            "typescript",
            "typescript-operations",
            "typescript-graphql-request",
          ],
          config: {
            avoidOptionals: {
              field: true,
              inputValue: false,
              object: false,
[...]

PR #9051, opened 2023-02-23
fix: regeneration should only happen for config document pattern

saihaj:
#9009 missed the fact that watch is running everywhere. Could not find a built in way with parcel but got a work around which I think is good enough for us.

closes https://github.com/dotansimha/graphql-code-generator/issues/9045

**Before**

https://user-images.githubusercontent.com/44710980/220997850-237a2aad-4068-4ba3-8550-8c7dd83319f6.mp4

**After**

https://user-images.githubusercontent.com/44710980/220997699-3ba4c7e3-1883-4626-921a-24f3a2b24242.mp4


packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -95,6 +96,12 @@ export const createWatcher = (
         // it doesn't matter what has changed, need to run whole process anyway
         await Promise.all(
           events.map(async ({ type: eventName, path }) => {
+            /**
+             * @parcel/watcher has no way to run watcher on specific files (https://github.com/parcel-bundler/watcher/issues/42)
+             * But we can use micromatch to filter out events that we don't care about
+             */
+            if (!mm.contains(path, files)) return;
```
brianmcd, line 103: @saihaj I believe there's still an issue here when a negated path is used.

My use case looks like this:

```yaml
documents:
  - './apps/frontend/src/app/**/*.graphql'
  - '!./apps/frontend/src/app/exclude-these/**/*.graphql'
```

I'm seeing the watcher run when any file in my source tree is changed, and I'm thinking it's due to `'!./apps/frontend/src/app/exclude-these/**/*.graphql'` causing the `mm.contains` call to return true.

Quick example with mm:
```javascript
const mm = require('micromatch');

// Prints false
console.log(mm.contains('anyfile.js', './some-dir/**/*.graphql'));

// Prints true
console.log(mm.contains('anyfile.js', '!./some-dir/**/*.graphql'));
```

ChristianIvicevic: I haven't checked the code, but have noticed an excessive amount of runs in watch mode since the update myself.

saihaj: If you can open issue with repro will be easier for me to debug and fix

ChristianIvicevic: I have downgraded to 3.1.0 and noticed that there is indeed a regression in newer versions causing the generator to run 5-10 times in a row (probably tracking changes in the `.next` folder). For now I had to remove all negated paths and be more specific with included directories which helped to reduce the amount of compilations.

Unfortunantely my project is not public and I don't necesarily have sufficient time to reduce it to a MWE, but I could consider sharing the repository with you @saihaj in case you'd be willing to look into it unless somebody has a decent reproduction.

ChristianIvicevic: A minor update: I found the `--debug` flag and confirmed that having negations causes them to be picked up causing the excess updates:
CleanShot 2023-03-11 at 16 46 27

djfarrelly: I did a similar investigation by using `--debug` and modifying the source directly and can confirm the same thing that Christian is seeing. I was able to remove the negated pattern and now things work as I want them do. 

Fortunately, our project doesn't need the negated pattern - I only added the negated pattern with v3.2.0 to try to fix the issue that this PR ended up fixing. I am sharing this to confirm the behavior that the other user is also seeing.

saihaj: That's a great find! Let's use this cause this PR notif got hidden in my inbox https://github.com/dotansimha/graphql-code-generator/issues/9177

milesrichardson: We have the exact same issue in our Next app, now tracking it here #9270 (instead of #9177)

Think I found the issue/root cause, working on a fix now

Issue #9055, opened 2023-02-23
Fragment names as "magic strings" (no hard references)


charpeni:
### Is your feature request related to a problem? Please describe.

I've been concerned by the fact that we can't embed fragments with template string interpolation, as it breaks TypeScript type-inference, as mentioned in the documentation of `gql-tag-operations`.

Currently, the suggested approach is to spread fragments as "magic strings", meaning there's no hard reference to the fragment itself. We need to find the right name (and make sure of it!), be aware that fragments are generated with a `Fragment` suffix, then spread it. All fragments are defined "globally" or within the same file (`graphql.ts`), so all fragments are accessible.

```tsx
const allFilmsWithVariablesQueryDocument = graphql(/* GraphQL */ `
  query allFilmsWithVariablesQuery($first: Int!) {
    allFilms(first: $first) {
      edges {
        node {
          ...FilmItem
        }
      }
    }
  }
`);
```

It is worth mentioning that spreading a non-existing fragment will lead to an error within the codegen step. However, not having hard references (e.g., importing the fragment) could hurt the application's maintainability.

Another concern is that name conflicts between operations or fragments are possible, but again, the codegen will warn us against it. My concern is mainly for grown applications, it could be possible to spread the wrong existing fragment via a typo (e.g., `WorkspacesCard` instead of `WorkspaceCard`).

I would like to have the ability the rely on hard references for fragments, I don't mean to make it mandatory, but having the ability to would be interesting.

### Describe the solution you'd like

Currently, there's no way to easily extract the name of a generated fragment.

One solution could be to make that possible so that we could at least rely on a hard reference on the name rather than the fragment definition (to mitigate the related issue). So implementing a helper to extract the fragment's name from the `DocumentNode` object via some AST utils.

E.g., with the following helper: `getFragmentName()`, return the name of the generated fragment as a string so we can compose fragments or queries with it.

```tsx
import { graphql, getFragmentName } from './gql';

export const FilmFragment = graphql(/* GraphQL */ `
  fragment FilmItem on Film {
    id
    title
    releaseDate
    producers
  }
`);

const fragmentName = getFragmentName(FilmFragment); // 'FilmItemFragment'
//    ^? const fragmentName = "FilmItemFragment"
```

> **Note**: There's still an unanswered question about how to expose fragment names?

We could either: 
- export it as a `const`
- expose fragments as before (via imports), and call `getFragmentName` from consumers
[...]

PR #9056, opened 2023-02-23
feat: implement `getFragmentName`

charpeni:
## Description

> **Warning**: 🚧 WIP 🚧 I still need to update the documentation.

Related #9055

Currently, there's no way to easily extract the name of a generated fragment.

This pull request adds the implementation for a helper function: `getFragmentName` to extract the fragment's name from the `DocumentNode` object via some AST utils.

It returns the name of the generated fragment as a string literal so we can compose fragments or queries with it:

```tsx
import { graphql, getFragmentName } from './gql';

export const FilmFragment = graphql(/* GraphQL */ `
  fragment FilmItem on Film {
    id
    title
    releaseDate
    producers
  }
`);

const fragmentName = getFragmentName(FilmFragment); // 'FilmItemFragment'
//    ^? const fragmentName = "FilmItemFragment"
```

There are two implementations introduced here:

First, we have the function itself to traverse the AST object and return the name of the `FragmentDefinitionNode`.

Then there's a specific TypeScript implementation so we could statically infer the fragment's name as a string literal from ` $fragmentName` field on the type. One advantage of that is that returning a `string` type isn't helpful, I would like to be able to have the fragment's name directly in my IDE (string literal) without having to open the generated code. To achieve this, I split `FragmentType` into a new type: `FragmentName` which is reused by `FragmentType` and the new helper function.

## Type of change

## How Has This Been Tested?

The code snippet above was tested in a generated example.

## Checklist:

### DIFF
--- a/dev-test/gql-tag-operations/graphql/fragment-masking.ts
+++ b/dev-test/gql-tag-operations/graphql/fragment-masking.ts
@@ -1,13 +1,21 @@
 import { ResultOf, TypedDocumentNode as DocumentNode } from '@graphql-typed-document-node/core';
+import { FragmentDefinitionNode } from 'graphql';
+
+export type FragmentName<TDocumentType extends DocumentNode<any, any>> = TDocumentType extends DocumentNode<
+  infer TType,
+  any
+>
+  ? TType extends { ' $fragmentName'?: infer TKey }
+    ? TKey
+    : never
+  : never;
 
 export type FragmentType<TDocumentType extends DocumentNode<any, any>> = TDocumentType extends DocumentNode<
   infer TType,
   any
 >
-  ? TType extends { ' $fragmentName'?: infer TKey }
-    ? TKey extends string
-      ? { ' $fragmentRefs'?: { [key in TKey]: TType } }
-      : never
+  ? FragmentName<TDocumentType> extends string
+    ? { ' $fragmentRefs'?: { [key in FragmentName<TDocumentType>]: TType } }
     : never
   : never;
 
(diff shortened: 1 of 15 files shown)


n1ru4l:
Hey @charpeni, first of all, thank you for all your contributions. 😍

Could you share any use-cases for this feature - might this be even something that could be implemented in user-land?

I also think that this conflicts with this option. 🤔  https://github.com/dotansimha/graphql-code-generator/blob/1edfa6ec8651c4e55b3d15d984430bcf1e908fa7/packages/presets/client/src/index.ts#L75-L81

charpeni:
My pleasure, thanks to all of you for this amazing project. ❤️ 

I've left more context and use cases in the associated issue: https://github.com/dotansimha/graphql-code-generator/issues/9055, but to quickly summarize the primary use case:

The `getFragmentName()` helper is helpful to retrieve the name of a giving fragment if you don't want to rely entirely on a "magic string" to consume/spread the fragment, e.g.:

```diff
+import { FilmItemFragment } from './FilmItem';

const allFilmsWithVariablesQueryDocument = graphql(/* GraphQL */ `
  query allFilmsWithVariablesQuery($first: Int!) {
    allFilms(first: $first) {
      edges {
        node {
-         ...FilmItem
+         ...${getFragmentName(FilmItemFragment)}
        }
      }
    }
  }
`);
```

Or, if you would like to follow [Apollo's naming convention](https://www.apollographql.com/docs/react/data/fragments#creating-colocated-fragments):

```tsx

FilmItem.fragments = {
  film: getFragmentName(FilmItemFragment),
};

// ...

const allFilmsWithVariablesQueryDocument = graphql(/* GraphQL */ `
  query allFilmsWithVariablesQuery($first: Int!) {
    allFilms(first: $first) {
      edges {
        node {
         ...${FilmItem.fragments.film}
        }
      }
    }
  }
`);
```

I totally agree with you, this is something that could be implemented in user-land—or even only generated behind a feature flag. I thought it could be handy for consumers that don't want to rely entirely on magic strings. Depending on how well GraphQL is integrated into the build tools, it could be harder to maintain fragment references without hard references, especially when the feedback loop is taking longer. Another concern would be a potential conflict between fragments, the codegen tool will prevent duplicated fragments, but it doesn't prevent you from typing the wrong one.

Quickly like this, I would see three options:

- We add `getFragmentName` to the default template for `fragment-masking.ts`
- We hide the generation of `getFragmentName` behind a feature flag / config option, e.g. `{ generateGetFragmentNameHelper: true }`
- We don't include this in the code at all, but we add a mention in the documentation

@n1ru4l let me know if you find this helpful and what your recommendation would be for this helper, I will be happy to adapt the pull request.

saihaj (approved):
thank you! I think we should have this as an option to generate instead being a default.

PR #9060, opened 2023-02-24
drop create-react-app + swc in-repository test

n1ru4l:
this replaces creatr-react-app examples with vite
It also adds an in-repository example for verifying that the swc plugin actually works.


JesseVelden:
The strange thing is that I cannot reproduce the issue locally. I will do some more digging

n1ru4l:
@JesseVelden @YassinEldeeb pinned this down to the rust version

JesseVelden:
What I also saw is that we need access to the namingConvention, as now the  `const allFilmsWithVariablesQueryDocument =` query example doesn't resolve to `AllFilmsWithVariablesQueryDocument` (note the uppercase A). 

I was assuming that you just need to do that. I guess we can provide it as an option to the SWC plugin, or warn the user instead. I'm not sure how the Babel plugin does that.

JesseVelden:
> @JesseVelden @YassinEldeeb pinned this down to the rust version

Oh lol, I wouldn't expect that. Great debugging work @YassinEldeeb !

packages/presets/swc-plugin/tests/fixtures/simple-uppercase-operation-name.other-dir.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */ //@ts-ignore
-import { CFragmentDoc } from "../graphql";
-import { BDocument } from "../graphql";
-import { ADocument } from "../graphql";
+import { CFragmentDoc } from "./../graphql";
+import { BDocument } from "./../graphql";
+import { ADocument } from "./../graphql";
```
n1ru4l, line 4: @YassinEldeeb The `./` is kind of redundant as `./../graphql` is the same as `../graphql` 🤔

n1ru4l: we can improve it in a follow-up PR

JesseVelden:
Great work guys! I hadn't much time to work on it this week

PR #9063, opened 2023-02-24
chore: add `build/` to `.prettierignore`

charpeni:
## Description

I noticed that Prettier was executed on `build` folders of generated examples, e.g.: `examples/react/apollo-client/build`.

We don't need to run Prettier on the generated builds from generated examples, especially since they are ignored by git.


dotansimha:
thank you @charpeni !

Issue #9043, opened 2023-02-22
resolver: Using {T} in a custom mapper loses Omits


ryanrhee:
### Which packages are impacted by your issue?

@graphql-codegen/cli, @graphql-codegen/typescript-resolvers

### Describe the bug

When using a custom mapper with `{T}`, the recursive definition is lost, and the default type from the typescript package is used instead.

https://github.com/dotansimha/graphql-code-generator/issues/1631 says that `{T}` should be the recursive definition.

### Your Example Website or App

https://codesandbox.io/s/reverent-snowflake-7npr0h?file=/codegen.ts

### Steps to Reproduce the Bug or Issue

1. Run `yarn generate` in a codesandbox terminal
2. Read `types.ts`'s `ResolverTypes["Post"]` (line 119), note the definition which is `ResolverTypeWrapper<Omit<Post, 'authorEdge'> & { authorEdge: ResolversTypes['UserEdge'] }>`
3. Go to `codegen.ts` and change the config by uncommenting line 11 so that `Post` gets a custom mapper
4. Re-run `yarn generate`
5. Read the same type in `types.ts`, and note that the new type is now `ResolverTypeWrapper<Omit<Post, 'tags'>>`.

### Expected behavior

At step 5, I expect the type to be:
```ts
ResolverTypeWrapper<Omit<Omit<Post, 'authorEdge'> & { authorEdge: ResolversTypes['UserEdge'] }, 'tags'>
```

### Platform

- OS: macOS
- NodeJS: 16.16.0
- `graphql` version: 16.6.0

```
"@graphql-codegen/cli": "2.16.3",
"@graphql-codegen/typescript": "2.7.1",
"@graphql-codegen/typescript-resolvers": "^2.7.3",
```

### Codegen Config File

```js
import { CodegenConfig } from "@graphql-codegen/cli";

const config: CodegenConfig = {
  schema: "schema.graphql",
  documents: "document.graphql",
  generates: {
    "types.ts": {
      config: {
        mappers: {
          UserEdge: "Omit<{T}, 'user'>"
          // Post: "Omit<{T}, 'tags'>"
          // Post is `ResolverTypeWrapper<Omit<Post, 'authorEdge'> & { authorEdge: ResolversTypes['UserEdge'] }>`
          // when mapper is not specified.
          //
          // I expect that whole type to replace `{T}`.
          //
          // However, the type after mapper specification is
[...]

PR #9069, opened 2023-02-26
[typescript-resolvers] Extract union types to ResolversUnionTypes

eddeee888:
## Description

As discussed in a proof of concept, we thought that by creating a `ResolversUnionTypes` type and use reference unions in `ResolversTypes` and `ResolversParentTypes` give us a bit more control in the future. 

As the result, the union types are no longer inlined in `ResolversTypes` and `ResolversParentTypes`

## Example use case (future)

In the future, we could make union members' `__typename` non-nullable whilst leaving the counterparts in `ResolversTypes` and `ResolversParentTypes` as optional:

```ts
// typescript plugin types
export type BookResult = {
  __typename?: "BookResult";
  result?: Maybe<Book>;
};

export type StandardError = {
  __typename?: "StandardError";
  error: string;
};

// typescript-resolvers types
export type ResolversUnionTypes = {
  BookPayload: (BookResult & { __typename: "BookResult" }) | (StandardError & { __typename: "StandardError" }); // Non-nullable typename are not implemented. This is here as proof of concept
};

  // ...
  BookPayload: ResolverTypeWrapper<ResolversUnionTypes["BookPayload"]>; // This forces `BookPayload` require __typename from its union members `BookResult` and `StandardError`
  BookResult: ResolverTypeWrapper<BookResult>; // __typename of BookResult is still optional
  StandardError: ResolverTypeWrapper<StandardError>;
  // ...
};

  // ...
  BookPayload: ResolversUnionTypes["BookPayload"]; // This means BookPayload.__resolveType has access to non-nullable __typename from its union members.
  BookResult: BookResult;
  StandardError: StandardError;
  // ...
};
```

## Type of change

- [x] Refactor: change how union types are referenced

## How Has This Been Tested?

- [x] Unit tests

**Test Environment**:

- OS: MacOS
- `@graphql-codegen/typescript-resolvers`:
- NodeJS: 16.15.0

## Checklist:


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -822,13 +824,7 @@ export class BaseResolversVisitor<
         // Make sure the inner type has no ResolverTypeWrapper
         const name = clearWrapper(isScalar ? this._getScalar(typeName) : prev[typeName]);
         const replaced = replacePlaceholder(this.config.defaultMapper.type, name);
-
-        // Don't wrap Union with ResolverTypeWrapper, each inner type already has it
-        if (isUnionType(schemaType)) {
-          prev[typeName] = replaced;
-        } else {
-          prev[typeName] = applyWrapper(replacePlaceholder(this.config.defaultMapper.type, name));
-        }
```
eddeee888, line 831: We now handle all union logic in `ResolversUnionTypes` so there's no need for special treatments here

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -880,6 +876,82 @@ export class BaseResolversVisitor<
     return `Array<${t}>`;
   }
 
+  protected createResolversUnionTypes(): Record<string, string> {
```
eddeee888, line 879: `createResolversUnionTypes` is the simplified version of `createResolversFields` targeted for Unions. 
I chose to create a new function because `createResolversFields` is quite large and rigid so it's hard to implement/maintain

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -880,6 +876,82 @@ export class BaseResolversVisitor<
...
+
+              return {
+                addOptionalSign: !this.config.avoidOptionals && !isNonNullType(field.type),
+                fieldName: field.name,
+                replaceWithType: wrapTypeWithModifiers(this.getTypeToUse(baseType.name), field.type, {
+                  wrapOptional: this.applyMaybe,
+                  wrapArray: this.wrapWithArray,
+                }),
+              };
+            })
```
eddeee888, line 925: This is exactly the same as the one in `createResolversFields`, will need to abstract to reuse

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -880,6 +876,82 @@ export class BaseResolversVisitor<
...
+          const isScalar = this.config.scalars[typeName];
+
+          if (hasDefaultMapper && hasPlaceholder(this.config.defaultMapper.type)) {
+            const finalTypename = isScalar ? this._getScalar(typeName) : unionMemberValue;
+            return replacePlaceholder(this.config.defaultMapper.type, finalTypename);
+          }
+
+          return unionMemberValue;
+        });
+        res[typeName] = referencedTypes.map(type => `( ${type} )`).join(' | '); // Must wrap every union member in explicit "( )" to separate the members
```
eddeee888, line 947: > Must wrap every union member in explicit "( )" to separate the members

Result usually looks like this:
```ts
export type ResolversUnionTypes =  {
  UnionThing: ( TypeA ) | ( TypeB )
}
```
Prettier should remove the round brackets automatically i.e. result will be `UnionThing: TypeA | TypeB`

I've added it this way so it's a bit easier to add types to the union members without ambiguity e.g.
```ts
UnionThing: TypeA & { __typename: 'TypeA' } | TypeB & { __typename: 'TypeB' } // This is ambiguous

// Prettier will try to turn it into this anyways:
UnionThing: ( TypeA & { __typename: 'TypeA' } ) | ( TypeB & { __typename: 'TypeB' } ) 
```

packages/plugins/typescript/resolvers/tests/ts-resolvers.spec.ts
```diff
@@ -1426,10 +1426,38 @@ export type ResolverFn<TResult, TParent, TContext, TArgs> = (
       union CCCUnion = CCCFoo | CCCBar
     `);
 
-    const tsContent = (await tsPlugin(testSchema, [], {}, { outputFile: 'graphql.ts' })) as Types.ComplexPluginOutput;
+    const tsContent = await tsPlugin(testSchema, [], {}, { outputFile: 'graphql.ts' });
```
eddeee888, line 1429: `tsContent` should be typed as `Types.ComplexPluginOutput` so we don't need to do `as Types.ComplexPluginOutput`.  

Happy to put this back if it was here for some particular reasons

packages/plugins/typescript/resolvers/tests/mapping.spec.ts
```diff
@@ -7,23 +7,48 @@ describe('ResolversTypes', () => {
...
+        Subscription: {};
+        Node: ResolversParentTypes['SomeNode'];
+        ID: Scalars['ID'];
+        SomeNode: SomeNode;
+        MyUnion: ResolversUnionTypes['MyUnion'];
+        MyScalar: Scalars['MyScalar'];
+        Int: Scalars['Int'];
+        Boolean: Scalars['Boolean'];
+      };
+    `);
```
eddeee888, line 51: Added `ResolversParentTypes` to all related tests in this file because I want to be sure that `ResolversUnionTypes` is being used correctly.

packages/plugins/typescript/resolvers/tests/mapping.spec.ts
```diff
@@ -7,23 +7,48 @@ describe('ResolversTypes', () => {
...
     expect(result.content).toBeSimilarStringTo(`
     export type ResolversTypes = {
       MyType: ResolverTypeWrapper<Omit<MyType, 'unionChild'> & { unionChild?: Maybe<ResolversTypes['ChildUnion']> }>;
       String: ResolverTypeWrapper<Scalars['String']>;
       Child: ResolverTypeWrapper<Child>;
       MyOtherType: ResolverTypeWrapper<MyOtherType>;
-      ChildUnion: ResolversTypes['Child'] | ResolversTypes['MyOtherType'];
+      ChildUnion: ResolverTypeWrapper<ResolversUnionTypes['ChildUnion']>;
```
eddeee888, line 22: Previously, this is `ResolversTypes['Child'] | ResolversTypes['MyOtherType']` which translate to:

```ts
ResolverTypeWrapper<Child> | ResolverTypeWrapper<MyOtherType>
```

Now, using `ResolverTypeWrapper<ResolversUnionTypes['ChildUnion']>` translate to:

```ts
ResolverTypeWrapper<Child | MyOtherType>
```

I think these 2 are the same semantically. Happy to hear explanation if they are not 🙂

packages/plugins/typescript/resolvers/tests/mapping.spec.ts
```diff
@@ -332,23 +465,51 @@ describe('ResolversTypes', () => {
...
 
     expect(result.prepend).toContain(`import { CustomPartial } from './my-wrapper';`);
+    // TODO: eddeee888 check if ResolversUnionTypes.MyUnion is supposed to be which?
+    //   - CustomPartial<Omit<MyType, 'unionChild'> & { unionChild?: Maybe<ResolversTypes['ChildUnion']> }>
+    //   OR
+    //   - CustomPartial<MyType>
+    expect(result.content).toBeSimilarStringTo(`
+      export type ResolversUnionTypes = {
+        ChildUnion: ( Omit<Child, 'parent'> & { parent?: Maybe<ResolversTypes['MyType']> } ) | ( MyOtherType );
+        MyUnion: ( CustomPartial<Omit<MyType, 'unionChild'> & { unionChild?: Maybe<ResolversTypes['ChildUnion']> }> ) | ( MyOtherType );
```
eddeee888, line 475: This could be related to https://github.com/dotansimha/graphql-code-generator/issues/9043 as well.

Ideally, the `MyType` is should be treated the same way as `ResolversParentTypes['MyType']`. However:
- The `ResolversParentTypes['MyType']` is using the type from typescript plugin. 
- The mentioned issue seems to suggest that we should use the "Omit type" instead 

Currently, `ResolversUnionTypes` is using "Omit type" to replace the `{ T }` placeholder here but happy to change to match whichever direction `ResolversParentTypes['MyType']` uses.

cc @saihaj @gilgardosh

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
eddeee888: This comment is no longer relevant due to change in implementation details

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -819,15 +788,15 @@ export class BaseResolversVisitor<
...
-        // Don't wrap Union with ResolverTypeWrapper, each inner type already has it
         if (isUnionType(schemaType)) {
-          prev[typeName] = replaced;
+          // Don't clear ResolverTypeWrapper from Unions
+          prev[typeName] = replacePlaceholder(this.config.defaultMapper.type, originalTypeName);
         } else {
-          prev[typeName] = applyWrapper(replacePlaceholder(this.config.defaultMapper.type, name));
+          const name = clearWrapper(originalTypeName);
+          const replaced = replacePlaceholder(this.config.defaultMapper.type, name);
+          prev[typeName] = applyWrapper(replaced);
```
eddeee888, line 799: Previous logic suggests that it's possible for a union to be in the scalar map 🤔 I'm unsure when/if it's possible. 

However, the updated logic should be backward compatible as it generates the same partial and wrapper order.

n1ru4l (approved):

packages/plugins/typescript/resolvers/tests/mapping.spec.ts
ryanrhee: FYI, according to #1631, the parent type should use the "Omit type", and #1677 implements it. Not sure when it broke between then and now, but I suspect it was unintentional.

gilgardosh: @eddeee888 awesome work!
I implemented a fix here #9110 , but it's conflicting. I'll adjust rebase it on top of yours after it's merged

gilgardosh (approved):

hector-del-rio:
This caused an issue in my project due to the fact that now the first argument for `__resolveType` resolver function definition is wrapped in `ResolverTypeWrapper` where before it wasn't, which makes it possibly a promise.

ryanrhee:
`type ResolverTypeWithoutPromise = Exclude<ResolverTypeWrapper, Promise<unknown>>` should do what you want.

ryanrhee:
you can also extract all the sub-types as well.

```ts
import type { ResolversTypes } from "./path/to/generated/graphql"

type ShallowTypes = {
  [Key in keyof ResolversTypes]: Exclude<ResolversTypes[Key], Promise<unknown>>
}
```

hector-del-rio:
@ryanrhee thanks for your answer. The issue is with the parent argument (first argument) of the resolver for `__resolveType` field.

```
export type ResolversParentTypes = ResolversObject<{
  ...
  MyType: ResolversUnionTypes['MyType'];
  AnotherType: ResolverTypeWrapper<Mapper>
  ...
}>;

export type ResolversUnionTypes = ResolversObject<{
  MyType: { contract: ResolversTypes['AnotherType'] }; // This now uses ResolversTypes which uses ResolverTypeWrapper
}>;
```

I guess the only way to fix it is to add this type to the mappers as well?

ryanrhee:
You could add it to the mappers as well.

Does `__resolveType()` allow async functions? If so, you could check if the first argument is a promise and then resolve it so that you are always operating on the underlying type.

```ts
{
  __resolveType: async (objOrPromise, ...) => {
    // obj is `Promise<T> | T`
    const obj = Promise.resolve(objOrPromise)
    // your code here; obj is now of type `T`
  }
}
```

hector-del-rio:
Thanks @ryanrhee. I needed to add `await`, but it worked:

```
{
  __resolveType: async (objOrPromise, ...) => {
    // objOrPromise is `Promise<T> | T`
    const obj = await Promise.resolve(objOrPromise)
    // your code here; obj is now of type `T`
  }
}
```

However, it is a bit concerning that this happened in a patch update from `@graphql-codegen/typescript-resolvers` `3.1.0` to `3.1.1`.

eddeee888:
Hi @hector-del-rio , 

Do you mind creating a small repo or sandbox with a small schema and minimal codegen config? This will help me understand the problem better 🙂

For context, this went in as a patch because we only move type references around. Maybe I missed a use case. Sorry!

eddeee888:
Hi @hector-del-rio, 

This should be fixed in https://github.com/dotansimha/graphql-code-generator/pull/9206 🙂
It'll go out in the next version. Thanks for reporting the issue 🙌

hector-del-rio:
Thank you @eddeee888 🙏

Issue #8862, opened 2023-01-30
Support `.mts` extension for codegen file / package.json "type": "module"


n1ru4l:
### Is your feature request related to a problem? Please describe.

It is currently impossible to have a `codegen.mts` config.

### Describe the solution you'd like

Support the `.mts` extension.

### Describe alternatives you've considered

Use `import()` instead of `require()` in order to properly support importing all kinds of modules.

n1ru4l:
See https://github.com/davidtheclark/cosmiconfig/issues/224

mirismaili:
AFAIK when `"type": "module"` we don't need `.mjs`/`.mts` extension.

My problem is I have `"type": "module"` in my "package.json" file. And I get this error:
```bash
> graphql-codegen --config codegen.ts

TypeScriptLoader failed to compile TypeScript:
Must use import to load ES Module: codegen.ts
require() of ES modules is not supported.
require() of codegen.ts from node_modules/cosmiconfig-typescript-loader/dist/cjs/index.js is an ES module file as it is a .ts file whose nearest parent package.json contains "type": "module" which defines all .ts files in that package scope as ES modules.
Instead change the requiring code to use import(), or remove "type": "module" from package.json.
```

So I was forced to use yaml config format.

xiaoxinghu:
Having exactly the same issue.

n1ru4l:
@xiaoxinghu For now the following workaround is recommended:

https://github.com/dotansimha/graphql-code-generator/issues/8509#issuecomment-1418270837

> for those running into this issue where a static YAML file is not a solution, you can rename your file to `codegen.cjs`
> 
> ```js
> /** @type {import('@graphql-codegen/cli').CodegenConfig} */
> const config = {
>   // ...
> }
> 
> module.exports = config
> ```
> 
> and then modify your `scripts` in `package.json` accordingly
> 
> ```json
> "scripts": {
>   "codegen": "graphql-codegen --config codegen.cjs",
> }
> ```

Issue #8809, opened 2023-01-11
Unmet peer dependencies from `cosmiconfig-typescript-loader` and `graphql-config`


mhassan1:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

I am seeing the following peer dependency warnings with Yarn Berry:
```txt
➤ YN0002: │ @graphql-codegen/cli@npm:2.16.4 doesn't provide @types/node, requested by cosmiconfig-typescript-loader
➤ YN0002: │ @graphql-codegen/cli@npm:2.16.4 doesn't provide typescript, requested by cosmiconfig-typescript-loader
➤ YN0002: │ @graphql-codegen/cli@npm:2.16.4 doesn't provide cosmiconfig-toml-loader, requested by graphql-config
```

### Your Example Website or App

N/A

### Steps to Reproduce the Bug or Issue

1. Use Yarn Berry
2. `yarn add @graphql-codegen/cli@2.16.4`

### Expected behavior

No peer dependency warnings

### Platform

- OS: MacOS
- NodeJS: 18.12.1
- `graphql` version: 16.6.0
- `@graphql-codegen/*` version(s): 2.16.4

aaronadamsCA:
pnpm reports an additional missing `@babel/core` peer dependency:

```
 WARN  Issues with peer dependencies found
.
└─┬ @graphql-codegen/cli 2.16.4
  ├── ✕ missing peer ts-node@>=10
  ├─┬ @graphql-tools/code-file-loader 7.3.16
  │ └─┬ @graphql-tools/graphql-tag-pluck 7.4.3
  │   └─┬ @babel/plugin-syntax-import-assertions 7.20.0
  │     └── ✕ missing peer @babel/core@^7.0.0-0
  ├─┬ @graphql-tools/prisma-loader 7.2.53
  │ └─┬ @graphql-tools/url-loader 7.17.2
  │   └─┬ @graphql-tools/executor-http 0.1.1
  │     └─┬ meros 1.2.1
  │       └── ✕ missing peer @types/node@>=13
  ├─┬ cosmiconfig-typescript-loader 4.3.0
  │ ├── ✕ missing peer @types/node@"*"
  │ └── ✕ missing peer ts-node@>=10
  └─┬ graphql-config 4.4.0
    └── ✕ missing peer cosmiconfig-toml-loader@^1.0.0
Peer dependencies that should be installed:
[...]

Issue #8509, opened 2022-10-21
ViteJS + GraphQL code generator = codegen.ts config bug


forestjava:
### Describe the bug

TypeScriptLoader failed to compile the Codegen Config File `codegen.ts`
Please find `codegen.ts`, `tsconfig.json` and the entire my-app.zip attached.

[codegen.ts.zip](https://github.com/dotansimha/graphql-code-generator/files/9839251/codegen.ts.zip)

[tsconfig.json.zip](https://github.com/dotansimha/graphql-code-generator/files/9839252/tsconfig.json.zip)

[my-app.zip](https://github.com/dotansimha/graphql-code-generator/files/9839258/my-app.zip)

### Your Example Website or App

npm run codegen

### Steps to Reproduce the Bug or Issue

1) create React Typescript application via
```bash
npm create vite@latest my-app
cd my-app
npm install
```
2) install yours
```bash
npm install graphql
npm install -D @graphql-codegen/cli
```
3) initialize Application built with React
```bash
npx graphql-code-generator init
npm install
```
4) try to call codegen
```bash
npm run codegen
```

### Expected behavior

As a user I expected code generation, but getting the error:
```
TypeScriptLoader failed to compile TypeScript:
Must use import to load ES Module: \my-app\codegen.ts
require() of ES modules is not supported.
require() of \my-app\codegen.ts from \my-app\node_modules\cosmiconfig-typescript-loader\dist\cjs\index.js is an ES module file as it is a .ts file whose nearest parent package.json contains "type": "module" which defines all .ts files in that package scope as ES modules.
Instead change the requiring code to use import(), or remove "type": "module" from \my-app\package.json.
```

### Platform

- OS: Windows 10
- NodeJS: 16.17.1
- `graphql` version: ^16.6.0
- `@graphql-codegen/*` version: 2.13.7
- `@graphql-codegen/client-preset` version: 1.0.7

### Codegen Config File
```typescript

import type { CodegenConfig } from '@graphql-codegen/cli';

const config: CodegenConfig = {
  overwrite: true,
  schema: "http://localhost:4000",
  documents: "src/**/*.tsx",
  generates: {
    "src/gql": {
      preset: "client",
      plugins: []
    }
  }
};

export default config;
```

gterras:
Same issue from a fresh svelltekit install : 

`error TS5095: Option 'preserveValueImports' can only be used when 'module' is set to 'es2015' or later.`

[...]

PR #9086, opened 2023-03-01
Load codegen.ts in ESM projects

beerose:
## Description

This PR handles loading `codegen.ts` in projects using ESM. Some of the approaches I tried:

- Changing `cosmiconfig`'s configuration (nothing relevant that would make this work out of the box)
- Registering TS with `ts-node` (I tried different configurations to be able to use await import)
- Using `eval`: `eval("import("pathname")"`

I originally went with `execSync(tsc ...)` to compile the config file to CommonJS and cache the results. But the caching part has a bunch of edge cases so I took a different approach and used `jiti` to load the file.

<details>
<summary><b>See benchmarks ⏳</b></summary>
<br />
<br />
**ESM path — tsc, cache miss:**
```
➜  vite-react-ts git:(vite-configts-fix) ✗ time yarn codegen  
yarn run v1.22.19
$ graphql-codegen --config codegen.ts
✔ Parse Configuration
✔ Generate outputs
✨  Done in 7.46s.
yarn codegen  7.81s user 0.69s system 111% cpu 7.649 total
```

**ESM path — tsc, cache hit:**
```
➜  vite-react-ts git:(vite-configts-fix) ✗ time yarn codegen
yarn run v1.22.19
$ graphql-codegen --config codegen.ts
✔ Parse Configuration
✔ Generate outputs
✨  Done in 3.69s.
yarn codegen  3.22s user 0.38s system 92% cpu 3.896 total
```

**ESM path — jiti:**
```
➜  vite-react-ts git:(vite-configts-fix) ✗ time yarn codegen
yarn run v1.22.19
$ graphql-codegen --config codegen.ts
✔ Parse Configuration
✔ Generate outputs
✨  Done in 3.91s.
yarn codegen  3.33s user 0.36s system 90% cpu 4.094 total
```

**.cts — jiti:**
```
➜  vite-react-cts git:(vite-configts-fix) ✗ time yarn codegen
yarn run v1.22.19
$ graphql-codegen --config codegen.cts
✔ Parse Configuration
✔ Generate outputs
✨  Done in 3.36s.
yarn codegen  3.21s user 0.34s system 99% cpu 3.558 total
```

**Non-ESM path:**
```
➜  apollo-client git:(vite-configts-fix) ✗ time yarn codegen      
yarn run v1.22.19
$ graphql-codegen --config codegen.ts
✔ Parse Configuration
✔ Generate outputs
✨  Done in 3.92s.
yarn codegen  3.22s user 0.38s system 87% cpu 4.116 total
```
</details>

Some of the next steps are:
- [x] Make the same change `graphql-config` as well (same issue) — PR: https://github.com/kamilkisiela/graphql-config/pull/1286

Related Closes https://github.com/dotansimha/graphql-code-generator/issues/8509, Closes https://github.com/dotansimha/graphql-code-generator/issues/8862, Closes https://github.com/dotansimha/graphql-code-generator/issues/8809

## Type of change

## How Has This Been Tested?

- [x] Created a new example based on reproductions steps in the issue

**Test Environment**:

- OS: macOS
- NodeJS: 19.x

## Checklist:


enisdenjo (commented):
Your code is very clean and easy to read - this looks great!

About the ESM detection, maybe I am too naive, but couldn't we just look for the `import` keyword in the config? We can use a regex for increased accuracy, something like `import (.* from|"|')`.

Also, a test case for this would be great.

beerose:
> About the ESM detection, maybe I am too naive, but couldn't we just look for the `import` keyword in the config? We can use a regex for increased accuracy, something like `import (.* from|"|')`.

Hmm, I'm not sure. You can have a TypeScript file with ESM syntax that technically is a CommonJS and can be _required_ by ts-node. Notice that this bug only occurred with `"type": "module"` in `package.json`, and the workaround was to use `.cjs` instead of `.ts` for the config — [link](https://github.com/dotansimha/graphql-code-generator/issues/8509#issuecomment-1418270837).

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -142,7 +186,16 @@ export async function loadCodegenConfig({
 }
 
 export async function loadContext(configFilePath?: string): Promise<CodegenContext> | never {
-  const graphqlConfig = await findAndLoadGraphQLConfig(configFilePath);
+  let graphqlConfig: GraphQLConfig | undefined | void;
+  try {
+    graphqlConfig = await findAndLoadGraphQLConfig(configFilePath);
+  } catch (err) {
+    if (isRequireESMError(err)) {
+      // TODO: This needs a fix in graphql-config
```
beerose, line 194: I don't think it's a blocker but something that we should fix sooner or later. What do you think?

enisdenjo: I agree, especially because graphql-config is quite popular.

packages/graphql-codegen-cli/package.json
```diff
@@ -59,12 +59,12 @@
...
     "chalk": "^4.1.0",
     "cosmiconfig": "^7.0.0",
-    "cosmiconfig-typescript-loader": "^4.3.0",
     "debounce": "^1.2.0",
     "detect-indent": "^6.0.0",
-    "graphql-config": "^4.4.0",
+    "graphql-config": "^4.5.0",
     "inquirer": "^8.0.0",
     "is-glob": "^4.0.1",
+    "jiti": "^1.17.1",
```
B2o5T, line 67: I guess you can remove it

enisdenjo (approved):
This looks great to me! I like that we dropped the cosmiconfig-typescript-loader especially.

We can go ahead and merge after the CI's all green.

packages/graphql-codegen-cli/package.json
```diff
@@ -93,7 +93,13 @@
...
   },
   "peerDependencies": {
-    "graphql": "^0.8.0 || ^0.9.0 || ^0.10.0 || ^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0 || ^16.0.0"
+    "graphql": "^0.8.0 || ^0.9.0 || ^0.10.0 || ^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0 || ^16.0.0",
+    "typescript": ">=3.0"
+  },
+  "peerDependenciesMeta": {
+    "typescript": {
+      "optional": true
+    }
```
enisdenjo, line 102: Is typescript a peer dep still though?

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -74,15 +74,16 @@ function customLoader(ext: 'json' | 'yaml' | 'js' | 'ts') {
     }
 
     if (ext === 'ts') {
-      // #8437: conflict with `graphql-config` also using TypeScriptLoader(), causing a double `ts-node` register.
-      const tsLoader = TypeScriptLoader({ transpileOnly: true });
-      return tsLoader(filepath, content);
+      const jiti = require('jiti')(__filename, {
```
B2o5T, line 77: You can use `import`

B2o5T:
> This looks great to me! I like that we dropped the cosmiconfig-typescript-loader especially.
> 
> We can go ahead and merge after the CI's all green.

@enisdenjo I think we can remove ts-node too now?

packages/graphql-codegen-cli/package.json
beerose: Yeah, we use it in `config.ts`

packages/graphql-codegen-cli/package.json
beerose: Removed

packages/graphql-codegen-cli/src/config.ts
beerose: Ah right. We do bundle to both. What's the best way to deal with it? Something like this: 
```
import { fileURLToPath } from 'node:url';
const __filename = fileURLToPath(import.meta.url)
```
?

B2o5T: @enisdenjo could we add banner with bob? Or should we bundle to Esm only instead?

B2o5T: @beerose yeah but we need to add it only to ESM build 😅

beerose:
> > I think we can remove ts-node too now?
> 
> Yes sir! I think it was only necessary for cosmiconf.

Is there any documentation that needs to be updated regarding this?

saihaj (approved):

saihaj:
> > > I think we can remove ts-node too now?
> > 
> > 
> > Yes sir! I think it was only necessary for cosmiconf.
> 
> Is there any documentation that needs to be updated regarding this?

I think we are good here. `ts-node` already caused issues https://github.com/dotansimha/graphql-code-generator/issues/8809 so I suppose now we also fixed issues for `yarn pnp` users

packages/graphql-codegen-cli/src/config.ts
beerose: @B2o5T this is what was generated in `graphql-config` in ESM out of the box:

```
const __filename = fileURLToPath(import.meta.url);
```

What's the difference between the build setup? Ie can we have this also in this package?

B2o5T: @beerose i added it manually in your PR 😅 https://github.com/kamilkisiela/graphql-config/blob/master/scripts/postbuild.ts

beerose: should we do the same thing here?

beerose: @saihaj, do you have any thoughts on that?

saihaj: I think our script is fine for simplicity. We do some binary fixing too with scripts. Ideally `bob` should have some `hooks` that we can use to call scripts as needed.

B2o5T: > should we do the same thing here?

if there is no other workaround..

beerose: Ok, so it looks like we can workaround it but not provide __filename at all — jiti resolves it to `process.cwd()` which I guess it's correct? https://github.com/unjs/jiti/blob/main/src/jiti.ts#L101

n1ru4l (approved):

PR #9088, opened 2023-03-01
Add `addToSchema` to Document Transform

kazekyo:
## Description
I have added `addToSchema` to Document Transform. This is based on  https://github.com/dotansimha/graphql-code-generator/discussions/8724.

With this change, you can extend a type in the schema using both `documents` and `schema` before transforming `documents` with a document transform.  You can generate type information for a local-only field in your document transform, thus avoiding validation errors.

If you want to learn more about the detailed background or different implementation methods along with their pros and cons, please refer to https://github.com/dotansimha/graphql-code-generator/discussions/8724.

## Type of change

Please see documentation for usage.

## How Has This Been Tested?

I added tests to the spec file.

## Checklist:

### DIFF
--- a/.changeset/pretty-peaches-lay.md
+++ b/.changeset/pretty-peaches-lay.md
@@ -1,0 +1,72 @@
+---
+'@graphql-codegen/plugin-helpers': minor
+'@graphql-codegen/core': minor
+'@graphql-codegen/cli': minor
+---
+
+Add `addToSchema` to Document Transform.
+
+Now, before transforming `documents` with a document transform, you can extend a type in the schema using both `documents` and `schema`. With this change, you can generate type information for a local-only field in your document transform, thus avoiding validation errors.
+
+For instance, the following is an example of a document transform that adds a field named `localOnlyFieldForMyTool` when a `@useMyTool` directive is used:
+
+```ts
+const directiveName = 'useMyTool';
+const localOnlyFieldName = 'localOnlyFieldForMyTool
+const documentTransform: Types.DocumentTransformObject = {
+  transform: ({ documents }) => {
+    return documents.map(documentFile => {
+      documentFile.document = visit(documentFile.document, {
+        Field: {
+          leave(fieldNode) {
+            if (!fieldNode.directives) return undefined;
+            const addFieldDirective = fieldNode.directives.find(
+              directive => directive.name.value === directiveName
+            );
+            if (!addFieldDirective) return undefine
+            const localOnlyField: FieldNode = {
+              kind: Kind.FIELD,
+              name: { kind: Kind.NAME, value: localOnlyFieldName },
+              directives: [{ kind: Kind.DIRECTIVE, name: { kind: Kind.NAME, value: 'client' } }],
+            };
+            return {
+              ...fieldNode,
+              selectionSet: {
+                ...fieldNode.selectionSet!,
+                selections: [...fieldNode.selectionSet!.selections, localOnlyField],
+              },
+            };
+          },
+        },
+      });
+      return documentFile;
+    });
+  },
+  addToSchema: ({ schemaAst, documents }) => {
+    const typeInfo = new TypeInfo(schemaAst);
+    const typeNames = [];
+    visit(
+      concatAST(documents.map(file => file.document)),
+      visitWithTypeInfo(typeInfo, {
+        Field: {
+          leave(fieldNode) {
+            if (!fieldNode.directives) return;
+            const addFieldDirective = fieldNode.directives.find(
+              directive => directive.name.value === directiveName
+            );
+            if (!addFieldDirective) retur
+            const type = typeInfo.getType();
+            if (isNonNullType(type) && isObjectType(type.ofType)) {
+              typeNames.push(type.ofType.name);
+            }
+          },
+        },
+      })
+    );
+    if (typeNames.length > 0) {
+      return typeNames.map(name => `extend type ${name} { ${localOnlyFieldName}: String! }`).j('\n');
+    }
+    return '';
+  },
+};
+```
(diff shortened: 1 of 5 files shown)


kazekyo:
@n1ru4l Could you review this or assign someone as a reviewer?

n1ru4l:
Hey @kazekyo I am not sure we need this redundant option as the dame could also be achieved by adding multiple schema entries in your config. 🤔 What benefit does this option bring over that?

kazekyo:
@n1ru4l I may not have clearly communicated to you that the key point is that it's the tools, not the users, that automatically add type information based on `schema` and `documents`. 
As I understand it, if a directive is specified within a query, and a document transform adds local-only fields according to that directive, users currently have to manually add the type information for those fields to the schema via schema entries in the config. 
I want to eliminate this user task and the associated stress.

I quote the following from https://github.com/dotansimha/graphql-code-generator/discussions/8724:

> # Background
> It is already possible to define a schema for local-only fields, as shown below:
> 
> ```tsx
> import type { CodegenConfig } from '@graphql-codegen/cli';
> 
> const config: CodegenConfig = {
>   schema: ['https://localhost:4000/graphql', 'extend type Hoge { myLocalOnlyField: String! }']
>   documents: ['src/**/*.tsx'],
>   generates: {
>     './src/gql/': {
>       preset: 'client',
>     },
>   },
> };
> export default config;
> ```
> 
> However, users must manually add the necessary types to the schema. I want the tool to automatically add the type to the schema when a specific directive is included in `documents` or when a specific type is present in the schema.
> 
> # Main Use Case
> My use case is for implementing something like Relay's `__id`. The `__id` is a unique identifier assigned to a `Connection` when a user is using it, which is then used as a cache key when adding or removing elements from that `Connection`.
> 
> In my implementation, if a `@pagination` directive (an alias for `@connection`) is present in `documents`, a `_connectionId` (an alias for `__id`) type is added to a `SomethingConnection`.
> 
> ```graphql
> extend type SomethingConnection {
>   _connectionId: String!
> }
> ```
> 
> Both `documents` and `schema` are required to generate the schema for this tool.
> 
> More generally, use cases involve adding local-only fields when a directive is used. For example, assuming that there is a `@iconInfo` directive.
> 
> ```graphql
> query MyQuery {
>   user @iconInfo {
>     id
>     name
>   }
> }
> ```
> 
> And there is a document transform that automatically adds an `iconUrl` field and an `iconDownloaded` field when the `@iconInfo` directive is included, resulting in the following document:
> 
> ```graphql
> query MyQuery {
>   user {
>     id
>     name
>     iconUrl # Added by the document transform
>     iconDownloaded @client # Added by the document transform
>   }
> }
> ```
> 
> In this case, I want to add the type for the local-only field `iconDownloaded` automatically. However, although the field is automatically added by the document transform, the schema information must be manually added by the user, which is inconvenient.
>

kazekyo:
@n1ru4l Did I explain that clearly?
This is a very important feature for my tool's users and me. If an existing feature can do the same thing, that's fine, so please tell me about it 🙏

kazekyo:
If you would like to choose a different method of implementation than this one, we have several methods. 
I quote the following from: https://github.com/dotansimha/graphql-code-generator/discussions/8724

> # Proposal **of API**
> Several methods are possible.
> 
> ## Extending the `schema` field
> To be able to add a function to the `schema` field that is executed after the documents and schema are read.
> 
> ```tsx
> import type { CodegenConfig } from '@graphql-codegen/cli';
> 
> const rootSchema = ['https://localhost:4000/graphql']
> const config: CodegenConfig = {
>   schema: rootSchema,
>   documents: ['src/**/*.tsx'],
>   generates: {
>     './src/gql/': {
>       preset: 'client',
>       documentTransforms: [something],
>       schema: [
>         ...rootSchema,
>         ({ schema, documents }) => {
>           return generateMyPartialSchemaString(schema, documents);
>         }
>       ],
>     },
>   },
> };
> export default config;
> ```
> 
> However, this will not work with YAML. It is also possible to specify the function's file name, but this may not be the best idea due to implementation complexity and user-friendliness because there are already many ways to specify strings. Also, users may not understand the order of loading, since this function must be executed after the loading of the other schemas.
> 
> ## Adding `addToSchema` to Document Transform
> As far as I understand use cases, adding `addToSchema` to Document Transform seems reasonable as many use cases are closely related to Document Transform. Since plugins already have `addToSchema`, it should be easy to understand.
> 
> ```tsx
> import type { CodegenConfig } from '@graphql-codegen/cli';
> 
> const config: CodegenConfig = {
>   schema: ['https://localhost:4000/graphql'],
>   documents: ['src/**/*.tsx'],
>   generates: {
>     './src/gql/': {
>       preset: 'client',
>       documentTransforms: [
>         {
>           transform: ({ pluginContext, documents }) => {
>             return documents;
>           },
>           addToSchema: ({ schema, documents }) => {
>             return generateMyPartialSchemaString(schema, documents);
>           },
>         },
>       ],
>     },
>   },
> };
> export default config;
> ```
> 
> Document Transform can also be specified in YAML using file names.
> 
> ## Creating Schema Transform
> It is possible to create a Schema Transform similar to Document Transform.
> 
> ```tsx
> import type { CodegenConfig } from '@graphql-codegen/cli';
> 
> const config: CodegenConfig = {
>   schema: ['https://localhost:4000/graphql'],
>   documents: ['src/**/*.tsx'],
>   generates: {
>     './src/gql/': {
>       preset: 'client',
>       documentTransforms: [
>         {
>           transform: ({ documents }) => {
>             return documents;
>           },
>         },
>       ],
>       schemaTransforms: [
>         {
>           transform: ({ schema, documents }) => {
>             return generateNewSchema(schema, documents);
>           },
>         },
>       ],
>     },
>   },
> };
> export default config;
> ```
> 
> This mechanism is powerful as it allows not only adding types but also editing or deleting types in the schema. However, it may be over-engineering.
> 
> ## My opinion
> Considering the necessary use cases at present, I think adding `addToSchema` to Document Transform is a good option.

n1ru4l:
Hey @kazekyo, I still feel that this should instead be provided via the already existing schema array option. You would provide these anyways for framework-specific stuff (e.g. the `@connection` directive) in order to get proper IDE integration with tools such as GraphQL VSCode.

Issue #8670, opened 2022-12-01
fragment not deduplicated by `client` preset


saihaj:
Hey 👋  I have also encountered this issue, and it's a bit of a potential blocker so I'm hopeful we can resolve it. I've created a minimal repro here: https://codesandbox.io/s/focused-silence-flkogf

It is a very minimal repro but I think shows the essence of the issue. If you run `npm run start`, the GraphQL query is printed showing two instances of the example `SharedComponentFragment` fragments I included. This is essentially how my query looks when seeing the error at the network level `There can only be one fragment named 'X'`, so hopefully serves as a suitable reproduction — let me know if not though!

_Originally posted by @lrholmes in https://github.com/dotansimha/graphql-code-generator/discussions/8617#discussioncomment-4277757_

lrholmes:
Thanks for creating this issue for us! I'm curious to ask if you have any pointers for how someone could start looking into diagnosing this issue? I've no experience contributing to this repo but this issue is becoming a bit of a blocker for me so am open to helping if I can 🙏

saihaj:
Hey @lrholmes I started with a repro from the codesanbox to this https://github.com/dotansimha/graphql-code-generator/pull/8671 I don't think we have a easy way to mock this. 

But to fix the issue we just need to make the default to be `true` here https://github.com/dotansimha/graphql-code-generator/blob/224a5362ca26952aeab6947db18623efa69eb96e/packages/presets/client/src/index.ts#L94

beerose:
Fixed in https://github.com/dotansimha/graphql-code-generator/pull/8971

PR #9123, opened 2023-03-07
Add a test case to ensure deduplicated fragments

beerose:
## Description

I was looking into [an issue with duplicated fragments](https://github.com/dotansimha/graphql-code-generator/issues/8670) and found out it was fixed in [a recent PR](https://github.com/dotansimha/graphql-code-generator/pull/8971) by always inlining fragments. So this PR only adds a test case to ensure that it works as expected (I followed @saihaj's repros mentioned in the issue).

However, should we revisit the inlining part? It causes [issues with huge files being generated](https://github.com/dotansimha/graphql-code-generator/pull/8971#issuecomment-1451997425).

Related https://github.com/dotansimha/graphql-code-generator/issues/8670

## Type of change

### DIFF
--- a/packages/presets/client/tests/client-preset.spec.ts
+++ b/packages/presets/client/tests/client-preset.spec.ts
@@ -1043,7 +1043,9 @@ export * from "./gql.js";`);
     } catch {}
     await fs.promises.mkdir(path.join(dir, 'out1'), { recursive: true });
     for (const file of result) {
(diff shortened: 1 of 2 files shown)


saihaj (approved):
Thank you!

saihaj:
> However, should we revisit the inlining part? It causes issues with huge files being generated.

Yeah I think the starter for that is if we stop including the schema part of all the operations generated which will help reduce size. The figuring out a way to not inline the operations

PR #9137, opened 2023-03-08
Add TypedDocumentNode string alternative

beerose:
🚨 Depends on [a PR in `graphql-typed-document-node`](https://github.com/dotansimha/graphql-typed-document-node/pull/154)

## Description

Related to: https://github.com/dotansimha/graphql-code-generator/issues/8296:
CleanShot 2023-03-09 at 14 46 45@2x

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Environment**:

- OS:
- `@graphql-codegen/...`:
- NodeJS:

## Checklist:

## Further comments


.changeset/@graphql-codegen_client-preset-9137-dependencies.md
```diff
@@ -0,0 +1,5 @@
+---
+"@graphql-codegen/client-preset": patch
+---
+dependencies updates:
+  - Updated dependency [`@graphql-typed-document-node/core@3.2.0-alpha-20230309104331-975d9d2` ↗︎](https://www.npmjs.com/package/@graphql-typed-document-node/core/v/3.2.0) (from `3.1.2`, in `dependencies`)
```
beerose, line 5: TODO: update with the latest version when a PR  to `@graphql-typed-document-node/core` is accepted and released

packages/plugins/typescript/typed-document-node/src/index.ts
```diff
@@ -40,15 +39,11 @@ export const plugin: PluginFunction<TypeScriptTypedDocumentNodesConfig> = (
...
-  documents: Types.DocumentFile[],
-  config,
+  _schema: GraphQLSchema,
+  _documents: Types.DocumentFile[],
+  _config,
   outputFile: string
 ) => {
-  if (config && config.documentMode === DocumentMode.string) {
-    throw new Error(`Plugin "typed-document-node" does not allow using 'documentMode: string' configuration!`);
-  }
```
beerose, line 50: Not sure about this

packages/presets/client/src/fragment-masking-plugin.ts
```diff
@@ -1,7 +1,7 @@
 import type { PluginFunction } from '@graphql-codegen/plugin-helpers';
 
 const fragmentTypeHelper = `
-export type FragmentType<TDocumentType extends DocumentNode<any, any>> = TDocumentType extends DocumentNode<
+export type FragmentType<TDocumentType extends DocumentTypeDecoration<any, any>> = TDocumentType extends DocumentTypeDecoration<
```
beerose, line 4: Added a new type `DocumentTypeDecoration` in this PR: https://github.com/dotansimha/graphql-typed-document-node/pull/154. In here we don't really care if it's `DocumentNode` or `string` — the `__apiType?: (variables: TVariables) => TResult;` part is the important bit.

examples/react/urql-string-document-mode/src/App.tsx
```diff
@@ -0,0 +1,62 @@
...
+        node {
+          ...FilmItem
+        }
+      }
+    }
+  }
+`);
+
+declare module 'urql' {
+  // @ts-expect-error this is just temporary until we update types in urql
```
beerose, line 22: So I guess we should make a PR to `urql` after `@graphql-typed-document-node/core` is released, but what other places should we update?

packages/presets/client/src/fragment-masking-plugin.ts
n1ru4l: I was even wondering whether we should use a helper type for this, but instead make it a spec for embedding type information in anything? https://github.com/dotansimha/graphql-typed-document-node/issues/143#issuecomment-1446106576 That would also make this kind of stuff more future-proof as adding a new generic argument would be a breaking change... (see https://github.com/dotansimha/graphql-typed-document-node/pull/148)

@beerose What do you think?

beerose: I'm not quite sure if I get you, but we could remove the generic constraint here or default all type parameters in `DocumentTypeDecoration` to `any`, so we won't have a breaking change the next time we add something.

Alternatively, instead of a function signature in the phantom field, we could attach an object with any number of properties:

```ts
import type { DocumentNode } from "graphql"

interface TypeInfo {
  result?: unknown,
  variables?: unknown,
  value?: unknown
}

type DocumentTypeDecoration<TTypeInfo extends TypeInfo> = {
  __graphqlTypes?: TTypeInfo
}

export type TypedDocumentStringOperation<TVars, TRes> = string & DocumentTypeDecoration<{
  variables: TVars,
  result: TRes,
}>

export type ResultOf<T> = T extends DocumentTypeDecoration<infer TTypeInfo> ? TTypeInfo['result'] : never;

export type VariablesOf<T> = T extends DocumentTypeDecoration<infer TTypeInfo> ? TTypeInfo['variables'] : never;
```

then we'd be free to add any additional property:

```ts
interface TypedDocumentNodeFragment<T> extends DocumentNode, DocumentTypeDecoration<{
  value: T
}> {}
```

beerose: [Playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAqgzhATgQQG4ENjsfJAxRdAcwFsIA7YKAXigG8oB9R0Sc9MgLigCJdEeAbijoMWHIgA23OMEQBLckSgBfKADJ6AKCi8oAEgBmhUhWAA5DhB4B+bnwQox2fgWJlKQrSsFatraH4AYWwAE1cTDypaAAo6HSYWcAore34vXXlQmTlFIl9ddi4oWQUlXzVNeN0eA2N3MwAlCEM4W24GfjRMF0c3U0puLuccPsizVQqASg16PSNxyksydt5gsIiGz0m-AKgAUQAPDjBJCCCAexIwC-IzAEUAVyQQGm1dZgCiiHsnl4yoI9HHYEroAD5QOKJL6pNaOWqVd66ZE1OqLYDNVqrTqOEKIcJjLbAIa4jaEgZUHyg5EqKbUiHkR6SSQVXxaAD07KgAFpeXz+QLBULhSLRWLubtklAAAqICDAYDyQwgAA8ABUAHxvapQADaAGkoIooABrCAgC6GKBqgC63DVBptFVmdCp-ilADV0JJnnAAPKGdVa2gOs0Wq22tl7GDkeS3NUXACSlCQCAAxorbiqYMHITAoBBDsAKKE4CJyK8bJCTUMZtQtagLlkoNw7qgkDNC8XyKXIQka0byIYkFBE1o6w2m6EElWx7pWxB24go1K8JJMN3+lE1ck4Cq8Lm8AWiyWywxagsiZi2nZB8PEFA8NfVDPH+uFRQt2Yd5A99TdLK8qKsqKr-sifokPIwD7gANPM9QUtetQQheCFRMs1hauoYEohBUGgSihHIjGcbkAmybFog6aZuQKpej6ED+oG5i3OYTLrgARmc+7XhqfEwThhGoeiGE8IJugamBkkotJLaPr4exrhuFB-IgIBfpQP6MSq9y5vcx7dr2DAGoOprmpaUD3Ha5avCor7GYaxphhZVncCx5Bscy6BcRA+7vpu6JaXu9yOnxL7zlAbZIL4nJQAABjGJDoHAJrNHATLAAGcVGnA5AAORUMAAAW0BgIgFyQA+8jXGcUSYKRUAAO7QI1FxMqEUBFaI0DGnFaaSPIZjcmVjHynFcHNXlzJQANZqSK8haQBmIjxVpoQACIXGmjzoRcoQQONUAcY8VDNSIcqAnAeQrVdNVKoNHVLldtydUg0DJU10ChLcBWdd1UDFRcCBGjVEB1TRZYlXKAB07qQLA5BJSlaUZQGQZvGqBmnnMwlXi0N7cOgFbhVAVZKR+5AacAQVBgk3Dk92qnqYFu5BmysXipzXPc5zkrwwzn7onAaqMcAACMbwC5TLO-tmpL4psFIajFXK6AAejYfhprcsgA6LUtU8Loti-T-mC0SNPrAr5JRLmOqfMk3xpPCAmZNkvBwFcEDclkPCu5FsI8J7ZDct8fsJKIPQSNIHte9ykfiNyjxSOHbqxQAsvIcBXUo8UJy4UjZYYg2SB1MTGFckWixAHVoWYdLa+QuvFrIBtCyLsgAEym8p0sW6zVsEvg6J2wksXImrMrlZVoBQHl+fR3lOVQJB2fXcaATQ1vCQO2wgfpP7WT2MH3u+-7TuxyHYcCW6ilm+QTMd+Lkv30zVM00cJxnJc1y3A8zxqWVhyVWUANZ80CIjZKqVGKo0MFpCWtBEpQJRpITKgZP6gx-jcO4lAmZAPHqAzWWhG7N1Fkg5GMDUEBhNgjJG0D0pUPQccTBVxsH-xeKPD4SQ97FB4EzcOuggRIA6NSXeKReEH2pEfS+p9QgCORBfIOcdr7UgXjAKQx844LyTinf2Kgb7sy5JnNeudvSSHikIxARcS6lmITrQqZDIEUIYWg7utDkGULQSqDBpxzisL-rggBIBOHQkdoHfh-tLEiJRLFSeWk54MDEYo9IwgVBLyzivLOOdlDFWgIYC4zILiNWumVCqSBFSMSgBXEgAMpR5USdw8RPw4RICEEad2pQ8jCAvp08oIgRjqJjr0-Iqg8rcCyHBb4cE1FSFEY05JLsEj6O8L4IAA)

n1ru4l: I meant more like that would be a breaking change for GraphQL Code Generator to support this. Because if a user is still using an older typed-document-node version, that has no 3rd generic parameter, and Code Generator would already generate code like that, it would result in TypeScript errors.

I think the function usage is there for a specific TypeScript behavior and quirk 🤔  https://github.com/dotansimha/graphql-typed-document-node/commit/6ee77c273422bea7df58f92607e54fe73eba673a

examples/react/urql-string-document-mode/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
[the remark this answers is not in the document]
n1ru4l: I think urql is not a good example because it actually requires the AST at runtime anyways 🤔 .
We can add or modify a react-query or similar example that does not require the AST.
What do you think?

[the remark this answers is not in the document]
beerose: > I think urql is not a good example because it actually requires the AST at runtime anyways 

Does it, though? It accepts both string and AST and seems to work fine (I modified the original urql example). Unless I'm missing something.

Regarding the `react-query` example — it currently uses `@graphql-tools/executor-http`, but I could replace it with `fetch`.

beerose:
> Also I saw that we embed fragments with ... here, would this cause the same issues as mentioned/fixed by https://github.com/dotansimha/graphql-code-generator/pull/8971 for document nodes?

Yeah, good point. Will work on that.

> I think we will also have issues with metadata within the string document mode 🤔 https://github.com/dotansimha/graphql-code-generator/pull/8757

Yep. Will work on that too. Exporting an object sounds good — I don't really have other ideas right now except for `class MyString extends String` which seems a bit ugly 😅 

Are there any other things like that (things that would cause problems with `string` documents)? I guess it would be a good idea to extend the tests (or some of them) to also run in `"string"` document mode. What do you think, @n1ru4l?

packages/plugins/other/visitor-plugin-common/src/client-side-base-visitor.ts
```diff
@@ -427,7 +427,23 @@ export class ClientSideBaseVisitor<
...
+
+        if (meta && this._omitDefinitions === true) {
+          return `{${`"__meta__":${JSON.stringify(meta)},`.slice(0, -1)}}`;
+        }
+      }
+      if (meta) {
+        return `new TypedDocumentString(\`${doc}\`, ${JSON.stringify(meta)})`;
+      }
+
+      return `new TypedDocumentString(\`${doc}\`)`;
```
beerose, line 446: Added support for `__meta__` properties

packages/plugins/typescript/typed-document-node/src/index.ts
```diff
@@ -33,22 +33,41 @@ export const plugin: PluginFunction<TypeScriptTypedDocumentNodesConfig> = (
...
+  constructor(private value: string, public __meta__?: { hash: string }) {
+    super(value);
+  }
+
+  toString(): string & DocumentTypeDecoration<TResult, TVariables> {
+    return this.value;
+  }
+}`,
+    ];
+  }
```
beerose, line 55: I'm not exactly sure about this — I didn't want to export it from `@graphql-typed-document-node/core` because there's no need for that IMO — we can inline it in the generated code. But not sure what is the best place in the code to handle that. Having this in `packages/plugins/typescript/typed-document-node/src/index.ts` works, but maybe there's a better approach? Any thoughts, @n1ru4l?

examples/react/urql-string-document-mode/.gitignore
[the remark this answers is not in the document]
n1ru4l: yeah, urql uses graphql internally and thus will just parse the string to the ast at runtime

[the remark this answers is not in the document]
n1ru4l: yeah, I am fine with replacing executor HTTP in react-query!

packages/plugins/typescript/typed-document-node/src/index.ts
n1ru4l: yeah I am fine with that place

website/src/pages/plugins/presets/preset-client.mdx
```diff
@@ -525,3 +526,82 @@ const nextConfig = {
 ```
 
 Note that you will need to provide the `artifactDirectory` path that should be the same as the one configured in your `codegen.ts`
+
+### DocumentMode
```
beerose, line 530: Is it the right place for this in the docs?

examples/persisted-documents/package.json
```diff
@@ -7,7 +7,7 @@
     "@graphql-yoga/plugin-persisted-operations": "1.7.2"
   },
   "devDependencies": {
-    "@graphql-typed-document-node/core": "3.1.2",
+    "@graphql-typed-document-node/core": "3.2.0-alpha-20230314161625-aafb8c1",
```
saihaj, line 10: would need to get a stable release out before we merge

website/src/pages/plugins/presets/preset-client.mdx
saihaj: yeah should be good here. Maybe we should have some FAQ section or something that will tell why one would prefer using this option 🤔

n1ru4l: @beerose Yeah, I agree, having a section for when to use this over typed document node makes sense!

examples/persisted-documents/package.json
n1ru4l: https://github.com/dotansimha/graphql-typed-document-node/pull/155 released it @beerose please update 😇

website/src/pages/plugins/presets/preset-client.mdx
beerose: I added a short section in the same file

dev-test/gql-tag-operations/graphql/fragment-masking.ts
```diff
@@ -1,48 +1,46 @@
...
->
-  ? TType extends { ' $fragmentName'?: infer TKey }
-    ? TKey extends string
-      ? { ' $fragmentRefs'?: { [key in TKey]: TType } }
+export type FragmentType<TDocumentType extends DocumentTypeDecoration<any, any>> =
+  TDocumentType extends DocumentTypeDecoration<infer TType, any>
+    ? TType extends { ' $fragmentName'?: infer TKey }
+      ? TKey extends string
+        ? { ' $fragmentRefs'?: { [key in TKey]: TType } }
+        : never
```
beerose, line 8: Is this considered a breaking change for the codegen or not? @n1ru4l @saihaj

n1ru4l: I don't think so, people shouldn't use this type with anything else than our generated stuff.

saihaj: agreed should be fine

n1ru4l: Okay, as I think about it a bit more, this change kinda requires people to upgrade their typed document node version. 🤔 

In graphql-yoga we treat `peerDependency` upgrades depending on the type of change.

e.g. 
`X.1.1` bump of a peer dependency would be a breaking change
`1.X.1` bump of a peer dependency would be a minor change
`1.1.X` bump of a peer dependency would be a patch change

So I think we should treat this the same here (minor change).

Here the special case is that it is not listed as a peer dependency, but theoretically, it could be one, as people will need to install it, otherwise, their code won't compile with strict package manages like pnpm and yarn that do no hoisting. 🤔 

Then, however, whether it should be a peer dependency or not also depends on the location where the code is generated.

So the safest option would be to ship this as a breaking change. I am sure people will open issues if we don't...

Issue #9102, opened 2023-03-03
add tests for `examples/typescript-graphql-request/package.json`


saihaj:
This will help avoid issues so the CI will error and we do not merge anything by mistake. Related https://github.com/dotansimha/graphql-code-generator/pull/9090#issuecomment-1451158283

beerose:
Closed in https://github.com/dotansimha/graphql-code-generator/pull/9148

PR #9148, opened 2023-03-10
Minor examples improvements

beerose:
When working on another PR, I noticed that `react/graphql-request` doesn't use `graphql-request` directly anymore (looks like it was changed here: https://github.com/dotansimha/graphql-code-generator/pull/8982/files#diff-d6f9b6ed3630e420901e36cc9bf3eca683d310807463b92520644240d4505129) — updated that. Also added a test to `typescript-graphql-request` example. 

Related https://github.com/dotansimha/graphql-code-generator/issues/9102


examples/react/graphql-request/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@graphql-tools/executor-http": "^0.1.8",
```
ardatan, line 6: Maybe this should be another example since `graphql-request` can be overkill in some cases
 @n1ru4l What do you think?

n1ru4l: Why overkill? I guess both libraries have different features and the tool executor has even more, since it also support subscriptions. So I guess we could also simply rename this example instead and have a separate graphql-request example (or none at all as we would rather advocate people to use our library).

beerose: So do you think I should just rename this example and revert the changes there?

n1ru4l (approved):

PR #9193, opened 2023-03-16
docs: handleAsSDL option

saihaj:
We get this quesiton time to time so worth documenting.

### DIFF
--- a/website/src/pages/docs/config-reference/schema-field.mdx
+++ b/website/src/pages/docs/config-reference/schema-field.mdx
@@ -172,7 +172,7 @@ import { CodegenConfig } from '@graphql-codegen/cli';
 const config: CodegenConfig = {
   schema: [
     {
-      'http://localhost:3000/graphql': {
+      'http://localhost:3000/sdl': {
         handleAsSDL: true,
       }
     }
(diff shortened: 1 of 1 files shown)


n1ru4l:
Hive is a bad example because we have the sdl.graphql endpoint 🤔

website/src/pages/docs/config-reference/schema-field.mdx
```diff
@@ -162,6 +162,26 @@ export default config;
...
+
+Handling the response as SDL will allow you to load schema from remote server that doesn't return a JSON introspection. Example use case is when using schema registry like [GraphQL Hive](https://docs.graphql-hive.com/features/high-availability-cdn#graphql-schema-sdl).
+
+```ts {7}
+import { CodegenConfig } from '@graphql-codegen/cli';
+
+const config: CodegenConfig = {
+  schema: [
+    {
+      'http://localhost:3000/graphql': {
```
ardatan, line 175:
```suggestion
      'http://localhost:3000/sdl': {
```
`/graphql` can be confusing.

ardatan:
@saihaj That one has `.graphql` at the end so UrlLoader enables that option automatically.

saihaj:
> @saihaj That one has `.graphql` at the end so UrlLoader enables that option automatically.

Not everyone would be exposing `.graphql` we can have that part of Hive docs on specifics?

PR #9247, opened 2023-03-30
eslint fixes

dimaMachina:
- [x] use stable version of guild eslint config

### DIFF
--- a/packages/graphql-codegen-cli/src/generate-and-save.ts
+++ b/packages/graphql-codegen-cli/src/generate-and-save.ts
@@ -34,7 +34,6 @@ export async function generate(
             debugLog(`Removed stale file: ${prettyFilename}`);
           }
         });
-        continue;
       }
     }
     previouslyGeneratedFilenames = filenames;


packages/graphql-codegen-cli/src/generate-and-save.ts
```diff
@@ -19,26 +19,29 @@ export async function generate(
...
-        return unlinkFile(filename, err => {
+        unlinkFile(filename, err => {
           const prettyFilename = filename.replace(`${input.cwd || process.cwd()}/`, '');
           if (err) {
             debugLog(`Cannot remove stale file: ${prettyFilename}\n${err}`);
           } else {
             debugLog(`Removed stale file: ${prettyFilename}`);
           }
         });
+        continue;
```
B2o5T, line 37:
```suggestion
```

B2o5T: we can even yet simplify

saihaj (approved):

Issue #9270, opened 2023-04-04
`--watch` mode does not filter out file change events that are negated as part of a set of patterns


milesrichardson:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

This is a follow-up to #9266 (which was fixed by #9667), but it's an entirely separate bug. That fix just happened to expose the bug in our repository.

In our case, we have a NextJS app which uses the generated code from `graphql-codegen`, which means that on most page reloads, NextJS rebuilds pages in the `.next` directory. We noticed that these changes were not being filtered out of watch mode, even though they didn't match any of our patterns:

```
Running lifecycle hook "onWatchTriggered" scripts...
[Watcher] triggered due to a file create event: /src/js/cloud/catalog/.next/cache/webpack/server-development/7.pack
Running lifecycle hook "onWatchTriggered" scripts...
[Watcher] triggered due to a file delete event: /src/js/cloud/catalog/.next/cache/webpack/server-development/7.pack_
Running lifecycle hook "onWatchTriggered" scripts...
[Watcher] triggered due to a file create event: /src/js/cloud/catalog/.next/cache/webpack/server-development/index.pack
Running lifecycle hook "onWatchTriggered" scripts...
[Watcher] triggered due to a file delete event: /src/js/cloud/catalog/.next/cache/webpack/server-development/index.pack_
```

These file names could not possibly match any of our patterns in `codegen.yml`, which are all either files ending with `*.graphql` or `*.ts`:

<details>

<summary>Our `codegen.yml` from our repository</summary>

This is not reproducible, but is just to show how there are no patterns that could possibly match the files triggering the rebuilds:

```yml
overwrite: true
config:
  scalars:
    Datetime: "string"
    JSON: "{ [key: string]: any }"
    UUID: "string"
  documentMode: graphQLTag

autogeneratedFileWarning: &autogeneratedFileWarning
  add: "/* DO NOT EDIT! This file is auto-generated by graphql-code-generator - see `codegen.yml` */"

generates:
  types/index.ts:
    schema:
      - "../gql-api/schema/unified_schema.graphql"
    documents:
      - "queries/*.graphql"
      # All .graphql files in catalog...
      - "../catalog/**/*.graphql"
      # ...except those that are compiled into types/search.tsx in next section
      - "!../catalog/search/**/*.graphql"
      - "../dynamic-forms/scripts/fetch-externals.ts"
    plugins:
      - *autogeneratedFileWarning
[...]

Issue #9272, opened 2023-04-04
CLI exits with code 1 without an error message if directory is listed as output instead of file


dallonf:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

If you accidentally specify a directory instead of a filepath as a key in `generates`, the `graphql-codegen` command will exit with an error code of 1, but provide no other messaging as to what's gone wrong.

### Your Example Website or App

https://stackblitz.com/edit/github-itj2kj?file=package.json

### Steps to Reproduce the Bug or Issue

In the StackBlitz example terminal, run: `npm run codegen && echo done`

Notice that "done" is not printed; this is because the `codegen` step exits with code 1.

If you change `output` to `"output/types.ts"` in the `codegen.ts` file and run the same command, you'll see that the operation completes and prints out "done".

### Expected behavior

As a user, I expected some sort of error message to indicate what I've done wrong here, but instead an exit code (and a lack of generated output) is my only indication that anything's gone wrong at all.

### Platform

- OS: macOS 12.6.3
- NodeJS: v14.21.3 (but StackBlitz is on v16.14.2)
- `graphql` version: v16.6.0
- `@graphql-codegen/*` version(s): 
    - @graphql-codegen/cli@3.3.0
    - @graphql-codegen/core@3.1.0
    - @graphql-codegen/introspection@3.0.1
    - @graphql-codegen/plugin-helpers@4.2.0
    - @graphql-codegen/schema-ast@3.0.1
    - @graphql-codegen/typescript-operations@3.0.3
    - @graphql-codegen/typescript-resolvers@3.2.0
    - @graphql-codegen/typescript@3.0.3
    - @graphql-codegen/visitor-plugin-common@3.1.0

### Additional context

From stepping through the CLI code, it looks like there's just no error reporting on the output step.

PR #9275, opened 2023-04-05
Trigger rebuilds in watch mode while respecting rules of precedence and negation

milesrichardson:
_**STATUS (Apr 8)**: ✅ Ready for review, rebased, description updated, self-review complete_

_**UPDATE (Apr 8):** This PR is ready for review (and IMO, ready to merge ;)). The code is rebased on latest `master`. I added a lot of test cases (see `watcher.spec.ts`), which I believe cover all of the specification in terms of the behavior we want. These tests surfaced some bugs and I fixed them. Also, I updated the description in this PR to reflect the latest information about testing._

_**UPDATE (Apr 7):** I've added a bunch of test infrastructure code, namely: making the watcher cancellable, mocking ParcelWatcher and event dispatching/handling, adding assertion helpers for creating data-driven assertions with DRY test code, and formatting test error output to clearly show what did/didn't trigger. The code in this PR is now relatively stable, and all I intend to do is add more test cases. You could start a review if you'd like._

_**DRAFT (Apr 5)**: The implementation is working as far as I can tell, and it's an improvement over the status quo of very aggressive rebuild triggers during watch mode (even just running `git status` triggers a rebuild because it changes `.git/index.lock`). But ideally I want to add some unit tests before merging this, although I'm a bit tired now so we'll see._

------

## Description

This PR implements an algorithm for deciding which file change events should trigger a rebuild during watch mode. It respects the precedence rules, both in terms of "global' (top level) config preceding "local" (per-output target) config, and in terms of "watch patterns" (both global and local) preceding "document" and "schema" patterns. 

The algorithm assumes that the config file defines at least one output target. The basic idea is that it creates a `matcher` for each local config, and then on a file change event, it calls those matchers until it finds one that returns `true`. The specification of the algorithm ensures that as long as there is at least one output target, then all rules of precedence and negation will be respected.

The algorithm is described in [this comment](https://github.com/dotansimha/graphql-code-generator/issues/9270#issuecomment-1496765045) of #9270. For convenience, I'm including the diagram of the flow chart here as well (to read it, start from the top left):

<details>

<summary>SVG diagram of flowchart describing the algorithm</summary>

should-path-trigger-rebuild-for-localconf

🔗 [Raw link to SVG image](https://user-images.githubusercontent.com/835921/229949311-2023528a-a46e-4900-ab59-124041d7b574.svg)

</details>

<details>
<summary>PNG diagram of flowchart describing the algorithm</summary>

should-path-trigger-rebuild-for-localconf

🔗 [Raw link to PNG image](https://user-images.githubusercontent.com/835921/229949381-56117b9f-ea83-49d5-968c-23d3f866d38b.png)

</details>

Related #9270

## Type of change

- [X] New feature (non-breaking change which adds functionality) _[not technically a new feature, but significant amount of code]_

### Unit Tests

I added 500 lines of unit tests, and also made `watcher` testable, namely by: making it cancellable, mocking ParcelWatcher, and adding assertion helpers for testing whether or not a change event triggers a build. See `watcher.spec.ts` (note it's collapsed in the diff view) for all the test cases, which I hope you'll find easily readable, thanks to the assertion helpers. :)

To run:

```bash
yarn build && yarn test watcher
```

Here's a picture of those tests passing, including the names of the cases which should give you an idea of what's covered:

image

### Crude Smoke Tests

I've tested it locally, and added some configs to `dev-test/codegen.ts` that could impact watch mode. You can run watch mode like this:

```
yarn build
yarn watch:examples
```

Then you can try changing a few files:

```
# Should _not_ trigger a rebuild
touch dev-test-outer-dir/githunt/nothing-should-use-this-query.graphql

# Should trigger a rebuild
touch dev-test/codegen.ts

# Should trigger a rebuild
touch dev-test/star-wars/ExcludeQueryAlpha.graphql
```

Here's a video of that:

https://user-images.githubusercontent.com/835921/230742974-3614ff79-5dd2-4078-85cd-978c2e776043.mp4

**Test Environment**:

- OS: MacOS
- `@graphql-codegen/...`: This PR
- NodeJS: v18.10.0

## Checklist:

- [X] _[not applicable? maybe should add the diagram somewhere]_ I have made corresponding changes to the documentation

## Further comments

There is a lot of code here, but don't be alarmed! It's not as bad as it looks. 😄 I also did a self-review and left a bunch of comments, but none of them are actionable; I just wanted to provide some background info on various code choices, but you can just ignore them too.

**If you only review two files, these are the most important:**

- `patterns.ts` (implements the algorithm in the flow chart)
- `watcher.spec.ts` (describes all the test cases)

**If you have a question about the behavior**

Check whether the case is covered in `watcher.spec.ts`. If it's not, it should be really easy to add one. All you need to do is describe a config, and then list which changes you think should or should not trigger a build! 😄 Please let me know if I missed any corner cases in the `watcher.spec.ts`.

**Here is the gist of the changes:**

* Add `patterns.ts`
    * Implements the algorithm described in the flow chart, specifically in the `makeShouldRebuild` function (which creates the `shouldRebuild` callback that `watcher.ts` calls) (please see #9270 for longer discussion/explanation of the algorithm)
* Add tests in `watcher.spec.ts`
    * Implement tests with  the mock watcher and assertion helpers (see below), which makes for really clean and readable test cases
    * Add tests with assertions about `watchDirectory` (testing behavior of #9266)
    * Add tests to cover corner cases of global/local precedence and negation/affirmative precedence, covering most behavior described in the spec
* Update `dev-test/codegen.ts`
    * Add some entries for testing watch mode negations when running `yarn watch:examples`
    * Note: This also changes some of the generated files, which makes this PR look bigger than it is. But the changes are all expected; read the early commit messages to see the explanation for them.
* Update `watcher.ts`
    * Call `shouldRebuild` from `patterns.ts`
    * Fix two semi-related minor bugs
        * Ensure relative paths and glob patterns passed to `ParcelWatcher.Options["ignore"]` are relative from `watchDirectory` (as expected by ParcelWatcher), not from `cwd`
        * Fix double-concatenated absolute `fullPath` to just use `path` when passed to `onWatchTriggered`
    * Make the watch server cancellable, which is necessary for testing it without leaving open handles between tests
* Create testing infrastructure in `tests/watcher-test-helpers`
    * Add `watcher-test-helpers/setup-mock-watcher.ts`
        * Add function for mocking `ParcelWatcher` that returns functions for making assertions on it:
        * to spy on the subscribe callback (and other options, like `ParcelWatcher.Options.ignore`) passed by implementing code to `ParcelWatcher.subscribe()`
        * to mock `onWatchTriggered` lifecycle hook, which is called when a build is triggered
        * to `dispatchChange(someFilePath)` events to the mock watcher, so they can then make assertions on the subscribe callback and `onWatchTriggered` callback, which together can be used to see if a file change did or did not trigger a build
    * Add `watcher-test-helpers/assert-watcher-build-triggers.ts`
        * Add `assertBuildTriggers()` helper function, which takes as arguments: the mock watcher, a list of paths that should trigger build, a list of paths that should not trigger build, a list of paths that should be passed to `ParcelWatcher.Options.ignore`, and a list of globs that should be passed to `ParcelWatcher.Options.ignore`
    * Add `watch-test-helpers/format-watcher-assertion-errors.ts`
        * Add formatting functions to format error messages from assertions to clearly show why the assertion failed, and what was expected (i.e. "file X should have triggered build, but did not). For `ParcelWatcher.Options.ignore`, it also pretty prints tables showing what paths were passed to `ignore` and what was expected to be there but not found.

Here are some examples of what the formatting looks like for test failures:

<details>
<summary>Expected file to trigger build, but it didn't</summary>

image

</details>

<details>
<summary>Expected file NOT to trigger build, but it did</summary>

image

</details>

<details>
<summary>Expected path to be passed to ParcelWatcher.Options.ignore, but it was not</summary>

See: [ParcelWatcher `ignore` option](https://github.com/parcel-bundler/watcher#options)

image

</details>

<details>
<summary>Expected glob to be passed to ParcelWatcher.Options.ignore, but it was not</summary>

See: [ParcelWatcher `ignore` option](https://github.com/parcel-bundler/watcher#options)

image

</details>


packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+      // Does path match any affirmative watch pattern?
+      if (
+        matchesAnyAffirmativePattern(path, [
+          ...globalPatternSet.watch.affirmative,
+          ...localPatternSet.watch.affirmative,
+        ])
+      ) {
+        // Immediately return true: Watch pattern takes priority, even if documents or schema would negate it
+        return true;
+      }
```
milesrichardson, line 78: For watch patterns, we separate the patterns into `negated` and `affirmative` patterns, and we evaluate theme each separately. That is, we use `micromatch` only to evaluate affirmative patterns (or the inverted negations). This is necessary only for the `watch` rules because we use them to short circuit the algorithm.

packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+
+      // Does path match documents patterns (without being negated)?
+      if (pathMatches(path, [...globalPatternSet.documents.patterns, ...localPatternSet.documents.patterns])) {
+        return true;
+      }
+
+      // Does path match schemas patterns (without being negated)?
+      if (pathMatches(path, [...globalPatternSet.schemas.patterns, ...localPatternSet.schemas.patterns])) {
+        return true;
+      }
```
milesrichardson, line 88: For `documents` and `schemas`, we rely on `micromatch`, and we pass it a list of patterns that can include both affirmative (non-negated) and negated patterns. 

Because `micromatch` is a bit finnicky, we can't use `micromatch.isMatch`, which is why we need to make this `pathMatches` function.

packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+   */
+  return ({ path: absolutePath }: { path: string }) => {
+    if (!isAbsolute(absolutePath)) {
+      throw new Error('shouldRebuild trigger should be called with absolute path');
+    }
+
+    const path = relative(process.cwd(), absolutePath);
+    const shouldRebuild = localMatchers.some(matcher => matcher(path));
+    return shouldRebuild;
+  };
```
milesrichardson, line 106: This is the callback that is called on each path change. It loops over the pre-constructed matchers, one for each "local" (per-output target) pattern set. This way we avoid too much overhead on every file change, as the functions have all already been created. That's also why we separate the patterns into affirmative/negated only once, when we call `sortPatterns`, and store them in a `PatternSet` struct, rather than separating them on each path change.

packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+    watch: sortPatterns([
+      ...(typeof config.watch === 'boolean' ? [] : normalizeInstanceOrArray<string>(config.watch ?? [])),
+      relative(process.cwd(), initialContext.filepath),
+    ]),
+    schemas: sortPatterns(makePatternsFromSchemas(normalizeInstanceOrArray<Types.Schema>(config.schema))),
+    documents: sortPatterns(
+      makePatternsFromDocuments(normalizeInstanceOrArray<Types.OperationDocument>(config.documents))
+    ),
+  };
+};
```
milesrichardson, line 129: We set the global "watch" patterns to any defined in `global.watch` (if it's a non-boolean), and also adds the path to the config file. Note that a relative path is used for the config file because we expect all patterns to use relative paths.

Note that we are "sorting" these patterns (grouping them into `patterns`/`affirmative`/`negated` only once, on initialization, so that we can avoid doing that during every path change when we check if a rebuild should be triggered.

packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  for (const s of schemas) {
+    const schema = s as string;
+    if (isGlob(schema) || isValidPath(schema)) {
+      patterns.push(schema);
+    }
+  }
+
+  return patterns;
+};
```
milesrichardson, line 184: This logic is exactly the same as it was previously. It's just been moved to this file.

packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+    throw new Error('pathMatches should only be called with relative path');
+  }
+
+  // We use the default micromatch export because it is the only function that
+  // implements the compound matching logic as desired (.isMatch does not remove negations).
+  // This means we don't get a nice boolean result and must check array length instead
+  const match = mm([relativeCandidatePath], patterns, { cwd: process.cwd() }).length === 1;
+
+  return match;
+};
```
milesrichardson, line 241: This is the matcher function we use when we need to match a set of patterns that can include negations. Ideally we could use `micromatch.isMatch`, but as described in #9270, this does not work as expected when negations are included (it treats a negation match as being `true` when it should instead omit it from the match set).

However, the [`micromatch` default export (source here)](https://github.com/micromatch/micromatch/blob/master/index.js#L27-L77) does what we want. It's just a bit awkward because it's not meant to be called as a boolean, so we need to give it a single-item list and then check whether the result has length 1.

Unfortunately, `micromatch` does not implement path normalization, and if patterns are relative, then path must be relative too; `micromatch` will not match an absolute path against relative patterns, nor will it match a relative path against absolute patterns. So, we assume that all patterns are relative paths.  _(Ideally, we should warn the user if any absolute patterns are detected, and I thought I had added code for that, but I just realized it was part of an earlier function since removed. I'll add that warning in a later commit.)_

Note that the thrown errors can only be thrown due to developer error (they check path, not patterns). This is true in general for all thrown errors in this PR; user errors can only result in warnings.

packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+      throw new Error(`onlyNegatedPatterns got a non-negated pattern: ${pattern}`);
+    }
+
+    // SAFETY: Defensive programming, it's possible some picomatch/micromatch logic
+    // could call a patter negated but without its prefix !. We don't know how to
+    // handle this, but it's not worth crashing over, so warn the user and leave the pattern untransformed
+    if (scanned.negated && scanned.prefix !== '!') {
+      log(`${logSymbols.warning} Got confusing pattern, appears negated but without ! prefix: ${pattern}`);
+      return pattern;
+    }
```
milesrichardson, line 282: I couldn't find any instance of a case where a pattern is both `negated` and does not have a prefix of `!`, but just to be safe I added this warning here and made it fallback to the untransformed pattern, rather than risking slicing the first character of the pattern. This means the worst failure mode will be some missed rebuilds, and at least the user will see a warning.

packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  // Developer error: This function is not intended to work with pattern sets including negations
+  if (affirmativePatterns.some(pattern => mm.scan(pattern).negated)) {
+    throw new Error('matchesAnyAffirmativePattern should only include affirmative patterns');
+  }
+
+  // micromatch.isMatch does not omit matches that are negated by negation patterns,
+  // which is why we require this function only examine affirmative patterns
+  return mm.isMatch(relativeCandidatePath, affirmativePatterns);
```
milesrichardson, line 307: Here we _can_ use `micromatch.isMatch`, because we explicitly checked that we're only matching on affirmative patterns. (Recall that `micromatch.isMatch` does not work as expected when patterns include a mix of negated and affirmative patterns.)

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -23,47 +22,24 @@ function emitWatching(watchDir: string) {
...
-    }
-  }
+  let config: Types.Config & { configFilePath?: string } = initialContext.getConfig();
 
-  for (const s of schemas) {
-    const schema = s as string;
-    if (isGlob(schema) || isValidPath(schema)) {
-      files.push(schema);
-    }
-  }
```
milesrichardson, line 57: Most of this logic is moved basically unchanged into `patterns.ts` (see earlier comment)

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -23,47 +22,24 @@ function emitWatching(watchDir: string) {
...
-  if (typeof config.watch !== 'boolean') {
-    files.push(...normalizeInstanceOrArray<string>(config.watch));
-  }
+  const shouldRebuild = makeShouldRebuild({ globalPatternSet, localPatternSets });
 
   let watcherSubscription: Awaited<ReturnType<typeof subscribe>>;
 
   const runWatcher = async () => {
-    const watchDirectory = await findHighestCommonDirectory(files);
+    const watchDirectory = await findHighestCommonDirectory(allPatterns);
```
milesrichardson, line 42: To ensure we keep the same behavior as from #9270, we need to flatten the list of patterns, which is the equivalent of what `files` was before (basically: all possible patterns). This will include a mix of negated and non-negated patterns, but it doesn't make any difference, because `mm.scan(pattern).base("!/foo/bar")` is `/foo/bar`.

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -99,22 +75,20 @@ export const createWatcher = (
...
+          // NOTE: @parcel/watcher always provides path as an absolute path
           events.map(async ({ type: eventName, path }) => {
-            /**
-             * @parcel/watcher has no way to run watcher on specific files (https://github.com/parcel-bundler/watcher/issues/42)
-             * But we can use micromatch to filter out events that we don't care about
-             */
-            if (!mm.contains(path, files)) return;
+            if (!shouldRebuild({ path })) {
+              return;
+            }
```
milesrichardson, line 82: This is the crux of the matter: we call the previously constructed `shouldRebuild()` function on change events.

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -124,7 +98,7 @@ export const createWatcher = (
               newParsedConfig.overwrite = config.overwrite;
               newParsedConfig.configFilePath = config.configFilePath;
               config = newParsedConfig;
-              initalContext.updateConfig(config);
+              initialContext.updateConfig(config);
```
milesrichardson, line 101: Fixed this typo in a few places (`inital -> initial`)

dev-test/codegen.ts
```diff
@@ -55,7 +55,11 @@ const config: CodegenConfig = {
     },
     './dev-test/githunt/graphql-declared-modules.d.ts': {
       schema: './dev-test/githunt/schema.json',
-      documents: ['./dev-test/githunt/**/*.graphql', './dev-test-outer-dir/githunt/**/*.graphql'],
+      documents: [
+        './dev-test/githunt/**/*.graphql',
+        './dev-test-outer-dir/githunt/**/*.graphql',
+        '!**/nothing-should-use-this-query.graphql',
```
milesrichardson, line 61: I added the `dev-test-outer-dir/githunt/nothing-should-use-this-query.graphql` because it would otherwise match the pattern `'./dev-test-outer-dir/githunt/**/*.graphql',` (which is only referenced once in this `codegen.ts`, making it a convenient choice), but it's negated by `'!**/nothing-should-use-this-query.graphql',`.

This is one of the crude smoke tests I used to see if the code is working, i.e. verifying that `touch dev-test-outer-dir/githunt/nothing-should-use-this-query.graphql` does **not** trigger a rebuild.

dev-test/star-wars/types.ts
```diff
@@ -248,6 +248,24 @@ export type CreateReviewForEpisodeMutation = {
...
+
+export type ExcludeQueryBetaQueryVariables = Exact<{
+  episode?: InputMaybe<Episode>;
+}>;
+
+export type ExcludeQueryBetaQuery = {
+  __typename?: 'Query';
+  hero?: { __typename?: 'Droid'; name: string } | { __typename?: 'Human'; name: string } | null;
+};
+
```
milesrichardson, line 268: The changes to multiple `star-wars` files are nothing to be alarmed about. They are just due to adding the `ExcludeQueryAlpha` and `ExcludeQueryBeta` queries, which I added because they are contra-variant in terms of their inclusion or exclusion in `types.excludeQueryAlpha.ts` and `types.excludeQueryBeta.ts`; see commit messages for details. (Although I realized this doesn't actually test the contravariance during watch mode, unless you manually edit `codegen.ts` to remove one of the targets, so maybe they could be removed - but they're harmless, and at least this is testing negation contravariance between those two output targets during build mode.)

milesrichardson:
Looks like the workflow [testing exports integrity failed](https://github.com/dotansimha/graphql-code-generator/actions/runs/4621003898/jobs/8171886495?pr=9275). I guess this is probably due to my adding `packages/graphql-codegen-cli/src/utils/patterns.ts`. 

Do I need to edit some `tsconfig.json` or something? I'll take a look myself, but taking a break for a bit now.

```bash
[log] Import of file './esm/index.js' failed with error:
Error:  Integrity check of '@graphql-codegen/cli' failed.
node:internal/errors:478
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/runner/work/graphql-code-generator/graphql-code-generator/packages/graphql-codegen-cli/dist/esm/utils/patterns' imported from /home/runner/work/graphql-code-generator/graphql-code-generator/packages/graphql-codegen-cli/dist/esm/utils/watcher.js
    at new NodeError (node:internal/errors:387:5)
    at finalizeResolution (node:internal/modules/esm/resolve:330:11)
    at moduleResolve (node:internal/modules/esm/resolve:907:10)
    at defaultResolve (node:internal/modules/esm/resolve:1115:11)
    at nextResolve (node:internal/modules/esm/loader:163:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:841:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:[18](https://github.com/dotansimha/graphql-code-generator/actions/runs/4621003898/jobs/8171886495?pr=9275#step:5:19))
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:40)
    at link (node:internal/modules/esm/module_job:75:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
  node:internal/errors:478
  ErrorCaptureStackTrace(err);
  ^
  
  Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'packages/graphql-codegen-cli/dist/esm/utils/patterns' imported from /home/runner/work/graphql-code-generator/graphql-code-generator/packages/graphql-codegen-cli/dist/esm/utils/watcher.js
  at new NodeError (node:internal/errors:387:5)
  at finalizeResolution (node:internal/modules/esm/resolve:330:11)
  at moduleResolve (node:internal/modules/esm/resolve:907:10)
  at defaultResolve (node:internal/modules/esm/resolve:1115:11)
  at nextResolve (node:internal/modules/esm/loader:163:28)
  at ESMLoader.resolve (node:internal/modules/esm/loader:841:30)
  at ESMLoader.getModuleJob (node:internal/modules/esm/loader:4[24](https://github.com/dotansimha/graphql-code-generator/actions/runs/4621003898/jobs/8171886495?pr=9275#step:5:25):18)
  at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:40)
  at link (node:internal/modules/esm/module_job:75:36) {
  code: 'ERR_MODULE_NOT_FOUND'
  }
  at node_modules/bob-the-bundler/dist/commands/check.js:179:[27](https://github.com/dotansimha/graphql-code-generator/actions/runs/4621003898/jobs/8171886495?pr=9275#step:5:28)
  at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

milesrichardson:
Changeset added and force pushed.

packages/graphql-codegen-cli/tests/watcher.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        // note: globs are tested for exact match with argument passed to subscribe options
+        'some-preset-bar/**/*.generated.tsx', // output of preset
+      ],
+      shouldNotTriggerBuild: [
+        './foo/bar/something.ts', // unrelated file
+        './foo/some-output.ts', // output file (note: should be ignored by parcel anyway)
+        '.git/index.lock',
+      ],
+    });
+  });
```
milesrichardson, line 141: Ok, so I got.... a bit carried away building the test infrastructure. 👀 I made a few minor changes to the watcher itself, but mostly just for supporting the test infrastructure. At this point the bulk of the PR is all test code.

The watch server is now fully mocked and testable. It can also be cleanly cancelled without any unhandled promises. And I added a whole bunch of helpers for making assertions on whether or not a file change event triggers a rebuild for a given config. That means we can write code like I've highlighted with this comment, which makes for really clean testing.

I think I'm now mostly done with the test infrastructure, and now I just need to add a bunch of cases and actually test it. 😅 But I've found and fixed a few bugs so far, so it's a good start.

milesrichardson: Here are some examples of what the error message formatting looks like (click to expand for screenshots):

<details>
<summary>Expected file to trigger build, but it didn't</summary>

image

</details>


<details>
<summary>Expected file NOT to trigger build, but it did</summary>

image


</details>


<details>
<summary>Expected path to be passed to ParcelWatcher.Options.ignore, but it was not</summary>

See: [ParcelWatcher `ignore` option](https://github.com/parcel-bundler/watcher#options)


image


</details>


<details>
<summary>Expected glob to be passed to ParcelWatcher.Options.ignore, but it was not</summary>

See: [ParcelWatcher `ignore` option](https://github.com/parcel-bundler/watcher#options)

image


</details>

packages/graphql-codegen-cli/src/utils/patterns.ts
milesrichardson: After writing test cases, it turns out this was not true - we cannot rely on micromatch - so I deleted the `patchMatches` shortcut, in favor of explicitly checking the affirmative matches, and then checking there are no negated matches. But I actually prefer it this way, because now the logic matches the flow chart exactly.

The reason for this is that **micromatch cares about order of rules**, and a negation will only override a match when it comes **after** the affirmative rule. However, we do not care about order - negations should always override affirmative matches.

See fix in b2328fc30c0b6ad519d7e0b7bb9300a57e9442f8 (but note the commit message body is inaccurate - the reason micromatch doesn't work has to do with ordering, not exact vs. glob match). 

This shows why it didn't work:

```js
// Given this config
globalPattern.watch =   '!**/exclude-doc-everywhere.graphql'
localPattern.watch =   'foo/global-beats-local/exclude-doc-everywhere.graphql'

// And this path:
const path = 'foo/global-beats-local/exclude-doc-everywhere.graphql'

// We would expect this NOT to match, because the global negation should win

// However, micromatch considers this a match:
true === require('micromatch')(
  [path],
  [
    '!**/exclude-doc-everywhere.graphql',
    'foo/global-beats-local/exclude-doc-everywhere.graphql'
     // Note: the same is true even with:
     // '**/exclude-doc-everywhere.graphql',
  ],
  { cwd: process.cwd() }
).length === 1

// But when changing the order, it does not consider it a match
false === require('micromatch')(
  [path],
  [
    'foo/global-beats-local/exclude-doc-everywhere.graphql',
     // Note: the same is true even with:
     // '**/exclude-doc-everywhere.graphql',
    '!**/exclude-doc-everywhere.graphql'
  ],
  { cwd: process.cwd() }
).length === 1
```

packages/graphql-codegen-cli/src/utils/patterns.ts
milesrichardson: After writing test cases, I did find an instance of where a pattern is both `negated` and has a prefix other than `!`:

```js
mm.scan("!./foo/bar/never-watch.graphql").prefix === '!./'
```

So I removed this defensive programming check in favor of slicing off the length of the prefix when the pattern is negated.

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -84,13 +79,17 @@ export const createWatcher = (
...
         const extension = entry.config.presetConfig?.extension;
         if (extension) {
-          ignored.push(join(entry.filename, '**', '*' + extension));
+          ignored.push(join(filenameRelativeFromWatchDirectory, '**', '*' + extension));
         }
       } else {
-        ignored.push(entry.filename);
+        ignored.push(filenameRelativeFromWatchDirectory);
       }
     }
```
milesrichardson, line 94: This fixes a bug that would have been introduced in #9266 (although anyone it affected wouldn't even have been able to even get as far as noticing it was a bug prior to that fix).

Specifically: `ParcelWatcher.Options["ignore"]` expects that any relative paths or globs are relative from the `watchDirectory`, but we were passing them relative to `process.cwd()` (since that's how they're defined in the config). But after #9266, it's possible for `watchDirectory` to be something other than `process.cwd()`, so we need to convert these paths to be relative from `watchDirectory`.

I added (a lot of...) test code to make assertions on which paths and globs are passed to `ParcelWatcher.Options.ignore` and added regression tests for this.

Related (not to watch mode, but similar idea): #9272

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -99,22 +98,20 @@ export const createWatcher = (
...
-            const fullPath = join(watchDirectory, path);
             // In ESM require is not defined
             try {
-              delete require.cache[fullPath];
+              delete require.cache[path];
             } catch (err) {}
 
-            if (eventName === 'update' && config.configFilePath && fullPath === config.configFilePath) {
+            if (eventName === 'update' && config.configFilePath && path === config.configFilePath) {
               log(`${logSymbols.info} Config file has changed, reloading...`);
```
milesrichardson, line 115: There was no need for `fullPath` here, because `path` is already absolute, and so concatenating it with `watchDirectory` was resulting in invalid paths (combining two absolute paths together). We can just use `path` instead. See f6ab51112f5fec9b6e8a7a0e5bed974bdc525420

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -136,32 +133,82 @@ export const createWatcher = (
...
       log(`Shutting down watch...`);
-      watcherSubscription.unsubscribe();
-      lifecycleHooks(config.hooks).beforeDone();
+
+      const pendingUnsubscribe = watcherSubscription.unsubscribe();
+      const pendingBeforeDoneHook = lifecycleHooks(config.hooks).beforeDone();
+
+      if (afterShutdown && typeof afterShutdown === 'function') {
+        Promise.allSettled([pendingUnsubscribe, pendingBeforeDoneHook]).then(afterShutdown);
+      }
```
milesrichardson, line 149: This adds an optional callback to the `shutdown` handler for it to call after it's completed its async tasks. In normal usage, this callback is not supplied, which makes sense because `process.once()` event handlers are not supposed to wait for any asynchronous code. And in that case, the logic is unchanged: each async operation is queued immediately and there is no waiting for them to complete (the program will just exit).

But in tests, we want to be able to cancel the watcher, and gracefully shutdown. So in that case we give the `shutdown` handler a callback, and we wait for the two asynchronous events (`.unsubscribe()`, and `.beforeDone`) to complete before we call the callback, so that we ensure no async tasks are remaining on the queue when we use that callback to resolve the `pendingShutdown` promise outside of this loop.

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -136,32 +133,82 @@ export const createWatcher = (
...
-    process.once('SIGTERM', shutdown);
+    abortSignal.addEventListener('abort', () => shutdown(abortSignal.reason));
+
+    process.once('SIGINT', () => shutdown());
+    process.once('SIGTERM', () => shutdown());
+  };
+
+  // Use an AbortController for shutdown signals
+  // NOTE: This will be polyfilled on Node 14 (or any environment without it defined)
+  const abortController = new AbortController();
```
milesrichardson, line 160: In tests, we make the watcher cancellable by using an `AbortSignal`. When running from the CLI, we don't use this, but we do still initialize `new AbortController()`.

Unfortunately, `AbortController` is not available in Node v14 (unless an experimental flag is supplied to Node), so this caused some GitHub Actions to fail. I fixed this by adding a polyfill for `AbortController`. It will only be imported in Node v14, and it will only be executed when running tests in Node v14.

packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -136,32 +133,82 @@ export const createWatcher = (
...
       .catch(err => {
         watcherSubscription.unsubscribe();
         reject(err);
+      })
+      .then(() => pendingShutdown)
+      .finally(() => {
+        debugLog('Done watching.');
+        resolve();
       });
   });
```
milesrichardson, line 202: There are some subtleties to this shutdown order, because the two promises rely on each other and there is some lazy property assignment. The gist of it is that `await stopWatching()` will only resolve after the `runningWatcher` has resolved, and the intermediate step of `runningWatcher` will wait for `pendingShutdown`, which is ultimately the callback passed to the `shutdown` handler, to complete.

I tested this extensively with a bunch of log messages, and every time I ran tests, I passed `--detectOpenHandles`, so I'm very confident that this shutdown logic works correctly. Namely: it's a graceful shutdown, and there are no events that leak after `stopWatching` has resolved.

packages/graphql-codegen-cli/tests/watcher-test-helpers/setup-mock-watcher.ts
```diff
@@ -0,0 +1,249 @@
...
+     * include, for example, the `ignore` key that contains ignored paths and globs that
+     * should never cause the subscription callback to be called.
+     *
+     * NOTE: This mock does _not_ implement the Parcel Watcher `shouldIgnore` check,
+     * because that's implemented by Parcel Watcher in C++ and there is no sense
+     * duplicating it in JS. So if you want to make assertions about ignored paths,
+     * you should limit it to assertions about which paths end up in `subscribeOpts.ignore`,
+     * and otherwise assume that Parcel Watcher will work as expected. For making
+     * these assertions, see the `assertBuildTriggers` helper.
+     */
```
milesrichardson, line 120: Just highlighting this point about the mock: we don't implement ParcelWatcher's `shouldIgnore`, so even if paths/globs are passed to the `ignore` pattern, the `subscribe` callback will still be executed in test mode. To make up for this, we have assertion helpers for making assertions about which paths/globs are passed to `ignore`, and we just assume `ParcelWatcher` will implement its behavior correctly.

milesrichardson (commented):
I've completed this PR, and it's ready for review (pending some edits I'm about to make to the PR description). I also rebased it on the latest changes.

I realize there is a lot of code here, but note that almost all of it is test code (both mocking/assertion infrastructure, and also tests themselves), and the actual bulk of the algorithm from the flow chart is implemented in `patterns.ts`.

I added a _lot_ of test cases, and fixed some bugs they revealed, and now I'm really confident this feature works as we want it to.

If you only review two files, the most important are: `patterns.ts` and `watcher.spec.ts`.

milesrichardson:
@saihaj This PR is complete and ready for review (and merge IMO 😄). Sorry it's so long. Most of it is test code or helpers for mocking and assertions. I updated the description with some bullet points. The most important files are `patterns.ts` (implements the algorithm described in the flow chart), and `watcher.spec.ts` (describes test cases).

I also did a self-review on all my code, which also makes the PR look even longer and more intimidating, but feel free to ignore those comments.

packages/graphql-codegen-cli/tests/watcher.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+
+    expect(mockWatcher.watchDirectory).toBe(join(process.cwd(), 'foo'));
+
+    await assertBuildTriggers(mockWatcher, {
+      shouldNotTriggerBuild: [
+        'foo/global-beats-local/exclude-watch-everywhere.graphql',
+        'foo/global-beats-local/exclude-doc-everywhere.graphql',
+        'foo/global-beats-local/exclude-schema-everywhere.graphql',
+        '!**/exclude-watch-doc-everywhere.graphql',
+        '!**/exclude-watch-schema-everywhere.graphql',
```
milesrichardson, line 198: oops, small typo in this assertion. Fixing now, and making my final commit 😅

milesrichardson: fixed & force pushed (the tests still pass)

milesrichardson: Gah, after this 2 line change, a [GHA workflow failed](https://github.com/dotansimha/graphql-code-generator/actions/runs/4647277515/jobs/8224186329?pr=9275) ("Testing / Examples (pull_request) Failing after 9m")

I think this is a flaky test unrelated to my changes. I'm going to force push again.

milesrichardson:
One thing, follow-up to earlier comment: https://github.com/dotansimha/graphql-code-generator/pull/9275#discussion_r1161153200

I said that we don't care about _order_ of micromatch patterns, which is why I stopped using micromatch for matching within each stanza, but I'm now wondering if that's true. The docs say:

> All provided glob expressions are evaluated together. The usage is similar to `.gitignore`.

I think `.gitignore` does respect order.

If we support this, it should only be within each array of match patterns (i.e. only within one `documents` array). However, we would need to decide/define how global patterns mix into that order. I imagine `globlConf.watch` and `localConf.watchPattern` would continue to precede it entirely. But when mixing `globalConf.documents` and `localConf.documents`, we'd need to decide whether to do `[...localConf.documents, ...globalConf.documents]`, or `[...globalConf.documents, ...localConf.documents]`. This choice could have some subtle differences, but I'd need to see some examples to be sure.

Let me know what you think. It's easy to add this behavior, but we just need to decide what the expectation is.

milesrichardson:
> I said that we don't care about _order_ of micromatch patterns, which is why I stopped using micromatch for matching within each stanza, but I'm now wondering if that's true. 

After pondering this a bit more, I now think that it's not important for us to consider order of rules, within a stanza or otherwise. What _is_ important is to give users a way to "default exclude everything, but then include this other thing." And we do give them that, by offering both local and global priority, where one (global) takes priority over the other. This is no different than `.gitignore` using priority to make one rule take precedence over another; it's just a different mechanism to accomplish the same goal.

So, it should be possible to take advantage of the fact that `globalConfig.watch` has the highest precedence to implement this. For example, **this test passes:**

```ts
test.only('exclude all, but include this one thing', async () => {
    const mockWatcher = await setupMockWatcher({
      filepath: './foo/some-config.ts',
      config: {
        schema: './foo/something.ts',
        watch: ['./foo/exclude-all-but-apple/apple.graphql'],
        generates: {
          ['./foo/some-output.ts']: {
            documents: ['!./foo/exclude-all-but-apple/*.graphql'],
          },
          ['./foo/some-other-output.ts']: {
            documents: ['!./foo/exclude-all-but-apple/*.graphql'],
          },
        },
      },
    });

    expect(mockWatcher.watchDirectory).toBe(join(process.cwd(), 'foo'));

    await assertBuildTriggers(mockWatcher, {
      shouldTriggerBuild: ['./foo/exclude-all-but-apple/apple.graphql'],
      shouldNotTriggerBuild: [
        './foo/exclude-all-but-apple/orange.graphql',
        './foo/exclude-all-but-apple/banana.graphql',
        './foo/exclude-all-but-apple/red-fruit.graphql',
      ],
    });
  });
```

But there are a few inconvenient things about this:

- You need to repeat the glob exclude pattern for each output
- You must put the override in `globalConf.watch`. Putting it in `globalConf.documents` instead causes the test to fail (I might consider this a bug).

Point is, it's possible to use the precedence order of global vs. local and watch vs. others to accomplish this pattern of "exclude all, but then make an exception for this one thing." And that's what's important, IMO.

However, I have no idea if this is a breaking change given what other people's configs look like, and if they were relying on order of precedence beforehand. Of course, it would only be a breaking change for watch mode, which is not the end of the world. But maybe we should add some heuristics to detect rules that look like they might depend on ordering precedence, and print a notice that it might not work as expected, and to use `globalConf.watch` instead. An update to the docs is probably also a good idea.

packages/graphql-codegen-cli/src/utils/abort-controller-polyfill.ts
```diff
@@ -0,0 +1,105 @@
...
+ * We don't actually use AbortController anywhere except in tests, but it
+ * still gets called in watcher.ts, so by polyfilling it we can avoid breaking
+ * existing installations using Node v14 without flag --experimental-abortcontroller,
+ * and we also ensure that tests continue to pass under Node v14 without any new flags.
+ *
+ * This polyfill was adapted (TypeScript-ified) from here:
+ * https://github.com/southpolesteve/node-abort-controller/blob/master/index.js
+ */
+
+class AbortSignalPolyfill implements AbortSignal {
```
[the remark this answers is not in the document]
milesrichardson: Awesome! I deleted the polyfill and replaced the import with one from `@whatwg-node/fetch`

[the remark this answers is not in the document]
milesrichardson: Looks like that caused it to pass on Node 14, but [fail on Node 16](https://github.com/dotansimha/graphql-code-generator/actions/runs/4668328338/jobs/8265544119)

The errors are timeouts in the watcher tests, which implies that `AbortController` is not working as expected. Any idea what's the issue? Should I revert the change and go back to the polyfill?

saihaj:
@milesrichardson any idea on the failing test. It says timeouts 🤔 idk if increasing timeout will make it work, probably there is a memory leak to look at with thew new watch stuff

milesrichardson:
> @milesrichardson any idea on the failing test. It says timeouts 🤔 idk if increasing timeout will make it work, probably there is a memory leak to look at with thew new watch stuff

yes, the tests started failing when I switched to using whatwg-fetch for the AbortController (see [resolved comment](https://github.com/dotansimha/graphql-code-generator/pull/9275#discussion_r1162364587)). I don't know what specifically causes that AbortController implementation to break, but if I revert that commit then I'm sure the tests will pass again. That would be my recommendation, personally - I don't think it's a big deal to have a polyfill for AbortController checked in that's used only in tests, and apparently there's something wrong with the implementation from whatwg-fetch anyway

milesrichardson:
I just reverted the commit using `AbortController` from `@whatwg-node/fetch`, to go back to using the custom polyfill. I also rebased on latest `master` (no conflicts). The reverted commit was 7a374c27ef7d1f1f02531b7395ff3d37ce12fa31 (prior to rebase), and 1e515738babaed4abee3195b3a1364d750513e29 (after the rebase).

Let's see if tests pass again. It's possible there will be a [failure in a flaky test as observed before](https://github.com/dotansimha/graphql-code-generator/pull/9275#discussion_r1161163490), but as mentioned in that comment, I'm pretty sure it's unrelated to this MR.

dotansimha:
@saihaj @ardatan @eddeee888 @beerose can you please take a look? :)

packages/graphql-codegen-cli/src/utils/patterns.ts
n1ru4l: > However, we do not care about order (right? or maybe we do care about order within local groups? let me know) - negations should always override affirmative matches.

I actually think we care about the order e.g. if we have the following order

```
foo/foo.graphql
!foo/foo/bar/**/*
foo/foo/bar/asdf/foo.graphql
```

This should still include `foo/foo/bar/asdf/foo.graphql` IMHO and `!foo/foo/bar/**/*` should not overwrite it. is this the current behavior?

packages/graphql-codegen-cli/src/utils/patterns.ts
n1ru4l: See https://github.com/dotansimha/graphql-code-generator/pull/9275#discussion_r1181497873

packages/graphql-codegen-cli/src/utils/patterns.ts
n1ru4l: See https://github.com/dotansimha/graphql-code-generator/pull/9275#discussion_r1181497873

n1ru4l:
@milesrichardson Thank you so much for this pull request. It feels like you put in a lot of thought and effort, which I highly appreciate!

My only minor concern is regarding the ruling order. In my opinion, the order should matter. But I am happy to get convinced otherwise. Do you know how other CLI tools are handling this? AFAIK in a `.gitignore` the order matters.

Happy to get this merged soon! We can also keep the `AbortSignal` polyfill here until @ardatan resolved the underlying issue in `whatwg-node`!

n1ru4l (commented):
I added one comment regarding negations and glob order.

dotansimha (approved):

milesrichardson:
@n1ru4l See https://github.com/dotansimha/graphql-code-generator/pull/9275#issuecomment-1502431304 for my latest thoughts on whether order is important. My thinking is that it's not order of precedence that matters, but giving the user a way to say "exclude everything, except for this one thing" - which we do allow them to do by using the precedence of global vs. local. Once you introduce per-stanza ordering rules, the issue becomes how the differing orders across stanzas interact with each other (see https://github.com/dotansimha/graphql-code-generator/pull/9275#issuecomment-1500994402).

What I would like to know is what the behavior of non-watch mode is when it comes to ordering precedence.

saihaj (approved):
thank you and sorry for the delays on this one

Issue #9273, opened 2023-04-05
Unsupported protocol: github


kamilbonna:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

Hi,

We are using `@graphql-codegen/cli` to generate multiple files based on `schema.graphql` which is stored in a private Github repository. Up to this point everything worked normally, but around two weeks ago, a following error when trying to load the schema from Github started to occur (I've redacted organization and repo name):

```
✔ Parse Configuration
⚠ Generate outputs
  ❯ Generate to src/graphql/generated/schemas.tsx
    ✖
      Failed to load schema from github:<organization>/<repo_name>#main:generated/schema.graphql:
      Failed to find any GraphQL type definitions in: github:<organization>/<repo_name>#main:generated/schema.graphql;
      - Unsupported protocol: github:
      - Unexpected token R in JSON at position 2
      AggregateError: Failed to find any GraphQL type definitions in: github:<organization>/<repo_name>#main:generated/schema.graphql;
      - Unsupported protocol: github:
      - Unexpected token R in JSON at position 2
      at loadFile (/home/.../node_modules/@graphql-codegen/cli/node_modules/@graphql-tools/load/cjs/load-typed…
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async /home/kamil_bonna/Desktop/.../node_modules/@graphql-codegen/cli/node_modules/@graphql-tools/load/cjs/load-typedefs/…
      GraphQL Code Generator supports:
      - ES Modules and CommonJS exports (export as default or named export "schema")
      - Introspection JSON File
      - URL of GraphQL endpoint
      - Multiple files with type definitions (glob expression)
      - String in config file
      Try to use one of above options and run codegen again.
    ◼ Load GraphQL documents
    ◼ Generate
error Command failed with exit code 1.
```

To run codegen we're using `yarn graphql-codegen --config codegen.yml` command and the content of the config file is:

```yml
schema:
  - github:<organization>/<repo_name>#main:generated/schema.graphql:
      token: ${PERSONAL_ACCESS_TOKEN} # Set environment variable
overwrite: true

generates:
  src/graphql/generated/schemas.tsx:
    plugins:
      - typescript

...
```

The problem appeared without any changes for version:
- `"@graphql-codegen/cli": "^2.1.1"`
- I also tested newest version `"@graphql-codegen/cli": "^3.3.0"` which didn't help
[...]

PR #9281, opened 2023-04-07
chore: update graphql tool loaders

saihaj:
Updating graphql tools GH loader.

Related #9273


eddeee888:
Should we update [the GitHub schema loader guide](https://github.com/dotansimha/graphql-code-generator/blob/5c115658ab6487dd9147974b2c1713fbf5e96e06/website/src/pages/docs/config-reference/schema-field.mdx?plain=1#L446-L457) with how we use the `token` option or the new `GITHUB_TOKEN` env behaviour :) ?

Also, I'm seeing this type issue (even though it works at runtime):

Screenshot 2023-04-08 at 8 45 22 am

I guess both can be in another PR?

---

I think this is also related: https://github.com/dotansimha/graphql-code-generator/issues/9007

eddeee888 (approved):
Looks good, thank you! 👍 

I've also tested this change and updated docs in https://github.com/dotansimha/graphql-code-generator/pull/9332

PR #9352, opened 2023-04-27
chore: sync node versions used by GitHub Actions

charpeni:
All workflows are using node 18 (LTS), except `main` workflow.

Let's make sure we're also using node LTS to do things on GitHub Actions.

Additionally, node 20 is out as current and will be the active LTS starting in October, let's start testing against it.


saihaj (approved):
thank you

PR #9375, opened 2023-05-07
Update typescript, typescript-operations and typescript-resolvers plugins Scalars input/output type

eddeee888:
### 1. Extend `Scalars` base type to have input/output types. 

This allows us to be able to better type the Scalar types as input type and output type could be different.

For example, on the server, ID input type is `string` and ID output type is `string | number`. 
On the other hand, on the client, input type is `string | number` and ID output type is `string`.

The default Scalar type will look like this:

```ts
export type Scalars = {
  ID: { 
    input: string | number; 
    output: string; 
  }
  String: { 
    input: string; 
    output: string; 
  }
  Boolean: { 
    input: boolean; 
    output: boolean; 
  }
  Int: { 
    input: number; 
    output: number; 
  }
  Float: { 
    input: number; 
    output: number; 
  }
};
```

### 2. Codegen config has been updated to support input/output types

A previous `typescript` config may look like this:

```ts
config: {
  scalars: {
    MyScalar: 'Date'
  }
}
```

This still works in the new implementation. This means `MyScalar` would be `Date` for both input and output.

Here's an example where input and output can be different:

```ts
config: {
  scalars: {
    MyScalar: {
      input: 'Date',
      output: 'string'
    }
  }
}
```

### 3. All Scalar-related references need to be updated to input/output types

- [x] GraphQL `input` types are updated to use input Scalars
- [x] GraphQL object types (a.k.a output types) are updated to use output Scalar
- [x] Document variables are updated to use input Scalars
- [x] Selection sets are updated to use output Scalars
- [x] Resolver types are updated to use output Scalars

### 4. Other breaking changes

- If a file path is used as Scalars, all scalars exported in this module must be updated to support input/output types.

```ts
config: {
  scalars: '/path/to/scalars'
}
```

Related #2588 

## Type of change

## How Has This Been Tested?

- [x] Unit test
- [x] Alpha version tested in https://github.com/eddeee888/poc-gcg/pull/2/files

**Test Environment**:

- OS: MacOS
- `@graphql-codegen/...`:
- NodeJS: 18

## Checklist:

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
@@ -529,17 +529,21 @@ export class BaseTypesVisitor<
   }
 
   public getScalarsImports(): string[] {
-    return Object.keys(this.config.scalars)
-      .map(enumName => {
-        const mappedValue = this.config.scalars[enumName];
+    return Object.keys(this.config.scalars).reduce((res, enumName) => {
+      const mappedValue = this.config.scalars[enumName];
 
-        if (mappedValue.isExternal) {
-          return this._buildTypeImport(mappedValue.import, mappedValue.source, mappedValue.default);
-        }
+      if (mappedValue.input.isExternal) {
+        res.push(this._buildTypeImport(mappedValue.input.import, mappedValue.input.source, mappedValue.input.default));
+      }
 
-        return null;
-      })
-      .filter(a => a);
+      if (mappedValue.output.isExternal) {
+        res.push(
+          this._buildTypeImport(mappedValue.output.import, mappedValue.output.source, mappedValue.output.default)
+        );
+      }
+
+      return res;
+    }, []);
   }
 
   public getDirectiveArgumentAndInputFieldMappingsImports(): string[] {
@@ -559,12 +563,20 @@ export class BaseTypesVisitor<
   public get scalarsDefinition(): string {
     if (this.config.onlyEnums) return '';
     const allScalars = Object.keys(this.config.scalars).map(scalarName => {
-      const scalarValue = this.config.scalars[scalarName].type;
+      const inputScalarValue = this.config.scalars[scalarName].input.type;
+      const outputScalarValue = this.config.scalars[scalarName].output.type;
       const scalarType = this._schema.getType(scalarName);
       const comment = scalarType?.astNode && scalarType.description ? transformComment(scalarType.description, 1) : '';
       const { scalar } = this._parsedConfig.declarationKind;
 
-      return comment + indent(`${scalarName}: ${scalarValue}${this.getPunctuation(scalar)}`);
+      return (
+        comment +
+        indent(
+          `${scalarName}: { input: ${inputScalarValue}${this.getPunctuation(
+            scalar
+          )} output: ${outputScalarValue}${this.getPunctuation(scalar)} }`
+        )
+      );
     });
 
     return new DeclarationBlock(this._declarationBlockConfig)
@@ -677,7 +689,7 @@ export class BaseTypesVisitor<
     if (this.config.onlyOperationTypes || this.config.onlyEnums) return '';
     const originalNode = parent[key] as UnionTypeDefinitionNode;
     const possibleTypes = originalNode.types
-      .map(t => (this.scalars[t.name.value] ? this._getScalar(t.name.value) : this.convertName(t)))
+      .map(t => (this.scalars[t.name.value] ? this._getScalar(t.name.value, 'output') : this.convertName(t)))
       .join(' | ');
 
     return new DeclarationBlock(this._declarationBlockConfig)
@@ -945,8 +957,8 @@ export class BaseTypesVisitor<
       .join('\n\n');
   }
 
-  protected _getScalar(name: string): string {
-    return `Scalars['${name}']`;
+  protected _getScalar(name: string, type: 'input' | 'output'): string {
+    return `Scalars['${name}']['${type}']`;
   }
 
   protected _getDirectiveArgumentNadInputFieldMapping(name: string): string {
@@ -968,11 +980,11 @@ export class BaseTypesVisitor<
     return type || null;
   }
 
-  protected _getTypeForNode(node: NamedTypeNode): string {
+  protected _getTypeForNode(node: NamedTypeNode, isVisitingInputType: boolean): string {
     const typeAsString = node.name as any as string;
 
     if (this.scalars[typeAsString]) {
-      return this._getScalar(typeAsString);
+      return this._getScalar(typeAsString, isVisitingInputType ? 'input' : 'output');
     }
     if (this.config.enumValues[typeAsString]) {
       return this.config.enumValues[typeAsString].typeIdentifier;
@@ -990,7 +1002,7 @@ export class BaseTypesVisitor<
   NamedType(node: NamedTypeNode, key, parent, path, ancestors): string {
     const currentVisitContext = this.getVisitorKindContextFromAncestors(ancestors);
     const isVisitingInputType = currentVisitContext.includes(Kind.INPUT_OBJECT_TYPE_DEFINITION);
-    const typeToUse = this._getTypeForNode(node);
+    const typeToUse = this._getTypeForNode(node, isVisitingInputType);
 
     if (!isVisitingInputType && this.config.fieldWrapperValue && this.config.wrapFieldDefinitions) {
       return `FieldWrapper<${typeToUse}>`;
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-processor/pre-resolve-types.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-processor/pre-resolve-types.ts
@@ -56,7 +56,7 @@ export class PreResolveTypesProcessor extends BaseSelectionSetProcessor<Selectio
           (this.config.namespacedImportName ? `${this.config.namespacedImportName}.` : '') +
           this.config.convertName(baseType.name, { useTypesPrefix: this.config.enumPrefix });
       } else if (this.config.scalars[baseType.name]) {
-        typeToUse = this.config.scalars[baseType.name];
+        typeToUse = this.config.scalars[baseType.name].output;
       }
 
       const wrappedType = this.config.wrapTypeWithModifiers(typeToUse, fieldObj.type);
@@ -87,7 +87,7 @@ export class PreResolveTypesProcessor extends BaseSelectionSetProcessor<Selectio
       }
       const fieldObj = schemaType.getFields()[aliasedField.fieldName];
       const baseType = getBaseType(fieldObj.type);
-      let typeToUse = this.config.scalars[baseType.name] || baseType.name;
+      let typeToUse = this.config.scalars[baseType.name]?.output || baseType.name;
 
       if (isEnumType(baseType)) {
         typeToUse =
--- a/packages/plugins/other/visitor-plugin-common/src/utils.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/utils.ts
@@ -284,8 +284,25 @@ export function buildScalars(
 ): ParsedScalarsMap {
   const result: ParsedScalarsMap = {};
 
+  function normalizeScalarType(type: string | { input: string; output: string }): { input: string; output: string } {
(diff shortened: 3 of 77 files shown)


packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -677,7 +689,7 @@ export class BaseTypesVisitor<
     if (this.config.onlyOperationTypes || this.config.onlyEnums) return '';
     const originalNode = parent[key] as UnionTypeDefinitionNode;
     const possibleTypes = originalNode.types
-      .map(t => (this.scalars[t.name.value] ? this._getScalar(t.name.value) : this.convertName(t)))
+      .map(t => (this.scalars[t.name.value] ? this._getScalar(t.name.value, 'output') : this.convertName(t)))
```
eddeee888, line 692: Scalars in union types are always output.

packages/plugins/other/visitor-plugin-common/src/selection-set-processor/pre-resolve-types.ts
```diff
@@ -56,7 +56,7 @@ export class PreResolveTypesProcessor extends BaseSelectionSetProcessor<Selectio
           (this.config.namespacedImportName ? `${this.config.namespacedImportName}.` : '') +
           this.config.convertName(baseType.name, { useTypesPrefix: this.config.enumPrefix });
       } else if (this.config.scalars[baseType.name]) {
-        typeToUse = this.config.scalars[baseType.name];
+        typeToUse = this.config.scalars[baseType.name].output;
```
eddeee888, line 59: This is used in selection set so they are always output.

packages/plugins/other/visitor-plugin-common/src/selection-set-processor/pre-resolve-types.ts
```diff
@@ -87,7 +87,7 @@ export class PreResolveTypesProcessor extends BaseSelectionSetProcessor<Selectio
       }
       const fieldObj = schemaType.getFields()[aliasedField.fieldName];
       const baseType = getBaseType(fieldObj.type);
-      let typeToUse = this.config.scalars[baseType.name] || baseType.name;
+      let typeToUse = this.config.scalars[baseType.name]?.output || baseType.name;
```
eddeee888, line 90: This is used in selection set so they are always output.

packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -297,28 +314,87 @@ export function buildScalars(
...
+            result[name] = {
+              input: {
+                isExternal: false,
+                type: JSON.stringify(mappedScalar),
+              },
+              output: {
+                isExternal: false,
+                type: JSON.stringify(mappedScalar),
+              },
+            };
```
eddeee888, line 372: Previously, this was stringifying any object:

```ts
result[name] = {
  isExternal: false,
  type: JSON.stringify(scalarsMapping[name]),
};
```

I cannot find tests related to this case. However, this behaviour is backward compatible if the mapper object does not have input/output fields.

packages/plugins/other/visitor-plugin-common/src/variables-to-object.ts
```diff
@@ -60,7 +60,7 @@ export class OperationVariablesToObject {
   protected getScalar(name: string): string {
     const prefix = this._namespacedImportName ? `${this._namespacedImportName}.` : '';
 
-    return `${prefix}Scalars['${name}']`;
+    return `${prefix}Scalars['${name}']['input']`;
```
eddeee888, line 63: Scalar types used in variables are always input.

packages/plugins/typescript/typescript/tests/typescript.spec.ts
```diff
@@ -934,10 +976,11 @@ describe('TypeScript', () => {
     });
 
     it('#1488 - Should generate readonly also in input types when immutableTypes is set', async () => {
-      const schema = buildSchema(`
-      input MyInput {
-        f: String!
-      }`);
+      const schema = buildSchema(/* GraphQL */ `
```
eddeee888, line 979: I added `/* GraphQL */` to some of the updated tests to help formatting

saihaj (approved):
thank you!

dobesv:
Any reference to where `ID` could be a number?  The docs seem to say it is always a `string`.

dobesv:
Hmm that doesn't specifically mandate support for number, and also mentions GUID and other things.  I wonder if the change here would have been less disruptive if it stuck with `string` as the default for backwards compatibility and offered people the chance to add alternative types if they are using them.

eddeee888:
Hi @dobesv 👋 Are you generating types for clients or servers?

For client types, I didn't think it would be a big change since `number` can now be sent as well as `string`.
For server types, ID input is always coerced into `string` when it gets to the resolvers, and resolvers can return `string | number` so the following config is required:

```ts
   // ...
   generates: {
     'path/to/file': {
       // plugins...
       config: {
         scalars: {
           ID: { 
             input: 'string', 
             output: 'string | number' 
           },
         }
       },
     },
   },
};
```

Alternatively, if you want the previous behaviour, this can be used:

```ts
   // ...
   generates: {
     'path/to/file': {
       // plugins...
       config: {
         scalars: {
           ID: 'string'
         }
       },
     },
   },
};
```

eddeee888:
Hi @mikeverf, thanks for providing the sandbox! It makes it much easier to understand the setup 🙌

In your case, you'd have to also bump `@graphql-codegen/typescript-operations` to v4.0.0 like this [example](https://codesandbox.io/p/sandbox/beautiful-sea-klmd9j?file=%2Fpackage.json%3A13%2C45)

Here's the updated [generated template with input reference](https://codesandbox.io/p/sandbox/beautiful-sea-klmd9j?file=%2Fsrc%2Fapi.hooks.ts%3A25%2C36):

```ts
export type GetItemQueryVariables = Types.Exact<{
  id: Types.Scalars['ID']['input'];
}>;
```

---

Please note that the core Scalar setup lives in `@graphql-codegen/visitor-plugin-common` v4. 

I'm seeing two packages bringing an older version (v2.13.1) of `@graphql-codegen/visitor-plugin-common` in the yarn.lock: [@graphql-codegen/near-operation-file-preset](https://codesandbox.io/p/sandbox/beautiful-sea-klmd9j?file=%2Fyarn.lock%3A1103%2C1-1114%2C1) and [@graphql-codegen/typescript-react-query](https://codesandbox.io/p/sandbox/beautiful-sea-klmd9j?file=%2Fyarn.lock%3A1171%2C1-1180%2C19).
In this case it's still resolving the correct package. However, it's best to bump those as well to ensure there's only one version of the package. They should live in the community repo:

- [React Query](https://github.com/dotansimha/graphql-code-generator-community/tree/main/packages/plugins/typescript/react-query)
- [Near Operation File Preset](https://github.com/dotansimha/graphql-code-generator-community/tree/main/packages/presets/near-operation-file)

Wytrykus:
Hi, in my case the new input/output structure of scalars does not work. I use the following codegen config:
```
     plugins:
      - 'typescript'
      - 'typescript-operations'
      - 'typescript-graphql-request'
    config:
      avoidOptionals: true
      skipTypename: true
      typesPrefix: 'Foundation'
```
In a query have a parameter (`tenantId`) which is GraphQL type `String` and in the generated code I have
```
export type Scalars = {
  ID: { input: string | number; output: string };
  String: { input: string; output: string };
  Boolean: { input: boolean; output: boolean };
  Int: { input: number; output: number };
  Float: { input: number; output: number };
  Date: { input: any; output: any };
};
```
I assigned my `tenantId` variable first only to `input`, then also to both `input` and `output` field of the parameter object in the generated code, but I always got an error response from the GraphQL server:
```
ClientError: Graphql validation error
```

eddeee888:
Hi @Wytrykus , do you mind creating an issue with a sandbox please? :)

dobesv:
FWIW the workaround here was easy enough, setting it explicitly to string for both, but I can imagine that making 99% of people using the tool make that change is probably unnecessary busywork since I assume the vast majority are in fact using strings.

I'm glad to see it changed to be backwards compatible again even though I already put in the fix in our codebase.

adevine:
Just a comment about this - this was a major breaking change for us, and it would be really helpful if the new Scalars behavior was more easily customizable to generate the old way.

We use things like `someProp: Scalars['ID'];` _all over the place_ in our code because it essentially provides better documentation. We have a lot of custom scalars for which we do addition validation (e.g. email addresses) and doing something like `emailAddress: Scalars['EmailString']` really helps inform the developer that the property should conform to our EmailString rules, even if under the covers it just resolves to `string`.

Is there any way to revert this to get the old behavior across the board?

dobesv:
> Is there any way to revert this to get the old behavior across the board?

Was it not reverted in #9501 ?

If that wasn't released you may want to just stay on an older version until it is released.

dobesv:
Or is it that you wanted to keep using `Scalars["ID"]` but now you would have to use `Scalars["ID"]["input"]` ?  Hmm, interesting.

adevine:
@dobesv , yes, the thing that was reverted was just that the input type for ID was just made to be `string` instead of `string | number`, but the input/output object structure was still kept.

In my opinion the faulty assumption here is that `Scalars` would only be used internally for properties that codegen also creates, but we use it extensively as-is throughout our entire codebase for better documentation. Also, in reality, I think ID is a bit of special case it that it basically just supports the number coercion to string as a sort of "helpful edge case", but now even that has just been reverted back to string, so I really question how much of a need it is to have separate input/output types for scalars.

adevine:
Just adding this note here in case anyone is in the same predicament we are. While for now we've reverted our code to a pre 4.x version, the following type definition basically gives you the Scalars definition in the "old" format, so we'll use this as a way to decouple from the changes made in this PR:

`type FixedScalars = { [Property in keyof Scalars]: Scalars[Property]['output'] };`

eddeee888:
Hello @adevine,

Short answer: scalar input/output type will not be reverted.

Long answer:

Scalar can always be used as either input or output in GraphQL. So, having only one type for both input/output does not allow types to be set correctly in many scenarios. For example, resolvers could never be typed to take `string` as ID input and return `string | number` for IDs until this change.

Some setups may not need this, but some setups do.

> Just a comment about this - this was a major breaking change for us

This was released as a major version. So, we expect consumers to have breaking changes.

> , and it would be really helpful if the new Scalars behavior was more easily customizable to generate the old way across the board (i.e. just primitives, no input/output).

Unfortunately, maintaining both approaches in the core plugins means two things:

1. A lot of maintenance for the core maintainers
2. A lot of maintenance for the community libraries that depend on the core plugins

Adding more config options and logic means more maintenance, which hinders our ability to deliver ongoing value to the community.

> In my opinion the faulty assumption here is that Scalars would only be used internally for properties that codegen also creates, but we use it extensively as-is throughout our entire codebase for better documentation. 

I can confirm that this assumption was _not_ made. I refer to the generated `Scalars` in professional and production settings as well (maybe in different use cases than what you may have). So, I’m aware that this is a breaking change.

Personally, I fixed my input/output references based on the scalar flow. But if you have the same input and output type, one option is to find all references to `Scalars['EmailString']` and change to `Scalars['EmailString']['output']` (or using your proposed approach)?

> Also, in reality, I think ID is a bit of special case in that it basically just supports the number coercion to string as a sort of “helpful edge case”, but now even that has just been reverted back to string, so I really question how much of a need it is to have separate input/output types for scalars.

There’s the use case of custom scalars. In general, custom scalars can take any number of incoming types and coerced the value into one type (regardless of whether the scalar is used as input or output).

Ultimately, I believe this is a necessary evil. We needed to make this breaking change to make types more flexible and correct.

adevine:
Thanks for the response @eddeee888 . And my apologies, I didn't mean to come across as overly harsh.

I fully understand the custom scalars issue (we use custom scalars heavily), I just think it would have been less detrimental to add a second `InputScalars` type while leaving the existing `Scalars` type as-is. That would have made it possible to support this requirement in a completely backwards-compatible fashion, and it also means you only have to worry about InputScalars if you actually need to have separate types for input and output (which is much less common than the default when they are the same), as the default would just be to have InputScalars essentially inherit from Scalars.

But in any case, it's always easy to gripe from the peanut gallery. We'll go forth using the `{ [Property in keyof Scalars]: Scalars[Property]['output'] }` approach to make our migration less painful.

PR #9418, opened 2023-05-16
Fix scalar input/output object bug when given external file or module pattern

eddeee888:
### 1. Fix a missed case when handling input/output scalar object
This fixes the issue originally implemented in https://github.com/dotansimha/graphql-code-generator/pull/9375

Previously, we print the input/output value without parsing it through mapper. This means with the following config:

```ts
scalars: {
  MyScalar: {
    input: '@org/scalars#Input',
    output: '@org/scalars#Output'
  }
}
```

It will print the following invalid TypeScript syntax 😓 :

```ts
export type Scalars = {
  MyScalar:  {
    input: @org/scalars#Input
    output: @org/scalars#Output
  }
}
```
---

Now, input/output scalar object should now handle file and module import mapper consistently i.e. import at the top and reference the imports in the `Scalars` object:

```ts
import { Input } from '@org/scalars';
import { Output } from '@org/scalars';

export type Scalars = {
  MyScalar:  {
    input: Input
    output: Output
  }
}
```

### 2. Ensure backward compatibility when using mapped Scalars

Previous implementation expects external mapped scalar cases have input/output object. Below is the config for these cases

```ts
scalars: {
  MyScalar: '@org/scalars#MyScalar',
  MyRelativeScalar: './scalars#MyRelativeScalar',
}
```

Previously the following would be generated

```ts
import { MyScalar } from '@org/scalars';
import { MyRelativeScalar } from './scalars';

export type Scalars = {
  MyScalar:  {
    input: MyScalar['input']
    output: MyScalar['output']
  }
  MyRelativeScalar: {
    input: MyScalar['input']
    output: MyScalar['output']  
  }
}
```

This is big breaking change for downstream users as all mapped scalar currently is a type instead of an input/output object. This PR updates it so mapped scalars will be used for both input and output:

```ts
import { MyScalar } from '@org/scalars';
import { MyRelativeScalar } from './scalars';

export type Scalars = {
  MyScalar:  {
    input: MyScalar
    output: MyScalar
  }
  MyRelativeScalar: {
    input: MyScalar
    output: MyScalar
  }
}
```

## Type of change

## How Has This Been Tested?

- [x] Unit test

## Checklist:


dotansimha (approved):

packages/plugins/typescript/typescript/tests/typescript.spec.ts
```diff
@@ -1997,6 +2009,9 @@ describe('TypeScript', () => {
           MyScalar: { input: MyScalar['input']; output: MyScalar['output']; }
           MyOtherScalar: { input: MyOtherScalar['input']; output: MyOtherScalar['output']; }
           MyAliasedScalar: { input: AliasedScalar['input']; output: AliasedScalar['output']; }
+          OrgScalar: { input: OrgScalar['input']; output: OrgScalar['output']; }
+          OrgOtherScalar: { input: OrgOtherScalar['input']; output: OrgOtherScalar['output']; }
+          OrgAliasedScalar: { input: OrgAliasedScalar['input']; output: OrgAliasedScalar['output']; }
```
eddeee888, line 2014: I think it would have a much better backward compatibility if we don't require mapped scalar from external sources to be an input/output object

Will push a commit to fix this.

eddeee888: Fixed in https://github.com/dotansimha/graphql-code-generator/pull/9418/commits/2665abf09ccd813dc994d824d015845315d203ff

AaronMoat: You've already done it, but agreed, this last commit is quite nice. Means the only code change we _had_ to do in our (internal) project was stop using Scalars[''] or add the ['input'] / ['output'] type.

n1ru4l (approved):

PR #9449, opened 2023-05-22
node 16 upgrade

n1ru4l:


B2o5T:
enable follow rules too https://github.com/dotansimha/graphql-code-generator/blob/master/.eslintrc.cjs#L27-L28

PR #9459, opened 2023-05-25
fix examples for generated docs

B2o5T:
https://guild-oss.slack.com/archives/CCVDM3NCD/p1682592308465249

fixes a weird bug reported by @eddeee888 

upstream https://github.com/YousefED/typescript-json-schema/issues/431


website/src/lib/plugin-get-static-props.ts
```diff
@@ -47,7 +47,5 @@ export const pluginGetStaticProps = (fileName: string) => async () => {
         compiledHeader: mdxHeader.result,
       },
     },
-    // The page will be considered as stale and regenerated every 24 hours.
-    revalidate: 60 * 60 * 24,
```
saihaj, line 51: do we not want to re-generate?

saihaj (approved):

website/src/lib/plugin-get-static-props.ts
B2o5T: It never worked because we use `next export` 😅

Issue #9104, opened 2023-03-04
client preset is ignoring a lot of configs


christiansany:
### Describe the bug

The Configs that we should be able to use with the new client preset don't do anything.
The client preset should use the typescript plugin underneath, but the config that can be used with the typescript plugin doesn't do anything.

Can we fix this?

In Addition, the client preset generates types for the Schema and the Queries. But in the index.ts file that is generated these types are not exported.

### Steps to Reproduce the Bug or Issue

1. Try it in any project that is using the client preset

### Expected behavior

I want the config for the typescript plugin to also work with the client preset

### Platform

- OS: [e.g. macOS, Windows, Linux]
- NodeJS: [e.g. 18.5.0]
- `graphql` version: [e.g. 16.3.0]
- `@graphql-codegen/*` version(s): [e.g. 2.6.2]

### Codegen Config File

```ts
import { CodegenConfig } from "@graphql-codegen/cli";

const config: CodegenConfig = {
  schema: "./__generated__/schema.graphql",
  documents: [
    "./**/*.js",
    "./**/*.vue",
    "./**/*.ts",
    "!./__generated__/schema.graphql",
    "!./**/node_modules/**/*",
    "!./**/*.test.*",
    "!./**/*.d.ts",
  ],
  generates: {
    "./__generated__/gql/": {
      preset: "client",
      presetConfig: {
        gqlTagName: "gql",
        fragmentMasking: { unmaskFunctionName: "getFragmentData" },
        enumsAsConst: true, // Doesn't do anything (probably should not work, but I tried it anyway)
        immutableTypes: true, // Doesn't do anything (probably should not work, but I tried it anyway)
      },
      config: {
        enumsAsConst: true, // Doesn't do anything
        immutableTypes: true, // Doesn't do anything
      },
    },
    "./__generated__/schema.ts": {
      plugins: ["typescript"],
      config: {
        enumsAsConst: true, // Works fine here
        immutableTypes: true, // Works fine here
      },
    },
  },
  ignoreNoDocuments: true,
};

export default config;
```

mogelbrod:
Here's another bug report related to this:
https://github.com/dotansimha/graphql-code-generator/issues/8576#issuecomment-1422205746

2snEM6:
Any news on this? This is key so for example https://the-guild.dev/graphql/codegen/plugins/typescript/relay-operation-optimizer can be used with `client-preset`

n1ru4l:
[...]

Issue #8576, opened 2022-11-04
`client-preset` does not forward `dedupeOperationSuffix`


ntungare:
### Describe the bug

The `client-preset` forwards a very small sub-set of configs to the lower level as can be seen [here](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/presets/client/src/index.ts#L85).

This is problematic since I'm not able to set some of the options supported by the the `typescript-operations` [plugin](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/operations/src/config.ts).

### Your Example Website or App

https://github.com/ntungare/graphql-codegen-example

### Steps to Reproduce the Bug or Issue

Use the codegen config attached, the effect of the `dedupeOperationSuffix` is not seen.

### Expected behavior

All configs are respected since the preset is just a combination of the plugin.

I'd like to not use the `gql-tag-operations-preset` but this is blocker for that. And the `gql-tag-operations-preset` is supposedly deprecated?

### Platform

- OS: [e.g. macOS, Windows, Linux]
- NodeJS: [e.g. 18.5.0]
- `"graphql": "^16.6.0"`
- `"graphql-tag": "^2.12.6"`
- `"@graphql-codegen/cli": "^2.13.8"`
- `"@graphql-codegen/gql-tag-operations-preset": "^1.7.0"`

### Codegen Config File

```ts
import type { CodegenConfig } from "@graphql-codegen/cli";

const config: CodegenConfig = {
    generates: {
        "src/generated/": {
            schema: "https://countries.trevorblades.com/",
            documents: "src/query/**/*.ts",
            plugins: [],
            preset: "gql-tag-operations-preset",
            presetConfig: {
                augmentedModuleName: "graphql-tag",
            },
            config: {
                dedupeFragments: true,
                dedupeOperationSuffix: true,
            },
        },
    },
};

export default config;
```

### Additional context

Using the above config the `dedupeFragments` is respected but not the `dedupeOperationSuffix` config.

mogelbrod:
There seems to be a bunch of options that aren't being forwarded right now 😞
While attempting to migrate to client-preset I couldn't find any way to apply the following options:
```yaml
config:
  typesPrefix: I
  enumPrefix: I
  enumsAsTypes: true
  documentVariableSuffix: Doc
  fragmentVariableSuffix: Fragment
  omitOperationSuffix: true
  dedupeFragments: true
  exportFragmentSpreadSubTypes: true
[...]

Issue #8562, opened 2022-11-02
[client-preset] allowed options and defaults


charlypoly:
This issue is the place to discuss allowed options and default values when using the `preset: 'client'` setup.

## Current supported options

- `scalars`
- `strictScalars`
- `namingConvention`
- `useTypeImports`
- `skipTypename`
- `enumsAsTypes`
- `arrayInputCoercion`
- `presetConfig.fragmentMasking`
- `presetConfig.gqlTagName`
- `presetConfig.unmaskFunctionName`
- `emitLegacyCommonJSImports`

## Requested options support
- `nonOptionalTypename` (https://github.com/dotansimha/graphql-code-generator/issues/8296#issuecomment-1287010974)
- `immutableTypes` (https://github.com/dotansimha/graphql-code-generator/issues/8296#issuecomment-1287010974)

## Current defaults
- `presetConfig.fragmentMasking: true`
- `inlineFragmentTypes: true` (if fragment masking enabled)

## Requested defaults
- `ignoreNoDocuments: true` (@charlypoly)
- should `presetConfig.fragmentMasking` be `false` by default? (@charlypoly)
- `dedupeFragments: true` (https://github.com/dotansimha/graphql-code-generator/issues/8103#issuecomment-1298692748)

n1ru4l:
`useTypeImports` should be on by default IMHO, as it is considered a best practice by me.

n1ru4l:
`nonOptionalTypename` is dangerous IMHO - It allows the wrong configuration.

E.g. the correct type for

```graphql
query MeQuery {
  me {
    id
  }
}
```

is 

```ts
type MeQuery = {
  me: {
    id: string
  }
}
```

not

```ts
type MeQuery = {
  me: {
    __typename?: "User",
    id: string
  }
}
```

or

```ts
type MeQuery = {
  me: {
    __typename: "User",
    id: string
  }
}
```

However, GraphQL clients such as apollo and urql ADD `__typename` selections to the documents before sending them to the GraphQL server. I am unsure whether I like this behavior. But if you look at this it makes sense why the defaults are like they are today.

Having that said, for me, the following makes the most sense:

- `__typename` is OMITTED from the generated types by default.

______

Adding the `__typename` property to the generated type via a configuration option by default, either as optional or mandatory - is something I would ideally try to avoid - but I can also see why people don't want to explicitly clutter their GraphQL Operation documents with `__typename` selections.

[...]

PR #9496, opened 2023-06-10
Fix options getting swallowed by client preset

jwueller:
## Description

This deals with #8576, #8993 and #9104 by passing through _all_ options by default instead of only allowing a small subset with no way to override that behavior. The original code doesn't give a rationale for doing this.

A few computed values are inserted into the config by the preset, and those will keep overriding the user options as of right now, because I assume that they required for the preset to work as expected.

This change should only be noticeable to users that were already passing options that didn't work/exist before. It basically has the same impact as adding a new option.

## Type of change

## How Has This Been Tested?

- [x] `yarn test` doesn't show additional failures (there were some pre-existing ones before this change though)
- [x] Added a test for using `dedupeOperationSuffix` with the client preset (failed before, passes after)

**Test Environment**:

- OS: Linux 6.3.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 30 May 2023 13:44:01 +0000 x86_64 GNU/Linux
- `@graphql-codegen/client-preset@4.0.0`
- NodeJS: v20.2.0

## Checklist:

### DIFF
--- a/packages/presets/client/src/index.ts
+++ b/packages/presets/client/src/index.ts
@@ -116,22 +116,6 @@ export const preset: Types.OutputPreset<ClientPresetConfig> = {
 
     const reexports: Array<string> = [];
 
-    // the `client` preset is restricting the config options inherited from `typescript`, `typescript-operations` and others.
-    const forwardedConfig = {
-      scalars: options.config.scalars,
-      defaultScalarType: options.config.defaultScalarType,
-      strictScalars: options.config.strictScalars,
-      namingConvention: options.config.namingConvention,
-      useTypeImports: options.config.useTypeImports,
-      skipTypename: options.config.skipTypename,
-      arrayInputCoercion: options.config.arrayInputCoercion,
-      enumsAsTypes: options.config.enumsAsTypes,
-      dedupeFragments: options.config.dedupeFragments,
-      nonOptionalTypename: options.config.nonOptionalTypename,
-      avoidOptionals: options.config.avoidOptionals,
-      documentMode: options.config.documentMode,
-    };
-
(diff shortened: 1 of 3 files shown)


jwueller:
I'm not certain whether I should be the one adding a changeset to this. This would probably be a "patch" version bump based on my impact assessment.

n1ru4l:
This is not a bug, but a strategic decision. We decided that we don’t allow all typescript plugin options, as we want to be more opinionated and later-on replace the internals of the preset anyways to no longer use the typescript and typescript-operations plugins.

mogelbrod:
@n1ru4l Where do you want us to give feedback? Should we open a github issue for each individual setting we're missing and utilize the thumbs up feature? In that case issues like https://github.com/dotansimha/graphql-code-generator/issues/8576 should probably be reopened.

jwueller:
@n1ru4l This is a very surprising decision, as it arbitrarily and severely reduces the usefulness of the preset.

I understand that you might want things to behave slightly differently in the future, but I don't see why it has to be _forbidden_ until that decision actually gets made. It's going to be a backwards-incompatible change either way.

This is compounded by issues with the opinionated approach seemingly not getting addressed for a few months (see #9104, #8993 and #8576), which is presumably due to you having more important things to deal with. I feel like this change could reduce your workload in that regard, because it doesn't require dealing with a new issue and PR every time someone needs another option. This is assuming the user even bothers making an issue for it.

If I'm not able to override the preset in any way, I will unfortunately have to replace it entirely, even though it has been very useful up to this point.

n1ru4l:
@mogelbrod https://github.com/dotansimha/graphql-code-generator/issues/8562

PR #9489, opened 2023-06-06
fix: SWC plugin

YassinEldeeb:
## Fix reproductions and close issues

- [ ] https://github.com/awinograd/graphql-codegen-client-preset-swc-oom
- [x] https://github.com/dotansimha/graphql-code-generator/issues/9145
- [x] https://github.com/dotansimha/graphql-code-generator/issues/9195
- [ ] https://github.com/dotansimha/graphql-code-generator/issues/9337
- [x] https://github.com/dotansimha/graphql-code-generator/issues/9445
- [ ] https://github.com/huv1k/nextjs-auth-prisma
- [x] https://github.com/dotansimha/graphql-code-generator/issues/9450
- [ ] https://github.com/dotansimha/graphql-code-generator/issues/9457

PR #9567, opened 2023-07-25
fix: swc plugin

CaptainN:
This is the (mostly) same as PR #9489 except it only includes the one commit.

I started with the relevant commit from that other PR, and I changed very few additional things, just enough to get things to compile - specifically, I updated swc_core to the latest version.

I did test this, and it didn't work. I'm making this request in the hope that it's easy for someone who knows a bit more about how this goes together to get it working without too much effort.


Daniel3711997:
any news here?

Issue #9584, opened 2023-08-04
TypeScript error TS6133: 'RefType' is declared but its value is never read


josendf-basetis:
### Is your feature request related to a problem? Please describe.

Given the following GraphQL schema, having types implementing an interface:

interface Node {
    id: ID!
}

type Country implements Node {
    id: ID!
    countryName: String!
}

type City implements Node {
    id: ID!
    cityName: String!
}

The following type declaration is generated:

/** Mapping of interface types */
export type ResolversInterfaceTypes<RefType extends Record<string, unknown>> = {
    Node: City | Country;
};

The generated type declaration above produces the following TypeScript error:

error TS6133: 'RefType' is declared but its value is never read.

### Describe the solution you'd like

Unused TypeScript type parameters should not be emitted.

eddeee888:
Hello 👋 

There's a similar issue and here's the recommendations: https://github.com/dotansimha/graphql-code-generator/issues/9498#issuecomment-1598550662

Basically, the errors reported are from TS's `noUnusedLocals` and `noUnusedParameters` settings. It's recommended to ignore linting rules in generated files if they get in the way 🙂

shellscape:
@eddeee888 those aren't "linting" errors. That is a literal and legitimate TypeScript configuration option. Also doesn't change the fact that the code generator is creating invalid code per valid TS config.

Antcharles22:
Hey ,
i'm creating a react project and I can't import my JS functions from other pages

skizzo:
Any updates on this?

diaes:
Problem is still there, is there any work planned on this?

dannyuttarwar:
I faced the same issue after updating the dependencies So I fixed using the below steps

1. Created a codegen.yml
2. Disabled the ts-nocheck in same yaml file Like below

```
schema: src/**/*.schema.ts
generates:
  ./src/generated/generatedTypes.ts:
    plugins:
      - add:
          content: |
            /* eslint-disable */
            // @ts-nocheck
      - typescript
      - typescript-resolvers
```

eddeee888:
Hi all 👋 

Could you please try this alpha version to see if it fixes your issue:

```
@graphql-codegen/typescript-resolvers@4.0.7-alpha-20240425115511-1ac284ac51b27d312ef3f9922218f228a3b468e3
```

pdavies:
[...]

PR #9593, opened 2023-08-16
Support Generating a Record Type

aidant:
## Description

When defining a GraphQL type of name `Record` the generated types reference the generated `Record` type rather than the `Record` type from TypeScript when trying to define a generic for `ResolversUnionTypes` and `ResolversInterfaceTypes`. 

## Type of change

## How Has This Been Tested?

The following patch has been applied in the repo with the effected types:

```
diff --git a/node_modules/@graphql-codegen/visitor-plugin-common/cjs/base-resolvers-visitor.js b/node_modules/@graphql-codegen/visitor-plugin-common/cjs/base-resolvers-visitor.js
index 4d4adf3..cdefaba 100644
--- a/node_modules/@graphql-codegen/visitor-plugin-common/cjs/base-resolvers-visitor.js
+++ b/node_modules/@graphql-codegen/visitor-plugin-common/cjs/base-resolvers-visitor.js
@@ -415,7 +415,7 @@ class BaseResolversVisitor extends base_visitor_js_1.BaseVisitor {
         return new utils_js_1.DeclarationBlock(this._declarationBlockConfig)
             .export()
             .asKind(declarationKind)
-            .withName(this.convertName('ResolversInterfaceTypes'), `<RefType extends Record<string, unknown>>`)
+            .withName(this.convertName('ResolversInterfaceTypes'), `<RefType extends { [key: string]: unknown }>`)
             .withComment('Mapping of interface types')
             .withBlock(Object.entries(this._resolversInterfaceTypes)
             .map(([typeName, value]) => (0, utils_js_1.indent)(`${typeName}: ${value}${this.getPunctuation(declarationKind)}`))
```

## Checklist:

### DIFF
--- a/dev-test/modules/types.ts
+++ b/dev-test/modules/types.ts
@@ -165,7 +165,7 @@ export type DirectiveResolverFn<TResult = {}, TParent = {}, TContext = {}, TArgs
 ) => TResult | Promise<TResult>;
 
 /** Mapping of union types */
-export type ResolversUnionTypes<RefType extends Record<string, unknown>> = {
+export type ResolversUnionTypes<RefType extends { [key: string]: unknown }> = {
   PaymentOption: CreditCard | Paypal;
 };
 
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -1137,7 +1137,7 @@ export class BaseResolversVisitor<
     return new DeclarationBlock(this._declarationBlockConfig)
       .export()
       .asKind(declarationKind)
-      .withName(this.convertName('ResolversUnionTypes'), `<RefType extends Record<string, unknown>>`)
+      .withName(this.convertName('ResolversUnionTypes'), `<RefType extends { [key: string]: unknown }>`)
       .withComment('Mapping of union types')
       .withBlock(
         Object.entries(this._resolversUnionTypes)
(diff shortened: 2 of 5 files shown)


eddeee888:
Hi @aidant , can you help me understand why the issue you are facing with `Record<string, unknown>` please? 🙂
Or maybe some resources explaining the difference between `Record<string, unknown>` and `{ [key: string]: unknown }`

gilgardosh:
If I understand correct, the aforementioned issue is dealing with a case when schema contains GraphQL type named `Record`, which then overrides TS's `Record` resulting in some issues, did I get that right @aidant ?

If so, we should add a mechanism to prevent this (and any other usage of preserved TS utility types names).
In any case, the approach shouldn't be avoiding utility types as in the suggested fix, but renaming the schema types with some consistent convention.

aidant:
@gilgardosh correct. I would be happy to avoid this issue with a prefix of all types by using the `typesPrefix` option. Which I did try at some point as we would like to have all of our generated types prefixed with `Api` but there are unrelated issues with the generated types when a prefix is used. I don't remember the exact details off the top of my head, but can open a separate issue once I have time. 

I understand your hesitations for avoiding utility types and am happy for a different fix to be implemented in the long term. But for an immediate fix this elevates the issue as `Record<string, unknown>>` and `{ [key: string]: unknown }` are equivalent types.

Something else to consider is this generic is not used in certain cases (including mine), which seems to be causing issues for people with what I suspect is `noUnusedLocals` and/or `noUnusedParameter` enabled, see #9584

Issue #9445, opened 2023-05-20
@graphql-codegen/client-preset-swc-plugin adds imports before "use client"


victorandree:
### Which packages are impacted by your issue?

@graphql-codegen/cli, @graphql-codegen/client-preset

### Describe the bug

`@graphql-codegen/client-preset-swc-plugin` inserts imports at the top of modules where you define queries. This breaks modules that are defined as React client components using the "use client" directive, which must appear at the top of the file - because the inserted imports end up before the directive.

This means you cannot define GraphQL queries in the same module as a client component when using the Next.js app directory.

Please see the example repository for a reproduction.

### Your Example Website or App

https://github.com/victorandree/graphql-code-generator-swc-app-dir

### Steps to Reproduce the Bug or Issue

1. Create a client component in a Next.js app using the "app routes", requiring the use of a `"use client"` directive at the top of the module.
2. Define a GraphQL query in this module.
3. Set up GraphQL code generator with the "client preset" and using the SWC plugin.
4. Try to build the Next.js app

The build will fail with a message like the following:

```shell
$ npm run dev

[...]
- error ./app/QueryComponent.tsx
ReactServerComponentsError:

The "use client" directive must be placed before other expressions. Move it to the top of the file to resolve this issue.

   ,-[/graphql-code-generator-swc-app-dir/app/QueryComponent.tsx:1:1]
 1 | 'use client';
   : ^^^^^^^^^^^^^
 2 |
 3 | import { useQuery } from '@apollo/client';
 4 | import { graphql } from '../gql';
   `----

Import path:
  ./app/QueryComponent.tsx
  ./app/page.tsx
```

### Expected behavior

Since ["use client" is required to be at the top of the file by the React Server Components RFC](https://github.com/reactjs/rfcs/pull/227), I expect the SWC plugin to insert its imports after this directive.

Expressed differently: I expect to be able to define GraphQL queries in client components.

### Platform

- OS: macOS
- NodeJS: 16.19.1
- `graphql` version: 16.6.0
- `@graphql-codegen/cli`: 3.3.1
- `@graphql-codegen/client-preset`: 3.0.1
- `@graphql-codegen/client-preset-swc-plugin`: 0.2.0

### Codegen Config File

```typescript
import { CodegenConfig } from '@graphql-codegen/cli';

const config: CodegenConfig = {
  schema: 'schema.graphql',
  documents: 'app/**/*.tsx',
  generates: {
    'gql/': {
      preset: 'client',
[...]

PR #9603, opened 2023-08-18
Add SWC plugin imports after "use client" directive (#9445)

victorandree:
## Description

When adding a "use client" directive to a module, it must come before any other expressions. Instead of always inserting import declarations as the first item, the SWC plugin now checks if the first expression in a module's body is a "use client" directive and inserts its imports after it, if that's the case.

Related: #9445

## Type of change

## How Has This Been Tested?

I've tested this by using it with a Next.js app directory project where I defined a query in the same module as a "use client" directive; verifying that it would not build before but would with this fix.

I have not updated the unit tests for the Rust plugin, to verify that imports are inserted after the "use client" directive. Should probably add that.

**Test Environment**:

- OS: macOS Ventura 13.5
- `@graphql-codegen/cli`: `^5.0.0`
- `@graphql-codegen/client-preset`: `^4.1.0`
- NodeJS: v20.5.1

## Checklist:

## Further comments

### DIFF
--- a/packages/presets/swc-plugin/src/lib.rs
+++ b/packages/presets/swc-plugin/src/lib.rs
@@ -6,6 +6,7 @@ use swc_core::{
     common::Span,
     ecma::{
         ast::*,
+        atoms::atom,
         utils::quote_ident,
         visit::{as_folder, FoldWith, VisitMut, VisitMutWith},
     },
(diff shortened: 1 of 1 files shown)


kdy1:
Can you move this PR to https://github.com/swc-project/plugins ?

victorandree:
@kdy1 Sorry for the late response—but it looks like that project already implements the fix in this PR, that is, it inserts imports after any "use " directives. At least, that's how I read https://github.com/swc-project/plugins/pull/517.

If that's the case, this PR seems irrelevant considering the plugin has been moved to another repository and the fix is implemented.

Could you clarify if there's anything in this PR that needs migrating?

victorandree:
I'm closing this PR, because it seems like the `@swc-contrib/plugin-graphql-codegen-client-preset` plugin implements this fix now, and it seems irrelevant to try and merge this. I'll publish a deprecation of my package on npm.

Issue #9526, opened 2023-07-03
"Longest common prefix () is not accessible" on Windows


srmagura:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

On Windows, running `graphql-codegen --watch` prints this warning:

```text
[Watcher] Longest common prefix () is not accessible
[Watcher] Watching current working directory (C:\Projects\oss\bugs\graphql-codegen-windows-repro\app) instead
```

This is problematic if your GraphQL schema file is outside of the directory in which `graphql-codegen` is run.

The issue does not occur on macOS or Linux.

### Your Example Website or App

https://github.com/srmagura/graphql-codegen-windows-repro

### Steps to Reproduce the Bug or Issue

See the README of the GitHub repository.

### Expected behavior

No warnings when running `graphql-codegen --watch`.

### Platform

- OS: Windows 11
- NodeJS: 18.16.0
- `graphql` version: 16.7.1
- `@graphql-codegen/*` version(s): 4.0.1

### Codegen Config File

```yaml
schema: ../schema.gql
documents:
  - "src/**/*.ts?(x)"
  - "!src/gql/**/*"
generates:
  src/gql/:
    preset: client
```

YassinEldeeb:
Hey, can you please try this build, this better handles Windows paths for the watcher, but I need to verify if it solved it or not, I don't have a Windows machine with me.

https://www.npmjs.com/package/@graphql-codegen/cli/v/4.0.2-alpha-20230706143738-f59fb0497

srmagura:
Thanks for looking into it! Unfortunately, I tried build 4.0.2-alpha-20230706143738-f59fb0497 and the behavior was unchanged.

YassinEldeeb:
do you get a different output or is it still the same?

srmagura:
Exact same output.

OriginalEXE:
For anyone facing this issue, my solution was to pass the `documents` array of strings in a config file through this map:
```
.map((path) => path.replace(/\\/g, '\\'))
```

Basically, the problem was that Windows was setting path to be something like this:
```
C:\\Users\\username\\project\\some-path\\**\\*.ts
```

but the watcher thinks the path separator is just one `\`, so it ends up splitting the path wrong when trying to find the common parent.

eddeee888:
[...]

PR #9698, opened 2023-10-04
fix(watcher): fix watcher unable to find highest common directory on Windows

OriginalEXE:
## Description

A dependency `micromatch`  has a `scan` method used for parsing a given file and returning glob details. This method does not work with Windows path separator (`\`), causing issues with the watcher on a Windows platform where the watcher would not be able to find a highest common directory, and fall back to watching the `process.cwd()` instead.

From my testing, this is the only place where Windows separator is not supported and the watcher works as expected with the change made in this PR, where we temporarily convert the separators to `/`.

Related #9526

## Type of change

## How Has This Been Tested?

We have a big monorepo at work in which I verified my changes. It would be useful for someone to verify on MacOS to make sure non-Windows platforms are not affected

**Test Environment**:

- OS: Windows
- `@graphql-codegen/cli`: 5.0.0
- NodeJS: any

## Checklist:

### DIFF
--- a/.changeset/stupid-onions-tap.md
+++ b/.changeset/stupid-onions-tap.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': patch
+---
+
+fix watcher unable to find highest common directory on Windows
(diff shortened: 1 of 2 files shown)


saihaj (approved):
thank you! 🙏🏼

PR #9709, opened 2023-10-17
fix(visitor-plugin-common): avoid reading from null values

frandiox:
## Description

tl;dr `visitor-plugin-common` tries to read from a variable that can be null in some situations. It's probably hard to reproduce since I'm using custom stuff. However, the change in this PR is just making the code more robust by moving 1 line that reads from a variable after an existing null check.

---

While working on codegen for Shopify/Hydrogen, I found that the following fragment breaks when using `typescript-operation` in `@graphql-codegen/cli` v4 and v5. It used to work in v3.3.1:

```ts
const CART_QUERY_FRAGMENT = `#graphql
  fragment CartLine on CartLine {
    id
    quantity
  }
  fragment CartApiQuery on Cart {
    id
    lines {
      nodes {
        ...CartLine
      }
    }
  }
`
```

It throws with `[Codegen] Cannot read properties of null (reading 'union')`, which comes from [this line](https://github.com/dotansimha/graphql-code-generator/blob/04d8781ff80b816a5735f86360559e1657108595/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts#L717):

```
const hasUnions = grouped[typeName].filter(s => typeof s !== 'string' && s.union).length > 0;
```

In the code above, `grouped` contains the following data:

```js
{
  grouped: {
    CartLine: [ "Pick<StorefrontAPI.CartLine, 'id' | 'quantity'>" ],
    ComponentizableCartLine: [ null ]
  }
}
```

The fragment for CartLine is correct, but the selection set for "nodes {...}" (ComponentizableCartLine here) has a null value which probably wasn't there in 3.3.1. Even though the type says it should be a string, truth is it can also be null because that's what is returned [here](https://github.com/dotansimha/graphql-code-generator/blob/04d8781ff80b816a5735f86360559e1657108595/packages/plugins/other/visitor-plugin-common/src/selection-set-processor/base.ts#L35).

It works when adding anything next to the fragment spread (the null value becomes a string):

```ts
const CART_QUERY_FRAGMENT = `#graphql
  fragment CartLine on CartLine {
    id
    quantity
  }
  fragment CartApiQuery on Cart {
    id
    lines {
      nodes {
        id # <=== anything here makes it work
        ...CartLine
      }
    }
  }
`
```

I'm not sure what changed since 3.3.1, perhaps it's the type of data that comes from other packages different from `visitor-plugin-common`, but I think it's not bad to make this package more robust and filter out null values.

The change in this PR is simply moving the offending line a few lines below after an existing check for null values.

## Type of change

## How Has This Been Tested?

Quite a manual test, but the code change is just making it more robust so it can't break anything else.

**Test Environment**:

- OS: macOS Ventura 13.5.2
- `@graphql-codegen/...`: ^5
- NodeJS: 18.12.0

## Checklist:

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -714,7 +714,6 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
     return (
       Object.keys(grouped)
         .map(typeName => {
-          const hasUnions = grouped[typeName].filter(s => typeof s !== 'string' && s.union).length > 0;
           const relevant = grouped[typeName].filter(Boolean);
 
           if (relevant.length === 0) {
@@ -728,6 +727,8 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
               return '(' + selectionObject.union.join(' | ') + ')';
             })
             .join(' & ');
+
+          const hasUnions = grouped[typeName].filter(s => typeof s !== 'string' && s.union).length > 0;
 
           return relevant.length > 1 && !hasUnions ? `( ${res} )` : res;
         })
--- a/.changeset/perfect-forks-flash.md
+++ b/.changeset/perfect-forks-flash.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+---
+
+Avoid reading from null values when selection sets only contain fragments.
--- a/dev-test/codegen.ts
+++ b/dev-test/codegen.ts
@@ -221,6 +221,17 @@ const config: CodegenConfig = {
       preset: 'client',
       presetConfig: { fragmentMasking: true },
     },
+    './dev-test/test-null-value/result.d.ts': {
+      schema: './dev-test/test-null-value/schema.graphql',
+      documents: ['./dev-test/test-null-value/query.ts'],
+      plugins: ['typescript', 'typescript-operations'],
+      config: {
+        // The combination of these two flags caused the following issue:
+        // https://github.com/dotansimha/graphql-code-generator/pull/9709
+        skipTypename: true,
+        mergeFragmentTypes: true,
+      },
+    },
   },
 };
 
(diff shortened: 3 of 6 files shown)


packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -714,7 +714,6 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
     return (
       Object.keys(grouped)
         .map(typeName => {
-          const hasUnions = grouped[typeName].filter(s => typeof s !== 'string' && s.union).length > 0;
           const relevant = grouped[typeName].filter(Boolean);
 
           if (relevant.length === 0) {
```
frandiox, line 719: `relevant.length === 0` will be `true` for `[null]` so it exists before declaring `hasUnions`.

frandiox:
Tests are passing but `pr / algolia / algolia-records-check (pull_request) ` is failing. I see it also fails in other unrelated PRs so I'm assuming it's something broken upstream 🤔

saihaj (commented):
Can you please add a failing test case?

saihaj:
> Tests are passing but pr / algolia / algolia-records-check (pull_request)  is failing


PR algolia breaks on forks because it needs write access. I wouldn’t worry about this

frandiox:
I think I've found the [combination](https://github.com/dotansimha/graphql-code-generator/pull/9709/commits/a74141c4d59df1bac249c5cde13bc9902f9966d0) that makes it fail. It works when you do at least one of these:
- In the schema file, remove the whole `ComponentizableCartLine` type.
- Turn off either `skipTypename` or `mergeFragmentTypes`.
- Add `id` or `quantity` in the selection field next to the `...CartLine` in the query file.

When this whole situation happens, the `hasUnions` check fails because the array is `[null]`. It used to be `[]` in this same situation in graphql-cli@3.3.1.

I don't have enough context to understand why this array now contains `null`, just that making this piece of code more robust makes it work again. Maybe this information will give you a hint on how to find the reason why `null` is added 🤞

frandiox:
For some reason, `yarn examples:build` is failing because `swc_plugin.wasm` is missing from `@graphql-codegen/client-preset-swc-plugin`. This seems to be happening in master branch as well so I guess it's unrelated to this PR.

saihaj (approved):
thank you!

PR #9737, opened 2023-10-26
fix(ci): pin nextjs version for nextjs-swc example

YassinEldeeb:


examples/react/nextjs-swr/codegen.ts
```diff
@@ -1,4 +1,5 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
+// @ts-ignore
```
YassinEldeeb, line 2: for some reason though, I faced some errors regarding:
```
Type error: Cannot find module '@graphql-codegen/cli' or its corresponding type declarations.

  1 | // eslint-disable-next-line import/no-extraneous-dependencies
> 2 | import { CodegenConfig } from '@graphql-codegen/cli'
```
when running yarn workspace example-react-nextjs-swr run build locally which I skipped with a `// @ts-ignore` for now, do you have any ideas on why is it throwing that?

YassinEldeeb:
not sure why End2End step fails @saihaj

examples/programmatic-typescript/src/gql.generated.ts
```diff
@@ -9,16 +9,16 @@ export type MakeEmpty<T extends { [key: string]: unknown }, K extends keyof T> =
...
-  ID: string;
-  String: string;
-  Boolean: boolean;
-  Int: number;
-  Float: number;
+  ID: { input: string; output: string };
+  String: { input: string; output: string };
+  Boolean: { input: boolean; output: boolean };
+  Int: { input: number; output: number };
+  Float: { input: number; output: number };
```
eddeee888, line 16: These were never updated because codegen/tests are executed sequentially... and this example was after some other failing tests 😓

examples/react/nextjs-swr/hooks/use-query.ts
```diff
@@ -20,7 +20,7 @@ export function useGraphQL<TResult, TVariables>(
       document.definitions.find(isOperationDefinition)?.name,
       variables,
     ] as const,
-    async (_key: string, variables: any) =>
+    async ([_key, variables]: any) =>
```
eddeee888, line 23: Looks like the current version of swr keeps the array as-is: https://swr.vercel.app/docs/arguments#multiple-arguments

examples/typescript-graphql-request/src/main.ts
```diff
@@ -46,6 +46,3 @@ export const getPeople = async (first?: number) => {
   }
   return res?.allPeople?.edges;
 };
-
-getPeople().then(res => console.log(res));
-getPeople(10).then(res => console.log(res));
```
eddeee888, line 51: These calls cause the Jest tests to fail. These also look like debugging statements that were never removed 🤔

scripts/print-example-ci-command.js
```diff
@@ -4,7 +4,7 @@ const fg = require('fast-glob');
 
 const packageJSON = fg.sync(['examples/**/package.json'], { ignore: ['**/node_modules/**'] });
 
-const ignoredPackages = ['example-react-nextjs-swr'];
+const ignoredPackages = [];
```
eddeee888, line 7: We can now remove the example from the ignored array 🎉 

I'll keep the current implementation in case we need to ignore tests in the future (hope not! 🤞 )

examples/programmatic-typescript/src/gql.generated.ts
YassinEldeeb: so what should we do about them? 👀

YassinEldeeb (commented):
looks good, @saihaj or @enisdenjo can you take a last look before merging?

saihaj (approved):
Thank you 🙏🏼

PR #9778, opened 2023-12-06
Update plugin-structure.mdx

PeteDuncanson:
Seems there is now a nested "plugins:" key missing from that listed in the docs so added it in, was throwing errors without it

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Related #9779 

This change requires a documentation update


saihaj (approved):
thank you!

Issue #9925, opened 2024-04-15
BUG codegen incorrectly changes order for persisted operations


wouterpen:
### Which packages are impacted by your issue?

@graphql-codegen/client-preset

### Describe the bug

https://the-guild.dev/graphql/yoga-server/docs/features/persisted-operations

gives the advise to use https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#persisted-documents

to create our persisted operations. the operation then generates a persisted.document.json like so: 

{
  "b2c3d4e5f6g7h8i9j0a1": "query Hello { hello }",
  "kae4fe7f6g7h8i9ej0ag": "mutation echo($msg: String!) { echo(message: $msg) }"
}

however persistedDocument.js is using printExecutableGraphQLDocument: https://www.npmjs.com/package/@graphql-tools/documents

this sorts all operations alphabetically. But this changes the way mutations work and operations should therefore not be sorted alphabetically see docs: https://spec.graphql.org/October2021/#sec-Mutation
 
### Your Example Website or App

https

### Steps to Reproduce the Bug or Issue

create an operation that fires multiple mutations in an order that is not alphabetical
generate trusted documents
see that the order of the mutation in persisted.document.json is different than the original mutations order

### Expected behavior

As a user, I expect the order of my mutations to be the same as in my .graphql file so that I can ensure the order in which my mutations are beeing fired. But I am seeing that on generation the order changes.

### Screenshots or Videos

```javascript

mutation x { 
b(test) {
 _id
 }
 a(test) {
 _id
 }
}
```

results in :
```javascript
  "kae4fe7f6g7h8i9ej0ag": "mutation x{ a(test){_id} b(test){_id} "
```

expects: 
```javascript
  "kae4fe7f6g7h8i9ej0ag": "mutation x{ b(test){_id} a(test){_id} "

```

### Platform

- NodeJS: v20.12.1
- "@graphql-codegen/client-preset": "4.2.4",

### Codegen Config File

 './libs/x/src/lib/': {
              preset: 'client',
              documents: ['./libs/x/*.graphql'],
              presetConfig: {
                persistedDocuments: {
                  hashAlgorithm: 'sha256'
                },
              },
            },

PR #9926, opened 2024-04-15
add useGraphqlPrint & removeClientSpecificFields options to persisted…

martyganz:
… documents in client-preset

## Description

Adds two options to trusted documents codegen:

```
 persistedDocuments: {
          useGraphqlPrint: true
          removeClientSpecificFields: true
        }
```

* useGraphqlPrint: boolean (default: false) - Wether to use print from graphql to convert the `DocumentNode` to string or to use the default: @graphql-tools/documents printExecutableGraphQLDocument
* removeClientSpecificFields: boolean (default: true) - Remove client specific fields from the `DocumentNode`

I've added both to be backwards compatible with what is currently in place, thus not having a breaking change but open to suggestions. 

It's related to below bug where using the @graphql-tools/documents print actually breaks our schema because it changes the order of fields within a mutation (where for us the order is actually important). This will allow us to fix it but if it is an actual bug which could affect more people it is probably better to change the printExecutable to graphql's print.

Related #9925 

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## How Has This Been Tested?

`yarn test packages/presets/client/tests/client-preset.spec.ts --watch`

- 1. Build locally and symlinked into our project
- 2. Generated trusted documents and verified output manually

Added unit tests

**Test Environment**:

- OS: MacOS
- `@graphql-codegen/...`: latest
- NodeJS: v18 (lts/hydrogen) & v20 (lts/iron)

## Checklist:

## Further comments

First PR here so open to any and all comments, feedback etc.

### DIFF
--- a/packages/presets/client/src/index.ts
+++ b/packages/presets/client/src/index.ts
@@ -70,29 +70,46 @@ export type ClientPresetConfig = {
   onExecutableDocumentNode?: (documentNode: DocumentNode) => void | Record<string, unknown>;
   /** Persisted operations configuration. */
   persistedDocuments?:
-    | boolean
-    | {
-        /**
-         * @description Behavior for the output file.
-         * @default 'embedHashInDocument'
-         * "embedHashInDocument" will add a property within the `DocumentNode` with the hash of the operation.
-         * "replaceDocumentWithHash" will fully drop the document definition.
-         */
-        mode?: 'embedHashInDocument' | 'replaceDocumentWithHash';
-        /**
-         * @description Name of the property that will be added to the `DocumentNode` with the hash of the operation.
-         */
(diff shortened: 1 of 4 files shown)


n1ru4l:
@martyganz It seems like the fix for the ordering should be performed within `@graphql-tools/documents`? Would you be open to send a pull request for fixing that there？

n1ru4l:
@martyganz Also, please avoid sending different features in a single pull request. The `useGraphqlPrint` and `removeClientSpecificFields` are not related.

martyganz:
@n1ru4l Thanks for the feedback, I opened a issue and PR on graphql-tools. See above linked issue and PR.

PR #9954, opened 2024-04-26
[Docs] Update server preset guide to reflect changes in v0.8

eddeee888:
This PR updates server preset guide with changes in server preset v0.8

### DIFF
--- a/website/src/pages/docs/guides/graphql-server-apollo-yoga-with-server-preset.mdx
+++ b/website/src/pages/docs/guides/graphql-server-apollo-yoga-with-server-preset.mdx
@@ -257,7 +257,7 @@ The server preset handles all resolver types and imports. So, you only need to i
 
 This means instead of updating the `codegen.ts` config file, you make changes in each module. This keeps the GraphQL API maintainable at any scale.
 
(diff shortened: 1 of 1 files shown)


website/src/pages/docs/guides/graphql-server-apollo-yoga-with-server-preset.mdx
```diff
@@ -374,11 +408,11 @@ export type ResolversParentTypes = {
 ```ts filename="src/schema/user/resolvers/User.ts" {3-5,7-10}
 import type { UserResolvers } from './../../types.generated'
-  fullName: () => {
+  fullName: async (_parent, _arg, _ctx) => {
     /* User.fullName resolver is required because User.fullName exists but UserMapper.fullName does not */
   },
 
-  isAdmin: ({ isAdmin }) => {
+  isAdmin: ({ isAdmin }, _arg, _ctx) => {
```
saihaj, line 415: do we need to this change? 🤔

saihaj (approved):

website/src/pages/docs/guides/graphql-server-apollo-yoga-with-server-preset.mdx
eddeee888: The server preset now generates this by default, so I think matching it here helps set the user expectation

PR #9961, opened 2024-05-06
[typescript-resolvers] Add meta field to typescript-resolvers plugin's output

eddeee888:
## Description

Currently, we don't know easily _what_ has been generated from the run without parsing through the output's content. 

This PR adds a new optional `meta` to allow plugins to report certain things about the run.

This PR also makes `@graphql-codegen/typescript-resolvers` report generated resolver type names - after applying `namingConvention` - to `meta`. This allows plugins/presets (e.g. Server Preset) that relies on `@graphql-codegen/typescript-resolvers` plugin to import the generated resolver types without trying to match the `namingConvention` logic manually.

## Type of change

- [x] New interal feature

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/.changeset/new-radios-flash.md
+++ b/.changeset/new-radios-flash.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+'@graphql-codegen/typescript-resolvers': patch
+---
+
+Update typescript-resolvers to report generated resolver types in the run to meta field in the output
--- a/.changeset/new-radios-flash.md
+++ b/.changeset/new-radios-flash.md
@@ -1,6 +1,6 @@
 ---
-'@graphql-codegen/visitor-plugin-common': patch
-'@graphql-codegen/typescript-resolvers': patch
+'@graphql-codegen/visitor-plugin-common': minor
+'@graphql-codegen/typescript-resolvers': minor
 ---
 
 Update typescript-resolvers to report generated resolver types in the run to meta field in the output
--- a/.changeset/curvy-lobsters-kneel.md
+++ b/.changeset/curvy-lobsters-kneel.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/plugin-helpers': patch
+---
+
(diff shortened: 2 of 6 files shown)


.changeset/new-radios-flash.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
```
dotansimha, line 2: I assume that `patch` because this is kinda considered an internal API so it will not effect other users?

eddeee888: Yes, internal API is also my thinking here. I'm happy to go either `minor` or `path` in this case though 🙂

saihaj (commented):
should we only output these in generated content if we have `DEBUG` enabled? or maybe just output based on some config file? 

To me doesn't make much difference cause I don't commit generated artifacts from codegen to a repo, but those who do it will create a big diff once they upgrade?

eddeee888:
Hi @saihaj , the intention is that `meta` will not be printed to any generated files, therefore users won't see any difference in their artifacts. 

The main use for these is to report metadata of the run for advanced use cases. For example, server preset needs to know the name of generated types so it can generate its import statements. Here's an example generated file from server preset:

```ts
import type { UserResolvers } from './../../types.generated';
  /* Implement User resolver logic here */
};
```

If user applies `namingConvention` to change generated type to snake case though, the above would fail as we need to import `user_resolvers` instead of `UserResolvers`. One option is for server preset to implement the same `namingConvention` logic, which is cumbersome. With the `meta` field, server preset can just look to see the generated type names to generate the imports line:

```ts
import type { user_resolvers } from './../../types.generated'; // Server Preset can look into the `meta` field to know `user_resolvers` is the name of the generated type
export const User: user_resolvers = {
  /* Implement User resolver logic here */
};
```

n1ru4l (approved):
Approved, but I would use minor instead of patch as @dotansimha suggested 😇

PR #9989, opened 2024-06-06
[federation][resolvers] Add `generateInternalResolversIfNeeded. __resolveReference` to generate `__resolveReference` only when resolvable

eddeee888:
## Description

From Apollo Federation doc: https://www.apollographql.com/docs/federation/federated-types/federated-directives/#resolvable: when there's no resolvable `@keys` on an object type in a subgraph, we cannot move to the object type in said subgraph.

This PR: 
- adds `generateInternalResolversIfNeeded.__resolveType`
- By default, `generateInternalResolversIfNeeded.__resolveType = false` generates optional `__resolveType` without checking whether there's at least one resolvable `@key`
- When `generateInternalResolversIfNeeded.__resolveType = true`:
  - avoids generating `__resolveReference` if there's no resolvable `@key`
  - generates non-required `__resolveReference` for resolvable `@key` and skip non-resolvable ones
  - report whether a type has reference resolver in `meta` so something like Server Preset can handle it correctly

## Type of change

## How Has This Been Tested?

- [x] Unit tests for federation and resolvers plugin


packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -275,7 +291,10 @@ export class ApolloFederation {
  * Checks if Object Type is involved in Federation. Based on `@key` directive
  * @param node Type
  */
-function isFederationObjectType(node: ObjectTypeDefinitionNode | GraphQLObjectType, schema: GraphQLSchema): boolean {
+export function checkObjectTypeFederationDetails(
+  node: ObjectTypeDefinitionNode | GraphQLObjectType,
+  schema: GraphQLSchema
+): { resolvableKeyDirectives: readonly DirectiveNode[] } | false {
```
eddeee888, line 297: I wanted to make the return type `{ resolvableKeyDirectives: readonly DirectiveNode[] } | undefined`, and make `undefined` represent the case when the node is not a object type with federation.

However, the returned value is never `undefined` (image attached) 🤔 So I've made it `false` so at least typecheck works

Screenshot 2024-08-25 at 11 41 12 PM

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -275,7 +291,10 @@ export class ApolloFederation {
  * Checks if Object Type is involved in Federation. Based on `@key` directive
  * @param node Type
  */
-function isFederationObjectType(node: ObjectTypeDefinitionNode | GraphQLObjectType, schema: GraphQLSchema): boolean {
+export function checkObjectTypeFederationDetails(
```
eddeee888, line 294: This function checks whether a node has is federation, and returns federation metadata for the node (for now, it returns an array of resolvable `@keys`)

I'm not very happy with the name `checkObjectTypeFederationDetails` as it reads a bit clumsy 🤔 Open for suggestion.

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -656,7 +681,6 @@ export class BaseResolversVisitor<
   protected _globalDeclarations = new Set<string>();
   protected _federation: ApolloFederation;
   protected _hasScalars = false;
-  protected _hasFederation = false;
```
eddeee888, line 659: Previously, whenever there's a type with `__resolveReference`, this will be set to `true`.

Instead of a `boolean`, we now want to track all the types with generated `__resolveReference`  types as output `meta`, so downstream consumers (e.g. Server Preset) can easily use it. 
The `ApolloFederation` class now tracks this metadata. 

I could just put it here as a property of the plugin e.g. `_federationMeta` but it seems like `ApolloFederation` is supposed to handle federation concerns. Happy to discuss if needed.

kamilkisiela (approved):

Issue #9994, opened 2024-06-09
[preset/client] Custom hash function for persisted operations


nahn20:
### Is your feature request related to a problem? Please describe.

I'd like to add more flexibility for persisted operations to allow developers to insert a custom hashing function instead of being limited to `sha1` and `sha256`. This helps to avoid any issues requesting additional algorithms in the future, as well as supports additional use-cases. For example, I have a use-case in which we sign operations with a specific secret, and I'd like to re-use that for the persisted operation hash. 

### Describe the solution you'd like

An additional configuration option allowing developers to pass in a `hashFunction`. I've put together a proof of concept [here](https://github.com/nahn20/graphql-code-generator/commit/50e4c413a7107a4a83ecb63cf161b397894422bd). (Also, I'm happy to create a pull request for this). 

### Describe alternatives you've considered

Currently, I'm using [this plugin](https://www.npmjs.com/package/graphql-codegen-on-operations) which supports the use-case of generating an additional file with the operations and the hashes. For convenience, simplicity, and bundle size, I'd like to attach the hash to the operation document itself.

n1ru4l:
Instead of adding a nee config option we can make allow the existing hash algorithm property to be a function. Please feel free to send a PR with tests and documentation updates!

PR #9996, opened 2024-06-09
Persisted document custom hash function

nahn20:
## Description

This features allows for customization of the `hashAlgorithm` configuration for persisted documents in the client preset. This mitigates the need to support additional future algorithms beyond `sha1` and `sha256` and gives developers full flexibility for generating document hashes, unlocking new use-cases. 

Related #9994

## Type of change

## How Has This Been Tested?

Added three new test cases to `client-preset.spec.ts`. Ran `yarn test client-preset`.

- `custom hash remove whitespace` - This test case demonstrates basic functionality in a human-readable manner. Since the output is just the operation without whitespace, it's easy to tell if the custom hash function is broken.
- `custom hash sha256` - This test case replicates the existing test case `hashAlgorithm="sha256"` to ensure identical results.
- `custom hash docs sha512` - This test case demonstrates a realistic, real-world use-case and was chosen as the example for the docs. 

**Test Environment**:

- OS: MacOS
- `@graphql-codegen/...`:
- NodeJS: v18.18.2

## Checklist:

### DIFF
--- a/.changeset/spicy-starfishes-press.md
+++ b/.changeset/spicy-starfishes-press.md
@@ -1,0 +1,6 @@
+---
+"@graphql-codegen/client-preset": patch
+"website": patch
+---
+
+Added configuration to allow for custom hash functions for persisted documents in the client preset
--- a/.changeset/spicy-starfishes-press.md
+++ b/.changeset/spicy-starfishes-press.md
@@ -4,3 +4,27 @@
 ---
 
 Added configuration to allow for custom hash functions for persisted documents in the client preset
+
+### Example
+```ts filename="codegen.ts" {10-12}
+import { type CodegenConfig } from '@graphql-codegen/cli'
+
+const config: CodegenConfig = {
+  schema: 'schema.graphql',
+  documents: ['src/**/*.tsx'],
+  generates: {
+    './src/gql/': {
+      preset: 'client',
+      presetConfig: {
+        persistedDocuments: {
+          hashAlgorithm: operation => {
+            const shasum = crypto.createHash('sha512')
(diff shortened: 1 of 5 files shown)


n1ru4l:
@nahn20 Can you please also add a changeset for these changes?

n1ru4l:
What would happen if the hash function returns the same hash for multiple operations? Should we assert this?

nahn20:
I think that considering how advanced of a feature that this is, any developer using a custom function should know how to design an effective hash function. I'd be comfortable leaving in this footgun for the sake of full customizability, to account for potential use-cases that we aren't thinking of. 

I also think that the cost of a bug here is minimal due to this potential issue being easy to debug. If a developer implements a poor hashing function, then they'll find their GraphQL calls aren't working. They would disable persisted documents, then see they are working again. From there, I have full confidence a developer can debug it down to multiple equal hash outputs by analyzing the mismatched operations. 

Additionally, I think this would add a false sense of security. Not all approved operations are guaranteed to be present at generation time (for example, a backend which remotely stores operations from previous client versions). 

This functionality is also easy for a developer to implement if they want the extra safety. I don't think this is necessary to build into the preset itself.

.changeset/spicy-starfishes-press.md
```diff
@@ -0,0 +1,6 @@
+---
+"@graphql-codegen/client-preset": patch
+"website": patch
+---
+
+Added configuration to allow for custom hash functions for persisted documents in the client preset
```
n1ru4l, line 6: Can you please provide a configuration example? Right now the changeset contains little information.

Issue #10004, opened 2024-06-14
#9962 causes issues with our schema: parts of the types in ResolversType map contradict other parts of the type (repro)


mx-bernhard:
### Which packages are impacted by your issue?

@graphql-codegen/visitor-plugin-common, @graphql-codegen/typescript-resolvers

### Describe the bug

With the update of @graphql-codegen/visitor-plugin-common to 5.2.0 and from that very likely this [PR](com/dotansimha/graphql-code-generator/pull/9962) we run into an issue with the generated types. To me, it looks like the resolver types are improved, by replacing the base interface with its concrete implementations. This does not seem to be generated correctly in all cases. 

Take the following _graphql.schema_ (also in the [Github repro repo](https://github.com/mx-bernhard/graphql-codegen-resolver-types)):

```gql
type Assignment {
  resource: HumanResource
}

interface Resource {
  id: ID!
  attributes: [Attribute!]!
}

type VehicleResource implements Resource {
  id: ID!
  attributes: [Attribute!]!
  plateNumber: String!
}

type HumanResource implements Resource {
  id: ID!
  attributes: [Attribute!]!
  fullName: String!
}

interface Attribute {
  id: ID!
}

type JSONAttribute implements Attribute {
  id: ID!
  json: String!
}

type TextAttribute implements Attribute {
  id: ID!
  text: String!
}

type Query {
  assignment: Assignment!
  humanResource: HumanResource!
  resource: Resource!
  attribute: Attribute!
}

```

It contains two levels of interface to type mappings. When you look at the type generated for the Assignment resolver, the resource points to the schema base type instead of the resolver type:

```ts
export type ResolversTypes = {

  // has resource: HumanResource instead of resource: ResolversTypes<'HumanResource'>
  Assignment: ResolverTypeWrapper<Assignment>; // !!!

  Attribute: ResolverTypeWrapper<ResolversInterfaceTypes<ResolversTypes>['Attribute']>;
  Boolean: ResolverTypeWrapper<Scalars['Boolean']['output']>;
  HumanResource: ResolverTypeWrapper<Omit<HumanResource, 'attributes'> & { attributes: Array<ResolversTypes['Attribute']> }>;
  ID: ResolverTypeWrapper<Scalars['ID']['output']>;
  JSONAttribute: ResolverTypeWrapper<JsonAttribute>;
  Query: ResolverTypeWrapper<{}>;
[...]

PR #10014, opened 2024-06-19
Fix abstract types not being handled when nested in object types

eddeee888:
## Description

This PR fixes a scenario where abstract types are not pointing to generated resolvers type correctly:
- an object type A...
- with a field that returns an object type B (could be self-referencing to A) that has an abstract type field.

In said case, the object type B must use generated resolver types, otherwise, it may have type conflicts

Related https://github.com/dotansimha/graphql-code-generator/issues/10004

## Type of change

## How Has This Been Tested?

- [x] Unit test
- [x] Integration test


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1835,3 +1854,84 @@ function normalizeResolversNonOptionalTypename(
     ...input,
   };
 }
+
+function checkIfObjectTypeHasAbstractTypesRecursively(
+  baseType: GraphQLObjectType,
+  result: {
+    isObjectWithAbstractType: boolean;
+    checkedTypesWithNestedAbstractTypes: Record<string, { hasNestedAbstractTypes: 'yes' | 'no' | 'checking' }>;
```
dimaMachina, line 1862: can it have a better name ? for me, anything which starts with `has*` should be a boolean

dotansimha (approved):

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
eddeee888: True! It started as `boolean` but then I saw the need to change to 3 different options, but didn't review the name.

Do you have suggestions? I'm thinking a few options but I'm unsure about them:
- `value`
- `result`
- `checkStatus`

packages/plugins/typescript/resolvers/tests/ts-resolvers.spec.ts
eddeee888, line 1: `resolversNonOptionalTypename` tests are moved to `ts-resolvers.config.ResolversNonOptionalTypename` to further split tests

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
dimaMachina: checkStatus For me is good

dylanwulf:
@eddeee888 Hi, are there any config options to revert this behavior? On `@graphql-codegen/typescript-resolvers` v4.2.1 I am getting an error during tsc type check which did not happen in v4.2.0:
`RangeError: Maximum call stack size exceeded`

eddeee888:
Hi @dylanwulf, could you please open an issue here with an example codegen config + snippet of your schema? https://github.com/dotansimha/graphql-code-generator/issues/new?assignees=&labels=&projects=&template=bug_report.yml?

dylanwulf:
> Hi @dylanwulf, I think the first step is to understand the issue. Could you please open an issue here with an example codegen config + snippet of your schema? https://github.com/dotansimha/graphql-code-generator/issues/new?assignees=&labels=&projects=&template=bug_report.yml

No problem, issue opened here: #10084

PR #10073, opened 2024-07-28
feat: reduce code output of client preset

n1ru4l:
We don't need all these types for the GraphQL schema for client preset. I added an option to omit them within `@graphql-codegen/visitor-plugin-common` and `@graphql-codegen/typescript` and enabled this option by default for the client preset.

For GraphQL Hive this reduces the [generated code for the client preset by over 50%](https://github.com/kamilkisiela/graphql-hive/pull/5286). Also, see the diff and removed code here from the fixtures/examples. :D

We could even reduce generated enums and scalars to only those used within the provided graphql documents. However, let's keep that for another time.


packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts
```diff
@@ -714,6 +716,7 @@ export class BaseTypesVisitor<
   }
 
   UnionTypeDefinition(node: UnionTypeDefinitionNode, key: string | number | undefined, parent: any): string {
+    if (this.config.onlyTypescriptOperationTypes) return '';
     if (this.config.onlyOperationTypes || this.config.onlyEnums) return '';
```
eddeee888, line 720: I can see `onlyOperationTypes` and `onlyEnums` seems to do the same thing. Could you please help me understand the differences in use cases between the three options? 🙂

n1ru4l: That is a very valid point... 😅 I think in the future `onlyTypescriptOperationTypes` could actually have the side effect of only emitting the scalars/enums used within the GraphQL operations provided. But since I want to keep that out of scope of this PR, just setting `onlyOperationTypes` and `onlyEnums` to `true` has the same effect.

n1ru4l: Adjusted in [`bdd83b3` (#10073)](https://github.com/dotansimha/graphql-code-generator/pull/10073/commits/bdd83b330efab9495a17e7cf0a8226c750fc1772)

eddeee888 (approved):
Nice one!

saihaj:
Released [@graphql-codegen/client-preset@4.4.0](https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1728313341014)

jacquesg:
This is a massive breaking change, upgrading from `4.3.X` to `4.4` the default should not have changed.

jacquesg:
Even undoing the patch locally results in the generated file being broken:

```
diff --git a/codegen.ts b/codegen.ts
index 28d74543..5fcad7cf 100644
--- a/codegen.ts
+++ b/codegen.ts
@@ -6,7 +6,11 @@ const config: CodegenConfig = {
   generates: {
     './src/api/deianira/generated/': {
       preset: 'client',
-      plugins: [],
+      plugins: [
+        {
+          [`typescript`]: {},
+        },
+      ],
```

```
src/api/deianira/generated/graphql.ts:5950:13 - error TS2300: Duplicate identifier 'UpdateZoneInput'.

5950 export type UpdateZoneInput = {
                 ~~~~~~~~~~~~~~~


Found 1148 errors in the same file, starting at: src/api/deianira/generated/graphql.ts:13
```

asgeo1:
Breaking my app also because I was actually using those types! @n1ru4l do you know how we can get the original behaviour back?

I can see `onlyEnumTypes: true`, `onlyOperationTypes:true` were added in the PR, but I don't get how we can revert them back to their original values in a `graphql-codgen.ts` config file.

piotr-czerwik:
Same for the app I am working on - we relied on these types. 

I've tried adding 
```js
onlyEnumTypes: false,
onlyOperationTypes: false,
```
to the configuration hoping it will 'revert' the current defaults to the previous ones, but it seems not having any effect...

bmulholland:
Filed as https://github.com/dotansimha/graphql-code-generator/issues/10167

nebbles:
PR https://github.com/dotansimha/graphql-code-generator/pull/10155 adds `onlyEnumTypes` and `onlyOperationTypes` to the overridable config.

PR #10077, opened 2024-08-01
Add root level resolvers support to `avoidOptionals` 

eddeee888:
## Description

Most (if not all of the time), we need to implement root-level resolvers, otherwise there'd be runtime errors.
This PR extends `avoidOptionals` to generate non-optional root-level resolvers. 

This will also be used in Server Preset to avoid having to use `NonNullable` as discussed [here](https://github.com/kamilkisiela/graphql-hive/pull/5182#discussion_r1672467181)

## Type of change

## How Has This Been Tested?

- [x] Unit test


packages/plugins/other/visitor-plugin-common/src/avoid-optionals.ts
```diff
@@ -1,21 +1,29 @@
-import { AvoidOptionalsConfig } from './types.js';
+import { AvoidOptionalsConfig, NormalizedAvoidOptionalsConfig } from './types.js';
 
-export const DEFAULT_AVOID_OPTIONALS: AvoidOptionalsConfig = {
+export const DEFAULT_AVOID_OPTIONALS: NormalizedAvoidOptionalsConfig = {
```
eddeee888, line 3: `NormalizedAvoidOptionalsConfig` is similar to `AvoidOptionalsConfig` but all the fields are non-optional, making it much easier to code with

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1307,7 +1314,7 @@ export class BaseResolversVisitor<
   }
 
   protected formatRootResolver(schemaTypeName: string, resolverType: string, declarationKind: DeclarationKind): string {
-    return `${schemaTypeName}${this.config.avoidOptionals ? '' : '?'}: ${resolverType}${this.getPunctuation(
+    return `${schemaTypeName}${this.config.avoidOptionals.resolvers ? '' : '?'}: ${resolverType}${this.getPunctuation(
```
eddeee888, line 1317: The old implementation can cause a scenario where an empty object is passed in, but it'd be treated as truthy so it'd treat it as non-optional.

In this new implementation, we use `resolvers` because it's the option that is used for all resolvers-related fields in the past.

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -682,7 +689,7 @@ export class BaseResolversVisitor<
       allResolversTypeName: getConfigValue(rawConfig.allResolversTypeName, 'Resolvers'),
       rootValueType: parseMapper(rawConfig.rootValueType || '{}', 'RootValueType'),
       namespacedImportName: getConfigValue(rawConfig.namespacedImportName, ''),
-      avoidOptionals: getConfigValue(rawConfig.avoidOptionals, false),
+      avoidOptionals: normalizeAvoidOptionals(rawConfig.avoidOptionals),
```
eddeee888, line 692: This plugin seems to be the only one that's not using `normalizeAvoidOptionals`. This makes it consistent.

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -78,6 +80,8 @@ export interface ParsedResolversConfig extends ParsedConfig {
   resolversNonOptionalTypename: ResolversNonOptionalTypenameConfig;
 }
 
+type FieldDefinitionPrintFn = (parentName: string, avoidResolverOptionals: boolean) => string | null;
```
eddeee888, line 83: I was REALLY confused why/how the field's name is a function. Found out it's because we do custom logic in `FieldDefinition` function. I'm extracting the type here to be able to type it instead of using `any`

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1463,10 +1467,6 @@ export class BaseResolversVisitor<
 
       const resolverType = isSubscriptionType ? 'SubscriptionResolver' : directiveMappings[0] ?? 'Resolver';
 
-      const avoidResolverOptionals =
-        typeof this.config.avoidOptionals === 'object'
-          ? this.config.avoidOptionals?.resolvers
-          : this.config.avoidOptionals === true;
```
eddeee888, line 1469: We now pass this from caller. I found that this function is called in Interface fields and Object type fields

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1532,15 +1532,31 @@ export class BaseResolversVisitor<
...
+      if (this.schema.getMutationType()?.name === typeName) {
+        return 'mutation';
+      }
+      if (this.schema.getSubscriptionType()?.name === typeName) {
+        return 'subscription';
+      }
+      return false;
+    })();
+
+    const fieldsContent = (node.fields as unknown as FieldDefinitionPrintFn[]).map(f => {
```
eddeee888, line 1549: I have to inline `node.fields as unknown as FieldDefinitionPrintFn[]` here because it's the easiest way to communicate where this function comes from. If there's a better way, let me know!

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1720,21 +1736,23 @@ export class BaseResolversVisitor<
     const allTypesMap = this._schema.getTypeMap();
     const implementingTypes: string[] = [];
 
-    this._collectedResolvers[node.name as any] = {
+    const typeName = node.name as any as string;
```
eddeee888, line 1739: Cleaning this up by limiting where we do `as any as string`.

packages/plugins/typescript/resolvers/tests/ts-resolvers.config.avoidOptionals.spec.ts
eddeee888, line 1: There's quite a few tests related to `avoidOptionals` so I'm creating this new file to keep the old ones and add new tests related to query, mutation and subscription.

saihaj (approved):

n1ru4l (approved):
👍 nice

saihaj:
Released [@graphql-codegen/visitor-plugin-common@5.4.0](https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1728313341014), [@graphql-codegen/typescript-operations@4.3.0](https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1728313341014), [@graphql-codegen/typescript-resolvers@4.3.0](https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1728313341014), [@graphql-codegen/typescript@4.1.0](https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1728313341014)

Issue #9490, opened 2023-06-06
codegen/cli `generate` with `watch` bug


TomKlaverAH:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

When using `watch` in the `generate` call from `@graphql-codegen/cli` you will get an error:

```
node:internal/errors:478
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "to" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:387:5)
    at validateString (node:internal/validators:162:11)
    at relative (node:path:1192:5)
    at makeGlobalPatternSet (/home/tomk/git/ah/ah-checkout/web/node_modules/@graphql-codegen/cli/cjs/utils/patterns.js:103:33)
    at createWatcher (/home/tomk/git/ah/ah-checkout/web/node_modules/@graphql-codegen/cli/cjs/utils/watcher.js:28:69)
    at generate (/home/tomk/git/ah/ah-checkout/web/node_modules/@graphql-codegen/cli/cjs/generate-and-save.js:100:47) {
  code: 'ERR_INVALID_ARG_TYPE'
}
error Command failed with exit code 1.
```

This is triggered from here:
https://github.com/dotansimha/graphql-code-generator/blob/2139f1db5704aa34a0cd89342736bf67ec556da3/packages/graphql-codegen-cli/src/utils/patterns.ts#L122

`filePath` is not always set (in fact it is not possible to set it at all using the plain object input of `generate`), see here:
https://github.com/dotansimha/graphql-code-generator/blob/2139f1db5704aa34a0cd89342736bf67ec556da3/packages/graphql-codegen-cli/src/config.ts#L365

Typescript in strict mode would have complained here, since the result of the RHS is then in fact `string | undefined`.

The constructor is called only with the `config` object, so `filepath` will always be `undefined`, triggering the error.
https://github.com/dotansimha/graphql-code-generator/blob/2139f1db5704aa34a0cd89342736bf67ec556da3/packages/graphql-codegen-cli/src/config.ts#L448

Why not use `strict: true` for this project? That would have caught the bug here.

Workaround for the current version:
Call `generate` with an instance of `CodegenContext`, instead of using the plain object.

### Your Example Website or App

private repo

### Steps to Reproduce the Bug or Issue

See description
import { generate } from '@graphql-codegen/cli';

generate({
   // ...config
  watch: true
});

[...]

PR #10093, opened 2024-08-15
fix: nullable filepath config

MatthewMaclean:
## Description

Filepath is an optionally provided param that is unnecessarily mandatory when the CLI is set to watch mode. This change reflects that in the typing and it's usage when setting watch patterns. This has been a problem since 4.x.

Related # (issue)
See https://github.com/dotansimha/graphql-code-generator/issues/9490

## Type of change

## How Has This Been Tested?

Not sure the easiest way to go about testing this locally. If I modify the node_module code to drop this reference to filepath, it works as expected. 

## Checklist:

## Further comments

I'm open to alternatives here. I also wonder if the `process.cwd()` that's used when filepath is provided should instead use the context's cwd. It's confusing to provide a param to override to cwd but then have it ignored. That change would definitely be breaking though, so I'm not inclined to lump it into this PR.

### DIFF
--- a/packages/graphql-codegen-cli/src/utils/patterns.ts
+++ b/packages/graphql-codegen-cli/src/utils/patterns.ts
@@ -119,7 +119,7 @@ export const makeGlobalPatternSet = (initialContext: CodegenContext) => {
   return {
     watch: sortPatterns([
       ...(typeof config.watch === 'boolean' ? [] : normalizeInstanceOrArray<string>(config.watch ?? [])),
-      relative(process.cwd(), initialContext.filepath),
+      ...(initialContext.filepath ? [relative(process.cwd(), initialContext.filepath)] : []),
     ]),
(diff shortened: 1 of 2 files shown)


packages/graphql-codegen-cli/src/utils/patterns.ts
```diff
@@ -119,7 +119,7 @@ export const makeGlobalPatternSet = (initialContext: CodegenContext) => {
   return {
     watch: sortPatterns([
       ...(typeof config.watch === 'boolean' ? [] : normalizeInstanceOrArray<string>(config.watch ?? [])),
-      relative(process.cwd(), initialContext.filepath),
+      ...(initialContext.filepath ? [relative(process.cwd(), initialContext.filepath)] : []),
```
MatthewMaclean, line 122: Admittedly, this is not the cleanest way to conditionally add the path to the array. I'm following the above precedence, but I'm happy to refactor it so that the conditions checked when building the array are easier to read.

PR #10113, opened 2024-08-21
Fix `build examples` CI step

eddeee888:
## Description

[This renovate PR](https://github.com/dotansimha/graphql-code-generator/pull/9431/commits/0ea01bf375981e93792e10e86bff55ed0170b790#diff-a5768ffdde6eee94e5956dd67b49469c54baf1f2858e83c675c2eb324bed58d2R12) seems to introduce unwanted changes to `@graphql-codegen/client-preset-swc-plugin`'s `package.json` as it starts to point to non-existent files.

This PR reverts changes to non-existent files and to be more inline with the rest of the other package.json changes

Also, convert Vite example to use ESM to avoid warning about CJS support deprecation


packages/presets/swc-plugin/package.json
```diff
@@ -27,32 +27,6 @@
     "swc_plugin.wasm"
   ],
   "type": "module",
```
eddeee888, line 29: `type: module` wasn't in the original package.json.
I know this'd mark this package as ESM, but I'm not too familiar with how it's supposed to work with WASM? 😵 

It seems to work with or without this value, so I'm leaving it here, please let me know if we should remove this instead.

PR #10130, opened 2024-09-05
Improve duplicate question for on feature request template

bmulholland:
The final question is a duplicate of the first. I've updated it to something a little bit better, but am still not happy with it and I'm not 100% on the intent. Just wanted to take a step toward an improvement.


n1ru4l (approved):

saihaj (approved):
Thanks!

PR #10138, opened 2024-09-11
[resolvers] Add `generateInternalResolversIfNeeded.__isTypeOf`

eddeee888:
## Description

This PR compliments https://github.com/dotansimha/graphql-code-generator/pull/9989 and should be merged after.
Previously, we always generate `__isTypeOf` type which means consumers may try to implement it, even when they don't have to.
This PR adds `generateInternalResolversIfNeeded.__isTypeOf` to conditionally generate this resolver only for implementing types or union members

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -394,7 +394,6 @@ export interface RawResolversConfig extends RawConfig {
    *        plugins: ['typescript', 'typescript-resolver', { add: { content: "import { DeepPartial } from 'utility-types';" } }],
    *        config: {
    *          defaultMapper: 'DeepPartial<{T}>',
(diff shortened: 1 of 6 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -675,8 +674,8 @@ export class BaseResolversVisitor<
   protected _hasReferencedResolversUnionTypes = false;
   protected _hasReferencedResolversInterfaceTypes = false;
   protected _resolversUnionTypes: Record<string, string> = {};
-  protected _resolversUnionParentTypes: Record<string, string> = {};
```
eddeee888, line 678: `_resolversUnionParentTypes` was used previously but not needed anymore. Removing in this PR as a drive-by

eddeee888:
Will be done as part of https://github.com/dotansimha/graphql-code-generator/pull/10283

Issue #10162, opened 2024-10-04
[typescript-operations] Provide compatibility with Apollo Client data masking feature


jerelmiller:
### Is your feature request related to a problem? Please describe.

We are adding data masking to Apollo Client that will mask fragments in the result. To achieve the TypeScript end of this, we are utilizing GraphQL Codegen's fragment masking output generated by passing `inlineFragmentTypes: "mask"` in the codegen config.

What is currently missing is support for the `@unmask` directive we are adding as a part of this feature. We are adding this both to provide a migration path for existing Apollo Client users, and provide an escape hatch in situations where you might want to interact with the full result. We would like fragments marked with `@unmask` to be available in the parent type rather than stripped away as these fields will be available at runtime.

### Describe the solution you'd like

We'd like the `typescript-operations` plugin mask mode to keep fields in fragments marked with `@unmask`.

### Describe alternatives you've considered

We could provide our own maintained GraphQL Codegen plugin, but it would have to be a replacement for `typescript-operations` and we'd rather get this into core. The request is to add a tiny piece to the existing functionality.

### Any additional important details?

See the [data masking RFC](https://github.com/apollographql/apollo-client/issues/11666) for more details on the feature in Apollo Client. My [alpha announcement comment](https://github.com/apollographql/apollo-client/issues/11666#issuecomment-2392409229) especially has some more details in regards to how we plan to utilize the TypeScript output from `typescript-operations` in Apollo Client.

eddeee888:
This is available in `@graphql-codegen/typescript-operations@4.4.0`

PR #10163, opened 2024-10-04
Keep fields specified by `@unmask` for use with Apollo Client data masking

jerelmiller:
## Description
Provide support for `@unmask` with `inlineFragmentTypes: "mask"` so that these fields are available in the TypeScript type. This provides compatibility with Apollo Client's [data masking feature](https://github.com/apollographql/apollo-client/issues/11666) that will be released in the upcoming 3.12 release.

Related #10162
Related https://github.com/apollographql/apollo-client/issues/12085

## Type of change

expected)

See the [data masking RFC](https://github.com/apollographql/apollo-client/issues/11666) for more details on the feature in Apollo Client.

## How Has This Been Tested?

- [x] Added unit tests

**Test Environment**:

- OS:
- `@graphql-codegen/...`:
- NodeJS:

## Checklist:

## Further comments

I'm opening this as a draft to discuss the idea with the maintainers. This is a continuation of our conversation in Slack 🙂.


packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -7423,5 +7423,108 @@ function test(q: GetEntityBrandDataQuery): void {
...
+        export type UserFragmentFragment = { __typename?: 'User', id: string } & { ' $fragmentName'?: 'UserFragmentFragment' };
+        export type UserFragment2Fragment = { __typename?: 'User', email: string } & { ' $fragmentName'?: 'UserFragment2Fragment' };
+      `);
+    });
+
+    it("'mask' with @unmask and masked fragments on overlapping fields yields correct types", async () => {
+      const ast = parse(/* GraphQL */ `
+        query {
+          me {
+            ...UserFragment @unmask
```
jerelmiller, line 7497: Its unclear to me whether adding this functionality should be an extra option or built into the core functionality. Please advise. I'd be happy to change this up if we'd like to put an option around this.

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -7423,5 +7423,108 @@ function test(q: GetEntityBrandDataQuery): void {
         export type UserFragmentFragment = { __typename?: 'User', id: string } & { ' $fragmentName'?: 'UserFragmentFragment' };
       `);
     });
+
+    it("'mask' with @unmask yields correct types", async () => {
+      const ast = parse(/* GraphQL */ `
+        query {
+          me {
+            ...UserFragment @unmask
```
jerelmiller, line 7431: Apollo Client will have support for `@unmask(mode: "migrate")` which will generate runtime warnings if you access a field that would otherwise be masked. I'd love for this to generate the field and mark it with `@deprecated` so that users get a strikethrough in their editor.

I struggled to add this however as it felt the change needed would be a fairly large change to accomodate the data needed to add this. I'd love to discuss viable ways to handle this, both from an implementation standpoint, and in a way that makes sense for GraphQL Codegen. 

I realize this is quite specific to Apollo Client so I'd be happy to discuss viable options for adding this functionality without shifting it too much toward Apollo Client specifically (and FYI, we are open to changing the name of the argument on our end if it makes sense to do so).

eddeee888:
Should we update the doc for `inlineFragmentTypes` to mention the `mask` option and how it might interact with Apollo's `@unmask`? https://github.com/dotansimha/graphql-code-generator/blob/1bed87b201b6c0879ef605ec41f5a1acdcbfe1e7/packages/plugins/other/visitor-plugin-common/src/base-visitor.ts#L362-L365

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
eddeee888: It could go either way I think. I'm checking in with the team but this shouldn't be blocking (as we can do another PR before releasing).

However, my personal opinion here is we don't need an option for now. This directive is unique to Apollo Client, so only Apollo Client users would try to use it. Other users won't notice any difference.

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
eddeee888: I think `@deprecated` is a nice idea. How much of a change would this be?

IIRC GraphQL Codegen's fragment masking feature is just faux type, and the props are still available at runtime, unlike Apollo Client's feature. So, `@unmask` _could_ make sense if the user tries to widen the type. Maybe the `mode` could be more generic like `'report-invalid-usage'` or something similar?

eddeee888:
Could you please run `yarn changeset` and follow the prompts? 
It'd publish an alpha version once that file is committed and pushed.

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
eddeee888: Hi @jerelmiller 

I've checked with the team, since the directive is custom, it could be better to add an option. We are thinking that all custom directives can be under an parent `customDirectives` config. For example:

```
customDirectives: {
  apolloUnmask: true,
  // This can be used for other custom directives e.g.
  // relaySemanticNonNull: true,
}
```

jerelmiller: Ok great! I feel better about that rather than forcing a behavior with our specific implementation of the directive, so I will go this route. Thanks for checking on this!

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
jerelmiller: > I think @deprecated is a nice idea. How much of a change would this be?

The trouble here is that I need to be able to traverse the selection set of the named fragment and add `/** @deprecated */` above the generated properties for all fields in that type, but there is no real way to hold onto that "state" so that the code further down in the plugin can add this correctly. It felt like I'd need a large change in this plugin to be able to track that somehow so that by the time the string is created, we know whether to add that comment or not. Its just that seems to happen much further down in this plugin IIRC. If you have any suggestions on how best to approach this, that would be much appreciated.

That said, this is purely a DX thing. I care more that the fields are available in the type when `@unmask` is used, which is already working. I just thought it would be a nice touch to also show a strikethrough in the editor by marking the field as deprecated so that we have a nice visual to which fields will be removed once you remove the `@unmask(mode: "migrate")` directive from the named fragment.

> IIRC GraphQL Codegen's fragment masking feature is just faux type, and the props are still available at runtime, unlike Apollo Client's feature. So, `@unmask` could make sense if the user tries to widen the type.

I believe you're right. If I remember looking at the generated `useFragment` function from codegen's fragment masking feature, its just an identity function. Apollo Client on the other hand will either hide the properties in the returned type, or make them available if it detects an `@unmask` directive at a particular named fragment. I'd love for the types to reflect that as well, hence this change.

With the suggestion to add the additional config option though, I'm less concerned that this feature will leak outside of Apollo's use case. If down the line you adopt this directive as part of the fragment masking feature, then great! But I think using that config option first to introduce this behavior makes most sense 🙂 

> Maybe the mode could be more generic like 'report-invalid-access' or something similar?

We called that argument `mode: "migrate"` to signal to the user that using is used for migration purposes. The feature, when used with that mode in Apollo Client, will also console warn at runtime. I'd like to keep the name unless we have a good reason to change it since there is that runtime behavior. We will have a ton of users with large apps that can't use data masking immediately, so we'll have a codemod that will go apply `@unmask(mode: "migrate")` to all named fragments app-wide so that users can start the process of migrating to masking without having to do it at once. We figured it would be nice to support that whole use case in the TypeScript story as well.

Again, since we will be going with the new config option, I think the argument name is less important now that it will be marked as an apollo feature.

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -600,7 +600,10 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
 
       if (this._config.inlineFragmentTypes === 'combine' || this._config.inlineFragmentTypes === 'mask') {
         fragmentsSpreadUsages.push(selectionNode.typeName);
-        continue;
+
+        if (!selectionNode.fragmentDirectives?.some(d => d.name.value === 'unmask')) {
+          continue;
+        }
```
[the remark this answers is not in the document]
jerelmiller: Yes I think so. `mask` is the only one that removes fields from generated type correct? 

I can probably be a bit more explicit here if need be that this should only be combined with `unmask`.

jerelmiller:
> Should we update the doc for inlineFragmentTypes to mention the mask option and how it might interact with Apollo's `@unmask`?

Yes definitely! I'll add that change as well.

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
eddeee888: > Again, since we will be going with the new config option, I think the argument name is less important now that it will be marked as an apollo feature.

Agreed 👍  Sorry, I left the `report-invalid-access` comment before chatting to the team about using a specific `apolloUnmask` config and was thinking "if we could use it for our useFragment, what may a good name be?"

.changeset/twelve-windows-give.md
```diff
@@ -0,0 +1,16 @@
+---
+'@graphql-codegen/typed-document-node': minor
```
jerelmiller, line 2: I marked these as `minor` since I viewed this as a feature, but feel free to correct this if this seems off!

packages/plugins/other/visitor-plugin-common/src/base-documents-visitor.ts
```diff
@@ -149,6 +150,30 @@ export interface RawDocumentsConfig extends RawTypesConfig {
...
+   *          customDirectives: {
+   *            apolloUnmask: true
+   *          }
+   *        },
+   *      },
+   *    },
+   *  };
+   *  export default config;
+   */
+  customDirectives?: CustomDirectivesConfig;
```
jerelmiller, line 176: Here is the new config option to enable the new `apolloUnmask` directive. Let me know if there is a better place for these.

packages/plugins/other/visitor-plugin-common/src/base-visitor.ts
```diff
@@ -362,6 +362,7 @@ export interface RawConfig {
    * @description Whether fragment types should be inlined into other operations.
    * "inline" is the default behavior and will perform deep inlining fragment types within operation type definitions.
    * "combine" is the previous behavior that uses fragment type references without inlining the types (and might cause issues with deeply nested fragment that uses list types).
+   * "mask" transforms the types for use with fragment masking. Useful when masked types are needed when not using the "client" preset e.g. such as combining it with Apollo Client's data masking feature.
```
jerelmiller, line 365: Suggestions welcome on improving this

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
jerelmiller: I'm tabling the `@unmask(mode: "migrate")` with the `@deprecated` output for now. I'd love to add this in a future improvement, but having some form of `@unmask` support is the biggest thing for me right now. Would love to revisit when I have more time to learn the codegen codebase!

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
jerelmiller: Added the new config option in [a9aee49](https://github.com/dotansimha/graphql-code-generator/pull/10163/commits/a9aee4957f4b59e693f5da5f26247114601d3884)

jerelmiller:
@eddeee888 apologies for taking so long to get back to this! I'm finally back from all my conference travel and vacation. This should be ready to go on my end so please let me know if anything else needs tweaked. I'm holding off on the support for `@deprecated` with `@unmask(mode: "migrate")` for now and hope to come back to it. At this point though I think its a bit more work than I anticipated (happy to be proven wrong) and its far less important than having support for `@unmask` on its own.

Looking forward to landing this sometime soon!

.changeset/twelve-windows-give.md
jerelmiller: Thanks for confirming! I figured the list was too much so thanks for the assist. Updated in [5790112](https://github.com/dotansimha/graphql-code-generator/pull/10163/commits/57901125dac87c8e77adb6dcc0e309eef60f2994)

jerelmiller:
@eddeee888 got the prettier issue fixed in [5035012](https://github.com/dotansimha/graphql-code-generator/pull/10163/commits/503501277e85aa55682abd3ca02180ac91b14183). I think that should be the last thing missing. Thanks for reviewing! Looking forward to having this in there 🙂

dotansimha (approved):

PR #10199, opened 2024-11-22
Fix ESLint issues

eddeee888:
## Description

This PR fixes lint step on `master` because of ESLint errors (image below shows the start of the ESLint errors)

Screenshot 2024-11-22 at 11 31 04 PM

When I try checking it locally, we had about 80 errors, probably because we updated the rules recently? 

Screenshot 2024-11-22 at 11 30 18 PM

This PR fixes or ignore these rules to make the linting CI step green.

### DIFF
--- a/packages/graphql-codegen-cli/src/cli.ts
+++ b/packages/graphql-codegen-cli/src/cli.ts
@@ -31,7 +31,7 @@ export async function runCli(cmd: string): Promise<number> {
 export async function ensureGraphQlPackage() {
   try {
     await import('graphql');
-  } catch (e) {
+  } catch {
     throw new Error(
       `Unable to load "graphql" package. Please make sure to install "graphql" as a dependency! \n
        To install "graphql", run:
--- a/packages/graphql-codegen-cli/tests/init.spec.ts
+++ b/packages/graphql-codegen-cli/tests/init.spec.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
 import { resolve } from 'path';
 import bddStdin from 'bdd-stdin';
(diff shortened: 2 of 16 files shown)


packages/graphql-codegen-cli/src/cli.ts
```diff
@@ -31,7 +31,7 @@ export async function runCli(cmd: string): Promise<number> {
 export async function ensureGraphQlPackage() {
   try {
     await import('graphql');
-  } catch (e) {
+  } catch {
```
eddeee888, line 34: It's complaining we have `e` but it's never used. I think we could do `_e` but removing the variable like this works too

packages/graphql-codegen-cli/tests/init.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
```
eddeee888, line 1: We got a few require in here but I think it's fine, we can update it if needed

packages/plugins/other/visitor-plugin-common/src/variables-to-object.ts
```diff
@@ -122,7 +122,7 @@ export class OperationVariablesToObject {
     return `${formattedFieldString}: ${formattedTypeString}`;
   }
 
-  public wrapAstTypeWithModifiers(_baseType: string, _typeNode: TypeNode, _applyCoercion?: Boolean): string {
+  public wrapAstTypeWithModifiers(_baseType: string, _typeNode: TypeNode, _applyCoercion?: boolean): string {
```
eddeee888, line 125: ESLint suggests using `boolean` over `Boolean`

dimaMachina (approved):

PR #10201, opened 2024-11-25
Bump Components to have a /docs link in nav instead of Sign up

hasparus:
## Type of change

Quick fix to the navigation confusion we overlooked before.

image

## How Has This Been Tested?

I clicked the link.

### DIFF
--- a/website/tailwind.config.ts
+++ b/website/tailwind.config.ts
@@ -32,6 +32,7 @@ const config: Config = {
       },
     },
   },
+  content: [...theGuildTailwindConfig.content, './theme.config.tsx'],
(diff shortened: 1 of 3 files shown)


website/tailwind.config.ts
```diff
@@ -32,6 +32,7 @@ const config: Config = {
       },
     },
   },
+  content: [...theGuildTailwindConfig.content, './theme.config.tsx'],
```
hasparus, line 35: `max-w-[1392px]` didn't work so the navigation on landing didn't align with the hero and looked kinda off on large screens

Issue #10206, opened 2024-11-28
Federation support


eddeee888:
### Which packages are impacted by your issue?

@graphql-codegen/typescript-resolvers

### Describe the bug

TODO

### Your Example Website or App

TODO

### Steps to Reproduce the Bug or Issue

TODO

### Expected behavior

TODO

### Platform

- OS: [e.g. macOS, Windows, Linux]
- NodeJS: [e.g. 18.5.0]
- `graphql` version: [e.g. 16.3.0]
- `@graphql-codegen/*` version(s): [e.g. 2.6.2]

LongLiveCHIEF:
It would be great if I could specify the federation version in the `codegen.ts`, and have that automatically add the `extends schema @link` with the correct federation version.

Is there a feature on the backlog for this?

eddeee888:
Hello @LongLiveCHIEF  
This is not planned at the moment, as this checklist is for gaps in federation types, and this sounds like a new feature.
Since this is my first time hearing it, could you help me understand how this helps users and why alternatives are not great? e.g. could the version be updated directly in the schema file/s?

LongLiveCHIEF:
I want to take a better pass through the docs and then if I still want to discuss, I would open a new issue as a feature request.

Basically, I want to reduce the boiler plate in each schema file, and federation versions can change quite often.

There could be a banner option or something that I'm missing that would suffice for this.

eddeee888:
Released in https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1757264511789

lworkman:
@eddeee888 Was the above mentioned `@inaccessible` behaviour ever added? I don't see any mention of it looking through the codebase.

eddeee888:
Hey @lworkman , you are right, there's no implementation of it 😅
It's been awhile, and I don't remember why it wasn't implemented. Maybe a mistake on my end.

Are you seeing issues? Could you please create an issue to help me track?

lworkman:
I don't know if it's an issue, more of a feature request. The problem is that using `inaccessible` enum values can cause GraphQL validation errors because the consuming schema does not consider them valid.

Happy to write something up, can you point me to contributor guidelines?

eddeee888:
Amazing @lworkman !

1. You can start by creating a feature request: [issues page](https://github.com/dotansimha/graphql-code-generator/issues) > Click on New > Feature Request, then fill out the form.
[...]

PR #10216, opened 2024-12-10
[resolvers][federation] Fix mapper being incorrectly used as the base type for reference

eddeee888:
[](https://www.youtube.com/watch?v=9mbWNVyIGm4)
👆 Video explaining this change: https://www.youtube.com/watch?v=9mbWNVyIGm4

## Description

`__resolveReference`'s first param (i.e. usually called "reference") can never be mappers because:

1. `__resolveReference` is only called by the gateway with data from `@key` and/or `@requires`
2. `__resolveReference` is never reached by the subgraph it is in, therefore it can never receive mapper

This PR creates a new `FederationTypes` to make it easier to target federation object types in entity types that need it.

For example, the following `User` type is an entity with `id` being the key
```graphql
type User @key(fields: "id") {
  id: ID!
  name: String
}
```
So, the `UserResolvers` would have some differences compared to normal types:
1. It has an extra generic: `FederationType extends FederationTypes['User'] = FederationTypes['User']` to target the base `User` entity 
2. `__resolveReference` uses `FederationType` to pick the key using `GraphQLRecursivePick`

## Other notes

As part of change from `ParentType` -> `FederationType`, we no longer need the `UnwrappedObject` added when `wrapFieldDefinitions = true`. This PR also removes this utility type and relevant codes

## Type of change

## How Has This Been Tested?

- [ ] Unit test
- [ ] Alpha version tested

### DIFF
--- a/packages/plugins/typescript/resolvers/tests/ts-resolvers.federation.mappers.spec.ts
+++ b/packages/plugins/typescript/resolvers/tests/ts-resolvers.federation.mappers.spec.ts
@@ -1,19 +1,5 @@
 import '@graphql-codegen/testing';
-import { codegen } from '@graphql-codegen/core';
-import { parse } from 'graphql';
-import { TypeScriptResolversPluginConfig } from '../src/config.js';
-import { plugin } from '../src/index.js';
-
-function generate({ schema, config }: { schema: string; config: TypeScriptResolversPluginConfig }) {
-  return codegen({
-    filename: 'graphql.ts',
-    schema: parse(schema),
-    documents: [],
-    plugins: [{ 'typescript-resolvers': {} }],
-    config,
-    pluginMap: { 'typescript-resolvers': { plugin } },
-  });
-}
+import { generate } from './utils';
 
 describe('TypeScript Resolvers Plugin + Apollo Federation - mappers', () => {
   it('generates FederationTypes and use it for reference type', async () => {
--- a/packages/plugins/typescript/resolvers/tests/utils.ts
+++ b/packages/plugins/typescript/resolvers/tests/utils.ts
@@ -1,0 +1,15 @@
+import { codegen } from '@graphql-codegen/core';
+import { parse } from 'graphql';
+import { TypeScriptResolversPluginConfig } from '../src/config.js';
+import { plugin } from '../src/index.js';
+
+export function generate({ schema, config }: { schema: string; config: TypeScriptResolversPluginConfig }) {
+  return codegen({
+    filename: 'graphql.ts',
+    schema: parse(schema),
+    documents: [],
+    plugins: [{ 'typescript-resolvers': {} }],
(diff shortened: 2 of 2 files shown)


packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -172,30 +174,32 @@ export class ApolloFederation {
 
       const { resolvableKeyDirectives } = objectTypeFederationDetails;
 
-      if (resolvableKeyDirectives.length) {
-        const outputs: string[] = [`{ __typename: '${parentType.name}' } &`];
+      if (resolvableKeyDirectives.length === 0) {
+        return federationTypeSignature;
+      }
```
eddeee888, line 179: Since we need to return `federationTypeSignature` instead of `parentTypeSignature` when there's all `@key` are `resolvable=false`, I'm changing to return early here to make it easier to read as well

packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -106,13 +106,6 @@ export type ResolverWithResolve<TResult, TParent, TContext, TArgs> = {
   const stitchingResolverUsage = `StitchingResolver<TResult, TParent, TContext, TArgs>`;
 
   if (visitor.hasFederation()) {
-    if (visitor.config.wrapFieldDefinitions) {
-      defsToInclude.push(`export type UnwrappedObject<T> = {
-        [P in keyof T]: T[P] extends infer R | Promise<infer R> | (() => infer R2 | Promise<infer R2>)
-          ? R & R2 : T[P]
-      };`);
```
eddeee888, line 113: For a given type:
```ts
export type ProductMapper = {
  id: () => Promise<string>;
};
```
This type seems to return something like `(() => Promise<string>) & string` which is unusual - How can a function intersect with a string?

Regardless, since we move from `ParentType` to `FederationType`, I don't think this is needed in the future.

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1223,6 +1223,28 @@ export class BaseResolversVisitor<
...
+      .withName(this.convertName('FederationTypes'))
+      .withComment('Mapping of federation types')
+      .withBlock(
+        Object.keys(federationMeta)
+          .map(typeName => {
+            return indent(`${typeName}: ${this.convertName(typeName)}${this.getPunctuation(declarationKind)}`);
+          })
+          .join('\n')
+      ).string;
+  }
```
eddeee888, line 1246: This `FederationTypes` type will be used by federation entities to refer to the base types, similar to other types like `ResolversParentTypes` and `ResolversTypes`

eddeee888:
Note: current Examples Tests in CI are broken because of an unrelated issue.

PR #10221, opened 2024-12-22
[resolvers][federation] Add `__resolveReference` to applicable `Interface` entities, fix Interface types having non-meta resolver fields

eddeee888:
TODO: video walkthrough

## Description

This PR ensures correct types for Interfaces:

### 1. `__resolveReference` are generated for Interface entities

Federation Checklist: https://github.com/dotansimha/graphql-code-generator/issues/10206

```graphql
interface User @key(fields: "id") { # must generate `__resolveReference` for User
  # User fields
}
```

https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/entities/interfaces#example-schemas

### 2. Interfaces do not have non-meta resolvers

Related PR: https://github.com/dotansimha/graphql-code-generator/pull/5666

```graphql
interface User { 
  # fields below should not be generated for User interface because they are never called.
  # Implementing types fields are called instead
  id: ID!
  name: String!
}
```

## Breaking Changes

As part of this work, we remove unnecessary config that could generate wrong types for both Federation and normal use cases:

- BREAKING CHANGES: Deprecate `onlyResolveTypeForInterfaces` because majority of use cases cannot implement resolvers in Interfaces.
- BREAKING CHANGES: Deprecate `generateInternalResolversIfNeeded.__resolveReference` because types do not have `__resolveReference` if they are not Federation entities or are not resolvable. Users should not have to manually set this option. This option was put in to wait for this major version.

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -1,4 +1,4 @@
-import { ApolloFederation, checkObjectTypeFederationDetails, getBaseType } from '@graphql-codegen/plugin-helpers';
+import { ApolloFederation, type FederationMeta, getBaseType } from '@graphql-codegen/plugin-helpers';
 import { getRootTypeNames } from '@graphql-tools/utils';
 import autoBind from 'auto-bind';
 import {
@@ -78,13 +78,15 @@ export interface ParsedResolversConfig extends ParsedConfig {
   allResolversTypeName: string;
   internalResolversPrefix: string;
   generateInternalResolversIfNeeded: NormalizedGenerateInternalResolversIfNeededConfig;
-  onlyResolveTypeForInterfaces: boolean;
   directiveResolverMappings: Record<string, string>;
   resolversNonOptionalTypename: ResolversNonOptionalTypenameConfig;
   avoidCheckingAbstractTypesRecursively: boolean;
 }
 
-type FieldDefinitionPrintFn = (parentName: string, avoidResolverOptionals: boolean) => string | null;
+type FieldDefinitionPrintFn = (
+  parentName: string,
+  avoidResolverOptionals: boolean
+) => { value: string | null; meta: { federation?: { isResolveReference: boolean } } };
 export interface RootResolver {
   content: string;
   generatedResolverTypes: {
@@ -584,20 +586,13 @@ export interface RawResolversConfig extends RawConfig {
   internalResolversPrefix?: string;
   /**
    * @type object
-   * @default { __resolveReference: false }
+   * @default {}
    * @description If relevant internal resolvers are set to `true`, the resolver type will only be generated if the right conditions are met.
    * Enabling this allows a more correct type generation for the resolvers.
    * For example:
    * - `__isTypeOf` is generated for implementing types and union members
-   * - `__resolveReference` is generated for federation types that have at least one resolvable `@key` directive
    */
   generateInternalResolversIfNeeded?: GenerateInternalResolversIfNeededConfig;
-  /**
-   * @type boolean
-   * @default false
-   * @description Turning this flag to `true` will generate resolver signature that has only `resolveType` for interfaces, forcing developers to write inherited type resolvers in the type itself.
-   */
-  onlyResolveTypeForInterfaces?: boolean;
   /**
    * @description Makes `__typename` of resolver mappings non-optional without affecting the base types.
    * @default false
@@ -700,7 +695,8 @@ export class BaseResolversVisitor<
     rawConfig: TRawConfig,
     additionalConfig: TPluginConfig,
     private _schema: GraphQLSchema,
-    defaultScalars: NormalizedScalarsMap = DEFAULT_SCALARS
+    defaultScalars: NormalizedScalarsMap = DEFAULT_SCALARS,
+    federationMeta: FederationMeta = {}
   ) {
     super(rawConfig, {
       immutableTypes: getConfigValue(rawConfig.immutableTypes, false),
@@ -714,7 +710,6 @@ export class BaseResolversVisitor<
         mapOrStr: rawConfig.enumValues,
       }),
       addUnderscoreToArgsType: getConfigValue(rawConfig.addUnderscoreToArgsType, false),
-      onlyResolveTypeForInterfaces: getConfigValue(rawConfig.onlyResolveTypeForInterfaces, false),
       contextType: parseMapper(rawConfig.contextType || 'any', 'ContextType'),
       fieldContextTypes: getConfigValue(rawConfig.fieldContextTypes, []),
       directiveContextTypes: getConfigValue(rawConfig.directiveContextTypes, []),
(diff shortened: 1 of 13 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -584,20 +583,13 @@ export interface RawResolversConfig extends RawConfig {
...
    * - `__isTypeOf` is generated for implementing types and union members
-   * - `__resolveReference` is generated for federation types that have at least one resolvable `@key` directive
    */
   generateInternalResolversIfNeeded?: GenerateInternalResolversIfNeededConfig;
-  /**
-   * @type boolean
-   * @default false
-   * @description Turning this flag to `true` will generate resolver signature that has only `resolveType` for interfaces, forcing developers to write inherited type resolvers in the type itself.
-   */
-  onlyResolveTypeForInterfaces?: boolean;
```
eddeee888, line 600: Related https://github.com/dotansimha/graphql-code-generator/issues/5648

`onlyResolveTypeForInterfaces` is only applicable for some rare use cases mentioned [here](https://github.com/dotansimha/graphql-code-generator/issues/5648#issuecomment-790151449). This option's implementation blocks `__resolveReference` being created for Interfaces when the option is turned on.

I think it's simpler just to deprecate this option as we are moving to a new major version. If anyone complains, we can put it back as a minor version

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -584,20 +583,13 @@ export interface RawResolversConfig extends RawConfig {
...
    * @type object
-   * @default { __resolveReference: false }
+   * @default {}
    * @description If relevant internal resolvers are set to `true`, the resolver type will only be generated if the right conditions are met.
    * Enabling this allows a more correct type generation for the resolvers.
    * For example:
    * - `__isTypeOf` is generated for implementing types and union members
-   * - `__resolveReference` is generated for federation types that have at least one resolvable `@key` directive
    */
   generateInternalResolversIfNeeded?: GenerateInternalResolversIfNeededConfig;
```
eddeee888, line 592: `generateInternalResolversIfNeeded` might be deprecated in this major version, it'll come in one of the next few PRs.
Since we are bumping a major version, it's worth generating these meta types only when required

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -28,14 +30,48 @@ export const federationSpec = parse(/* GraphQL */ `
...
+export type FederationMeta = { [typeName: string]: TypeMeta };
+
 /**
- * Adds `__resolveReference` in each ObjectType involved in Federation.
+ * Adds `__resolveReference` in each ObjectType and InterfaceType involved in Federation.
  * @param schema
  */
-export function addFederationReferencesToSchema(schema: GraphQLSchema): GraphQLSchema {
-  return mapSchema(schema, {
+export function addFederationReferencesToSchema(schema: GraphQLSchema): {
```
eddeee888, line 43: We are transforming the schema to have `__resolveReference` fields to prepare for the resolvers plugin. Because of this pattern, it's worth capturing meta here so, and use the meta instead of running `checkTypeFederationDetails` in the plugin.

Because of this, `checkTypeFederationDetails` is now only used in this file i.e. it's no longer exported and called in resolvers plugin -> better logic encapsulation

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -28,14 +30,67 @@ export const federationSpec = parse(/* GraphQL */ `
...
- * Adds `__resolveReference` in each ObjectType involved in Federation.
+ * Adds `__resolveReference` in each ObjectType and InterfaceType involved in Federation.
+ * We do this to utilise the existing FieldDefinition logic of the plugin, which includes many logic:
+ * - mapper
+ * - return type
  * @param schema
  */
-export function addFederationReferencesToSchema(schema: GraphQLSchema): GraphQLSchema {
-  return mapSchema(schema, {
+export function addFederationReferencesToSchema(schema: GraphQLSchema): {
```
eddeee888, line 47: We now collect Federation meta in this function because this is called before calling visitor.
This means we no longer have to collect meta as we find entities during visitor -> simpler, more predictable, and more performant

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -78,13 +78,15 @@ export interface ParsedResolversConfig extends ParsedConfig {
...
   directiveResolverMappings: Record<string, string>;
   resolversNonOptionalTypename: ResolversNonOptionalTypenameConfig;
   avoidCheckingAbstractTypesRecursively: boolean;
 }
 
-type FieldDefinitionPrintFn = (parentName: string, avoidResolverOptionals: boolean) => string | null;
+type FieldDefinitionPrintFn = (
+  parentName: string,
+  avoidResolverOptionals: boolean
+) => { value: string | null; meta: { federation?: { isResolveReference: boolean } } };
```
eddeee888, line 89: Unfortunately, we don't know if a field is `__resolveReference` or not until we have generated it. This is because we are injecting `__resolveReference` as real fields in `addFederationReferencesToSchema`. Therefore, we need to add `meta` to have more info about the generated field

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1545,30 +1543,22 @@ export class BaseResolversVisitor<
...
-
-          if (!federationDetails || federationDetails.resolvableKeyDirectives.length === 0) {
-            return '';
-          }
-          signature.modifier = ''; // if a federation type has resolvable @key, then it should be required
+        if (!this._federation.getMeta()[parentType.name].hasResolveReference) {
+          return { value: '', meta };
         }
-
-        this._federation.setMeta(parentType.name, { hasResolveReference: true });
```
eddeee888, line 1560: We no longer check and collect meta as we hit federation fields any more because we are doing it before running visitor. [Relevant comment](https://github.com/dotansimha/graphql-code-generator/pull/10221/files?diff=split&w=1#r1912379037).
Therefore, we can check federation meta directly here

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1545,30 +1543,22 @@ export class BaseResolversVisitor<
...
         signature.type = 'ReferenceResolver';
-        if (signature.genericTypes.length >= 3) {
-          signature.genericTypes = signature.genericTypes.slice(0, 3);
-        }
```
eddeee888, line 1564: It's a bit odd to do the slice, may as well be explicit here!

packages/plugins/typescript/resolvers/tests/ts-resolvers.federation.spec.ts
```diff
@@ -498,50 +372,6 @@ describe('TypeScript Resolvers Plugin + Apollo Federation', () => {
...
-      },
-    });
-
-    expect(content).toBeSimilarStringTo(`
-    export type PersonResolvers<ContextType = any, ParentType extends ResolversParentTypes['Person'] = ResolversParentTypes['Person']> = {
-      __resolveType: TypeResolveFn<'User' | 'Admin', ParentType, ContextType>;
-      age?: Resolver<Maybe<ResolversTypes['Int']>, { __typename: 'User' | 'Admin' } & GraphQLRecursivePick<ParentType, {"name":{"first":true,"last":true}}>, ContextType>;
-    };
-    `);
-  });
```
eddeee888, line 543: This Interface test is now covered in `ts-resolvers.federation.interface.spec.ts`

dotansimha (approved):

Issue #10196, opened 2024-11-20
loadSchema function cannot accept array of pointers in Typescript declaration files


Nnigmat:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

Function `loadSchema` should support array of pointers as an argument. The function support this behaviour, but the types are not.

https://github.com/dotansimha/graphql-code-generator/blob/b73ed8f18cf55d00640617f7bc5c33d5cfc88663/packages/graphql-codegen-cli/src/load.ts#L31

The same problem in method `loadSchema` in `CodegenContext` class. It should be Types.Schema[].
https://github.com/dotansimha/graphql-code-generator/blob/b73ed8f18cf55d00640617f7bc5c33d5cfc88663/packages/graphql-codegen-cli/src/config.ts#L425

`loadSchema` from `@graphql-tools/load` supports array of pointers too.

https://github.com/ardatan/graphql-tools/blob/66b6a98d7e134e932f28e7788b4884d2c0461b12/packages/load/src/schema.ts#L45

### Your Example Website or App

-

### Steps to Reproduce the Bug or Issue

`loadSchema(['src/schema-1.graphql', 'src/schema-2.graphql']` shows error

### Expected behavior

`loadSchema(['src/schema-1.graphql', 'src/schema-2.graphql']` don't show error

### Platform

- OS: [e.g. macOS, Windows, Linux]
- NodeJS: [e.g. 18.5.0]
- `graphql` version: [e.g. 16.3.0]
- `@graphql-codegen/*` version(s): [e.g. 2.6.2]

PR #10227, opened 2025-01-03
Fix load schema pointer type

eddeee888:
## Description

This fixes a type issue in functions used in programmatic scenarios which have the wrong type for schema param. We can take a single schema or an array. This PR adds the array part to it

Related # https://github.com/dotansimha/graphql-code-generator/issues/10196

## Type of change

## How Has This Been Tested?

- [x] Unit test


packages/graphql-codegen-cli/src/load.ts
```diff
@@ -28,7 +28,7 @@ export const defaultDocumentsLoadOptions = {
 };
 
 export async function loadSchema(
-  schemaPointers: UnnormalizedTypeDefPointer,
+  schemaPointers: UnnormalizedTypeDefPointer | UnnormalizedTypeDefPointer[],
```
ardatan, line 31: Should we add tests for this use case?

eddeee888: Good idea! This is done in https://github.com/dotansimha/graphql-code-generator/pull/10227/commits/9efc60b7e075fce6ef74b79d4bb27803ce4115ec

Note that instead of testing this `load.ts` directly, I'm passing multiple schemas into Codegen Context which eventually passes those schemas to `load.ts`

PR #10242, opened 2025-01-15
Switch to use stable SWAPI endpoint in examples

eddeee888:
## Description

https://github.com/graphql/swapi-graphql recently moved to a new endpoint, so this PR updates the endpoint references. 
Some tests were failing because they were using the old endpoint

## Type of change

- [x] Fix CI

### DIFF
--- a/examples/react/apollo-client-swc-plugin/codegen.ts
+++ b/examples/react/apollo-client-swc-plugin/codegen.ts
@@ -2,7 +2,7 @@
 import { CodegenConfig } from '@graphql-codegen/cli';
 
 const config: CodegenConfig = {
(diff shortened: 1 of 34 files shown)


saihaj (dismissed):
ohh i would move to this is the stable endpoint. Folks try to point to the neltify thing but that's internal only (yes people find it) but this endpoint you can be certain will be there :) 

```
https://graphql.org/graphql/
```

eddeee888:
Thanks @saihaj ! I've updated to `https://graphql.org/graphql/`, and we worked through the endpoint issues outside of this PR. This is now working!

I'll merge this first to unblock CI & rebase a few PRs - if there's other comments, happy to implement as subsequent PRs

Issue #10245, opened 2025-01-16
`apolloUnmask` in client preset


16patsle:
### Which packages are impacted by your issue?

@graphql-codegen/client-preset

### Describe the bug

In `@graphql-codegen/typescript-operations` version 4.4.0, support for the `@unmask` directive from Apollo's data masking was added. This requires settings `inlineFragmentTypes: "mask"` and `customDirectives: { apolloUnmask: true }` in the config, and seems to work as expected. However, `@graphql-codegen/client-preset` (version 4.5.1) does not seem to support this. [Apollo's docs suggest](https://www.apollographql.com/docs/react/data/fragments#with-the-client-preset) to set `customDirectives` in `presetConfig`, but neither config nor presetConfig seems to work for this combination.

### Your Example Website or App

https://stackblitz.com/edit/16patsle-codegen-apollo-unmask

### Steps to Reproduce the Bug or Issue

1. Set up codegen to support Apollo data masking (like in the example)
2. Have a query with a fragment and `@unmask`
3. Find the generated type for the Query
4. Notice how it does not include the fields from the fragment, despite the `@unmask` directive

```ts
export type AllPostsQueryQuery = { __typename?: 'Query', allPost: Array<(
    { __typename?: 'Post', id: string }
    & { ' $fragmentRefs'?: { 'PostFragmentFragment': PostFragmentFragment } }
  ) | null> };
```

### Expected behavior

I would expect the `@unmask` directive to work and give output similar to this:

```ts
export type AllPostsQueryQuery = { __typename?: 'Query', allPost: Array<(
    { __typename?: 'Post', id: string, title: string, excerpt?: string | null }
    & { ' $fragmentRefs'?: { 'PostFragmentFragment': PostFragmentFragment } }
  )> };
```

Notice that this includes other fields that are present in the PostFragment fragment in the example code

I guess the current behavior is in one way be expected given that `customDirectives` is not one of the options forwarded by client-preset. However, I think it would be desired for the client-preset to support customDirectives, and it does indeed work if I add customDirectives to `forwardedConfig` in the preset code.

### Platform

- OS: macOS, Linux
- NodeJS: v22.12.0, v20.17.0
- `graphql` version: 16.10.0
- `@graphql-codegen/*` version(s): `@graphql-codegen/cli@5.0.3`, `@graphql-codegen/client-preset@4.5.1`,
    `@graphql-typed-document-node/core@3.2.0`

eddeee888:
[...]

PR #10268, opened 2025-01-22
Forward customDirectives in client preset to support Apollo unmask directive

eddeee888:
## Description

This is a continuation of https://github.com/dotansimha/graphql-code-generator/pull/10163

Related issue https://github.com/dotansimha/graphql-code-generator/issues/10245

The previous PR added support for `unmask` directive in `typescript-operations`. However, we need to forward it to the client preset for it to take effect. This PR adds that and unit test that covers the recommended config:

```
presetConfig: { fragmentMasking: false },
config: {
  inlineFragmentTypes: 'mask',
  customDirectives: { apolloUnmask: true },
},
```

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/.changeset/great-dancers-pump.md
+++ b/.changeset/great-dancers-pump.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/client-preset': minor
+---
+
+Forward customDirectives to support Apollo unmask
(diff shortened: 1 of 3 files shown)


jerelmiller (approved):
Looks good from my side! Thanks for fixing this.

Issue #10281, opened 2025-02-06
Schema URL (not a filesystem path) forces the watcher to scan the root directory


oprypkhantc:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

Hey. 

E.g. given this config:
```ts
import type { CodegenConfig } from '@graphql-codegen/cli';

const config: CodegenConfig = {
	overwrite: true,
	schema: 'https://api.tc.loc/graphql',
	documents: ['src/**/*.graphql'],
	generates: {
		'src/api/schema.graphql': {
			plugins: ['schema-ast'],
		},
		'src/api/types/': {
			preset: preset,
		},
		'src/': {
			preset: 'near-operation-file',
			plugins: [
				'typescript-operations',
				'typescript-apollo-angular',
				'fragment-matcher',
			],
		},
	},
};

export default config;
```

and this project structure:

```
.
├── .angular/
│   └── cache
├── src/
│   ├── source.ts
│   └── api.graphql
├── node_modules/
│   ├── dep1
│   └── dep2
├── graphql-codegen.ts
├── package.json
└── yarn.lock
```

The watcher (`graphql-codegen --watch`) would be watching the entire root, including `.angular` and `node_modules`, which degrades the performance by a lot. It will still not trigger the generation itself (because of filtering that happens later), but `@parcel/watcher` internally with receive the project root as the directory being watched.

This is because `graphql-codegen` watches the "lowest common denominator" directory of all paths. In this case, the watched paths are:
- `graphql-codegen.ts`
- `src/**/*.graphql`
- `https://api.tc.loc/graphql`

For which the common directory is the project root (`/`). This is caused by two things:
- graphql-codegen config is always forcefully watched (meaning if the config is in the project root, the project root is also always watched, no matter what)
- `https://api.tc.loc/graphql` is treated as a file path (`process.cwd() + `/https://api.tc.loc/graphql`), which also forces the project root to always be watched

The easiest fix for both is:
- moving `graphql-codegen.ts` under `src/`
- add a condition [here](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/utils/patterns.ts#L178) to exclude URLs from being watched by `@parcel/watcher`

A more proper fix is to subscribe to each watched directory separately, and stop trying to detect the common root of all paths.

### Your Example Website or App

https://stackblitz.com/edit/github-fmcayzpe-gjq6kh9t

[...]

PR #10282, opened 2025-02-07
fix: Don't treat schema URLs as relative file paths for the watcher

oprypkhantc:

## Description

Partial fix for #10281. This PR makes sure that schema URLs are __not__ being treated as file paths when creating a watcher for `--watch`. This allows the common watched directory to be a nested directory, which in turn dramatically increases performance of the watcher as it doesn't have to watch all the extra directories (like `node_modules`).

## Type of change

## How Has This Been Tested?

- [X] Unit tests

**Test Environment**:

- OS: macOS
- `@graphql-codegen/...`: master
- NodeJS: 20.15.1

## Checklist:

- [X] I have made corresponding changes to the documentation
- [X] Any dependent changes have been merged and published in downstream modules

## Further comments

As I said, it's only a partial fix. This resolves the issue of schema URL being treated as a file, which forces the watcher to watch the project root. However, this still does not allow the watcher to be efficient and performant when the graphql-codegen config is located in the project root, or when watched files are in two different locations.

A proper fix would be a refactor of the `watcher.ts` - one that utilizes multiple watcher subscriptions per each watched path OR one that automatically specifies all directories except relevant as ignored. I'm not sure which solution would give better performance, and I'm not sure that this project will accept a PR making those changes. Please let me know if this refactor would be possible.

### DIFF
--- a/.changeset/hungry-ties-press.md
+++ b/.changeset/hungry-ties-press.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/cli': patch
+---
+
+Fix watcher watching project root when schema URL is used
--- a/packages/graphql-codegen-cli/src/utils/helpers.ts
+++ b/packages/graphql-codegen-cli/src/utils/helpers.ts
@@ -1,0 +1,8 @@
+export function isURL(str: string): boolean {
+  try {
+    const url = new URL(str);
+    return !!url;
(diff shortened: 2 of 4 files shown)


eddeee888 (approved):
Thank you!

PR #10283, opened 2025-02-08
Ensure `__isTypeof` is only generated for implementing types (of Interfaces) or Union members

eddeee888:
## Description

Only implementing types (of Interfaces) or Union members need `__isTypeOf` to help pick the right type name for the responses. This PR implements this behaviour by only generating `__isTypeOf` for said object types.

This PR also introduces `_parsedSchemaMeta` to hold metadata about the schema. This can help with performance and reduces complexity because we only parse through the schema _once_ to collect meta about the types, and use the collected meta for subsequent generation.

---

Note: This is an re-implementation for https://github.com/dotansimha/graphql-code-generator/pull/10138, because this will be released as a breaking change instead of a new config.

## Type of change

## How Has This Been Tested?

- [x] Unit Test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -6,9 +6,11 @@ import {
   DirectiveDefinitionNode,
   EnumTypeDefinitionNode,
   FieldDefinitionNode,
+  GraphQLInterfaceType,
   GraphQLNamedType,
   GraphQLObjectType,
   GraphQLSchema,
+  GraphQLUnionType,
   InputValueDefinitionNode,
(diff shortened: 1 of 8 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -679,7 +694,6 @@ export class BaseResolversVisitor<
   protected _hasReferencedResolversUnionTypes = false;
   protected _hasReferencedResolversInterfaceTypes = false;
   protected _resolversUnionTypes: Record<string, string> = {};
-  protected _resolversUnionParentTypes: Record<string, string> = {};
```
eddeee888, line 682: This is no longer used, so we can remove it now.

dotansimha (approved):

PR #10287, opened 2025-02-10
[resolvers][federation] Fix fields or types being wrong generated when marked with @external

eddeee888:
## Description

When a field or type is marked with `@external`, the resolver should not be generated (unless part of `@provides`). This PR takes care of the following scenarios:

- [x] If a field is marked as `@external`, do not generate it
- [x] If any field is marked with `@external`, the object type's resolvers signature must still use `ParentType` for its first param
- [x] If all fields in an object type is marked as `@external`, do not generate the type
- [x] If the object type is marked as `@external`, do not generate the type
- [x] Question: if a field returns object type and not marked with `@external`, but the object type itself is marked with `@external`, what happens? (`PlaceOfBirth` in the test setup)
     - The return type of `User.placeOfBirth` points to the base `PlaceOfBirth`, i.e. we don't need to generate `PlaceOfBirthResolvers` because it is unused.
- [x] Question: same as above but with `@provides` in play? (`Company` in the test setup)
     - It should work the same way as normal `@provides`

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -86,8 +86,6 @@ type FieldDefinitionPrintFn = (
   parentName: string,
   avoidResolverOptionals: boolean
 ) => { value: string | null; meta: { federation?: { isResolveReference: boolean } } };
-export type FieldDefinitionResult = [{ node: FieldDefinitionNode }, FieldDefinitionPrintFn];
-
 export interface RootResolver {
   content: string;
   generatedResolverTypes: {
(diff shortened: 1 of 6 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1519,122 +1524,133 @@ export class BaseResolversVisitor<
...
-      const meta: ReturnType<FieldDefinitionPrintFn>['meta'] = {};
+    const original: FieldDefinitionNode = parent[key];
 
-      if (this._federation.skipField({ fieldNode: original, parentType })) {
-        return { value: null, meta };
-      }
+    return {
+      node: original,
+      printContent: (parentNode, avoidResolverOptionals) => {
+        const parentName = parentNode.name as unknown as string;
```
eddeee888, line 1536: I see a lot of type cast like `as any as string` across this plugin. 
Actually, I'm not sure what's causing it. 
I vaguely remember seeing _something_ changing the type (Maybe new vs old visitor?) but can't exactly remember what. 😅 
Would appreciate some pointers

packages/plugins/typescript/resolvers/tests/ts-resolvers.config.customDirectives.spec.ts
```diff
@@ -62,7 +62,6 @@ describe('customDirectives.sematicNonNull', () => {
         nonNullableListWithNonNullableItemLevel0?: Resolver<Array<ResolversTypes['String']>, ParentType, ContextType>;
         nonNullableListWithNonNullableItemLevel1?: Resolver<Array<ResolversTypes['String']>, ParentType, ContextType>;
         nonNullableListWithNonNullableItemBothLevels?: Resolver<Array<ResolversTypes['String']>, ParentType, ContextType>;
-        __isTypeOf?: IsTypeOfResolverFn<ParentType, ContextType>;
```
eddeee888, line 65: This is a new test from `master`. 
`__isTypeOf` is not required here after this PR in the feature branch: https://github.com/dotansimha/graphql-code-generator/pull/10283

packages/plugins/typescript/resolvers/tests/ts-resolvers.federation.spec.ts
```diff
@@ -520,16 +546,41 @@ describe('TypeScript Resolvers Plugin + Apollo Federation', () => {
...
+
+    // `PlaceOfBirthResolvers` should not be generated because the type is marked with @external, even if `User.placeOfBirth` is not marked with @external
+    expect(content).not.toBeSimilarStringTo('export type PlaceOfBirthResolvers');
+
+    // FIXME: `CompanyResolvers` should only have taxCode resolver because it is part of the `@provides` directive in `Book.editor`, even if the whole `Company` type is marked with @external
+    // expect(content).toBeSimilarStringTo(`
+    //   export type CompanyResolvers<ContextType = any, ParentType extends ResolversParentTypes['Company'] = ResolversParentTypes['Company']> = {
+    //     taxCode?: Resolver<ResolversTypes['String'], ParentType, ContextType>;
+    //   };
+    // `);
```
eddeee888, line 583: I feel this edge case is quite hard to implement, will come back in another PR, or maybe after the major version bump.

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -345,19 +382,15 @@ export class ApolloFederation {
...
-  }
-
-  private isExternal(node: FieldDefinitionNode): boolean {
+  private isExternal(node: FieldDefinitionNode | ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode): boolean {
     return getDirectivesByName('external', node).length > 0;
   }
 
-  private hasProvides(objectType: ObjectTypeDefinitionNode | GraphQLObjectType, node: FieldDefinitionNode): boolean {
-    const fields = this.providesMap[isObjectType(objectType) ? objectType.name : objectType.name.value];
+  private hasProvides(node: ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode, fieldName: string): boolean {
```
eddeee888, line 389: Note for the future: I find allowing both types (e.g. `GraphQLObjectType`) and nodes (e.g. `ObjectTypeDefinitionNode`) can get confusing very quickly. 

Maybe we should just only accept nodes in the future?

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1519,122 +1527,133 @@ export class BaseResolversVisitor<
     return `ParentType extends ${parentType} = ${parentType}`;
   }
 
-  FieldDefinition(node: FieldDefinitionNode, key: string | number, parent: any): FieldDefinitionPrintFn {
+  FieldDefinition(node: FieldDefinitionNode, key: string | number, parent: any): FieldDefinitionResult {
     const hasArguments = node.arguments && node.arguments.length > 0;
     const declarationKind = 'type';
 
-    return (parentName, avoidResolverOptionals) => {
```
eddeee888, line 1526: This function is used to print field definition. We need more details when we handle fields.

So, this now returns the original node, etc. as part of the result object. We might return more things in the future.

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
eddeee888: Put a FIXME here: https://github.com/dotansimha/graphql-code-generator/pull/10287/files#r2086739962

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -433,7 +437,14 @@ function getDirectivesByName(
     astNode = node;
   }
 
-  return astNode?.directives?.filter(d => d.name.value === name) || [];
+  return (
+    astNode?.directives?.filter(d => {
+      // A ObjectTypeDefinitionNode's directive looks like `{ kind: 'Directive', name: 'external', arguments: [] }`
+      // However, other directives looks like `{ kind: 'Directive', name: { kind: 'Name', value: 'external' }, arguments: [] }`
+      // Therefore, we need to check for both `d.name.value` and d.name
+      return d.name.value === name || (d.name as unknown as string) === name;
```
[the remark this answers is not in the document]
eddeee888: Related comment: https://github.com/dotansimha/graphql-code-generator/pull/10287/files#r2086739962

dotansimha (approved):

packages/utils/plugins-helpers/src/federation.ts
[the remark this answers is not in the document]
dotansimha: This works! We discussed this separately, it's a codegen issue, in one of the internal components that does `visit` and converts `NamedNode` into a `string`. The TS typechecker can't really know about this.

Issue #10299, opened 2025-02-19
Issue with type generation and fragment masking in codegen


ashutoshlodhi09:
### Which packages are impacted by your issue?

Not able to use apollo client masking [feature](https://www.apollographql.com/blog/more-resilient-code-with-data-masking-in-apollo-client-3-12) 

### Describe the bug

**Description**
I'm trying to implement an incremental adoption approach for apollo data masking :

**PostDetailsFragment** is **masked**, and data is accessed using `useFragment` .
**ActivityDetailsFragment** is **unmasked**, meaning data is accessed directly from prop.

Component code 
```tsx

/**
 * Root Query to fetch data
 * Post details fragment with masking enabled
 * Activity details fragment with masking disabled
 */

const GET_DATA_QUREY = gql`
  query GetDetails {
    posts {
      id
      ...PostDetailsFragment
    }
    activities {
      ...ActivityDetailsFragment @unmask(mode: "migrate")
    }
  }
  ${POST_DETAILS_FRAGMENT}
  ${Activity_DETAILS_FRAGMENT}
`;

/**
 * Component to display Post and Activity details
 * Fetching data using useQuery hook
 */
export default function PostsAndActivities() {
  const { data } = useQuery(GET_DATA_QUREY);
  return (
    <div>
      {data.posts.map((post) => (
        <>
        <PostDetails key={post.id} post={post} />
        </>
      ))}
       {data.activities.map((activity) => (
        <>
        <ActivityDetails  key={activity.id} activity={activity} />
        </>
      ))}
    </div>
  );
}

/**
 * Post component fragment
 */
export const POST_DETAILS_FRAGMENT = gql`
  fragment PostDetailsFragment on Post {
    title
    shortDescription
    publishedAt
  }
`;
/**
 * Post component with props post as FragmentType typeOf POST_DETAILS_FRAGMENT for type saftey
 */
function PostDetails({ post } : { post: FragmentType<typeof  POST_DETAILS_FRAGMENT> } ) {
  const { data, complete } = useFragment({
    fragment: POST_DETAILS_FRAGMENT,
    from: post,
  });
  return (
    <section>
      <h1>{data.title}</h1>
      <p>{data.shortDescription}</p>
    </section>
  );
}

/**
 * Activity component fragment
 */
export const Activity_DETAILS_FRAGMENT = gql`
  fragment ActivityDetailsFragment on Activity {
    id
    title
    shortDescription
    publishedAt
  }
`;
/**
[...]

PR #10302, opened 2025-02-20
[operations] CODEGEN-784 - Fix Apollo @unmask fragmentRefs

eddeee888:
## Description

When using Apollo masking features, an `@unmask` fragment is the same as inlined i.e.
- should not have `$fragmentRefs`
- selection node fields are inlined

Reference: https://github.com/dotansimha/graphql-code-generator/issues/10299

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -548,6 +548,7 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
     selectionNodes = [...selectionNodes];
     let inlineFragmentConditional = false;
     for (const selectionNode of selectionNodes) {
+      // 1. Handle Field or Directtives selection nodes
       if ('kind' in selectionNode) {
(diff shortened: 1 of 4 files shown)


packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -548,6 +548,7 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
     selectionNodes = [...selectionNodes];
     let inlineFragmentConditional = false;
     for (const selectionNode of selectionNodes) {
+      // 1. Handle Field or Directtives selection nodes
```
jerelmiller, line 551:
```suggestion
      // 1. Handle Field or Directives selection nodes
```

Noticed a small typo 🙂

jerelmiller (approved):
I know my approval doesn't mean much here, but this looks right to me. Thanks so much for helping out! Much simpler than I realized this would be 🙌

dotansimha (approved):

PR #10159, opened 2024-10-01
feat(plugin/typescript-resolvers): Add allowSemanticNonNull option

Quramy:
## Description

Related #10151 .

Added a new option for plugin/typescript-resolvers, `allowSemanticNonNull`.

Several GraphQL frameworks work with `@semanticNonNull` schema directive:

- https://relay.dev/docs/guides/semantic-nullability/#proposed-solution
- https://www.apollographql.com/docs/kotlin/advanced/nullability/#semanticnonnull

If `allowSemanticNonNull` option is set to `true`, field resolver annotated by `@semanticNonNull` gets prohibited to return nullish.

For example:

```gql
type User {
  name: String @semanticNonNull
}
```

The above schema and `allowSemanticNonNull: true` generates the following resolver type:

```ts
export type UserResolvers<ContextType = BaseContext, ParentType extends ResolversParentTypes['User'] = ResolversParentTypes['User']> = {
  name?: Resolver<ResolversTypes['String'], ParentType, ContextType>;
  __isTypeOf?: IsTypeOfResolverFn<ParentType, ContextType>;
};
```

```ts
import type { UserResolvers } from "./_gql_gen_generated_/"

export const User = {
  name:  () => {
    // @ts-expect-error
    return null;
  }
} satisfies UserResolvers
```

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Screenshots/Sandbox (if appropriate/relevant):

Adding links to sandbox or providing screenshots can help us understand more about this PR and take action on it as appropriate

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Unit Testing

**Test Environment**:

- OS: macos Sonoma
- NodeJS: v20.14.0

## Checklist:

- [x] I have followed the [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

Issue #10151, opened 2024-09-25
Support `@semanticNonNull`


Quramy:
### Is your feature request related to a problem? Please describe.

Relay v18.0.0 introduces a new feature "Semantic Non Null", https://relay.dev/docs/guides/semantic-nullability/

> The proposal works by allowing the schema to specify a new type of nullability of "null only on error". If a client sees this type, and the client has some strategy for handling field errors out-of-band, it may treat the field that is exposed to user code as non-nullable.
>
> The full spec change will likely require adding additional syntax to GraphQL's schema definition language, but in the meantime, various GraphQL servers and clients have collaborated on a temporary directive [@semanticNonNull](https://specs.apollo.dev/nullability/v0.2/) that can be used to experiment with this idea.

The `@semanticNonNull` schema directive means "this field is not null unless error" .
So, field resolver annotated this directive must not return null and I want types generated by typescript-resolver to check this constraint. 

For example:

```gql
type User {
  name: String @semanticNonNull
}
```

```ts
import type { UserResolvers } from "./_gql_gen_generated_/"

export const User = {
  name:  () => {
    // @ts-expect-error
    return null;
  }
} satisfies UserResolvers
```

### Describe the solution you'd like

I want a new typescript-resolver plugin option to enable `@semanticNonNull` . 

```ts
 import type { CodegenConfig } from '@graphql-codegen/cli';
 
 const config: CodegenConfig = {
   // ...
   generates: {
     'path/to/file.ts': {
       plugins: ['typescript', 'typescript-resolvers'],
       config: {
         allowSemanticNonNullType: true // default false
       },
     },
   },
 };
 export default config;
```

If `allowSemanticNonNullType` set `true`, codegen generates resolver types like the above `UserResolvers` .

erikwrede:
+1 on this. 
Generally, I see two areas of support on this: Server and client. 

The server part is already well-described by @Quramy. For the client part, I have the following comments:
Ideally, the `typed-document-ast` would also support @catch directives / `@throwOnFieldError` fragments to infer the correct nullability condition. 

See here for a quick overview on the topic: 

- How Apollo Kotlin implements it: https://www.apollographql.com/docs/kotlin/advanced/nullability#semanticnonnull
- Quick overview of the concept & `@throwOnFieldError` https://graphql.org/conf/2024/schedule/8daaf10ac70360a7fade149a54538bf9/
- Nullability Discussion: https://github.com/graphql/nullability-wg/discussions/58

[...]

PR #10315, opened 2025-03-16
[resolvers] CODEGEN-500 Support semantic non null

eddeee888:
## Description

Based on https://github.com/dotansimha/graphql-code-generator/pull/10159
Related to https://github.com/dotansimha/graphql-code-generator/issues/10151

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -1449,8 +1449,6 @@ export class BaseResolversVisitor<
 
     return (parentName, avoidResolverOptionals) => {
       const original: FieldDefinitionNode = parent[key];
-      const baseType = getBaseTypeNode(original.type);
-      const realType = baseType.name.value;
       const parentType = this.schema.getType(parentName);
 
       if (this._federation.skipField({ fieldNode: original, parentType })) {
@@ -1458,11 +1456,6 @@ export class BaseResolversVisitor<
       }
 
       const contextType = this.getContextType(parentName, node);
-
-      const typeToUse = this.getTypeToUse(realType);
-      const mappedType = this._variablesTransformer.wrapAstTypeWithModifiers(typeToUse, original.type);
-      const subscriptionType = this._schema.getSubscriptionType();
-      const isSubscriptionType = subscriptionType && subscriptionType.name === parentName;
 
       let argsType = hasArguments
         ? this.convertName(
@@ -1499,15 +1492,34 @@ export class BaseResolversVisitor<
         parentType,
         parentTypeSignature: this.getParentTypeForSignature(node),
       });
-      const mappedTypeKey = isSubscriptionType ? `${mappedType}, "${node.name}"` : mappedType;
 
-      const directiveMappings =
-        node.directives
-          ?.map(directive => this._directiveResolverMappings[directive.name as any])
-          .filter(Boolean)
-          .reverse() ?? [];
+      const { mappedTypeKey, resolverType } = ((): { mappedTypeKey: string; resolverType: string } => {
+        const baseType = getBaseTypeNode(original.type);
+        const realType = baseType.name.value;
+        const typeToUse = this.getTypeToUse(realType);
+        const mappedType = this._variablesTransformer.wrapAstTypeWithModifiers(typeToUse, original.type);
 
-      const resolverType = isSubscriptionType ? 'SubscriptionResolver' : directiveMappings[0] ?? 'Resolver';
+        const subscriptionType = this._schema.getSubscriptionType();
+        const isSubscriptionType = subscriptionType && subscriptionType.name === parentName;
+
+        if (isSubscriptionType) {
+          return {
+            mappedTypeKey: `${mappedType}, "${node.name}"`,
+            resolverType: 'SubscriptionResolver',
+          };
+        }
+
+        const directiveMappings =
+          node.directives
+            ?.map(directive => this._directiveResolverMappings[directive.name as any])
+            .filter(Boolean)
+            .reverse() ?? [];
+
+        return {
+          mappedTypeKey: mappedType,
+          resolverType: directiveMappings[0] ?? 'Resolver',
+        };
+      })();
 
       const signature: {
         name: string;
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -1628,7 +1628,7 @@ export class BaseResolversVisitor<
   }
 
   /**
-   * Function that takes in a TypeScript type and apply extra transformation to it based on config options
+   * Function to apply extra transformation to a FieldDefinitionNode's TypeScript type based on config options
    * e.g. takes `Maybe<ResolversTypes['String']>`, and returns the type without the `Maybe<>` wrapper.
    */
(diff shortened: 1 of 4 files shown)


eddeee888:
Hi @alf, could you please try the following alpha versions:

```
@graphql-codegen/typescript-resolvers@4.5.0-alpha-20250316110831-3f757aba5365f95cb2ea64e6f2e73cd4ddd534a2
```

dotansimha (approved):

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1499,15 +1523,41 @@ export class BaseResolversVisitor<
...
+            ?.map(directive => this._directiveResolverMappings[directive.name as any])
+            .filter(Boolean)
+            .reverse() ?? [];
 
-      const resolverType = isSubscriptionType ? 'SubscriptionResolver' : directiveMappings[0] ?? 'Resolver';
+        return {
+          mappedTypeKey: this.modifyFieldDefinitionNodeTransformedType({ node: original, originalType: mappedType }),
+          resolverType: directiveMappings[0] ?? 'Resolver',
+        };
+      })();
```
eddeee888, line 1560: I refactored this bit to bring related variables together, in the same scope - so it's easier to read

martinbonnin:
> if the type is a list, we make the list non-nullable, each item in the list is not affected

FWIW, the  [`levels` parameter](https://specs.apollo.dev/nullability/v0.4/#@semanticNonNull) allows changing the nullability of items:

```graphql
type Query {
  a: [String] @semanticNonNull                # List<String?>
  b: [String] @semanticNonNull(levels: [1])   # List<String>?
  c: [String] @semanticNonNull(levels: [0,1]) # List<String>
}
```

BenjieGillam:
Using TypeScript syntax:

```
type Query {
  a: [String] @semanticNonNull                # Array<String | null>
  b: [String] @semanticNonNull(levels: [1])   # Array<String> | null
  c: [String] @semanticNonNull(levels: [0,1]) # Array<String>
}
```

You could also use `semanticToNullable` or `semanticToStrict` from graphql-sock to convert the schema appropriately, then you'd only need to change a line or three in codegen itself:

https://github.com/graphile/graphql-sock/blob/fdffa2e117fc1ca0fcb5894c4d8d37372b7ccd0f/src/cli.ts#L76-L82

BenjieGillam:
It's also important to note that you should _not_ use a non-nullable type for semantic non-nullable _unless_ the client is an error handling client; i.e. the client must either:

1. discard entire requests if any errors occur,
2. throw when an errored field is read (throw-on-error - see [graphql-toe](https://github.com/graphile/graphql-toe))
3. use `@catch` or similar directives to influence how errors are handled (but if this is the case, you wouldn't use graphql-codegen since the behavior would change on a per-field basis based on client configuration).

To put it another way: users must opt-in to these positions being non-nullable, otherwise the directive should be ignored and the default nullable should be carried through.

benjie:
I meant that

```graphql
type Query {
  bar: String @semanticNonNull
}
```

With query

```graphql
query Q {
  bar
}
```

Should come out as:

```ts
type Q = {
  bar: string | null
}
```

unless the user _opts in_ to having it come out as non-nullable (e.g. via a configuration parameter, choice of preset, etc), at which point you'd generate:

```ts
type Q = {
  bar: string
}
```

This is because query `Q` would always return a `bar` field, and it can be null if an error occurs, e.g.:

```json
{
  "data": { "bar": null },
  "errors": [{"message": "Blah blah blah", "path": ["bar"]}]
}
```

e.g. strictly the field is still `string | null`. It's only when the client prevents a user from _reading_ that null where you know it can be `string` alone - e.g. if you have throw-on-error then you can never read that `null` (since an error will be returned if you attempt it) and thus the only value you can read from it is `string`.

If you're doing pure type generation then you have no control over this. If you're doing code generation too, then you can - you can choose to throw the errors either when the user reads from the field that's broken (e.g. using a "getter" as we do in [`graphql-toe`](https://github.com/graphile/graphql-toe)) or, _iff_ you have fragment isolation (e.g. via masking), you could do it when the user uses the fragment if the non-masked part of the fragment contain an error. In this latter case though, you would have to ensure that there's no way around this, otherwise the user could still read a null from this position.

Personally I'd recommend using graphql-toe, it ensures that no matter how you read the data (once it has gone through TOE) whenever you read an errored field, an error will be thrown. I also put quite a bit of effort into ensuring it was as efficient as possible, only creating new types/adding getters where it had to, and carrying through the source data otherwise.

martinbonnin:
> `bar: String @semanticNonNull` should come out as:

```graphql
type Q = {
  bar: string | null
}
```

> unless the user opts in to having it come out as non-nullable

Disagree on this one. If someone takes the time to annotate their schema with `@semanticNonNull`, I would expect codegen to generate the field as non-null by default. It's on the user to do the correct truntime thing (by using `graphql-toe` or other). 

If the client is not an error handling client, I'm fine with an option to ignore `@semanticNonNull` altogether but if we want to raise awareness around this, I'd rather have graphql-code-generate default to the new "error-handling" mode.

benjie:
Strong disagree; the team producing the schema (e.g. adding `@semanticNonNull` to it) and the team consuming the schema (e.g. the client frontend team) are often different. Over time we'd like to move most people over to using error handling clients, but right now the vast majority of clients are _not_ "error handling clients" (i.e. they don't have "throw on error" or similar behavior preventing the user from reading an "error null").

If we were to generate non-nulls from _semantic_ non-null right now, then the types would be a lie - e.g. someone using a semantic non-null marked schema with Apollo Client would be able to read `null` from positions that the types say are non-nullable, which could produce unexpected `TypeError`, or worse: subtle rendering bugs.

If it's the client team adding `@semanticNonNull` to their schema, then they should already know to configure codegen into "error handling" mode, it's just one extra line of config for them.

benjie:
@eddeee888 I've just pushed out a new version of `graphql-sock` with better documentation, including documenting the library mode. I think you'd just need to add a line wherever you build/source the GraphQL schema:

```ts
import { semanticToStrict, semanticToNullable } from 'graphql-sock';

//...

const schema = codegenOptions.errorHandlingClient ? semanticToStrict(sourceSchema) : semanticToNullable(sourceSchema);
```

Note that graphql-sock supports both the `@semanticNonNull` directive _and_ the `GraphQLSemanticNonNull` schema type, which means it supports [all proposed solutions](https://github.com/graphql/graphql-wg/blob/main/rfcs/SemanticNullability.md#-possible-solutions) to the semantic nullability problem; though you'd need to be running the relevant version of graphql.js with the syntax support for the solution you're using (except solutions 5 and 6 which will always work). E.g. to support the syntax for solution 1, you'd need to be running `graphql@canary-pr-4192`.

benjie:
Good catch; I forgot that was why I only had CLI docs and not library docs! Should be feasible for it to `import * as graphql from 'graphql'` and use GraphQLSemanticNonNull iff it exists - let me get back to you on that :+1:

benjie:
@eddeee888 I've released `graphql-sock@1.0.0` which is now tested against v15, v16, v17 and `graphql@canary-pr-4192` in CI; it also splits the library code from the CLI code so there's no imports of anything but `graphql` in library mode, so should be safe to bundle and run on the web too.

Oh, also it now exports a `convertFieldConfig` method to just convert an individual `GraphQLFieldConfig` rather than the entire schema :+1:

PR #10323, opened 2025-03-21
[operations][client-preset] CODEGEN-500 - Support semantic non-null for clients

eddeee888:
## Description

This PR supports `@semanticNonNull` support for client use cases:

- [x] `typescript-operations`
- [x] `client-preset`

Introducing `semanticNonNull` config option. This will be the target option for all semantic non null options in the future.
For now, it includes an option called `errorHandlingClient` so users can declare whether they are using an error-handling client or not:
- If an error handling client is used, all semantic non-null fields are treated as non-null because accessing these values will result in a thrown error 
- If an error handling client is _not_ used, all semantic non-null fields are treated as null, which means when a value is null, it could be "data null" or "error null"

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/typescript/operations/src/config.ts
+++ b/packages/plugins/typescript/operations/src/config.ts
@@ -294,7 +294,7 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
   allowUndefinedQueryVariables?: boolean;
 
   /**
-   * @description Options related to `@semanticNonNull` directive
+   * @description Options related to handling nullability
    * @exampleMarkdown
    * ## `errorHandlingClient`
    * When using error handling clients, a semantic non-nullable field can never be `null`.
@@ -317,7 +317,7 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
    *     'path/to/file.ts': {
    *       plugins: ['typescript', 'typescript-operations'],
    *       config: {
-   *         semanticNonNull: {
+   *         nullability: {
    *           errorHandlingClient: true
    *         }
    *       },
@@ -327,7 +327,7 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
    * export default config;
    * ```
    */
-  semanticNonNull?: {
+  nullability?: {
     errorHandlingClient: boolean;
   };
 }
--- a/packages/plugins/typescript/operations/src/config.ts
+++ b/packages/plugins/typescript/operations/src/config.ts
@@ -292,4 +292,42 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
    */
 
   allowUndefinedQueryVariables?: boolean;
+
+  /**
+   * @description Options related to `@semanticNonNull` directive
+   * @exampleMarkdown
+   * ## `errorHandlingClient`
+   * When using error handling clients, a semantic non-nullable field can never be `null`.
+   * If a field is read and its value is `null`, there must be a respective error. The error handling client will throw in this case, so the `null` value is never read.
+   *
+   * To enable this option, install `graphql-sock` peer dependency:
+   *
+   * ```sh npm2yarn
+   * npm install -D graphql-sock
+   * ```
+   *
+   * Now, you can enable support for error handling clients:
+   *
+   * ```ts filename="codegen.ts"
+   * import type { CodegenConfig } from '@graphql-codegen/cli';
+   *
+   * const config: CodegenConfig = {
+   *   // ...
+   *   generates: {
+   *     'path/to/file.ts': {
+   *       plugins: ['typescript', 'typescript-operations'],
+   *       config: {
+   *         semanticNonNull: {
+   *           errorHandlingClient: true
+   *         }
+   *       },
+   *     },
+   *   },
+   * };
+   * export default config;
+   * ```
+   */
+  semanticNonNull?: {
+    errorHandlingClient: boolean;
+  };
 }
--- a/packages/plugins/typescript/operations/src/index.ts
+++ b/packages/plugins/typescript/operations/src/index.ts
@@ -7,14 +7,14 @@ import { TypeScriptDocumentsVisitor } from './visitor.js';
 export { TypeScriptDocumentsPluginConfig } from './config.js';
 
 export const plugin: PluginFunction<TypeScriptDocumentsPluginConfig, Types.ComplexPluginOutput> = async (
-  schema: GraphQLSchema,
+  inputSchema: GraphQLSchema,
   rawDocuments: Types.DocumentFile[],
   config: TypeScriptDocumentsPluginConfig
 ) => {
-  const transformedSchema = config.semanticNonNull?.errorHandlingClient ? await semanticToStrict(schema) : schema;
+  const schema = config.nullability?.errorHandlingClient ? await semanticToStrict(inputSchema) : inputSchema;
 
   const documents = config.flattenGeneratedTypes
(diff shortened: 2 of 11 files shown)


packages/plugins/typescript/operations/src/config.ts
```diff
@@ -292,4 +292,42 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
...
+   *       },
+   *     },
+   *   },
+   * };
+   * export default config;
+   * ```
+   */
+  semanticNonNull?: {
+    errorHandlingClient: boolean;
+  };
```
eddeee888, line 332: I'm thinking this can be extended to handle other related `semanticNonNull` directives e.g. `@catch`

packages/presets/client/src/index.ts
```diff
@@ -114,6 +114,10 @@ export const preset: Types.OutputPreset<ClientPresetConfig> = {
       );
     }
     const isPersistedOperations = !!options.presetConfig?.persistedDocuments;
+    if (options.config.semanticNonNull?.errorHandlingClient) {
+      options.schemaAst = await semanticToStrict(options.schemaAst!);
+      options.schema = options.schemaAst as any; // FIXME: `schemaAst` _seems_ to be able to used as `schema`, but not sure if it has any side effects?
```
eddeee888, line 119: I'm a bit unsure about this:
- `schema` is a `DocumentNode`
- `shemaAst` is a `GraphQLSchema`
- We can forcefully override `schema` with `schemaAst` ... and it works 🤔 

Maybe when loading `schema` in each `generates` block, we do a check? That's my best guess though, and if anyone has more context, please let me know!

eddeee888:
Hi @benjie @martinbonnin, this is the client handling of `semanticNonNull`, I've left how it works in the PR description

packages/plugins/typescript/operations/src/config.ts
```diff
@@ -292,4 +292,42 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
...
+   *           errorHandlingClient: true
+   *         }
+   *       },
+   *     },
+   *   },
+   * };
+   * export default config;
+   * ```
+   */
+  semanticNonNull?: {
```
benjie, line 330: Perhaps, and this is just a suggestion, rename this to "semantic nullability" as a more generic term. If we were to accept a solution that used a document directive to enable the syntax, this is the likely name that we would use for that syntax, and it doesn't tie the solution to one specific implementation. (GraphQL-SOCK can be updated to work with other solutions without any effort on your part.)

```suggestion
  semanticNullability?: {
```

packages/plugins/typescript/operations/src/index.ts
```diff
@@ -6,13 +6,15 @@ import { TypeScriptDocumentsVisitor } from './visitor.js';
 
 export { TypeScriptDocumentsPluginConfig } from './config.js';
 
-export const plugin: PluginFunction<TypeScriptDocumentsPluginConfig, Types.ComplexPluginOutput> = (
+export const plugin: PluginFunction<TypeScriptDocumentsPluginConfig, Types.ComplexPluginOutput> = async (
   schema: GraphQLSchema,
   rawDocuments: Types.DocumentFile[],
   config: TypeScriptDocumentsPluginConfig
 ) => {
+  const transformedSchema = config.semanticNonNull?.errorHandlingClient ? await semanticToStrict(schema) : schema;
```
benjie, line 14: I personally would rename the `schema` variable to `inputSchema` so that `schema`, the variable you'd use by default, is the correct variable to use:

```suggestion
  inputSchema: GraphQLSchema,
  rawDocuments: Types.DocumentFile[],
  config: TypeScriptDocumentsPluginConfig
) => {
  const schema = config.semanticNonNull?.errorHandlingClient ? await semanticToStrict(inputSchema) : inputSchema;
```

(This would also mean no other changes were needed in the function.)

packages/plugins/typescript/operations/src/index.ts
```diff
@@ -64,3 +66,14 @@ export const plugin: PluginFunction<TypeScriptDocumentsPluginConfig, Types.Compl
...
 
 export { TypeScriptDocumentsVisitor };
+
+const semanticToStrict = async (schema: GraphQLSchema): Promise<GraphQLSchema> => {
+  try {
+    const sock = await import('graphql-sock');
+    return sock.semanticToStrict(schema);
+  } catch {
+    throw new Error(
+      "To use the `customDirective.semanticNonNull` option, you must install the 'graphql-sock' package."
```
benjie, line 76:
```suggestion
      "To use the `semanticNonNull.errorHandlingClient` option, you must install the 'graphql-sock' package."
```

packages/presets/client/src/index.ts
benjie: Maybe what you want here is this?

```suggestion
      options.schema = parse(printSchema(options.schemaAst));
```

Just guessing from the types you mention, I'm not familiar with this codebase.

website/src/pages/plugins/presets/preset-client.mdx
```diff
@@ -58,6 +58,7 @@ The `client` preset allows the following `config` options:
 - [`nonOptionalTypename`](/plugins/typescript/typescript#nonoptionaltypename): Automatically adds `__typename` field to the generated types, even when they are not specified in the selection set, and makes it non-optional.
 - [`avoidOptionals`](/plugins/typescript/typescript#avoidoptionals): This will cause the generator to avoid using TypeScript optionals (`?`) on types.
 - [`documentMode`](#documentmode): Allows you to control how the documents are generated.
+- [`semanticNonNull`](/plugins/typescript/typescript-operations#semanticnonnull): Handles `@semanticNonNull` directive based on client-side requirements.
```
benjie, line 61:
```suggestion
- [`semanticNullability`](/plugins/typescript/typescript-operations#semanticnullability): Indicate the client capabilities to get stronger types with [semantic nullability](https://github.com/graphql/graphql-wg/blob/main/rfcs/SemanticNullability.md)-enabled schemas.
```

benjie (changes requested):
LGTM! I've left a few suggestions, but really excited to see this get merged! :rocket:

martinbonnin (commented):
Looks great!! A couple documentation/terminology comments but otherwise looking forward to play with this 🚀 !!

website/src/pages/plugins/presets/preset-client.mdx
benjie: By using `graphql-sock` this supports not only `@semanticNonNull` the directive but all the proposed solutions (assuming GraphQL.js support) to the semantic nullability RFC. Theoretically this should already work just fine with the `*` syntax when using `graphql@canary-pr-4192` for example (I've not tested it though). If we were to come up with an alternative solution (e.g. `GraphQLStrictNonNull` to represent the old "kills parent on exception" style of non-null) then "semantic nullability" would still cover this solution, but "semanticNonNull" would not.

benjie: I'd support renaming the option to simply `nullability` though, as you proposed separately.

benjie (approved):

eddeee888:
Thanks @benjie @martinbonnin for your great suggestions! 

I've manually applied the suggestions because changes need to happen across multiple files:
- Renamed the config to `nullability.errorHandlingClient`: `nullability` proposal makes a lot of sense here and I like the generic reasoning of it
- Minor code logic based on suggestions

eddeee888:
Merging this so we can try out the alpha versions! We can make changes before the final release if needed 🙂

Issue #9740, opened 2023-10-30
stop truncating error file path


n1ru4l:
### Is your feature request related to a problem? Please describe.

It is hard to get the full path of the file that caused graphql code generator to error in case the terminal is to small

When the terminal is wide enough:

<img width="836" alt="image" src="https://github.com/dotansimha/graphql-code-generator/assets/14338007/1f97e820-613f-4c16-ba0b-56365d56d9b8">

```
at /Users/laurinquast/Projects/graphql-hive-3/packages/libraries/client/src/internal/operations-store.ts:3:5
```

If the terminal is not wide enough:

<img width="671" alt="image" src="https://github.com/dotansimha/graphql-code-generator/assets/14338007/05710fb2-5cf7-4d59-9a71-5cc2634997bb">

```
at /Users/laurinquast/Projects/graphql-hive-3/packages/libraries/client/src/interna…
```

### Describe the solution you'd like

Always print the full path even if the terminal is not wide enough.

U-4-E-A:
If you config file: -

`verbose: true`

simhnna:
We get the truncated output in IDE terminals although they are wide enough.
A workaround I've found is to redirect stdout to a file. If you do that the complete error is printed to stderr.

The command should really not truncate errors.

eddeee888:
Hello 👋 
The latest version `@graphql-codegen/cli@5.0.6` has been [released here](https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1746539042593) with a fix to stop truncating.

However, if your terminal is too narrow, the text will break into multiple lines. This is the behaviour of the underlying library that renders text in the terminal. However, in my mind this should still be an improvement over truncating.

Thanks for your suggestions! I'll close this for now, let me know if it doesn't fix your issue.

Issue #9172, opened 2023-03-15
No error output whatsoever when parsed GraphQL documents have syntax errors


mogelbrod:
### Which packages are impacted by your issue?

@graphql-codegen/cli, @graphql-codegen/core, @graphql-codegen/gql-tag-operations-preset

### Describe the bug

Likely related issues:
- https://github.com/dotansimha/graphql-code-generator/issues/6749
- https://github.com/dotansimha/graphql-code-generator/issues/8841

Running `graphql-codegen` in a project containing **invalid GraphQL queries in typescript files** wrapped with `gql()` has some really strange and inconsistent behaviour with regards to error reporting:
1. If only src/index.ts **OR** src/nested/directory/file.ts contains an invalid query it doesn't get reported
   ```
	$ graphql-codegen
	✔ Parse Configuration
	✔ Parse Configuration
	✔ Generate outputs
   ```
2. If both src/index.ts **AND** src/nested/directory/file.ts contain invalid queries both errors are reported, but with no information regarding which file, line, or query
   ```
	$ graphql-codegen
	✔ Parse Configuration
	⚠ Generate outputs
	  ❯ Generate to src/graphql/
	    ✔ Load GraphQL schemas
	    ✖ Failed to find any GraphQL type definitions in: src/**/*.{graphql,ts,tsx,js,jsx};
	      - Syntax Error: Expected Name, found "}".
	      - Syntax Error: Expected Name, found "}".
	    ◼ Generate
   ```

Running the `graphql-codegen` command with `--debug` and/or `--verbose` flags changes nothing, but prepending the command with `DEBUG=1` does result in an error being logged in both cases, albeit with a bunch of noise surrounding it:

<details>
<summary><code>DEBUG=1 graphql-codegen --verbose</code> output</summary>

```   
[STARTED] Parse Configuration
[SUCCESS] Parse Configuration
[STARTED] Generate outputs
[STARTED] Generate to src/graphql/
[STARTED] Load GraphQL schemas
@graphql-tools/load: normalizePointers [object Object]: 0.058ms
@graphql-tools/load: normalizePointers: 0.172ms
@graphql-tools/load: collectDocumentString https://rickandmortyapi.com/graphql: 0.283ms
@graphql-tools/load: collectSources https://rickandmortyapi.com/graphql: 0.408ms
@graphql-tools/load: loadFile https://rickandmortyapi.com/graphql: 213.892ms
[...]

PR #10333, opened 2025-04-02
CODEGEN-334 - Improve error logging

eddeee888:
## Description

Related https://github.com/dotansimha/graphql-code-generator/issues/9172, https://github.com/dotansimha/graphql-code-generator/issues/9740

This PR aims to address two things:

- Error message getting truncated by Listr: This is fixed by using `wrap` renderer option. This is still not perfect because when the error wraps, jump-to-definition feature may not recognise the full path. However, this is an improvement.
- GraphQL syntax error are displayed with native GraphQLError message: i.e. file path and error location 

Note: this does not fix the issue where `ignoreNoDocument` may swallow errors. That will be done in a follow-up PR.

## Type of change


packages/graphql-codegen-cli/src/load.ts
```diff
@@ -52,22 +52,17 @@ export async function loadSchema(
...
+        `Failed to load schema from ${Object.keys(schemaPointers).join(',')}:`,
+        printError(e),
+        '\nGraphQL Code Generator supports:',
+        '\n- ES Modules and CommonJS exports (export as default or named export "schema")',
+        '- Introspection JSON File',
+        '- URL of GraphQL endpoint',
+        '- Multiple files with type definitions (glob expression)',
+        '- String in config file',
+        '\nTry to use one of above options and run codegen again.\n',
+      ].join('\n')
```
eddeee888, line 65: I find joining strings like this is easier to read and formatted better in the terminal

packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -254,4 +254,103 @@ describe('generate-and-save', () => {
...
+        );
+        // We can only use partial file path to the error file, because the error contains absolute path on the host machine
+        expect(consoleErrorMock.mock.calls[0][0]).toBeSimilarStringTo(
+          '/tests/test-files/schema-dir/error-schema.graphql:2:15'
+        );
+        expect(consoleErrorMock.mock.calls[0][0]).toBeSimilarStringTo(`
+          [FAILED] 1 | type Query {
+          [FAILED] 2 |   foo: String!!
+          [FAILED]   |               ^
+          [FAILED] 3 | }
```
eddeee888, line 311: GraphQLError Formatting looks correct in `--verbose` scenario, if we use the Listr default renderer, it may look like this in the terminal:

Screenshot 2025-04-04 at 12 17 24 am

It looks like the pad left doesn't work on the line with the pointer `^`. This is a bit annoying but we can't easily fix this unless we fix `graphql-js`'s error logging (high effort, not much impact right now IMO)

packages/graphql-codegen-cli/src/load.ts
dotansimha: YES!

packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
dotansimha: nice!!!!!

dotansimha (approved):

PR #10338, opened 2025-04-13
CODEGEN-334 - Fix ignoreNoDocuments swallowing all errors, not just no document errors

eddeee888:
## Description

Related #9172

This PR fixes the issue where `ignoreNoDocuments` flag catches and swallows all errors such as syntax errors, and not just when no documents can be found. This uses feature implemented in https://github.com/ardatan/graphql-tools/pull/7093

## Type of change

## How Has This Been Tested?

- [x] Unit test


packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -77,22 +77,6 @@ describe('generate-and-save', () => {
...
-      errorsOnly: true,
-      overwrite: true,
-      profile: true,
-      require: [],
-      silent: false,
-      watch: false,
-    });
-
-    await generate(config, false);
-  });
```
eddeee888, line 94: This is moved to the section handling errors below

packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -255,11 +239,11 @@ describe('generate-and-save', () => {
     expect(writeSpy).toHaveBeenCalled();
   });
 
-  describe('Syntax errors when loading pointers', () => {
+  describe('Errors when loading pointers', () => {
```
eddeee888, line 242: I'm changing this section to handle all errors, not just syntax errors

packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -324,6 +309,7 @@ describe('generate-and-save', () => {
     });
 
     test('Document syntax error - should print native GraphQLError', async () => {
+      expect.assertions(4);
```
eddeee888, line 312: `expect.assertions(4);` is needed here to avoid the scenario where `await generate` somehow passes, which means it'd run just the `try` block. `try` block doesn't have an expect so it'd pass the test.

packages/graphql-codegen-cli/tests/test-files/graphql.config.json
eddeee888, line 1: This is turned into `graphql.config.no-doc.js` and `graphql.config.no-doc-ignored.js` to be more specific.

dotansimha (approved):

eddeee888:
In the end, I couldn't pin `graphql-config` to later version (>5.1.4) to work for this PR because of the current Node 16 requirement/test. 😢 
These later versions have the same `@graphql-tools/load` version `^8.1.0`, which helps making lock files cleaner. However, the `graphql-config` version doesn't support Node 16, which Codegen is still supporting. 
This is a nice-to-have though. So, we can just merge this PR as-is. 

Note: 
- We will soon release a breaking change to drop Node 16 and 18 support (Both EOL by the time this PR is merged)
- Consumers that are on Node 16 (EOL in 11 Sep 2023) and installs the latest `graphql-config` _may_ experience runtime errors. But in my testing, I can't install latest Codegen with Node 16 because of a dep requirement, so I feel this is a very rare scenario, and we can suggest consumers to update to at least Node 18.

PR #10366, opened 2025-06-04
[resolvers][federation] Fix federation @requires type

eddeee888:
## Description
- Creates `FederationReferenceTypes` to declare the type of references of each object. This is used by all generated types, and can be used in mappers. [More details in comment](https://github.com/dotansimha/graphql-code-generator/pull/10366#discussion_r2135735298)
- Correctly finds all combinations of selection sets declared by `@requires` in different fields and put them in `FederationReferenceTypes`. [More details in comment](https://github.com/dotansimha/graphql-code-generator/pull/10366#discussion_r2142881641)

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
@@ -818,12 +818,8 @@ export class BaseResolversVisitor<
       shouldInclude: namedType => !isEnumType(namedType),
       onNotMappedObjectType: ({ typeName, initialType }) => {
         let result = initialType;
-        const federationReferenceTypes = this._federation.printReferenceSelectionSets({
-          typeName,
-          baseFederationType: `${this.convertName('FederationTypes')}['${typeName}']`,
-        });
-        if (federationReferenceTypes) {
-          result += ` | ${federationReferenceTypes}`;
+        if (this._federation.getMeta()[typeName]?.referenceSelectionSetsString) {
+          result += ` | ${this.convertName('FederationReferenceTypes')}['${typeName}']`;
         }
         return result;
       },
@@ -1306,6 +1302,33 @@ export class BaseResolversVisitor<
       ).string;
   }
 
+  public buildFederationReferenceTypes(): string {
+    const federationMeta = this._federation.getMeta();
+
+    if (Object.keys(federationMeta).length === 0) {
+      return '';
+    }
+
+    const declarationKind = 'type';
+    return new DeclarationBlock(this._declarationBlockConfig)
+      .export()
+      .asKind(declarationKind)
+      .withName(this.convertName('FederationReferenceTypes'))
+      .withComment('Mapping of federation reference types')
+      .withBlock(
+        Object.entries(federationMeta)
+          .map(([typeName, { referenceSelectionSetsString }]) => {
+            if (!referenceSelectionSetsString) {
+              return undefined;
+            }
+
+            return indent(`${typeName}: ${referenceSelectionSetsString}${this.getPunctuation(declarationKind)}`);
+          })
+          .filter(v => v)
+          .join('\n')
+      ).string;
+  }
+
   public get schema(): GraphQLSchema {
     return this._schema;
   }
@@ -1586,13 +1609,6 @@ export class BaseResolversVisitor<
           }
         }
 
-        const parentTypeSignature = this._federation.transformFieldParentType({
-          fieldNode: original,
-          parentType,
-          parentTypeSignature: this.getParentTypeForSignature(node),
-          federationTypeSignature: 'FederationType',
-        });
-
         const { mappedTypeKey, resolverType } = ((): { mappedTypeKey: string; resolverType: string } => {
           const baseType = getBaseTypeNode(original.type);
           const realType = baseType.name.value;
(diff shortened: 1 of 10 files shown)


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1306,6 +1302,33 @@ export class BaseResolversVisitor<
...
+            if (!referenceSelectionSetsString) {
+              return undefined;
+            }
+
+            return indent(`${typeName}: ${referenceSelectionSetsString}${this.getPunctuation(declarationKind)}`);
+          })
+          .filter(v => v)
+          .join('\n')
+      ).string;
+  }
```
eddeee888, line 1330: Having `buildFederationReferenceTypes` to build standalone `FederationReferenceTypes` allows us to easily refer to the complex `reference` param types in different scenarios:

1. When using mappers:
```ts
import type { FederationReferenceTypes } from "./types.generated";
export type UserMapper = DatabaseUser | FederationReferenceTypes["User"]
```
This allows us to return reference as-is without type errors:
```ts
  __resolveReference: (ref) => {
    return ref;
  },
}
```
---
2. When not using mappers:
```ts
  User: User | FederationReferenceTypes["User"];
};
```

Similar to the mapper case, the parent of each normal resolver can receive a normal type or reference type

---

3. And for object resolvers:
```ts
export type UserResolvers<
  ContextType = ServerContext,
  ParentType extends
    ResolversParentTypes["User"] = ResolversParentTypes["User"],
  FederationReferenceType extends
    FederationReferenceTypes["User"] = FederationReferenceTypes["User"],
> = {
  __resolveReference?: ReferenceResolver<
    Maybe<ResolversTypes["User"]> | FederationReferenceType,
    FederationReferenceType,
    ContextType
  >;
};
```
Again, `FederationReferenceTypes["User"]` is used:
- As `reference` type (2nd generic param) 
- and `__resolveReference` must return: (1) `User` value normally, or (2) `FederationReferenceType`

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -1586,13 +1609,6 @@ export class BaseResolversVisitor<
           }
         }
 
-        const parentTypeSignature = this._federation.transformFieldParentType({
-          fieldNode: original,
-          parentType,
-          parentTypeSignature: this.getParentTypeForSignature(node),
-          federationTypeSignature: 'FederationType',
-        });
```
eddeee888, line 1594: Since federation reference type is extracted to [FederationReferenceTypes](https://github.com/dotansimha/graphql-code-generator/pull/10366/files#r2135735298), we don't need complex, inline reference types. 
So, `transformFieldParentType` can be completely removed 🎉

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -412,43 +505,6 @@ export class ApolloFederation {
...
-          return this.printReferenceSelectionSet({
-            referenceSelectionSet,
-            typeName: baseFederationType,
-          });
-        });
-
-        return result.length > 1 ? `( ${result.join(' | ')} )` : result.join(' | ');
-      })
-      .join('\n    & ')} )`;
-  }
```
eddeee888, line 450: This is now moved to `addFederationReferencesToSchema` because it's only used there.
I'm thinking to create a `addFederationReferencesToSchema.ts` file to shorten this file.

`addFederationReferencesToSchema` started as the function to mutate the schema (to re-use a lot of core plugin's functionalities). 
However, since it is run once at the start on each interface and object, it is convenient to gather metadata (e.g. reference type) and do stuff with said metadata (e.g. turn reference metadata to TypeScript type)

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -38,12 +37,15 @@ export const federationSpec = parse(/* GraphQL */ `
...
+ * - fields declared in `@requires`
  */
-interface ReferenceSelectionSet {
+interface DirectiveSelectionSet {
   name: string;
-  selection: boolean | ReferenceSelectionSet[];
+  selection: boolean | DirectiveSelectionSet[];
 }
 
+type ReferenceSelectionSet = Record<string, boolean>; // TODO: handle nested
```
eddeee888, line 47: I now understand `extractReferenceSelectionSet` better. It is a series of functions that turn GraphQL selection set -> `DirectiveSelectionSet` -> `ReferenceSelectionSet`

The previously named `ReferenceSelectionSet` is now correctly named `DirectiveSelectionSet` (It's only the immediary step)

packages/utils/plugins-helpers/src/federation.ts
```diff
@@ -99,25 +103,138 @@ export function addFederationReferencesToSchema(schema: GraphQLSchema): {
...
+        result.push({ ...currentSelectionSet, ...selectionSet });
+      }
+    }
+
+    const next = selectionSets.slice(1, selectionSets.length);
+
+    if (next.length > 0) {
+      findAllSelectionSetCombinations(next, result);
+    }
+  }
```
eddeee888, line 192: `findAllSelectionSetCombinations` is core to `@requires` (based on my understanding of this directive, feel free to correct if I misunderstood it):

 If a field is marked with `@requires`: 
   - and a client requests for said field, then the reference object contains fields in the selection set declared by `@requires`
   - otherwise, if the client doesn't request for said field, the declared selection set won't be in the reference object

This means when there are multiple fields with `@requires`, the selection set fields can appear in any combination.
  
---

For example:

```graphql

type Foo @key(fields: "id") {
  id: ID!

  a: String @external
  aRequires: String @requires(fields: "a")


  b: String @external
  bRequires: String @requires(fields: "b")

  c: String @external
  cRequires: String @requires(fields: "c")
}
```

- If a client doesn't request for `aRequires`, `bRequires`, or `cRequires`, then the reference shape is `{ __typename: 'Foo', id: 'something' }`
- If a client requests for `aRequires`, then the reference shape is `{ __typename: 'Foo', id: 'something', a: 'a-value' }`
- If a client requests for `aRequires` and `bRequires`, then the reference shape is `{ __typename: 'Foo', id: 'something', a: 'a-value', b: 'b-value' }`
- If a client requests for `aRequires` and `cRequires`, then the reference shape is `{ __typename: 'Foo', id: 'something', a: 'a-value', c: 'c-value' }`
- etc.

`findAllSelectionSetCombinations` will find all the combination of selection sets declared for `aRequires`, `bRequires` and  `cRequires`

eddeee888:

Issue #9701, opened 2023-10-06
Any errors stop generation with multiple outputs


lachtanek:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

Note: I am filing this as a bug, because it seems to me like this was working differently in an old release, but I do not know the history of this tool and I am quite a new user, so I am sorry if this should be "Feature request" instead.

When there are multiple `generates` outputs, if any one of them fails, no output is generated from either of them, even though the CLI output suggests something was generated (see image below).

This happens because if an error happens in any of the sub-tasks, an exception is thrown [here](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/codegen.ts#L436), while file generation happens later based on `results` variable (see [here](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/graphql-codegen-cli/src/generate-and-save.ts#L136-L138)).

It looks like this behavior was introduced in this release (quite a while ago): [https://github.com/dotansimha/graphql-code-generator/commit/f13d3554ed838047b62285281d0f87fd3ab20035](https://github.com/dotansimha/graphql-code-generator/commit/f13d3554ed838047b62285281d0f87fd3ab20035)

---

The last 3 lines would suggest that `schemas/api.graphql`, `components/types.generated.ts` and `components/` succeeded in generating, but no files are changed:

<img width="606" alt="image" src="[https://github.com/dotansimha/graphql-code-generator/assets/16037322/d9603169-ae5d-4722-82c6-a17accde572e](https://github.com/dotansimha/graphql-code-generator/assets/16037322/d9603169-ae5d-4722-82c6-a17accde572e)">

### Your Example Website or App

[https://codesandbox.io/p/sandbox/magical-tree-nm4fs9?file=%2Fcodegen.ts%3A15%2C5](https://codesandbox.io/p/sandbox/magical-tree-nm4fs9?file=%2Fcodegen.ts%3A15%2C5)

### Steps to Reproduce the Bug or Issue

See linked Codesandbox example

### Expected behavior

I'd either expect for the indicated generations to actually happen, or at least for the CLI output to be changed to reflect nothing was actually changed on disk.

Optionally, there could be a config flag choosing which behavior to use.

### Screenshots or Videos

*No response*

### Platform

* `@graphql-codegen/cli` version: 5.0.0
* OS: MacOS/Linux, but irrelevant
* NodeJS: irrelevant
[...]

PR #10376, opened 2025-06-27
CODEGEN-834 - [cli] Handle partial generation success

eddeee888:
## Description

Introduce `allowPartialOutputs` to allow users to write partial success to files.

This is a breaking change because we change how `executeCodegen` function works (exposed and can be imported): previously if any `generates` fails, it'd throw. Now, it will return the error and output files as result and let the caller decide what to do with them.

The main caller of `executeCodegen` is the CLI and it (1.) throws if there is complete failure or (2) use `allowPartialOutputs` to write on partial success or not

Related https://github.com/dotansimha/graphql-code-generator/issues/9701

## Type of change

- [x] Breaking change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/website/src/pages/docs/config-reference/codegen-config.mdx
+++ b/website/src/pages/docs/config-reference/codegen-config.mdx
@@ -102,7 +102,7 @@ Here are the supported options that you can define in the config file (see [sour
 
   - **`skipDocumentsValidation.skipDuplicateValidation`** - A flag to disable the validation for duplicate documents
 
-- **`allowPartialOutputs`** - Allows partial outputs to be written to files if one or more `generates` blocks have errors. If all `generates` blocks fail, treat as a complete failure and return non-zero to the terminal. Default is `false`.
+- **`allowPartialOutputs`** - Allows partial outputs to be written to files if one or more `generates` blocks have errors. Default is `false`.
 
 ## Environment Variables
 


packages/graphql-codegen-cli/src/utils/watcher.ts
```diff
@@ -78,7 +78,10 @@ export const createWatcher = (
     const debouncedExec = debounce(() => {
       if (!isShutdown) {
         executeCodegen(initialContext)
-          .then(onNext, () => Promise.resolve())
+          .then(
+            ({ result }) => onNext(result),
```
eddeee888, line 82: Previously, partial success in watch mode doesn't write to files.

With this new no-throw-on-error implementation in watch mode, partial success always write to files.

Users primarily use watch mode only during development, so I think this is fine. They are already seeing something like "✔ Generate to ./src/graphql/types.generated.ts", which can be interpreted as something has been generated (check image below)

Screenshot 2025-07-04 at 12 13 36 am

website/public/config.schema.json
eddeee888, line 1: I had an issue and thought running the command to update this would help 🤔 
I think this file is only updated at build time before publishing the website. Maybe we don't have to track it? 

Anyways, this should keep it up-to-date until we decide to ignore it (or not)

website/src/pages/docs/config-reference/codegen-config.mdx
```diff
@@ -40,7 +40,7 @@ export default config
 
 ## Configuration options
 
-Here are the supported options that you can define in the config file (see [source code](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-helpers/src/types.ts#L92)):
+Here are the supported options that you can define in the config file (see [source code](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/utils/plugins-helpers/src/types.ts#L388)):
```
eddeee888, line 43: I noticed the line number changed on master, so this updates it to point to the right `Config` type

eddeee888:

PR #10377, opened 2025-06-29
[resolvers] Refactor to remove NameNode override and simplify federation functions

eddeee888:
## Description

To prepare for next major version, we want to fix a few long-standing type cast issues for `typescript-resolvers` plugin and federation support:
- Previously ,`NameNode` was implemented to change how a node name can be accessed. Instead of `node.name.value`, it just becomes `node.name`.  However, the type of this node (from `graphql` package) does not know that this is changed at runtime, so we always need to type case `as any as string` or `as unknown as string` to satisfy TypeScript. This results in confusion for contributors, and normally requires runtime debug to understand/correctly use the right value.
   - This PR removes `NameNode` function to avoid typecasts later
- Some functions in `federation.ts` take both AST nodes and types. This adds more complexity as there are more code paths to think about.
   - This PR simplifies and make it more consistent by requiring functions in federation support to use AST nodes.

## Type of change

- [x] Refactor

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/utils/plugins-helpers/src/federation.ts
+++ b/packages/utils/plugins-helpers/src/federation.ts
@@ -9,7 +9,6 @@ import {
   GraphQLObjectType,
   GraphQLSchema,
   InterfaceTypeDefinitionNode,
-  isInterfaceType,
   isObjectType,
   ObjectTypeDefinitionNode,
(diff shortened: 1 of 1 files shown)


eddeee888:

PR #10392, opened 2025-08-05
Drop Node 18 support

eddeee888:
Node 18 is has reached EOL so we should avoid using it or test it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,8 +29,6 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       - name: Setup env
         uses: the-guild-org/shared-config/setup@main
-        with:
-          nodeVersion: 18
       - name: Prettier Check
         run: yarn prettier:check
   dev-tests-old:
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -29,8 +29,6 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       - name: Setup env
         uses: the-guild-org/shared-config/setup@main
-        with:
-          nodeVersion: 18
```
eddeee888, line 33: This defaults to Node 22, so removing explicit version is simpler & will get advantage when the shared workflows gets updated

.github/workflows/main.yml
```diff
@@ -139,10 +129,10 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest] # remove windows to speed up the tests
-        node_version: [16, 18, 20]
+        node_version: [20, 22, 24]
         graphql_version: [15, 16]
         include:
-          - node-version: 14
+          - node-version: 20
```
eddeee888, line 135: We should at least test Node 20 on Windows. Node 14 is a bit too old

PR #10403, opened 2025-08-12
Update tsconfig for Node 20

eddeee888:
## Description

This is based on https://github.com/tsconfig/bases/blob/main/bases/node20.json

Note:
- `bob` has mapping to `es2022` when building for ESM, so we cannot set `module` or `moduleResolution` like the base Node 20 example


babel.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   presets: [
     ['@babel/preset-env', { targets: { node: process.versions.node.split('.')[0] } }],
-    '@babel/preset-typescript',
+    ['@babel/preset-typescript', { allowDeclareFields: true }],
```
eddeee888, line 4: We use `babel-jest`, so we need this to understand `declare` keyword needed by tsconfig version of Node 20

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -691,7 +691,7 @@ export class BaseResolversVisitor<
   TRawConfig extends RawResolversConfig = RawResolversConfig,
   TPluginConfig extends ParsedResolversConfig = ParsedResolversConfig
 > extends BaseVisitor<TRawConfig, TPluginConfig> {
-  protected _parsedConfig: TPluginConfig;
+  protected declare _parsedConfig: TPluginConfig;
```
eddeee888, line 694: Without this, `declare` keyword, TS will throw error:
```
Property '_parsedConfig' will overwrite the base property in 'BaseVisitor<TRawConfig, TPluginConfig>'. If this is intentional, add an initializer. Otherwise, add a 'declare' modifier or remove the redundant declaration.ts(2612)
```

PR #10404, opened 2025-08-13
Bump packages

eddeee888:
## Description

This PR bumps deps to next major versions:

- nock v14
- dependency-graph to v1
- cosmiconfig v9
- debounce v2
- jiti v2.3

These have lower risk so I'm grouping them together

## Type of change

- [x] Version update


packages/graphql-codegen-cli/package.json
```diff
@@ -78,7 +78,6 @@
   "devDependencies": {
     "@graphql-tools/merge": "9.0.6",
     "@parcel/watcher": "^2.1.0",
-    "@types/debounce": "1.2.4",
```
eddeee888, line 81: `debounce` v2 comes with types so we don't need this type package anymore 🕺

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -75,8 +75,8 @@ function customLoader(ext: 'json' | 'yaml' | 'js' | 'ts' | 'mts' | 'cts'): Codeg
     }
 
     if (ext === 'ts') {
-      const jitiLoader = jiti('', { interopDefault: true });
-      return jitiLoader(filepath);
+      const jitiLoader = createJiti('');
+      return jitiLoader.import(filepath, { default: true });
```
eddeee888, line 79: In `jiti@2.3`, `default: true` is the equivalent of `interopDefault: true`

PR #10405, opened 2025-08-16
[CLI] Bump deps for next major version

eddeee888:
## Description

This PR bump CLI deps and handle breaking changes:
- listr2
- inquirer

## Tests

- [x] Using alpha package

### Listr2

https://github.com/user-attachments/assets/084925ca-da58-4d64-9c19-348da35e34ae

### Inquirer

https://github.com/user-attachments/assets/a6d157a0-49b6-4b9b-a762-ffc3e20553e6


packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -414,13 +423,13 @@ export async function executeCodegen(
     {
       rendererOptions: {
         clearOutput: false,
-        collapse: true,
+        collapseSubtasks: true,
```
eddeee888, line 426: `collapse` is renamed to `collapseSubtasks` since [listr2 v6](https://github.com/listr2/listr2/releases/tag/v6.0.0)

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -414,13 +423,13 @@ export async function executeCodegen(
...
         formatOutput: 'wrap',
         removeEmptyLines: false,
       },
       renderer: config.verbose ? 'verbose' : 'default',
       ctx: { errors: [] },
-      rendererSilent: isTest || config.silent,
+      silentRendererCondition: isTest || config.silent,
```
eddeee888, line 432: `rendererSilent` is renamed to `silentRendererCondition` since [listr2 v6](https://github.com/listr2/listr2/releases/tag/v6.0.0)

packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -242,31 +242,26 @@ describe('generate-and-save', () => {
...
     const originalNodeEnv = process.env.NODE_ENV;
 
-    let consoleErrorMock: jest.Mock;
+    let outputErrorMock: jest.SpyInstance;
 
     beforeEach(() => {
-      // Mock common console functions to avoid noise in the terminal
-      global.console.log = jest.fn();
-      global.console.warn = jest.fn();
-      global.console.error = jest.fn();
```
eddeee888, line 254: `listr2` doesn't call `console.*`. Instead, the output goes to `std*`

packages/graphql-codegen-cli/src/init/questions.ts
eddeee888, line 1: BIG API change from `inquirer` (legacy) and `@inquirer/prompts`.

We now call the questions functions (e.g. `checkbox`, `input`, etc.) instead of providing an object with said type.
Main changes:
- `when` no longer exists. We now just use the previous answers to decide whether to call the next
- `suffix` no longer exists. We have to construct the strings manually
- `select.default` (`select` is formerly `list`) must point to the value of the item for comparison to work correctly. [Source](https://github.com/SBoudrias/Inquirer.js/blob/ee16061a1e3f99a6cc714a3d473f7cd12b06a3f1/packages/select/src/index.ts#L146-L147)
- `default` are not functions anymore. For our use case, we can find the default values using the variables created above
-

packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -281,37 +276,34 @@ describe('generate-and-save', () => {
...
-          [FAILED] - ES Modules and CommonJS exports (export as default or named export "schema")
-          [FAILED] - Introspection JSON File
-          [FAILED] - URL of GraphQL endpoint
-          [FAILED] - Multiple files with type definitions (glob expression)
-          [FAILED] - String in config file
-          [FAILED]
-          [FAILED] Try to use one of above options and run codegen again.
+        const cwd = process.cwd(); // cwd is different for every machine, remember to replace local path with this after updating snapshot
+        expect(outputErrorMock.mock.calls[0][0]).toMatchInlineSnapshot(`
+          "[31m[FAILED][39m Failed to load schema from ./tests/test-files/schema-dir/error-schema.graphql:
```
eddeee888, line 281: `listr2` prints verbose messages with colour now.
The `[31m[FAILED][39m` makes `FAILED` red in the terminal.
To make it a bit simpler to test the output, I've just use `toMatchInlineSnapshot`

packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -281,37 +276,34 @@ describe('generate-and-save', () => {
...
+          [31m[FAILED][39m Syntax Error: Expected Name, found "!".
+
+          [31m[FAILED][39m ${cwd}/tests/test-files/schema-dir/error-schema.graphql:2:15
```
eddeee888, line 284: `cwd` is overwritten to the machine's actual path if we update the snapshot tests of this file 
i.e. the person updating this test would have to replace this manually. This could be annoying.

However, this is rarely updated so let's try this for now.

PR #10410, opened 2025-08-19
Migrate to Vitest

eddeee888:
## Description

Jest does not work well for ESM packages.
This mega PR is to migrate everything off Jest and to Vitest

### DIFF
--- a/packages/graphql-codegen-cli/vitest.setup.ts
+++ b/packages/graphql-codegen-cli/vitest.setup.ts
@@ -1,4 +1,5 @@
 import { vi } from 'vitest';
 
+process.env.NO_COLOR = '1'; // When running listr2 tests in generate-and-save.spec.ts, do not add colors to avoid failure in CI as different OS may treat colours differently
 vi.spyOn(process, 'cwd').mockImplementation(() => __dirname);
 vi.mock('some-fetch', () => require('./tests/__mocks__/some-fetch.cjs'));
--- a/packages/utils/plugins-helpers/tests/resolve-external-module-and-fn.spec.ts
+++ b/packages/utils/plugins-helpers/tests/resolve-external-module-and-fn.spec.ts
@@ -1,7 +1,7 @@
 import path from 'path';
 import { resolveExternalModuleAndFn } from '../src/resolve-external-module-and-fn.js';
 
-describe.only('resolveExternalModuleAndFn', () => {
+describe('resolveExternalModuleAndFn', () => {
   describe('Issues', () => {
     it('#6553 - Cannot find module', () => {
       const relativePathToSelf = path.relative(process.cwd(), path.join(__dirname, './fixtures/externalModuleFn.js'));


packages/utils/graphql-codegen-testing/src/index.ts
```diff
@@ -1,18 +1,18 @@
...
-  }
+interface CustomMatchers<R = unknown> {
+  /**
+   * Normalizes whitespace and performs string comparisons
+   */
+  toBeSimilarStringTo(expected: string): R;
+}
+declare module 'vitest' {
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
+  interface Matchers<T = any> extends CustomMatchers<T> {}
```
eddeee888, line 13: Type config is [here](https://vitest.dev/guide/extending-matchers.html)

packages/plugins/typescript/resolvers/vitest.config.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export default mergeConfig(
+  sharedConfig,
+  defineProject({
+    test: {
+      name: 'typescript-resolvers',
+      include: ['**/*.spec.ts'],
+    },
+  })
+);
```
eddeee888, line 12: Project-based config like this adds the name and ✨colour✨ to the terminal. 
We can run tests by project name as well.

tsconfig.spec.json
eddeee888, line 1: Having a separate tsconfig file for tests allows us to customise for it
e.g. only test files would have `vitest/globals` types applied, meaning normal files will see TS errors if they try to use `vi.`

packages/presets/client/tests/client-preset.spec.ts
```diff
@@ -1250,7 +1250,7 @@ export * from "./gql.js";`);
   });
 
   it('should dedupe fragments in a "string" document mode', async () => {
-    const dir = path.join(__dirname, 'tmp/duplicate-fragments');
+    const dir = path.join(__dirname, 'tmp/duplicate-fragments-string');
```
eddeee888, line 1253: There are 2 tests in here writing to the same path, which causes a race-condition.
We can just create a separate dir for each test to avoid this issue.

packages/presets/graphql-modules/vitest.setup.ts
```diff
@@ -0,0 +1,3 @@
+import { vi } from 'vitest';
+
+vi.spyOn(process, 'cwd').mockImplementation(() => __dirname);
```
eddeee888, line 3: This is a project-wide setup to replace `useMonorepo` functionalities.
Seems to work for this project. Hopefully it works for others too.

I needed to change over to this approach because by the time the `cwd` is mocked, it's already called in the tests. i.e. `useMonorepo` doesn't work.

There could be a better way to do it when we need to, but for now, I'll try this approach first.

packages/graphql-codegen-cli/tests/__mocks__/fs.cjs
eddeee888, line 1: This follows the [vitest example to mock fs](https://vitest.dev/guide/mocking.html#file-system)

packages/graphql-codegen-cli/tests/init.spec.ts
```diff
@@ -1,17 +1,17 @@
...
 import { guessTargets } from '../src/init/targets.js';
 import { Tags } from '../src/init/types.js';
 
-jest.mock('../src/utils/get-latest-version.ts', () => {
+vi.mock('../src/utils/get-latest-version.ts', () => {
   return { getLatestVersion: () => Promise.resolve('1.0.0') };
 });
 
-jest.mock('fs');
+vi.mock('fs', () => require('./__mocks__/fs.cjs'));
```
eddeee888, line 14: `__mocks__` isn't at the right place so we have to manually do this.
I _think_ it'd need to be at the project root 🤔 

Either way, this is fine as it's clear anyways

packages/graphql-codegen-cli/tests/init.spec.ts
```diff
@@ -78,66 +78,64 @@ const packageJson = {
...
   afterEach(() => {
-    require('fs').__resetMockFiles();
-    jest.clearAllMocks();
+    vi.clearAllMocks();
   });
 
   describe('guessTargets()', () => {
     it('should guess angular projects', async () => {
-      require('fs').__setMockFiles(resolve(process.cwd(), 'package.json'), packageJson.withAngular);
+      vol.fromJSON({ ['package.json']: packageJson.withAngular }, process.cwd());
```
eddeee888, line 91: `memfs` blocks writing to actual fs for us, which is great!
`vol` is used to prepare a file, so we can read it virtually.

packages/utils/graphql-codegen-testing/src/index.ts
```diff
@@ -68,42 +66,6 @@ expect.extend({
...
-      let spyProcessCwd: jest.SpyInstance;
-      beforeEach(() => {
-        spyProcessCwd = jest.spyOn(process, 'cwd').mockReturnValue(cwd);
-      });
-      afterEach(() => {
-        spyProcessCwd.mockRestore();
-      });
-    },
-  };
-}
```
eddeee888, line 105: These can go as we are mocking `cwd` in the projects that require it in `vitest.setup.ts`

package.json
```diff
@@ -14,7 +14,8 @@
     "postbuild": "yarn fix-bins",
     "watch-build": "npx tsc-watch --project tsconfig.json --onSuccess \"bob build\"",
     "rebuild": "bob build --incremental",
-    "test": "jest --forceExit --no-watchman",
+    "test": "vitest --no-watch",
+    "test:watch": "vitest",
```
eddeee888, line 18: This is added to be able to run in watch mode:
- Vitest runs in watch mode by default
- So, to run once (i.e. `test` existing behaviour), we add `--no-watch`
- But with `--no-watch`, we cannot do `yarn test -w` because `-w` conflicts and is ignored

packages/graphql-codegen-cli/tests/custom-loaders/custom-schema-loader-with-context.cjs
eddeee888, line 1: This file (and the next few below) are changed to `.cjs` because `.js` is loaded as ESM by Vitest by default.
`.cjs` forces it to load as CommonJS

packages/graphql-codegen-cli/tests/watcher-test-helpers/setup-mock-watcher.ts
eddeee888, line 1: This file is simplified and made compatible with Vitest [here](https://github.com/dotansimha/graphql-code-generator/pull/10410/files#diff-290aa52b408aeb4c2a1fb1aae1a310317ff36d74ced3556c7b0e303705835692R9-R31)

PR #10417, opened 2025-08-28
[resolvers] Report if an object can have `__isTypeOf` resolver to output meta

eddeee888:
## Description

This PR ensures `typescript-resolvers` plugin reports whether user defined objects have `__isTypeOf` or not.
This helps downstream plugins and presets (such as Server Preset) handle correct typing without parsing the schema again.

## Type of change

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - master
+      - federation-fixes # FIXME: Remove this line after the PR is merged
 
 env:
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - master
+      - federation-fixes # FIXME: Remove this line after the PR is merged
```
eddeee888, line 10: Putting this back temporarily so we run tests in CI

Issue #10438, opened 2025-09-08
graphql-modules-preset requires __isTypeOf


mzl-md:
### Which packages are impacted by your issue?

@graphql-codegen/graphql-modules-preset

## Problem

After updating the codegen libs, our typescript typecheck fails with `Property '__isTypeOf' is missing in type`

Package | Change
-- | -- 
@graphql-codegen/cli (source) | 5.0.7 -> 6.0.0 | 
@graphql-codegen/graphql-modules-preset (source) | 4.0.17 -> 5.0.0 |  
@graphql-codegen/typescript (source) | 4.1.6 -> 5.0.0 | 
@graphql-codegen/typescript-resolvers (source) | 4.5.2 -> 5.0.0 | 

## Workaround 

When we add `__isTypeOf: undefined` to the object instances, it works again.

## Suspected Problem

typescript-resolvers now generates resolver types without `__isTypeOf` instead of optional, but graphql-modules-preset still generates the type `Pick<Types.MyTypeResolvers, DefinedFields['MyType'] | '__isTypeOf'>`. So the module types now require a `__isTypeOf: any` attribute.

### Your Example Website or App

none

### Steps to Reproduce the Bug or Issue

1. Generate types for any schema with a query using the given dependencies

### Expected behavior

As a user I expect the generated module types to not include __isTypeOf anymore if they are not part of the generated resolver types.

### Platform

- OS: macOS
- NodeJS: v20.19.5
- `graphql` version: 16.11.0
- `@graphql-codegen/*` version(s): 5.0.0 (see description)

### Codegen Config File

import type {CodegenConfig} from '@graphql-codegen/cli'

const config: CodegenConfig = {
    generates: {
        './src/schema/': {
            config: {
                avoidOptionals: true,
                inputMaybeValue: 'T | null | undefined',
                makeResolverTypeCallable: true,
                mappers: {},
                maybeValue: 'T | null | undefined',
                optionalInfoArgument: true,
                optionalResolveType: true,
                useIndexSignature: true
            },
            plugins: [
                {
                    add: {
                        content: '/* eslint-disable */'
                    }
                },
                'typescript',
                'typescript-resolvers'
            ],
            preset: 'graphql-modules',
            presetConfig: {
                baseTypesPath: '../generated-types/graphql.ts',
                filename: 'generated-module-types.ts',
                useGraphQLModules: false
            }
        }
    },
    schema: 'src/schema/**/schema.graphql'
}

export default config

[...]

PR #10447, opened 2025-09-23
CODEGEN-859 [graphql-modules-preset] Fix __isTypeOf getting picked from from all objects

eddeee888:
## Description

This PR fixes `__isTypeOf` getting picked from all objects.
With https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1757264511789 release, `__isTypeOf` is only generated for implementing types and union members.

This PR applies the same logic to `graphql-modules-preset` to make sure it's correctly picked.

Related https://github.com/dotansimha/graphql-code-generator/issues/10438

## Type of change

## How Has This Been Tested?

- [x] Unit test


packages/presets/graphql-modules/src/builder.ts
```diff
@@ -433,6 +442,10 @@ export function buildModule(
 
       case Kind.UNION_TYPE_DEFINITION: {
         defined.unions.push(name);
+
+        for (const namedType of node.types || []) {
+          pushUnique(withIsTypeOf.objects, namedType.name.value);
```
eddeee888, line 447: I'm using `pushUnique` for union type here (unlike `.push` on other type definition) because a union may contain a union member that already exists on another union declaration.

PR #10459, opened 2025-10-06
[client-preset] Forward immutableTypes option to client preset

eddeee888:
## Description

`immutableTypes` is a good option to pass through because it forces users handle responses as immutable. Majority of users would use a client such as Apollo Client which manages its own caching which shouldn't be mutated in any way.

Related: https://github.com/dotansimha/graphql-code-generator/issues/8296#issuecomment-1287010974

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/.changeset/gold-zoos-show.md
+++ b/.changeset/gold-zoos-show.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/client-preset': minor
+---
+
+Forward immutableTypes to client preset config


yarn.lock
eddeee888, line 1: I think I didn't commit `yarn.lock` correctly in the last PR, so when running `yarn` in this PR, it cleans up the lockfile like this.

Issue #10464, opened 2025-10-10
Autogenerated `graphql.ts` file is deleted and not restored when unknown properties are declared in query


MarkLeMerise:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

The `graphql-codegen` command is deleting autogenerated files when unknown properties are specified in a query. However, these files are **not** being restored once the issue is fixed.

### Your Example Website or App

https://stackblitz.com/edit/gql-codegen-v6-file-deletion

### Steps to Reproduce the Bug or Issue

1. In the [linked StackBlitz](https://stackblitz.com/edit/gql-codegen-v6-file-deletion), run `npm run codegen-watch` in a terminal.
2. Add an unknown property to either GraphQL query file. (Observer the autogenerated `gql/gql.ts` and `gql/graphql.ts` are deleted).
3. Undo the change so all queries are valid again.

### Expected behavior

As a dev, I would expect the `gql/gql.ts` and `gql/graphql.ts` files to be restored after fixing any issues in my GQL queries.

Note that the behavior of v6 differs from v5, where specifying unknown properties would not delete any files.

### Screenshots or Videos

https://github.com/user-attachments/assets/70c374f0-2508-46f3-9058-63c0a3a051b0

### Platform

- OS: macOS Tahoe 26.0.1
- NodeJS: 24.9.0 (as of writing this issue, StackBlitz uses Node 20.19.1)
- `graphql` version: 16.11.0
- `@graphql-codegen/cli` version(s): 6.0.0

### Codegen Config File

```ts
{
  schema: './schema.gql',
  documents: ['./src/**/*.ts'],
  generates: {
    './gql/': {
      preset: 'client',
    },
  },
}
```

### Additional context

Running `npm run codegen-watch` when a GQL query _already_ has an issue will initially generated the expected files.

I also tried `allowPartialOutputs: true` in the config, but this did not change the observed behavior.

eddeee888:
Hi @MarkLeMerise , 
Thanks for reporting this!

> Note that the behavior of v6 differs from v5, where specifying unknown properties would not delete any files.

In v5, if there are any errors during watch mode, Codegen won't write to files.
However, in v6, we introduced a new mode called `allowPartialOutputs`, but did not handle watch mode correctly.

---

I've created a quick patch to revert the watch behaviour closer to v5 temporarily. Could you please try this alpha version:

```
[...]

PR #10468, opened 2025-10-20
CODEGEN-866 [CLI] Fix generated files disappearing in watch mode

eddeee888:
## Description

Previously, on partial or full failure, watch mode still writes to output. However, since the output'd be an empty array, it will then call `removeStaleFiles` internally to remove all previously generated files.

This patch puts a temporary fix to avoid writing output on any failure to fix the described behaviour.

This also means the `config.allowPartialOutputs` does not work in watch mode for now.

Related https://github.com/dotansimha/graphql-code-generator/issues/10464

## Type of change

- [x] Manual test

## How Has This Been Tested?

- [x] Unit test
- [ ] Test B

### DIFF
--- a/packages/graphql-codegen-cli/tests/watcher.spec.ts
+++ b/packages/graphql-codegen-cli/tests/watcher.spec.ts
@@ -30,7 +30,7 @@ const setupMockWatcher = async (
   const dispatchChange = async (path: string) => subscribeCallbackMock(undefined, [{ type: 'update', path }]);
 
   // createWatcher doesn't set up subscription immediately, so we wait for a tick before continuing
-  await new Promise(resolve => setTimeout(resolve, 10));
+  await new Promise(resolve => setTimeout(resolve, 100));
 
   return { stopWatching, dispatchChange };
 };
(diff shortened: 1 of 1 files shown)


packages/graphql-codegen-cli/tests/watcher.spec.ts
```diff
@@ -771,4 +776,49 @@ describe('Watch targets', () => {
...
+        },
+      },
+      onNextMock
+    );
+
+    // Because document has error, onNext shouldn't be called
+    expect(onNextMock).not.toHaveBeenCalled();
+  });
+
+  test.todo('on watcher subsequent codegen run, it does not call onNext on error');
```
eddeee888, line 827: We currently cannot test subsequent calls because we are currently mocking the `.subscribe` above.

I'll look to remove the mock to see if we can test the watch behaviour better.

packages/graphql-codegen-cli/tests/watcher-test-helpers/assert-watcher-build-triggers.ts
eddeee888, line 1: Some of these are drive-bys but I'm keen to keep them to make future contributions easier

PR #10474, opened 2025-10-27
[CLI] Create watcher.run.spec to test actual running behaviour

eddeee888:
## Description

Current `watcher.spec.ts` mocks out the main subscribe functionalities, so whilst it's great to test whether a watch pattern would trigger the re-run, it's not great to test the behaviour of the watcher _after_ the reruns trigger.

This PR:
- adds `watcher.run.spec.ts` to test the run, re-runs behaviour
- renames `watcher.spec.ts` to `watcher.patterns.spec.ts` to communicate that it's for testing watcher trigger patterns

## Known issues

Currently, Node v22 and v24 GitHub test runner cannot import the watcher. Works fine on my machine (Mac) 😎™️ 

So:
- Real watcher run behaviour will only be tested in Node v20 on Ubuntu and Windows.
- I've raised an [issue in @parchel/watcher](https://github.com/parcel-bundler/watcher/issues/218).

### DIFF
--- a/examples/programmatic-typescript/package.json
+++ b/examples/programmatic-typescript/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@graphql-codegen/core": "5.0.0",
     "@graphql-codegen/plugin-helpers": "6.0.0",
-    "@graphql-codegen/typed-document-node": "6.0.2",
+    "@graphql-codegen/typed-document-node": "6.1.0",
     "@graphql-codegen/typescript": "5.0.2",
     "@graphql-codegen/typescript-operations": "5.0.2",
     "@graphql-codegen/typescript-resolvers": "5.1.0",
--- a/packages/plugins/typescript/typed-document-node/CHANGELOG.md
+++ b/packages/plugins/typescript/typed-document-node/CHANGELOG.md
@@ -1,5 +1,15 @@
 # @graphql-codegen/typed-document-node
 
+## 6.1.0
+
+### Minor Changes
+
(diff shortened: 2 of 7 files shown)


packages/graphql-codegen-cli/tests/watcher.spec.ts
```diff
@@ -776,53 +774,4 @@ describe('Watch targets', () => {
       }
     );
   });
-
```
eddeee888, line 779: Moved to `watcher.run.spec.ts` to better test the actual behaviour, instead of mocking

packages/graphql-codegen-cli/tests/watcher.spec.ts
```diff
@@ -10,8 +10,6 @@ const unsubscribeMock = vi.fn();
 const subscribeMock = vi.fn();
 let subscribeCallbackMock: Mock<SubscribeCallback>;
 
-// FIXME: this mocks out the main functionality which is triggering the codegen
-// This is not great because we cannot test the actual watch functionality
```
eddeee888, line 14: This test file now is used to test trigger patterns. 

It's fine to mock the main behaviour in this "patterns" test because it's easier and faster.

packages/graphql-codegen-cli/tests/watcher.run.spec.ts
eddeee888, line 1: The easiest to test file watcher is .... to create real files for watcher to watch!
In this file, we will create files in a local `temp` folder and tell watcher to watch them.

Note: the `temp` folder is ignored by git and is used by another test already. So we can continue use it for testing purposes

.github/workflows/main.yml
```diff
@@ -148,17 +148,49 @@ jobs:
         run: node ./scripts/match-graphql.js ${{matrix.graphql_version}}
       - name: Install Dependencies
         run: yarn
-      - name: Cache Vitest
-        uses: actions/cache@d4323d4df104b026a6aa633fdb11d772146be0bf # v4
```
eddeee888, line 152: Vitest cache only stores run metadata to make re-runs focus on those that failed or is long. We don't need to cache it as it increases CI time and doesn't give values.

PR #10508, opened 2025-11-18
[typescript-operations] Generate enums referenced in operation Variables

eddeee888:
## Description

This PR adds logic to generate enums in the generated file if used in operations Variables. This uses the enum conversion logic in https://github.com/dotansimha/graphql-code-generator/pull/10520

This is a breaking change because previous setup with `typescript` plugin will break as both plugins now generate the same enum.

Note: there are some unfinished tests. I want to get the "used input types logic detection" into the feature branch first.

## Type of change

## How Has This Been Tested?

- [x] Unit test (partial, more test cases are coming in the following PR )

### DIFF
--- a/packages/plugins/typescript/operations/src/visitor.ts
+++ b/packages/plugins/typescript/operations/src/visitor.ts
@@ -171,7 +171,7 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
     };
   }
 
-  EnumTypeDefinition(node: EnumTypeDefinitionNode): string {
+  EnumTypeDefinition(node: EnumTypeDefinitionNode): string | null {
     const enumName = node.name.value;
     if (!this._usedNamedInputTypes[enumName]) {
       return null;


packages/plugins/typescript/operations/tests/ts-documents.standalone.spec.ts
```diff
@@ -157,3 +163,13 @@ describe('TypeScript Operations Plugin - Standalone', () => {
...
+
+describe('TypeScript Operations Plugin - Enum', () => {
+  it.todo('does not generate unused enum in variables and result');
+  it.todo('handles native numeric enum correctly');
+  it.todo('handles const enum correctly');
+  it.todo('handles native const enum correctly');
+  it.todo('handles native enum correctly');
+  it.todo('handles EnumValues correctly');
+  // Bring over tests from https://github.com/dotansimha/graphql-code-generator/blob/accdab69106605241933e9d66d64dc7077656f30/packages/plugins/typescript/typescript/tests/typescript.spec.ts
+});
```
eddeee888, line 175: These tests will be tested in a follow-up PR. I want to get the logic of detecting used input types into the feature branch first.

PR #10512, opened 2025-11-19
fix: allow mappers to override root types (Query, Mutation, Subscription)

### LINKED ISSUE #10511: Mappers don't work for root types (Query, Mutation, Subscription)
### Describe the bug

When configuring a mapper for a root type (Query, Mutation, or Subscription), the mapper is ignored and the default `rootValueType` (which defaults to `Record<PropertyKey, never>`) is used instead.

This happens because the code checks if a type is a root type **before** checking for mappers, causing mappers to be skipped for root types.

### Proposed solution

Move the mapper check before the root type check, allowing mappers to override `rootValueType` when configured. This maintains backward compatibility (if no mapper is configured, `rootValueType` is still used) while enabling the expected behavior when mappers are configured.

### Steps to Reproduce the Bug or Issue

1. Create a GraphQL schema with a root `Subscription` type that is also a regular object type:
```graphql
type Subscription {
  id: ID!
  status: String!
  # ... other fields
}
```

2. Configure codegen with a mapper:
```yaml
mappers:
  Subscription: ./helpers/mapSubscriptionToGraphQL#ExtendedSubscription
```

3. Run codegen

4. Observe that `ResolversTypes['Subscription']` is `Record<PropertyKey, never>` instead of `ExtendedSubscription`

## Minimal reproduction

**Schema:**
```graphql
type Subscription {
  id: ID!
  status: String!
}

type Query {
  subscription: Subscription!
}
```

**codegen.yml:**
```yaml
schema: schema.graphql
generates:
  types.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      mappers:
        Subscription: ./helpers#ExtendedSubscription
```

**helpers.ts:**
```typescript
export interface ExtendedSubscription {
  id: string;
  status: string;
}
```

**Expected output in types.ts:**
```typescript
  Subscription: ResolverTypeWrapper<ExtendedSubscription>;
};
```

**Actual output:**
```typescript
  Subscription: ResolverTypeWrapper<Record<PropertyKey, never>>;
};
```

### Expected behavior

When a mapper is configured for a root type, it should be used instead of `rootValueType`. For example:

```yaml
mappers:
  Subscription: ./helpers/mapSubscriptionToGraphQL#ExtendedSubscription
```

Should generate:
```typescript
  Subscription: ResolverTypeWrapper<ExtendedSubscription>;
  // ...
};
```

### Actual behavior

Even with the mapper configured, it generates:
```typescript
  Subscription: ResolverTypeWrapper<Record<PropertyKey, never>>;
  // ...
};
```

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- `@graphql-codegen/cli`: 6.0.2
- `@graphql-codegen/typescript-resolvers`: 5.1.2
- `@graphql-codegen/visitor-plugin-common`: 6.1.2
- NodeJS: 24.x

### Codegen Config File

_No response_

### Additional context

This is particularly problematic when:
- A GraphQL schema has a root `Subscription` type that represents a domain object (e.g., Stripe subscription)
- You want to use custom mapper types for better type safety
- You need consistent resolver typing across your codebase

The root cause is in `packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts` where root types are checked before mappers (around line 974).

realies:
This PR fixes #10511 where mappers configured for root types (Query, Mutation, Subscription) were being ignored. Previously, the code checked if a type was a root type before checking for mappers, which meant that even when a mapper was configured, the default `rootValueType` (which defaults to `Record<PropertyKey, never>`) would be used instead.

This is particularly problematic when a GraphQL schema has a root `Subscription` type that is also a regular object type (e.g., representing a Stripe subscription), and you want to use a custom mapper type instead of the default empty type.

**The fix:** Move the mapper check before the root type check, allowing mappers to override `rootValueType` when configured.

## Type of change

N/A

## How Has This Been Tested?

**Test Case:**
1. Configure a mapper for a root type (e.g., `Subscription`) in `codegen.yml`:
   ```yaml
   mappers:
     Subscription: ./helpers/mapSubscriptionToGraphQL#ExtendedSubscription
   ```
2. Run codegen
3. Verify that `ResolversTypes['Subscription']` uses `ExtendedSubscription` instead of `Record<PropertyKey, never>`

**Before the fix:**
```typescript
  Subscription: ResolverTypeWrapper<Record<PropertyKey, never>>;
  // ...
};
```

**After the fix:**
```typescript
  Subscription: ResolverTypeWrapper<ExtendedSubscription>;
  // ...
};
```

**Test Environment**:
- OS: Linux
- `@graphql-codegen/visitor-plugin-common`: 6.1.2
- `@graphql-codegen/typescript-resolvers`: 5.1.2
- NodeJS: 24.x

## Checklist:

## Further comments

**Why this change was needed:**

In GraphQL, root types (Query, Mutation, Subscription) are regular object types that can have fields. When a schema has a root `Subscription` type that represents a domain object (like a Stripe subscription), developers may want to use custom mapper types instead of the default `rootValueType`.

The previous implementation checked root types before mappers, which prevented this use case. This fix allows mappers to take precedence over `rootValueType`, which is the expected behavior - if a developer explicitly configures a mapper, it should be used.

**Alternatives considered:**

1. **Keep the current behavior** - This would require developers to use explicit `Resolver<...>` types everywhere, which is verbose and inconsistent with the rest of the codebase.

2. **Change `rootValueType` to be nullable** - This wouldn't solve the problem and would break existing code that relies on the default behavior.

3. **Add a new config option** - This would add unnecessary complexity when the existing mapper system should handle this case.

The chosen solution is the simplest and most intuitive: mappers should work for all types, including root types.

### DIFF
--- a/.changeset/few-worms-talk.md
+++ b/.changeset/few-worms-talk.md
@@ -1,0 +1,7 @@
+---
+"@graphql-codegen/visitor-plugin-common": patch
+---
+
+fix: allow mappers to override root types (Query, Mutation, Subscription)
+
+Previously, mappers configured for root types were ignored because root types were checked before mappers. This fix moves the mapper check before the root type check, allowing mappers to override `rootValueType` when configured.


packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -971,15 +971,20 @@ export class BaseResolversVisitor<
       const isScalar = this.config.scalars[typeName];
       const hasDefaultMapper = !!this.config.defaultMapper?.type;
 
+      // Check for mappers first, even for root types, to allow overriding rootValueType
+      if (isMapped && this.config.mappers[typeName].type && !hasPlaceholder(this.config.mappers[typeName].type)) {
+        this.markMapperAsUsed(typeName);
+        prev[typeName] = applyWrapper(this.config.mappers[typeName].type);
+
+        return prev;
```
eddeee888, line 979: This `return prev` early could change the existing logic. 
Previously, the code would continue to run after this `if` block. There are a few blocks of code after this if/else AFAICT

realies: The early return is safe because:

1. We only return early when `!hasPlaceholder(...)`, so the mapper is complete and doesn't need further processing.
2. The root type check already returns early (line 985), so this keeps behavior consistent.
3. Subsequent code (lines 1022-1063) only processes placeholders or unmapped types, which we've already excluded.

For root types, returning early is necessary to prevent `rootValueType` from being applied after the mapper is set. The original code would set `prev[typeName]` and continue, but subsequent code wouldn't modify it for mapped types without placeholders anyway.

PR #10525, opened 2025-11-27
[typescript-operations] Add missing enum tests & fix enum generating logic for standalone approach

eddeee888:
## Description

This is the continuation of https://github.com/dotansimha/graphql-code-generator/pull/10508. We merged that PR early to unblock the enum-in-input scenario

Related https://github.com/dotansimha/graphql-code-generator/pull/10496

## Type of change

- [x] Add tests

## How Has This Been Tested?

- [x] Unit test

PR #10527, opened 2025-12-02
Generate input types and output enums into target file

ikusakov2:
## Description

Generates input types into target files.
Generates output enums into target files.

Related # https://github.com/dotansimha/graphql-code-generator/pull/10496

## Type of change

## How Has This Been Tested?

- [x] Unit testing

### DIFF
--- a/packages/plugins/typescript/operations/src/visitor.ts
+++ b/packages/plugins/typescript/operations/src/visitor.ts
@@ -242,23 +242,11 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
     _path?: Array<string | number>,
     ancestors?: Array<TypeDefinitionNode>
   ): string {
-    const oneOfDetails = (function parseOneOf(
-      schema: GraphQLSchema
-    ): { isOneOfInputValue: true; realParentDef: TypeDefinitionNode } | { isOneOfInputValue: false } {
-      const realParentDef = ancestors?.[ancestors.length - 1];
-      if (realParentDef) {
-        const parentType = schema.getType(realParentDef.name.value);
-        if (isOneOfInputObjectType(parentType)) {
-          if (node.type.kind === Kind.NON_NULL_TYPE) {
-            throw new Error(
-              'Fields on an input object type can not be non-nullable. It seems like the schema was not validated.'
-            );
-          }
-          return { isOneOfInputValue: true, realParentDef };
-        }
-      }
-      return { isOneOfInputValue: false };
-    })(this._schema);
+    const oneOfDetails = parseOneOfInputValue({
+      node,
+      schema: this._schema,
+      ancestors,
+    });
 
     // 1. Flatten GraphQL type nodes to make it easier to turn into string
     // GraphQL type nodes may have `NonNullType` type before each `ListType` or `NamedType`
@@ -270,34 +258,11 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
     // - [Thing!]
     // - [Thing]!
     // - [Thing!]!
-    const typeNodes: Array<
-      { type: 'ListType'; isNonNullable: boolean } | { type: 'NamedType'; isNonNullable: boolean; name: string }
-    > = [];
-    (function collectAndFlattenTypeNodes({
-      currentTypeNode,
-      isPreviousNodeNonNullable,
-    }: {
-      currentTypeNode: TypeNode;
-      isPreviousNodeNonNullable: boolean;
-    }): void {
-      if (currentTypeNode.kind === Kind.NON_NULL_TYPE) {
-        const nextTypeNode = currentTypeNode.type;
-        collectAndFlattenTypeNodes({ currentTypeNode: nextTypeNode, isPreviousNodeNonNullable: true });
-      } else if (currentTypeNode.kind === Kind.LIST_TYPE) {
-        typeNodes.push({ type: 'ListType', isNonNullable: isPreviousNodeNonNullable });
-
-        const nextTypeNode = currentTypeNode.type;
-        collectAndFlattenTypeNodes({ currentTypeNode: nextTypeNode, isPreviousNodeNonNullable: false });
-      } else if (currentTypeNode.kind === Kind.NAMED_TYPE) {
-        typeNodes.push({
-          type: 'NamedType',
-          isNonNullable: isPreviousNodeNonNullable,
-          name: currentTypeNode.name.value,
-        });
-      }
-    })({
+    const typeNodes: Parameters<typeof collectAndFlattenTypeNodes>[0]['typeNodes'] = [];
+    collectAndFlattenTypeNodes({
       currentTypeNode: node.type,
       isPreviousNodeNonNullable: oneOfDetails.isOneOfInputValue, // If the InputValue is part of @oneOf input, we treat it as non-null (even if it must be null in the schema)
+      typeNodes,
     });
 
     // 2. Generate the type of a TypeScript field declaration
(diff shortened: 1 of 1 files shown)


packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -168,6 +180,7 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
     );
     this._declarationBlockConfig = {
       ignoreExport: this.config.noExport,
+      enumNameValueSeparator: ' =',
```
ikusakov2, line 188: I experimented with different enumType, this value appears to be required for other enumTypes to work.

packages/plugins/typescript/operations/src/index.ts
```diff
@@ -25,25 +25,14 @@ export const plugin: PluginFunction<TypeScriptDocumentsPluginConfig, Types.Compl
 
   const operationsResult = oldVisit(allAst, { leave: visitor });
 
-  let operationsContent = operationsResult.definitions.join('\n');
+  const operationsDefinitions = operationsResult.definitions;
```
ikusakov2, line 29: I rearranged/simplified the code here - please check.

packages/plugins/typescript/operations/src/index.ts
```diff
@@ -25,25 +25,14 @@ export const plugin: PluginFunction<TypeScriptDocumentsPluginConfig, Types.Compl
...
       if ('name' in d) {
-        exportConsts.push(`export declare const ${d.name.value}: import("graphql").DocumentNode;`);
+        operationsDefinitions.push(`export declare const ${d.name.value}: import("graphql").DocumentNode;`);
       }
     }
-
-    operationsContent = operationsResult.definitions.concat(exportConsts).join('\n');
-  }
-
-  if (config.globalNamespace) {
```
ikusakov2, line 43: I moved this block to the end - to include all the generated code.

packages/plugins/typescript/operations/tests/ts-documents.standalone.spec.ts
```diff
@@ -127,6 +139,260 @@ describe('TypeScript Operations Plugin - Standalone', () => {
...
+
+      export type UsersWithScalarInputQuery = { __typename?: 'Query', users:
+          | { __typename?: 'UsersResponseOk', result: Array<{ __typename: 'User' }> }
+          | { __typename: 'ResponseError' }
+         };
+      "
+    `);
+  });
+
+  it('try different ways to generate enums', async () => {
```
eddeee888: Oops sorry! I should have mentioned I'm working on the tests here: https://github.com/dotansimha/graphql-code-generator/pull/10525

Since similar tests have been moved to `standalone.enum.spec.ts`, shall we remove the duplicates in this PR?

packages/plugins/typescript/operations/src/visitor.ts
eddeee888: Yes you are right! I've got a similar fix in my follow-up Enum PR here: https://github.com/dotansimha/graphql-code-generator/pull/10525/files#diff-d0b5088bd638fefba8ba49f14b488769f115ef1daff777afbe718a4c5ecdf111R172

eddeee888:
Sorry! I caused some conflicts in this PR after merging https://github.com/dotansimha/graphql-code-generator/pull/10525 🙏

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -196,6 +209,92 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
...
+      .withContent(`\n` + (node.fields || []).join('\n  |'));
+  }
+
+  private isValidVisit(ancestors: any): boolean {
+    const currentVisitContext = this.getVisitorKindContextFromAncestors(ancestors);
+    const isVisitingInputType = currentVisitContext.includes(Kind.INPUT_OBJECT_TYPE_DEFINITION);
+    const isVisitingEnumType = currentVisitContext.includes(Kind.ENUM_TYPE_DEFINITION);
+
+    return isVisitingInputType || isVisitingEnumType;
+  }
```
eddeee888: With utility functions that can be shared between `typescript-operations` and `base-types-visitor` like these, I wonder if we should abstract them into `visitor-plugin-common` now or in a following PR? 🤔

Having them in `visitor-plugin-common` helps ensure they'd behave the same all the time. I've done the same for Enum utilities here : https://github.com/dotansimha/graphql-code-generator/pull/10525/files#r2585220345

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -243,6 +343,24 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
       },
     });
 
+    // Collect output enums
+    const typeInfo = new TypeInfo(schema);
+    visit(
+      documentNode,
+      visitWithTypeInfo(typeInfo, {
```
eddeee888, line 473: Huh, this is my first time learning about this. Could you help me with the TLDR; of `TypeInfo` and how it helps us here?

packages/plugins/typescript/operations/src/visitor.ts
ikusakov2: Sure, I'll move them to `visitor-plugin-common`!

Maybe after I fix all the bugs that I discover while running over the yelp codebase?
I am working on a quite tricky bug now with enum not being rendered from a fragment in a separate document.

packages/plugins/typescript/operations/src/visitor.ts
ikusakov2: I am no expert on this as well. I relied on `claude` AI for this piece of code.
Having said that, here is what I think this code is doing:

* It looks like an operation’s AST doesn’t include field types - only names. 
* `TypeInfo` is a stateful helper that tracks typing context while you walk the AST (https://www.graphql-js.org/api-v16/utilities/#typeinfo).
* `visitWithTypeInfo` wires that context into a visitor. I have not found a working docs link on `visitWithTypeInfo`, I guess our best bet is the docs in the source code which is located at the same place as `TypeInfo`: https://github.com/graphql/graphql-js/blob/next/src/utilities/TypeInfo.ts.

packages/plugins/typescript/operations/tests/ts-documents.standalone.spec.ts
ikusakov2: Sure! Extra tests removed.
We can probably rearrange tests in a better way - to avoid this file getting cluttered.

packages/plugins/typescript/operations/src/visitor.ts
eddeee888: Sure! No rush. 
I also put a `FIXME` or `TODO` on these utility functions in my PRs to refactor later

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -198,6 +214,186 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
...
+    }
+
+    return new DeclarationBlock(this._declarationBlockConfig)
+      .asKind('type')
+      .withName(this.convertName(node))
+      .withComment(node.description?.value)
+      .withBlock((node.fields || []).join('\n')).string;
+  }
+
+  InputValueDefinition(
```
eddeee888, line 238: In `typescript` plugin, the logic to generate a Input Value string is spread across multiple methods:
- `InputObjectTypeDefinition`
- `NamedType`
- `ListType`
- `NonNull`
- And a couple of private methods

That was very confusing to both maintainers and contributors, and results in `as any as string` in many places.

This version brings all logic into one place method. 
I'll see if this could be re-used between both plugins

ikusakov2: Excellent! 
I like that you replace all these functions with just one!

However, I am bit concerned about IIFE that you do here. Because this code runs in a hot loop, there is a chance for this code to cause performance issue. The chance is low (it should be optimized by JIT), but still...

Is there a way to write this code without IIFE? In the worst case - we can split it into 3 functions (it is still better than the original version).

eddeee888: Good call! I've replaced the IIFEs with pure functions in https://github.com/dotansimha/graphql-code-generator/pull/10527/changes/9cea149eca3ab6523fe52a634e7ae9f6bac51c3c

ikusakov2: Great! Yes, this look safer. 
Thanks!

Issue #10502, opened 2025-11-14
Preset: near-operation-file `extractAllFieldsToTypes` handling of nested interfaces in type names 


Georgegriff:
### Which packages are impacted by your issue?

@graphql-codegen/typescript-operations

### Describe the bug

I've been trying to workthrough the hell that is migrating off the apollo codegen, we can get most of the way there with `near-operation-file` adjusting some imports and using `extractAllFieldsToTypes` but i think i've found a bug with `extractAllFieldsToTypes` and how it handles nested interfaces - types that have the same fields present.

I've created a reproduction repo: [https://github.com/Georgegriff/graphql-codegen-repro/tree/main](<https://github.com/Georgegriff/graphql-codegen-repro/tree/main>)

Example query

```
export const GET_ANIMALS = gql`
  query GetAnimals {
    animals {
      name
      age
      species
      owner {
        name
        license
      }
    }
  }
```

In this example animal has two types a Cat and Dog (same fields) and Owner has two types Trainer & Veterinarian also with the same fields present.

I am deliberately `not` using `.. on` inline fragments to show the issue.

The types generated end up like this

```ts
export type GetAnimals_animals_Cat_owner_Trainer = { __typename: 'Trainer', name: string, license: string };

export type GetAnimals_animals_Cat_owner_Veterinarian = { __typename: 'Veterinarian', name: string, license: string };

export type GetAnimals_animals_Cat_owner = GetAnimals_animals_Cat_owner_Trainer | GetAnimals_animals_Cat_owner_Veterinarian;

export type GetAnimals_animals_Cat = { __typename: 'Cat', name: string, age: number, species: string, owner: GetAnimals_animals_Cat_owner };

export type GetAnimals_animals_Dog = { __typename: 'Dog', name: string, age: number, species: string, owner: GetAnimals_animals_Cat_owner };

export type GetAnimals_animals = GetAnimals_animals_Cat | GetAnimals_animals_Dog;

export type GetAnimals_Query = { animals: Array<GetAnimals_animals> | null };
```

See how `GetAnimals_animals_Dog` owner field is `GetAnimals_animals_Cat_owner` this is strange because that type is not specific to Cats its a general owner type that is common between the two.

I wonder if in this case the Interface name should be used in the typename? or no typename at all, i'm not sure.

### Your Example Website or App

[https://github.com/Georgegriff/graphql-codegen-repro](<https://github.com/Georgegriff/graphql-codegen-repro>)

[...]

PR #10530, opened 2025-12-04
fix invalid check for inline fragments

Georgegriff:

Fixes bugs introduced by https://github.com/dotansimha/graphql-code-generator/issues/10502

## Description

I realised my previous fix to detect inline fragments was... terrible
image
And i didn't notice because all the tests (now updated in snapshots) had Fragment in name 🤦 

I think this fix is much better and i've added some extra test cases to help be sure that i'm testing

- when fragments are not used
- when inline fragments are used
- when named fragments are used

Because the last fix went out as a major and this just fixes a bug i introduced i think this is fine as a bug fix if the code is good now.

I jumped ahead and opened a pr since its fixing a previous fix.

The bug:
I wasn't correctly looking for fragments so you end up with this 
image
image

duplicate type definitions because they incorrectly try to use the parent interface name which is the behaviour i added but was not supposed to happen when using fragments. this fix rectifies that.

Related # [(issue)](https://github.com/dotansimha/graphql-code-generator/issues/10502)

## Type of change

## How Has This Been Tested?

Bug originally found in private repo, added new specs in pr to cover fixes

**Test Environment**:

- OS:
- `@graphql-codegen/...`:
- NodeJS:

## Checklist:

## Further comments

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -964,7 +964,7 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
 
     const schemaType = this._schema.getType(typeName);
 
-    // Check if current selection set has inline fragments (e.g., "... on AppNotification")
+    // Check if current selection set has fragments (e.g., "... AppNotificationFragment" or "... on AppNotification")
     const hasFragment =
       this._selectionSet?.selections?.some(
         selection => selection.kind === Kind.INLINE_FRAGMENT || selection.kind === Kind.FRAGMENT_SPREAD


packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -962,16 +962,19 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
       return parentName;
     }
 
-    // When the parent schema type is an interface, use the interface name instead of the concrete type
-    // BUT only if we're not inside a fragment (fragments explicitly target specific types)
     const schemaType = this._schema.getType(typeName);
-    const isInFragment = parentName.includes('Fragment');
```
Georgegriff, line 968: i feel bad this super dodgy fix, sorry!

packages/plugins/typescript/operations/tests/extract-all-types.spec.ts
```diff
@@ -433,35 +433,35 @@ describe('extractAllFieldsToTypes: true', () => {
 
       export type ConversationTalkInteractionFragment = { __typename: 'TalkInteraction', channel: string, type: CallType };
 
-      export type ConversationConversationEventFragment_BrokenConversationEvent_originatedFrom_EmailInteraction = { __typename: 'EmailInteraction', originalEmailURLPath: string };
```
Georgegriff, line 436: i now understand why these were changing originally in my original PR and made me panic... they are all wrong, see how they are using a concrete type BrokenConversationEvent for shared types if you look in the schema for this test it doesn't use fragments, so these tests were showing the bug i originally (half) fixed

Georgegriff: This isn't a breaking change on top of my existing breaking change because my "fix" (which caused the bug) didn't happen if you happened to have Fragment in the parent type name 🤦

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
eddeee888: All good! I also missed it 😅

packages/plugins/typescript/operations/tests/extract-all-types.spec.ts
eddeee888: `BrokenConversationEvent` implements both `ConversationEvent` and `ConnectionNode`. I'm wondering if it's showing `ConversationEvent` because it's the first implemented interface? 🤔

eddeee888: Maybe it's ok for now since it could be a rare case 🙂

PR #10531, opened 2025-12-05
fix duplicate typescript keys

Georgegriff:

## Description

See issue https://github.com/dotansimha/graphql-code-generator/issues/10532 for a description

The problem here is that two maps are used for tracking keys `primitiveAliasFields` and `primitiveFields` so they we'ren't aware if you would end up with duplicate keys.

The simplest fix seems to be to just ensure we don't add duplicate fields to the type, since they should always be the same ts type here (i think?) it should be fine... i hope, this does seem like an edge case.. although one we've hit in our production code

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Related # https://github.com/dotansimha/graphql-code-generator/issues/10532

## Type of change

## How Has This Been Tested?

Unit tests

- [ ] Test A
- [ ] Test B

**Test Environment**:

- OS:
- `@graphql-codegen/...`:
- NodeJS:

## Checklist:

## Further comments

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -972,16 +972,19 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
       return parentName;
     }
 
-    // When the parent schema type is an interface, use the interface name instead of the concrete type
-    // BUT only if we're not inside a fragment (fragments explicitly target specific types)
     const schemaType = this._schema.getType(typeName);
-    const isInFragment = parentName.includes('Fragment');
-    if (
-      isObjectType(schemaType) &&
-      this._parentSchemaType &&
-      isInterfaceType(this._parentSchemaType) &&
-      !isInFragment
-    ) {
+
+    // Check if current selection set has fragments (e.g., "... AppNotificationFragment" or "... on AppNotification")
+    const hasFragment =
+      this._selectionSet?.selections?.some(
+        selection => selection.kind === Kind.INLINE_FRAGMENT || selection.kind === Kind.FRAGMENT_SPREAD
+      ) ?? false;
+
+    // When the parent schema type is an interface:
+    // - If we're processing inline fragments, use the concrete type name
+    // - If we're processing the interface directly, use the interface name
+    // - If we're in a named fragment, always use the concrete type name
+    if (isObjectType(schemaType) && this._parentSchemaType && isInterfaceType(this._parentSchemaType) && !hasFragment) {
       return `${parentName}_${this._parentSchemaType.name}`;
     }
 
--- a/packages/plugins/typescript/operations/tests/extract-all-types.spec.ts
+++ b/packages/plugins/typescript/operations/tests/extract-all-types.spec.ts
@@ -433,35 +433,35 @@ describe('extractAllFieldsToTypes: true', () => {
 
       export type ConversationTalkInteractionFragment = { __typename: 'TalkInteraction', channel: string, type: CallType };
 
-      export type ConversationConversationEventFragment_BrokenConversationEvent_originatedFrom_EmailInteraction = { __typename: 'EmailInteraction', originalEmailURLPath: string };
+      export type ConversationConversationEventFragment_ConversationEvent_originatedFrom_EmailInteraction = { __typename: 'EmailInteraction', originalEmailURLPath: string };
 
-      export type ConversationConversationEventFragment_BrokenConversationEvent_originatedFrom_CustomChannelInteraction = { __typename: 'CustomChannelInteraction', externalId: string, timestamp: string, resourceType: string };
+      export type ConversationConversationEventFragment_ConversationEvent_originatedFrom_CustomChannelInteraction = { __typename: 'CustomChannelInteraction', externalId: string, timestamp: string, resourceType: string };
 
-      export type ConversationConversationEventFragment_BrokenConversationEvent_originatedFrom_TalkInteraction = { __typename: 'TalkInteraction' };
+      export type ConversationConversationEventFragment_ConversationEvent_originatedFrom_TalkInteraction = { __typename: 'TalkInteraction' };
 
-      export type ConversationConversationEventFragment_BrokenConversationEvent_originatedFrom_NativeMessagingInteraction = { __typename: 'NativeMessagingInteraction', conversationId?: string | null };
+      export type ConversationConversationEventFragment_ConversationEvent_originatedFrom_NativeMessagingInteraction = { __typename: 'NativeMessagingInteraction', conversationId?: string | null };
 
-      export type ConversationConversationEventFragment_BrokenConversationEvent_originatedFrom_WhatsAppInteraction = { __typename: 'WhatsAppInteraction', conversationId?: string | null };
+      export type ConversationConversationEventFragment_ConversationEvent_originatedFrom_WhatsAppInteraction = { __typename: 'WhatsAppInteraction', conversationId?: string | null };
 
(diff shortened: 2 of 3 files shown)


packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -583,11 +583,21 @@ export class SelectionSetToObject<Config extends ParsedDocumentsConfig = ParsedD
             }
             linkFieldSelectionSets.set(fieldName, linkFieldNode);
           } else if (selectionNode.alias) {
-            primitiveAliasFields.set(selectionNode.alias.value, selectionNode);
+            const outputName = selectionNode.alias.value;
+            // Check if a field with the same output name already exists
+            // (either as an alias or as a regular field)
+            if (!primitiveFields.has(outputName)) {
```
Georgegriff, line 589: the two maps primitiveFields and primitiveAlias fields are the root cause of the problem here i've updated them so they are now aware of each other which fixes the bug

packages/plugins/typescript/operations/tests/extract-all-types.spec.ts
```diff
@@ -1747,4 +1747,76 @@ describe('extractAllFieldsToTypes: true', () => {
...
+          id
+          name
+          ...FragmentA
+          ...FragmentB
+        }
+      }
+      fragment FragmentA on Item {
+        owner {
+          id: realId
+          name
```
eddeee888, line 1779: I _think_ this'd result in a GraphQL Validation error, so I wonder if we need to cater for it?

Image

If my test setup is wrong, let me know how to fix it 🙂

Georgegriff: let me try out schema and double check if you're right and our code is just letting us do somethig invalid, its either that or my reproduction case isn't quite right. Hopefully you can see from the code itself there might be a problem though where variable names get duplicated, i'll work on the repro case

Georgegriff: Yeah, i just looked out the output of our deprecated Apollo codegen and frustratingly, they let us get away with this 
image 
with a warning but fixed the types silently

I guess this likely would with this library:
- have no warning
- break the types (not so silently but not obviously)


both aren't ideal, is there a way to get the warnings from validation for the presets?

Georgegriff: I have skip validation config setting on, that makes sense why this isn't breaking - guess the question is if this library should only generate valid typescript when validation is enabled, or should try to ensure it always generates valid code?

Georgegriff: interestingly, i cannot get a validation error for this specific error in this tooling, it does not detect it:
- old tooling detects it
- graphiql detect it
- codegen tooling doesn't pick up the validation error

Georgegriff: I've reproduced a bug where the validation errors are swallowed in the codegen core. I'll raise a separate issue for that when I have identified it.

As for this issue @eddeee888, Given `you can disable schema validation in the config settings` and still generate code, i think it might be worth fixing the bug still, because you could disable validation and then generate invalid typescript

eddeee888: I'm also some detection issue in VSCode LSP as well 😅 

| Error detected | Error not detected |
| -- | -- |
| Screenshot 2025-12-10 at 11 05 37 pm | Screenshot 2025-12-10 at 11 08 09 pm |

It can detect the error if the fragments are declared in the same block as the operation.

---

In general, Codegen should only try to generate for valid documents. If we generated for invalid documents, it may give users a false sense of safety and may cause errors.

I'm really interested in the swallowed error though 👀

Georgegriff: @eddeee888 i'm not sure your example quite shows the issue i ran into, the issue i can reproduce the fragments are present in the document, not a seperate document, same with the new test i wrote. This test does indeep have validation errors, but i'm just unsure that because you can disable graphql validation rules, how should the typescript codegen behave (apollo codegen allowed this - with warnings - as the dodgy invalid code is from our codebase - i'm trying to fix the validation errors)

As for validation errors that get missed by the code gen code

Try this:



```
type User {
  id: ID
  uuid: String
  name: String
}

type Task {
  id: ID
  assignedTo: User
  isAssignedToMe: Boolean
}

type Workflow {
  id: ID
  status: String
  nextTask: Task
}

type Project {
  id: ID
  title: String
  myWorkflow: Workflow
}

type Query {
  project(id: ID!): Project
}

```

```
query GetProject($id: ID!) {
  project(id: $id) {
    id
    ...ProjectData
  }
}

fragment ProjectData on Project {
  id
  myWorkflow {
    id
    nextTask {
      id
      assignedTo {
        id: uuid
        name
      }
    }
  }
  ...GetActionButtonsData
}

fragment GetActionButtonsData on Project {
  id
  ...IsAssignedToMeData
}

fragment IsAssignedToMeData on Project {
  id
  myWorkflow {
    id
    nextTask {
      id
      assignedTo {
        id
        name
      }
    }
  }
}
```

```
schema: scripts/test-schema.graphql
documents: scripts/test-query.graphql
generates:
  scripts/test-output.ts:
    plugins:
      - typescript
      - typescript-operations
config:
  skipDocumentsValidation: false
```

Hopefully it's right, I had to anonymise it from an example in our code. graphql() validae picks it up, so does `@graphql-tools/utils` i think it something about how the documents are passed in

image

In the above example if you stop using the intermediary fragment `GetActionButtonsData` and put `IsAssignedToMeData` directly in `ProjectData` you get an error

image

hopefully i dont have the example wrong, i was still working on it

eddeee888: Sorry for the delay, I'm looking at this now

eddeee888: Hi @Georgegriff , I'm trying the `skipDocumentsValidation: false` option but it doesn't seem to work 🤔 

Screenshot 2025-12-16 at 11 26 45 pm

I've created a PR for reprod here: https://github.com/eddeee888/graphql-server-template/pull/142
Could you help me find what I'm doing wrong?

Georgegriff: @eddeee888  That's odd, try using the file operation preset from community maybe it hits a different path than client preset or the typescript one 

Failing that I can try to refine the repro case it could be I broke it when I copy pasted and anonymised it from our internal schema

eddeee888: I've set that up here: https://github.com/eddeee888/graphql-server-template/pull/142, but it's still not reproducible 🙂

Georgegriff: @eddeee888 WEIRD, I also cannot reproduce it now, I'm very confused, apologies for the time wasting, if I can somehow try to get back to where I was with the issue, I'll let you know, but can't anymore. I was pretty sure i had a reliable replication case but i can't get it to happen now.

I guess this just leaves the question of the original intent of this PR:

- You can enable `skipDocumentsValidation: true`. It seems this is at odds with the bug in this PR, where it's possible to generate invalid typescript code. If the CLI is only intended to work with valid queries, then I'd question the purpose of `skipDocumentsValidation`. 
Happy either way though, let me know and i can close if need be

eddeee888: Thanks for helping me confirm the reprod repo @Georgegriff ! 
Regarding `skipDocumentsValidation`, I agree it's an odd config 😅 I'll take a look in the future to see if we could drop it.

eddeee888: I'll close this for now 🙂

PR #10536, opened 2025-12-12
Fix website deployment

eddeee888:
## Description

Website deployment has been broken for a while because we were getting rate-limited by NPM:
- Whenever a build happens, every plugin page hits NPM at the same time. (~50 plugins)
- NPM doesn't like it so they gives us "please no spam error" (429 too many requests)

The fix:
- Generate the data for packages before building
- Every call happens sequentially, to avoid NPM rate limiter
- When building, use the cached packages info, instead of fetching from NPM

### DIFF
--- a/website/scripts/generate-packages-info.ts
+++ b/website/scripts/generate-packages-info.ts
@@ -1,4 +1,5 @@
 import { writeFileSync } from 'node:fs';
+import path from 'node:path';
 import semver from 'semver';
 import { PACKAGES } from '../src/lib/plugins/packages.js';
 
@@ -59,6 +60,10 @@ for (let [identifier, pkg] of Object.entries(PACKAGES)) {
   })({ minMs: 100, maxMs: 5000 });
 }
 
-writeFileSync('../src/lib/npm/packages-info.generated.ts', `export const packagesInfo= ${JSON.stringify(result)}`, {
-  encoding: 'utf8',
-});
+writeFileSync(
+  path.join(process.cwd(), 'src', 'lib', 'packages-info.generated.ts'),
+  `export const packagesInfo= ${JSON.stringify(result)}`,
+  {
+    encoding: 'utf8',
+  }
+);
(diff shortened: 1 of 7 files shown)


website/src/components/plugins-marketplace-search.tsx
```diff
@@ -18,24 +19,30 @@ export type Plugin = {
...
-        description,
-        linkHref: `/plugins/${category}/${identifier}`,
-        weeklyNPMDownloads,
-        icon,
-        tags,
-      };
-    })
-  );
+  const plugins: Plugin[] = Object.entries(PACKAGES).map(([identifier, { title, icon, tags }]) => {
+    const packageInfo = packagesInfo[identifier as keyof typeof packagesInfo];
```
eddeee888, line 24: It's a bit dodgy to use `identifier as keyof typeof packagesInfo` here, 
but the identifier comes from the same place, 
and we do a nullish check after, so it is ok use `as ...` here IMO

website/scripts/generate-packages-info.ts
```diff
@@ -0,0 +1,55 @@
...
+            return semver.lte(curr.version, latestVersion);
+          }
+        })?.readme ||
+      '',
+    createdAt: time.created,
+    updatedAt: time.modified,
+    description,
+    weeklyNPMDownloads: downloads,
+  };
+};
```
eddeee888, line 46: As FIXME explains, this is a duplicate of the shared version. I cannot use the shared version in a simple script like this because it has SVG imports and a simple script like this won't be able to deal with that.

website/src/lib/plugins/index.ts
```diff
@@ -106,320 +108,3 @@ export const ALL_TAGS = [
 ] as const;
 
 export type Tag = (typeof ALL_TAGS)[number];
-
-export const PACKAGES: Record<
```
eddeee888, line 110: This is moved to another file because this file contains `.svg` file imports, and the new script `generate-packages-info.ts` won't be able to handle.

PR #10539, opened 2025-12-13
Revert "chore(deps): update dependency @theguild/eslint-config to v0.13.4"

eddeee888:
Reverts dotansimha/graphql-code-generator#10486


eddeee888:
Latest version seems to break Eslint because of flat config 🤔 
Reverting for now.

Issue #10551, opened 2025-12-17
`dedupeFragments` was removed, causing duplicate fragments... any alternatives?


ItaiYosephi:
### Which packages are impacted by your issue?

@graphql-codegen/cli, @graphql-codegen/typed-document-node

### Describe the bug

`dedupeFragments` options was removed [here](https://github.com/dotansimha/graphql-code-generator/pull/10408), causing fragments to appear twice on the document node (`doc.loc?.source.body`)

<img width="908" height="660" alt="Image" src="https://github.com/user-attachments/assets/aaa378f3-d826-4cb8-a914-dae8d487e53a" />

### Your Example Website or App

https://codesandbox.io/p/devbox/inspiring-burnell-l7snd2?workspaceId=ws_3d3S1CmTc7BCJtbnRB7BQC

### Steps to Reproduce the Bug or Issue

1. go to https://codesandbox.io/p/devbox/inspiring-burnell-l7snd2?workspaceId=ws_3d3S1CmTc7BCJtbnRB7BQC
2. run `npm run codegen && npm run print`
3. printing the doc body shows twice the same fragment

### Expected behavior

fragments should be deduped

### Platform

- OS: [e.g. macOS, Windows, Linux]
- NodeJS: [e.g. 18.5.0]
- `graphql` version: [e.g. 16.3.0]
- `@graphql-codegen/*` version(s): [e.g. 2.6.2]

ItaiYosephi:
Hi @eddeee888,
Do you have any updates on this issue? Since `dedupeFragments` was removed, I'm wondering if there's an alternative solution available or if there's a fix planned.
Thanks!

eddeee888:
Hi @ItaiYosephi , I'll take a look. Sorry about the delay 🙏

eddeee888:
Hi @ItaiYosephi , 
Sorry this is a bug that I missed! The intention is to enable the behaviour by default, instead of removing the config option and the behaviour.

Here's the PR to fix that: https://github.com/dotansimha/graphql-code-generator/pull/10554

You can try these alpha version together:
```
@graphql-codegen/visitor-plugin-common@6.2.3-alpha-20251226140101-7572bb12f904056cb4f4e91c2244e9d893cfd9d2
@graphql-codegen/typed-document-node@6.1.6-alpha-20251226140101-7572bb12f904056cb4f4e91c2244e9d893cfd9d2
```

ItaiYosephi:
[...]

PR #10554, opened 2025-12-26
CODEGEN-899 - Dedupe fragments when documentMode=graphQLTag

eddeee888:
## Description

We recently removed a deprecated `dedupeFragments` config option  [here](https://github.com/dotansimha/graphql-code-generator/pull/10408). 
However, I believe the intention was "to enable it by default, therefore the option is not required" 😅

This PR re-applies the deduping logic, without the need of an option. I've also added a test to demonstrate the behaviour for next time.

Related # https://github.com/dotansimha/graphql-code-generator/issues/10551

## Type of change

## How Has This Been Tested?

- [ ] Unit test

### DIFF
--- a/.changeset/@graphql-codegen_client-preset-10554-dependencies.md
+++ b/.changeset/@graphql-codegen_client-preset-10554-dependencies.md
@@ -1,6 +1,0 @@
----
-"@graphql-codegen/client-preset": patch
----
-dependencies updates:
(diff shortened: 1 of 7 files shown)


packages/plugins/typescript/typed-document-node/package.json
```diff
@@ -41,7 +41,7 @@
   },
   "dependencies": {
     "@graphql-codegen/plugin-helpers": "^7.0.0",
-    "@graphql-codegen/visitor-plugin-common": "^7.0.0",
+    "@graphql-codegen/visitor-plugin-common": "workspace:^",
```
eddeee888, line 44: We'll need to do this for other deps as well, will do in another PR to keep changes low here.

PR #10556, opened 2025-12-30
[typescript-operations] Clean up selection set processor that depends on `typescript` plugin (`preResolveTypes: false`)

eddeee888:
## Description

`ts-selection-set-processor.ts` is an old selection set processor that picks from `typescript` plugin types.
Since we no longer have this dependency, it should be safe to clean this up in this PR.

We can also update and remove relevant tests

## Type of change

## How Has This Been Tested?

- [x] Unit test
- [x] Integration test

### DIFF
--- a/packages/plugins/typescript/operations/tests/ts-documents.spec.ts
+++ b/packages/plugins/typescript/operations/tests/ts-documents.spec.ts
@@ -6194,15 +6194,19 @@ function test(q: GetEntityBrandDataQuery): void {
         }
       `);
 
-      const { content } = await plugin(
-        schema,
-        [{ location: '', document: fragment }],
-        { inlineFragmentTypes: 'mask' },
-        { outputFile: 'graphql.ts' }
-      );
+      const content = mergeOutputs([
+        await plugin(
+          schema,
+          [{ location: '', document: fragment }],
+          { inlineFragmentTypes: 'mask' },
+          { outputFile: 'graphql.ts' }
+        ),
+      ]);
 
       expect(content).toMatchInlineSnapshot(`
-        "export type WidgetFragmentFragment = { widgetCount: number, widgetPreference: string } & { ' $fragmentName'?: 'WidgetFragmentFragment' };
+        "type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
+        export type Incremental<T> = T | { [P in keyof T]?: P extends ' $fragmentName' | '__typename' ? T[P] : never };
+        export type WidgetFragmentFragment = { widgetCount: number, widgetPreference: string } & { ' $fragmentName'?: 'WidgetFragmentFragment' };
 
         export type FoodFragmentFragment = { favoriteFood: string, leastFavoriteFood: string } & { ' $fragmentName'?: 'FoodFragmentFragment' };
 


packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -156,58 +108,6 @@ describe('TypeScript Operations Plugin', () => {
...
-        export type TestQueryVariables = Exact<{ [key: string]: never; }>;
-
-
-        export type TestQuery = { f: Types.E | null, user: { id: string, f: Types.E | null, j: any | null } };
-        "
-      `
-      );
-
-      await validate(content, '', [`Cannot find namespace 'Types'.`]);
-    });
```
eddeee888, line 209: This is replaced by new tests such as this: https://github.com/dotansimha/graphql-code-generator/pull/10496/changes#diff-1c0111b8c7f697f38ef4de6e516943d6f07284cfd52aa19305f6a2933d8d4ba7R151-R293

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -34,63 +34,17 @@ describe('TypeScript Operations Plugin', () => {
...
-            )
-            | (
-              Pick<Types.ImageNotification, 'imageUrl' | 'id'>
-              & { metadata: { created: Types.ImageMetadata['createdBy'] } }
-            )
-          > };
-        "
-      `);
-      await validate(content, '', [`Cannot find namespace 'Types'.`]);
-    });
```
eddeee888, line 92: This is replaced by new tests such as this: https://github.com/dotansimha/graphql-code-generator/pull/10496/changes#diff-1c0111b8c7f697f38ef4de6e516943d6f07284cfd52aa19305f6a2933d8d4ba7R151-R293

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
eddeee888, line 214: TODO: should this be

```
nicknames?: Array<string | 'specialType'> ? 
```

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -365,36 +264,6 @@ describe('TypeScript Operations Plugin', () => {
...
-        }
-      `);
-      const config = { preResolveTypes: false };
-      const { content } = await plugin(testSchema, [{ location: 'test-file.ts', document: doc }], config, {
-        outputFile: '',
-      });
-      expect(content).toContain(`Pick<User, 'id' | 'joinDate'>`);
-      await validate(content);
-    });
-  });
```
eddeee888, line 396: This entirely tests the `preResolveTypes: false`  approach, so it's ok to remove.

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -1217,38 +988,18 @@ export type Q2Query = { search: Array<
...
-    it('Should add __typename as optional when its not specified', async () => {
-      const ast = parse(/* GraphQL */ `
-        query {
-          dummy
-        }
-      `);
-      const config = { preResolveTypes: false };
-      const { content } = await plugin(schema, [{ location: 'test-file.ts', document: ast }], config, {
-        outputFile: '',
-      });
```
eddeee888, line 1246: This test is no longer the expected behaviour. So it can be removed.

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -1261,25 +1012,25 @@ export type Q2Query = { search: Array<
...
-          { __typename: 'Query' }
-          & Pick<Query, 'dummy'>
-        );
+        export type Unnamed_1_Query = { __typename: 'Query', dummy: string | null };
         "
       `);
       await validate(content);
     });
 
-    it('Should add __typename correctly when unions are in use', async () => {
```
eddeee888, line 1282: Changing this test to test correct `__typename` generation instead. 

Previously, `__typename` would be optional so it'd appear in the generated template

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
```diff
@@ -1293,24 +1044,27 @@ export type Q2Query = { search: Array<
...
+            | { __typename: 'User', id: string }
+            | { __typename: 'Profile', age: number | null }
+           | null };
         "
       `);
       await validate(content);
     });
 
-    it('Should add __typename correctly when interfaces are in use', async () => {
+    it('Should add __typename correctly when interfaces are in use and nonOptionalTypename=true', async () => {
```
eddeee888, line 1066: Similar to https://github.com/dotansimha/graphql-code-generator/pull/10556/changes#r2656325472

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
eddeee888: This matches previous test output, so keeping as-is

PR #10560, opened 2026-01-05
Adjust config inheritance to reflect usage and type-hint better

eddeee888:
## Description

This PR re-adjusts a few config options that sits on different visitors incorrectly.

We have 4 main types of visitors:
1. base-visitor: This is the parent visitor that the following 3 extend. Every config option here ideally should be usable to the other 3. An example i naming convention of types, so the other 3 can have a consistent way of referring to the same schema type if they need to be used together.
2. documents-visitor: This is the visitor for client use cases e.g. operations.
3. types-visitor: This is the visitor to generate schema types. Usable with resolvers-visitor
4. resolvers-visitor: This is the visitor that uses the types from types-visitor for further extension

Note: documents-visitor includes `base-documents-visitor` and `typscript-operations` visitor. I've rolled these two up into one type here for simplicity

The updated config options are:
- `skipTypename` (turned into `addTypename` internally): previously lives in the base-visitor. However, it is only a types-visitor concern. documents-visitor has its own way to enforce typename, and resolvers-visitor has its own .`resolversTypename` already.
   - This option has been moved from base-visitor to types-visitor
- `nonOptionalTypename`: previously lives in the base-visitor. However, it is a documents-visitor and types-visitor concern.
   - Moved from base-visitor there.
- `extractAllFieldsToTypes`: is a documents-visitor concern as it is only used to generate operation types
   - Moved from base-visitor to documents-visitor
- `enumPrefix` and `enumSuffix`: need to be in base-visitor as all 3 types of visitors need this to correctly sync the enum type names. 
     - Logic related to this is moved from 3 children to the base visitor
- `ignoreEnumValuesFromSchema`: is a documents-visitor and types-visitor concern. 
     - Moved from base-visitor there.
- `globalNamespace`: is a documents-visitor concern. 
     - Moved from base-visitor there

Other refactors:
- documents-visitor no longer extends types-visitor _option types_ as they have two distinct usages now. The types now extend base-visitor types. This is now consistent with documents-visitor extending base-visitor
- Classes now handle config parsing and types at the same level 
     - e.g. if typescript-operations plugin parses configOne, then the types for configOne must be in that class, rather than in base-documents-visitor

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/.changeset/brown-things-jump.md
+++ b/.changeset/brown-things-jump.md
@@ -1,0 +1,27 @@
+---
+'@graphql-codegen/visitor-plugin-common': major
+'@graphql-codegen/typescript-operations': major
+'@graphql-codegen/typescript': major
+'@graphql-codegen/typescript-resolvers': major
+---
+
+BREAKING CHANGE: visitors' config option are moved based on their use case
+
+- addTypename/skipTypename: is only a types-visitor concern. This is moved to types-visitor from base-visitor
+- nonOptionalTypename: is a documents-visitor and types-visitor concern. Moved from base-visitor there
+- extractAllFieldsToTypes: is a documents-visitor concern. Moved from base-visitor there
+- enumPrefix and enumSuffix: need to be in base-visitor as all 3 types of visitors need this to correctly sync the enum type names. This is moved to base visitor
+- ignoreEnumValuesFromSchema: is a documents-visitor and types-visitor concern. Moved from base-visitor there.
+- globalNamespace: is a documents-visitor concern. Moved from base-visitor there
+
+Refactors
+
+- documents-visitor no longer extends types-visitor _option types_ as they have two distinct usages now. The types now extend base-visitor types. This is now consistent with documents-visitor extending base-visitor
+- Classes now handle config parsing and types at the same level e.g. if typescript-operations plugin parses configOne, then the types for configOne must be in that class, rather than in base-documents-visitor
+
+Note: These visitors are rolled up into one type for simplicity
+
+- base-visitor: includes `base-visitor`
+- documents-visitor: includes `base-documents-visitor` and `typescript-operations` visitor
+- types-visitor: includes `base-types-visitor` and `typescript` visitor
+- resolvers-visitor: includes `base-resolvers-visitor` and `typescript-resolvers` visitor


packages/plugins/other/visitor-plugin-common/src/base-documents-visitor.ts
```diff
@@ -380,3 +400,15 @@ export class BaseDocumentsVisitor<
...
+  switch (operation) {
+    case 'query':
+      return schema.getQueryType();
+    case 'mutation':
+      return schema.getMutationType();
+    case 'subscription':
+      return schema.getSubscriptionType();
+  }
+  throw new Error(`Unknown operation type: ${operation}`);
+}
```
eddeee888, line 414: I drived-by moved this function here to make the first bit of this file the related to visitor. This is done fairly consistently across the codebase now to help readability and navigation

packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts
```diff
@@ -795,8 +741,6 @@ export class BaseResolversVisitor<
     super(rawConfig, {
       immutableTypes: getConfigValue(rawConfig.immutableTypes, false),
       optionalResolveType: getConfigValue(rawConfig.optionalResolveType, false),
-      enumPrefix: getConfigValue(rawConfig.enumPrefix, true),
-      enumSuffix: getConfigValue(rawConfig.enumSuffix, true),
```
eddeee888, line 799: As mentioned in the description. This is moved to `base-visitor` since all 3 children visitors need it.

packages/plugins/other/visitor-plugin-common/src/selection-set-processor/base.ts
```diff
@@ -10,8 +10,8 @@ export type ProcessResult = null | Array<NameAndType | string>;
 export type SelectionSetProcessorConfig = {
   namespacedImportName: string | null;
   convertName: ConvertNameFn<any>;
-  enumPrefix: boolean | null;
-  enumSuffix: boolean | null;
+  enumPrefix: boolean;
+  enumSuffix: boolean;
```
eddeee888, line 14: `typescript-operations` (the main user of this file)  now coerce null -> false, so we don't have to handle null case here.

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -502,9 +502,6 @@ export class SelectionSetToObject<
         // The keys here will be used to generate intermediary
         // fragment names. To avoid blowing up the type name on large
         // unions, calculate a stable hash here instead.
-        //
-        // Also use fragment hashing if skipTypename is true, since we
-        // then don't have a typename for naming the fragment.
```
eddeee888, line 507: This is no longer relevant as we no longer have `skipTypename` for client use cases.

packages/plugins/typescript/operations/src/config.ts
```diff
@@ -186,7 +186,6 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
    * ```
    */
   noExport?: boolean;
-  globalNamespace?: boolean;
```
eddeee888, line 189: This is already set in `base-documents-visitor` so we don't have to re-declare it here.

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -140,10 +142,10 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
       },
       printFieldsOnNewLines: this.config.printFieldsOnNewLines,
     };
-    const processor = new PreResolveTypesProcessor(processorConfig);
+
     this.setSelectionSetHandler(
       new SelectionSetToObject(
-        processor,
```
eddeee888, line 146: Inlining the processor here, since we no longer have the complicated ternary.

packages/plugins/typescript/operations/tests/ts-documents.spec.ts
eddeee888, line 1: Most test changes  are related to `skipTypename` being a config option that no longer works for client use cases. 

Note that since [we inlined the config option in the last PR](https://github.com/dotansimha/graphql-code-generator/pull/10556), we start to see TypeScript complaining about unexpected config options. This makes cleanups like this a lot easier.

packages/plugins/other/visitor-plugin-common/src/base-documents-visitor.ts
```diff
@@ -7,43 +7,31 @@ import {
...
-      return schema.getQueryType();
-    case 'mutation':
-      return schema.getMutationType();
-    case 'subscription':
-      return schema.getSubscriptionType();
-  }
-  throw new Error(`Unknown operation type: ${operation}`);
-}
-
-export interface ParsedDocumentsConfig extends ParsedTypesConfig {
```
eddeee888, line 30: Previously, `documents-visitor` extends `types-visitor` config types, however, `BaseDocumentsVisitor` extends `BaseVisitor`, instead of `BaseTypesVisitor`.

This may have created some confusion (at least for me) of which option is used for which use case. This change correctly reflects the inheritance path for the visitors.

PR #10564, opened 2026-01-07
Remove unused utility types and functions

eddeee888:
## Description

This PR removes unused utitlity types from `typescript` plugin because they were only used in `typescript-operations` 
- `MakeOptional`
- `MakeMaybe`
- `MakeEmpty`
- `Incremental`

Note: `MakeOptional`, `MakeMaybe` and `MakeEmpty` were used in a code path that has been removed from `typescript-operations`, so we also don't add them there.

Also: remove utility function `getRootTypeNames` as it's not used in this codebase directly and is availabel in `@graphql-tools/utils`

### DIFF
--- a/.changeset/twelve-trams-pump.md
+++ b/.changeset/twelve-trams-pump.md
@@ -1,0 +1,13 @@
+---
+'@graphql-codegen/visitor-plugin-common': major
+'@graphql-codegen/typescript': major
+---
+
+BREAKING CHANGE: Remove unused utility types from `typescript` plugin as they were previously used for `typescript-operations` plugin:
+
+- `MakeOptional`
+- `MakeMaybe`
+- `MakeEmpty`
+- `Incremental`
+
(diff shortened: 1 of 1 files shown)


packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -418,13 +418,6 @@ function isStringValueNode(node: any): node is StringValueNode {
   return node && typeof node === 'object' && node.kind === Kind.STRING;
 }
 
-// will be removed on next release because tools already has it
-export function getRootTypeNames(schema: GraphQLSchema): string[] {
-  return [schema.getQueryType(), schema.getMutationType(), schema.getSubscriptionType()]
-    .filter(t => t)
-    .map(t => t.name);
-}
```
eddeee888, line 426: Driveby remove as nothing is using this

PR #10565, opened 2026-01-07
Bugfix: enums from external fragments were not generated along with operations

ikusakov2:
## Description

This PR includes 2 fixes:
1) enums from external fragment were not copied to the generated file along with operations. 
2) extractAllFieldsToTypesCompact config option in certain conditions generate duplicate names

Additionally, an integration test is added that mimics the Apollo-tooling setup:
`dev-test-apollo-tooling`

## Testing

- [x] Unit tests


packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -121,7 +121,13 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
       ...(config.externalFragments || []),
     ];
 
-    this._usedNamedInputTypes = this.collectUsedInputTypes({ schema, documentNode });
+    // Create a combined document that includes external fragments for enum collection
+    const documentWithExternalFragments: DocumentNode = {
+      ...documentNode,
+      definitions: [...documentNode.definitions, ...(config.externalFragments || []).map(f => f.node)],
+    };
```
eddeee888: I wonder if the functionality in the above statement used to create `allFragments` could be used for our fragments here somehow? 🤔 
It _seems_ to do a similar function of collecting fragments, either from documents or from config.

dev-test-alpha/.gitignore
```diff
@@ -0,0 +1,2 @@
+**/__generated__/**
```
eddeee888: I feel we should unignore the generation 🙂
This could be a good test that can be kept as snapshot test, and an advanced use case of codegen

dev-test-alpha/cli/index.ts
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env ts-node
```
ikusakov2, line 1: I created the `dev-test-alpha` just as a temporary package to demonstrate our company issues.
Do you think it would be worth keeping it? 
Then:
- I'll need to add some tests which would get invoked when global tests are running. 
- We need to come up with some better name than `dev-test-alpha`

dev-test-alpha/src/__generated__/Component.ts
```diff
@@ -0,0 +1,58 @@
+/* tslint:disable */
+/* eslint-disable */
+// @generated
+// This file was automatically generated and should not be edited.
+
+import type * as Types from '../../graphql-code-generator/dev-test-alpha/src/__generated__/globalTypes';
```
ikusakov2: I think we already worked on removing this line in `typescript-operations` plugin.
But it probably comes from `near-operation-file` preset.
Shall we ignore it until we do the next version of `near-operation-file`?

dev-test-alpha/src/__generated__/Component.ts
```diff
@@ -0,0 +1,58 @@
...
+
+type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
+export type Incremental<T> = T | { [P in keyof T]?: P extends ' $fragmentName' | '__typename' ? T[P] : never };
+export enum UserManagerRoleType {
```
ikusakov2: With the recent fix - it adds the enum in the correct file.

packages/plugins/typescript/operations/src/visitor.ts
ikusakov2: Good call! 
Check the new code - I combined it with allFragments.

dev-test-alpha/.gitignore
ikusakov2: OK, generated files added!

dev-test-alpha/cli/index.ts
eddeee888: Yes, I think so! I find this test is very valuable, as it can be used to catch unexpected bugs, at least in the initial stage after rolling out.

I feel we might be able to simplify a few things in the setup. And worst case, if its usefulness is overshadowed by maintenance, we can decide to drop it (or further simplify) later 🙂

ikusakov2: OK, I renamed `dev-test-alpha` to `dev-test-apollo-tooling` - to give a more descriptive name to what this integration test it about. I also added another (hopefully last) fix - so you can run `yarn install && yarn build` from the root directory: `graphql-code-generator` and - it should work.

dev-test-alpha/src/__generated__/Component.ts
ikusakov2: I did a research and - yes, it comes from `near-operation-file`.

dev-test-apollo-tooling/cli/index.ts
```diff
@@ -0,0 +1,97 @@
...
+// This file was automatically generated and should not be edited.
+`;
+
+/**
+ * The following GraphQL Codegen config matches as closely as possible
+ * to the old apollo-tooling codegen
+ * @see https://github.com/apollographql/apollo-tooling/issues/2053
+ * */
+const GRAPHQL_CODEGEN_CONFIG = {
+  skipTypename: false,
```
eddeee888: This config doesn't exist anymore 🙂

dev-test-apollo-tooling/cli/index.ts
```diff
@@ -0,0 +1,97 @@
...
+  useTypeImports: true,
+  preResolveTypes: true, // Simplifies the generated types
```
eddeee888: `preResolveTypes: true` is the only option now, and it's been removed entirely.
So we can omit it (when we are ready to merge)

dev-test-apollo-tooling/src/__generated__/Helper.ts
```diff
@@ -0,0 +1,35 @@
+/* tslint:disable */
+/* eslint-disable */
+// @generated
+// This file was automatically generated and should not be edited.
+
+import type * as Types from '../unused';
+
+type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
```
eddeee888: (Context/braindump, not a comment on this PR)

I made a TODO about `Exact` being generated but not used in files with only fragments in the [feature PR](https://github.com/dotansimha/graphql-code-generator/pull/10496). Some users' tsconfig.json may have strict types so it'd fail `tsc`.

I believe we are generating `Exact` all the time, but we should only do it if operations have been found in a document. I can take a look into that in a different PR.

package.json
```diff
@@ -84,7 +85,8 @@
   },
   "resolutions": {
     "typescript": "5.5.4",
-    "eslint-plugin-promise": "7.1.0"
+    "eslint-plugin-promise": "7.1.0",
+    "graphql": "16.9.0"
```
eddeee888: I think resolving a particular version of `graphql` is usually done in CI, when we test multiple versions of GraphQL in different actions.

Do we need this for this PR? 🙂

dev-test-apollo-tooling/cli/index.ts
ikusakov2: removed

dev-test-apollo-tooling/cli/index.ts
ikusakov2: removed

dev-test-apollo-tooling/src/__generated__/Helper.ts
eddeee888: This is done in https://github.com/dotansimha/graphql-code-generator/pull/10571

package.json
ikusakov2: No, we don't need it here. I added this dep while experimenting with making the `dev-test-apollo-tooling` to compile as part of the monorepo. But now I realize it is not needed.

yarn.lock
eddeee888, line 1: Image

👆  The previous commit had package issues because there were lots of changes (+4k -2k LOC changes) in `yarn.lock`, which I _think_ caused some package versions to be conflicting.

I've fixed this issue in this [commit](https://github.com/dotansimha/graphql-code-generator/pull/10565/changes/eec25ef25667b8241c77646d3f1017203591409a)
```
git checkout b55828d4ea802db1476fb7eaada41592b9c80985 -- yarn.lock // get working yarn.lock of a stable commit
yarn // re-install packages whilst checking that minimal line changes happened and tests can run
```

PR #10566, opened 2026-01-08
Make `unknown` instead of `any` the default scalar type

eddeee888:
## Description

`any` disables any typechecks and feedback from the TypeScript language server.
`unknown` forces the user to check at runtime, so it's better than `any`.

This PR: 
- changes the default scalar type to `unknown` for the reasons above.
- Input and Variables fields may take `undefined` values if it is nullable

## Type of change

### DIFF
--- a/dev-test/gql-tag-operations/gql/gql.ts
+++ b/dev-test/gql-tag-operations/gql/gql.ts
@@ -15,12 +15,12 @@ import { TypedDocumentNode as DocumentNode } from '@graphql-typed-document-node/
  */
 type Documents = {
   '\n  query Foo {\n    Tweets {\n      id\n    }\n  }\n': typeof types.FooDocument;
-  '\n  fragment Lel on Tweet {\n    id\n    body\n  }\n': typeof types.LelFragmentDoc;
+  '\n  fragment Lel on Tweet {\n    id\n    body\n    date\n  }\n': typeof types.LelFragmentDoc;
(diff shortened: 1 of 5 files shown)


packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -269,7 +267,7 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
         }
 
         typePart = usedInputType.tsType; // If the schema is correct, when reversing typeNodes, the first node would be `NamedType`, which means we can safely set it as the base for typePart
-        if (usedInputType.tsType !== 'any' && !typeNode.isNonNullable) {
```
eddeee888, line 273: When `unknown` and `any` is used, `| null | undefined` is not required.
However, the logic could be complex, so I'll bring back this condition and tests in another PR.

packages/plugins/typescript/operations/src/ts-operation-variables-to-object.ts
```diff
@@ -8,7 +8,7 @@ export const SCALARS = {
   Boolean: 'boolean',
 };
 
-const MAYBE_SUFFIX = ' | null';
+const MAYBE_SUFFIX = ' | null | undefined';
```
eddeee888: @ikusakov2 I'm thinking to handle this in `inputMaybeValue` so it's configurable e.g. some users may want to use `T | null`, others may want to use `T | null | undefined`.

Issue #10506, opened 2025-11-17
`os.cpus()` returns empty array causing CLI generation to fail in some sandbox envs


harvitronix:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

When running `graphql-codegen` in the CLI, the return value of `os.cpus().length` is passed into `task.newListr`. However, in some Sandbox environments, `os.cpus()` returns an empty array, resulting in passing the value `0` into the `concurrent` property of `newListr`. This in turn results in the generation hanging.

The Node JS documentation mentions:
> os.cpus().length should not be used to calculate the amount of parallelism available to an application. Use [os.availableParallelism()](https://nodejs.org/api/os.html#osavailableparallelism) for this purpose.

### Your Example Website or App

https://stackblitz.com/edit/github-ntbvauxb?file=package.json

### Steps to Reproduce the Bug or Issue

1. In the StackBlitz example, I've added a shim to mimic the sandbox response for `os.cpus()`
2. Run `yarn generate` to see the error
3. Change the assignment in `force-no-cpus.js` to `os.cpus = () => ["x"];` and re-run `yarn generate` and you'll see that generation now works, given that `["x"].length` is equal to `1`

### Expected behavior

I expect that the `concurrency` value is set to a minimum of 1, and/or uses the newer `availableParallelism()` function.

### Platform

- OS: macOS in Codex CLI
- NodeJS: v20.19.2
- `graphql` version: 16.9.0
- `@graphql-codegen/cli` version: 5.0.7

### Additional context

I'm happy to submit a PR to address this, if it's agreeable to do so.

etr2460:
@eddeee888 are you planing to take this on? If not i'm happy to make a PR here to address the issue

eddeee888:
Hi @etr2460! 
Please feel free to take this on! This is on my TODO list, but I'm currently working on something else at the moment.
Thank you!

eddeee888:
This is released here: https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1768129895852

Thank you @etr2460 !

PR #10569, opened 2026-01-11
fix(graphql-codegen-cli): Don't hang when 0 CPUs are found

etr2460:
## Description

Fixes generation when 0 CPUs are returned by `os.cpus()`. This happens in sandbox environments. Note that I did not replace `os.cpus()` with `os.availableParallelism()` as that was added in v18 of node and the `package.json` specifies v16 as the min version. This change matches how node v18 and up implements availableParallelism ([source](https://github.com/libuv/libuv/blob/f420c2bd57ba79d21e1797324ed3ca18e8d92f89/src/unix/core.c#L2104-L2107))

Related #10506

## Type of change

## Further comments

As this is a rather trivial change, I didn't add a new unit test for it (would involve mocking `os.cpus` which didn't seem especially useful). I'm happy to add a test if folks feel it's valuable though!

### DIFF
--- a/.changeset/ready-ravens-swim.md
+++ b/.changeset/ready-ravens-swim.md
@@ -1,0 +1,7 @@
+---
+'@graphql-codegen/cli': patch
+---
+
+fix(graphql-codegen-cli): Don't hang when 0 CPUs are found
+
+Fixes generation when 0 CPUs are returned by os.cpus(), which occurs in sandbox environments.


packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -423,7 +423,7 @@ export async function executeCodegen(
             };
           });
 
-          return task.newListr(generateTasks, { concurrent: cpus().length });
+          return task.newListr(generateTasks, { concurrent: cpus().length || 1 });
```
eddeee888, line 426: Good point about having unit tests here may not be useful 🤔 
Even if we mocked `cpus()`, I can't think of a good (or meaningful) way to assert that it works. 

Agree with merging this first, thank you!

PR #10571, opened 2026-01-13
Only generate `Exact` utility type if documents have operations

eddeee888:
## Description

`Exact` utility is only used to wrap variables types for operations (queries, mutations and subscriptions) if they exist in the document. `Exact` is never used for when there are fragments.

This is important to conditionally generate as users may use very strict tsconfig that will fail compiling if there are unused types.

This PR ensures `Exact` is conditionally generated: only generated when there are operations.

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/typescript/operations/tests/ts-documents.standalone.spec.ts
+++ b/packages/plugins/typescript/operations/tests/ts-documents.standalone.spec.ts
@@ -838,22 +838,12 @@ describe('TypeScript Operations Plugin - Standalone', () => {
       }
     `);
 
-    const result = mergeOutputs([
-      await plugin(
-        schema,
-        [{ document }],
-        {
-          skipTypeNameForRoot: true,
-          nonOptionalTypename: true,
-        },
-        { outputFile: '' }
-      ),
-    ]);
+    const result = mergeOutputs([await plugin(schema, [{ document }], {}, { outputFile: '' })]);
 
     expect(result).toMatchInlineSnapshot(`
       "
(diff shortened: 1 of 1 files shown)


packages/plugins/typescript/operations/tests/ts-documents.standalone.spec.ts
```diff
@@ -90,6 +98,23 @@ describe('TypeScript Operations Plugin - Standalone', () => {
...
+      subscription UserChanges {
+        makeUserAdmin(id: "100") {
+          ...UserFragment
+        }
+      }
+
+      fragment UserFragment on User {
+        id
+        role
+      }
```
eddeee888, line 117: Adding these to make sure we test many things in the default test. 
This gives more confidence  overall.
e.g. in this PR I want to make sure `Exact` is used for mutation and subscription variables.

packages/plugins/typescript/operations/tests/ts-documents.standalone.spec.ts
```diff
@@ -712,17 +749,6 @@ describe('TypeScript Operations Plugin - Standalone', () => {
...
-      type ResponseError {
-        error: ResponseErrorType!
-      }
-
-      enum ResponseErrorType {
-        NOT_FOUND
-        INPUT_VALIDATION_ERROR
-        FORBIDDEN_ERROR
-        UNEXPECTED_ERROR
-      }
```
eddeee888, line 724: Extraneous and never used in this test, so removed.

PR #10572, opened 2026-01-15
Decouple `typescript-operations` plugin from `typescript` plugin

eddeee888:
## Description

Previously, `TypeScriptOperationVariablesToObject` from `typescript-operations` was extending from `typescript` plugin. This made it (1) very hard to read, as we need to jump from base class <-> typescript class <-> typescript-operations class to understand the flow and (2) very hard to evolve the two independently (which is the point of this work).

Since there's not much shared logic anyways, it's simpler to extend the `typescript-operations` class from the base class directly.

## Type of change

- [x] Breaking change

### DIFF
--- a/.changeset/@graphql-codegen_typescript-operations-10572-dependencies.md
+++ b/.changeset/@graphql-codegen_typescript-operations-10572-dependencies.md
@@ -1,0 +1,5 @@
+---
+"@graphql-codegen/typescript-operations": patch
+---
+dependencies updates:
+  - Removed dependency [`@graphql-codegen/typescript@^5.0.7` ↗︎](https://www.npmjs.com/package/@graphql-codegen/typescript/v/5.0.7) (from `dependencies`)


packages/plugins/typescript/operations/package.json
```diff
@@ -15,7 +15,6 @@
   "dependencies": {
     "@graphql-codegen/plugin-helpers": "^6.1.0",
     "@graphql-codegen/schema-ast": "^5.0.0",
-    "@graphql-codegen/typescript": "^5.0.7",
```
eddeee888, line 18: Bye! 🫡 

`@graphql-codegen/typescript` served us well for the last 10+ years. But with goal of creating the independence of `typescript` for `typescript-operations` plugin for this work, we can now remove this dep.

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -169,9 +169,7 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
         this.config.enumPrefix,
         this.config.enumSuffix,
         this.config.enumValues,
-        this.config.arrayInputCoercion,
-        undefined,
-        undefined
```
eddeee888, line 174: Since we are always passing undefined through, we can just use the default.

packages/plugins/typescript/operations/src/ts-operation-variables-to-object.ts
```diff
@@ -10,7 +17,36 @@ export const SCALARS = {
...
+    _scalars: NormalizedScalarsMap,
+    _convertName: ConvertNameFn,
+    private _avoidOptionals: NormalizedAvoidOptionalsConfig,
+    private _immutableTypes: boolean,
+    _namespacedImportName: string | null,
+    _enumNames: string[],
+    _enumPrefix: boolean,
+    _enumSuffix: boolean,
+    _enumValues: ParsedEnumValuesMap,
+    _applyCoercion: boolean
```
eddeee888, line 31: There are lots of default values here from the `typescript` version of this class. 

Since we are really dedicating this implementation for `typescript-operations`, it's a lot easier to force them to be non-optional since we only have one use case to cater for.

We event removed the last param `_maybeType` since we don't use it in `typescript-operations` any more!

packages/plugins/typescript/operations/src/ts-operation-variables-to-object.ts
```diff
@@ -10,7 +17,36 @@ export const SCALARS = {
...
+    super(
+      _scalars,
+      _convertName,
+      _namespacedImportName,
+      _enumNames,
+      _enumPrefix,
+      _enumSuffix,
+      _enumValues,
+      _applyCoercion,
+      {}
```
eddeee888, line 42: This is `_directiveArgumentAndInputFieldMappings`. 

We were receiving `undefined` all the time from the only callsite, so it's ok to just pass through the default value.

PR #10573, opened 2026-01-15
Re-implement `inputMaybeValue` to allow changing Variables and Input nullable types

eddeee888:
## Description

This PR re-implements `inputMaybeValue` for `typescript-operations` use case. 
This was implemented for both Input and Variables previously, but temporarily removed from `typescript-operations` plugin whilst we refactor and decouple the plugins. 

Main change: Instead of using `InputMaybe` wrapper, we now inline it:
- Previously: `InputMaybe<string>`
- Now: `string | null | undefined` (if `type InputMaybe = T | null | undefined`)

Note: This may create a foot-gun situation since users could misuse the option. However, `typescript-operations` is a low-level plugin, and this is not supported by `client-preset`, so this is ok to keep IMO.

## Type of change

### DIFF
--- a/.changeset/fuzzy-poets-doubt.md
+++ b/.changeset/fuzzy-poets-doubt.md
@@ -1,0 +1,5 @@
+---
+'@graphql-codegen/typescript-operations': patch
+---
+
+Re-implement inputMaybeValue


packages/plugins/typescript/operations/src/config.ts
```diff
@@ -251,26 +253,38 @@ export interface TypeScriptDocumentsPluginConfig extends RawDocumentsConfig {
...
+   *         inputMaybeValue: 'T | null'
+   *       }
+   *     }
+   *   }
+   * }
+   * export default config
    * ```
    */
-  maybeValue?: string;
+  inputMaybeValue?: string;
```
eddeee888, line 287: I was debating whether we need to support `inputMaybeValue` as it can be misused and cause runtime errors for users. 

However, I've decided to keep it:
- Since `typescript-operations` is a low-level plugin, we could allow flexibilities for advanced users
- It's previously supported
- It's still not supported by client-preset, which is the main thing users should reach for, so the chance of foot-gunning is lower.

PR #10574, opened 2026-01-17
[typescript-operations] Handle printing "Maybe" types consistently when types are `any` and `unknown`

eddeee888:
## Description

When a TypeScript type is `any` and `unknown`, and we try to add other union options e.g. `any | null | undefined`, they'd be collapsed down to the `any` or `unknown`.

This PR adds `printTypeScriptMaybeType` to consistently print TypeScript Maybe types with these scenarios in mind.

## Type of change

### DIFF
--- a/packages/plugins/typescript/operations/tests/ts-documents.standalone.default-scalar-types.spec.ts
+++ b/packages/plugins/typescript/operations/tests/ts-documents.standalone.default-scalar-types.spec.ts
@@ -88,7 +88,7 @@ describe('TypeScript Operations Plugin - Default Scalar types', () => {
       }>;
 
 
-      export type UserQuery = { user: { id: string, nonNullableDate: unknown, nullableDate: unknown | null } | null };
+      export type UserQuery = { user: { id: string, nonNullableDate: unknown, nullableDate: unknown } | null };
       "
     `);
 
@@ -181,7 +181,7 @@ describe('TypeScript Operations Plugin - Default Scalar types', () => {
       }>;
 
 
-      export type UserQuery = { user: { id: string, nonNullableDate: any, nullableDate: any | null } | null };
+      export type UserQuery = { user: { id: string, nonNullableDate: any, nullableDate: any } | null };
       "
     `);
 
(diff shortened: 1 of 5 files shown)


packages/plugins/typescript/operations/tests/ts-documents.standalone.default-scalar-types.spec.ts
```diff
@@ -70,25 +70,25 @@ describe('TypeScript Operations Plugin - Default Scalar types', () => {
...
         dateArray4: Array<unknown>;
       };
 
       export type UserQueryVariables = Exact<{
         nonNullableDate: unknown;
-        nullableDate?: unknown | null | undefined;
-        dateArray1?: Array<unknown | null | undefined> | unknown | null | undefined;
-        dateArray2: Array<unknown | null | undefined> | unknown;
+        nullableDate?: unknown;
+        dateArray1?: Array<unknown> | unknown | null | undefined;
```
eddeee888, line 83: Ideally this is `unknown` since the `Array<unknown>`, `null` and `undefined` are all collapsed down to unknown. 

However, it's a bit hard to achieve without changing the old implementation. Let's leave this for now, since TypeScript will handle it.

packages/plugins/typescript/operations/tests/ts-documents.standalone.default-scalar-types.spec.ts
```diff
@@ -163,25 +163,25 @@ describe('TypeScript Operations Plugin - Default Scalar types', () => {
...
 
       export type UserQueryVariables = Exact<{
         nonNullableDate: any;
-        nullableDate?: any | null | undefined;
-        dateArray1?: Array<any | null | undefined> | any | null | undefined;
-        dateArray2: Array<any | null | undefined> | any;
+        nullableDate?: any;
+        dateArray1?: Array<any> | any | null | undefined;
+        dateArray2: Array<any> | any;
         dateArray3?: Array<any> | any | null | undefined;
```
eddeee888, line 178: Similar to https://github.com/dotansimha/graphql-code-generator/pull/10574/changes#r2702333885

PR #10579, opened 2026-01-21
[typescript-operations] Cut down `avoidOptionals` to only support `object`, `inputValue` and `defaultValue` (+ minor drive-by fixes)

eddeee888:
## Description

This PR updates `avoidOptionals` config option in `typescript-operations` to only support `object`, `inputValue` and `defaultValue`.

Strictly speaking, this is a type change because the rest of the options did not have any effect on the plugin. However, since it's an interface change, it's a breaking change and it's mentioned so in the changeset

I've also did a couple of drove-by fixes:
- for docs
- for Changeset setup to ignore `dev-test-apollo-tooling`

## Type of change

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -8,6 +8,7 @@
   "ignore": [
     "website",
     "example-*",
+    "dev-test*",
     "@graphql-codegen/client-preset-swc-plugin",
     "example-apollo-client-swc-plugin",
     "example-react-nextjs-swr"


.changeset/config.json
```diff
@@ -8,6 +8,7 @@
   "ignore": [
     "website",
     "example-*",
+    "dev-test*",
```
eddeee888, line 11: Image

Similar to `example` folders, the modules in `dev-test` (which includes the example for Apollo tooling migration), should be ignored, otherwise it'd show up in Changeset interface

packages/plugins/typescript/operations/src/config.ts
```diff
@@ -1,15 +1,13 @@
...
   type EnumValuesMap,
   RawDocumentsConfig,
 } from '@graphql-codegen/visitor-plugin-common';
+import type { AvoidOptionalsConfig } from './config.avoidOptionals';
 
 /**
  * @description This plugin generates TypeScript types based on your GraphQLSchema _and_ your GraphQL operations and fragments.
  * It generates types for your GraphQL documents: Query, Mutation, Subscription and Fragment.
- *
- * Note: In most configurations, this plugin requires you to use `typescript as well, because it depends on its base types.
```
eddeee888, line 12: Removing this makes me feel we are so close now!

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -175,9 +174,6 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
     this.setVariablesTransformer(
       new TypeScriptOperationVariablesToObject(
         {
-          // FIXME: this is the legacy avoidOptionals which was used to make Result fields non-optional. This use case is no longer valid.
-          // It's also being used for Variables so people could already be using it.
-          // Maybe it's better to deprecate and remove, to see what users think.
```
eddeee888, line 180: Let's keep it for now to keep the scope of this project manageable

PR #10582, opened 2026-01-22
[typescript-operations] Rename `avoidOptionals.object` to `avoidOptionals.variableValue` to better reflect the target

eddeee888:
## Description

This PR:
- renames `avoidOptionals.object` to `avoidOptionals.variableValue` to match the format of other options e.g. `inputValue`
- adds tests for `avoidOptionals` for `typescript-operations` (I think this was using tests from `typescript` plugin i.e. it was untested)

## Type of change

### DIFF
--- a/packages/plugins/typescript/operations/tests/ts-documents.standalone.config.avoidOptionals.spec.ts
+++ b/packages/plugins/typescript/operations/tests/ts-documents.standalone.config.avoidOptionals.spec.ts
@@ -12,7 +12,9 @@ describe('TypeScript Operations Plugin - config.avoidOptionals', () => {
 
       input UserInput {
         id: ID!
+        id2: ID! = "default-id"
         legacyId: ID
+        legacyId2: ID = "default-legacy-id"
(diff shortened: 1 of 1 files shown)


packages/plugins/typescript/operations/src/config.avoidOptionals.ts
```diff
@@ -1,9 +1,9 @@
 /**
- * This version of AvoidOptionalsConfig is a cut down version of the type of the same name in `@graphql-codegen/visitor-plugins-common`
+ * This version of AvoidOptionalsConfig is an alternative and cut down version of the type of the same name in `@graphql-codegen/visitor-plugins-common`
  * This version only deal with types available in client use cases.
  */
 export interface AvoidOptionalsConfig {
-  object?: boolean;
+  variableValue?: boolean;
```
eddeee888, line 6: `object` didn't mean much. However, we know that it is used to handle optionality of variables, so I've renamed it `variableValue`

PR #10584, opened 2026-01-25
[typescript-operations] Fix external custom scalars not getting imported

eddeee888:
## Description

This PR fix an issue where external custom scalars are not imported correctly into files:
- if a single file setup is used, external scalars import lines will be generated if used in Input, Variables or Result.
- if multiple file setup is used:
     - operation file will have import lines for said scalars used in Variables or Result 
     - shared type file will have import lines for said scalars used in Input

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/typescript/operations/src/visitor.ts
+++ b/packages/plugins/typescript/operations/src/visitor.ts
@@ -435,10 +435,13 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
   }
 
   public getScalarsImports(): string[] {
-    const fileType: 'shared-type-file' | 'operation-file' =
-      this.config.importSchemaTypesFrom || this.config.generatesOperationTypes
-        ? 'operation-file' // For `operation-file` here means either (1) operation file in a multiple file setup or (2) the main file in a single file setup
-        : 'shared-type-file';
+    const fileType: 'multi-file-shared-type-file' | 'multi-file-operation-file' | 'single-file-operation-file' = this
+      .config.importSchemaTypesFrom
+      ? 'multi-file-operation-file'
+      : this.config.generatesOperationTypes
+      ? 'single-file-operation-file'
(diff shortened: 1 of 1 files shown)


packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -411,6 +420,97 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
...
+      if (this._usedNamedInputTypes[enumName]) {
+        usedEnumMap[enumName] = enumDetails;
+      }
+    }
+
+    return getEnumsImports({
+      enumValues: usedEnumMap,
+      useTypeImports: this.config.useTypeImports,
+    });
+  }
```
eddeee888, line 435: I moved this up here as a drive-by to keep similar imports methods closer together.

packages/plugins/typescript/operations/src/visitor.ts
```diff
@@ -436,15 +544,38 @@ export class TypeScriptDocumentsVisitor extends BaseDocumentsVisitor<
...
+      }
+
+      usedInputTypes[node.name] = scalarType;
       return;
     }
 
     // GraphQLInputObjectType
+    if (usedInputTypes[node.name]) {
+      return;
+    }
```
eddeee888, line 585: `GraphQLInputObjectType` and `GraphQLEnumType` don't have special cases like `GraphQLScalarType` where it may need special handling depending on where it's used (tracked by `useCases`). Therefore, we can return early for the former two if we've seen them previously

packages/plugins/typescript/operations/tests/ts-documents.standalone.default-scalar-types.spec.ts
eddeee888, line 1: This is renamed to `ts-douments.standalone.scalars.spec.ts`

PR #10659, opened 2026-03-31
[cli] externalDocuments implemented, with `typescript-operations` support

ikusakov2:
## Description

`documentsReadOnly` implemented, for safe handling of fragment includes between packages in the monorepo

Related # (10658)[https://github.com/dotansimha/graphql-code-generator/issues/10658]

## Type of change

## How Has This Been Tested?

- [ ] Unit testing

### DIFF
--- a/.changeset/social-worms-report.md
+++ b/.changeset/social-worms-report.md
@@ -1,0 +1,14 @@
+---
+'@graphql-codegen/gql-tag-operations': minor
+'@graphql-codegen/visitor-plugin-common': minor
+'@graphql-codegen/typescript-operations': minor
+'@graphql-codegen/plugin-helpers': minor
+'@graphql-codegen/cli': minor
+'@graphql-codegen/client-preset': minor
+---
+
+Add support for `externalDocuments`
+
+`externalDocuments` declares GraphQL documents that will be read but will not have type files generated for them. These documents are available to plugins for type resolution (e.g. fragment types), but no output files will be generated based on them. Accepts the same formats as `documents`.
+
+This config option is useful for monorepos where each project may want to generate types for its own documents, but some may need to read shared fragments from across projects.
--- a/packages/graphql-codegen-cli/src/codegen.ts
+++ b/packages/graphql-codegen-cli/src/codegen.ts
@@ -14,7 +14,7 @@ import {
   normalizeOutputParam,
   Types,
 } from '@graphql-codegen/plugin-helpers';
-import { NoTypeDefinitionsFound } from '@graphql-tools/load';
+import { NoTypeDefinitionsFound, type UnnormalizedTypeDefPointer } from '@graphql-tools/load';
 import { mergeTypeDefs } from '@graphql-tools/merge';
 import { CodegenContext, ensureContext } from './config.js';
 import { getDocumentTransform } from './documentTransforms.js';
@@ -86,6 +86,7 @@ export async function executeCodegen(
   let rootConfig: { [key: string]: any } = {};
   let rootSchemas: Types.Schema[];
   let rootDocuments: Types.OperationDocument[];
+  let rootExternalDocuments: Types.OperationDocument[];
   const generates: { [filename: string]: Types.ConfiguredOutput } = {};
 
   const cache = createCache();
@@ -135,6 +136,11 @@ export async function executeCodegen(
 
     /* Normalize root "documents" field */
     rootDocuments = normalizeInstanceOrArray<Types.OperationDocument>(config.documents);
+
+    /* Normalize root "externalDocuments" field */
+    rootExternalDocuments = normalizeInstanceOrArray<Types.OperationDocument>(
+      config.externalDocuments,
+    );
 
     /* Normalize "generators" field */
     const generateKeys = Object.keys(config.generates || {});
@@ -228,13 +234,15 @@ export async function executeCodegen(
                 let outputSchemaAst: GraphQLSchema;
                 let outputSchema: DocumentNode;
                 const outputFileTemplateConfig = outputConfig.config || {};
-                let outputDocuments: Types.DocumentFile[] = [];
+                const outputDocuments: Types.DocumentFile[] = [];
                 const outputSpecificSchemas = normalizeInstanceOrArray<Types.Schema>(
                   outputConfig.schema,
                 );
                 let outputSpecificDocuments = normalizeInstanceOrArray<Types.OperationDocument>(
                   outputConfig.documents,
                 );
+                let outputSpecificExternalDocuments =
+                  normalizeInstanceOrArray<Types.OperationDocument>(outputConfig.externalDocuments);
 
                 const preset: Types.OutputPreset | null = hasPreset
                   ? typeof outputConfig.preset === 'string'
@@ -246,6 +254,10 @@ export async function executeCodegen(
                   outputSpecificDocuments = await preset.prepareDocuments(
                     filename,
                     outputSpecificDocuments,
+                  );
+                  outputSpecificExternalDocuments = await preset.prepareDocuments(
+                    filename,
+                    outputSpecificExternalDocuments,
                   );
                 }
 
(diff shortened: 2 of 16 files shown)


packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -308,41 +320,97 @@ export async function executeCodegen(
...
+                          ];
+
+                          const readOnlyPointerMap = populateDocumentPointerMap(
+                            allReadOnlyDenormalizedPointers,
+                          );
 
-                              throw error;
+                          const readOnlyHash = JSON.stringify(readOnlyPointerMap);
+                          const outputDocumentsReadOnly = await cache(
+                            'documents',
```
eddeee888, line 380: A readonly document is also just a document for hashing purposes. 
This ensures there is only one document in the final merged array, because we use the hash to detect the file uniqueness

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -464,18 +465,22 @@ export class CodegenContext {
     return addHashToSchema(loadSchema(pointer, config));
   }
 
-  async loadDocuments(pointer: Types.OperationDocument[]): Promise<Types.DocumentFile[]> {
+  async loadDocuments(
+    pointer: UnnormalizedTypeDefPointer,
```
eddeee888, line 478: The only caller of this function will be passing in `UnnormalizedTypeDefPointer`, and nothing else. May as well narrow the type to make it easier to code.

packages/graphql-codegen-cli/src/config.ts
```diff
@@ -502,24 +507,27 @@ function addHashToSchema(schemaPromise: Promise<GraphQLSchema>): Promise<GraphQL
...
 
-  if (doc.document) {
-    return hashContent(print(doc.document));
-  }
+    if (doc.document) {
+      return hashContent(print(doc.document));
+    }
 
-  return null;
-}
```
eddeee888, line 524: `hashDocument` is is only used by one function, so I moved it inline to avoid jumping around.

packages/graphql-codegen-cli/src/load.ts
```diff
@@ -69,7 +70,7 @@ export async function loadSchema(
 export async function loadDocuments(
   documentPointers: UnnormalizedTypeDefPointer | UnnormalizedTypeDefPointer[],
   config: Types.Config,
-): Promise<Types.DocumentFile[]> {
+): Promise<Source[]> {
```
eddeee888, line 73: When this function finishes, we'd only get the base `Source`. `Types.DocumentFile` is what we get after processing the `Source` (adding `hash` and `type`)

packages/plugins/other/visitor-plugin-common/src/optimize-operations.ts
```diff
@@ -14,6 +14,7 @@ export function optimizeOperations(
   );
 
   return newDocuments.map((document, index) => ({
+    ...documents[index],
```
eddeee888, line 17: Context for future me:

Previously we throw away `hash` when optimising the doc.
When we introduce `type` (`read-only` or `standard`), we need it for further processing, so we need to keep it like this.

Maybe we should not just add these custom Codegen meta directly to the node, since `optimizeDocuments` implementation may remove it in the future.

Maybe we can add a special key/value to reduce this risk e.g. `__meta: { hash: 'abc', types: 'standard' }` 🤔

packages/utils/plugins-helpers/src/types.ts
```diff
@@ -34,7 +34,8 @@ export namespace Types {
   };
 
   export interface DocumentFile extends Source {
-    hash?: string;
+    hash: string;
+    type: 'standard' | 'read-only';
```
eddeee888: `Source` is the default from parsing the docs. 
`DocumentFile` is the "processed" source, with Codegen metadata.

We may want to group them to avoid edge cases mentioned in: https://github.com/dotansimha/graphql-code-generator/pull/10659/changes#r3032421307

packages/utils/graphql-codegen-testing/src/operations.ts
eddeee888, line 1: Documents now have `type` being `'standard' | 'read-only'`.
If we don't supply that, the plugin will generate nothing.

This utility helps by creating `Types.DocumentFile` with `hash` and `type` hardcoded. If we need to make the params partial with override in the future, we could,.

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -308,41 +320,97 @@ export async function executeCodegen(
...
+                            const pointer: UnnormalizedTypeDefPointer = {};
+                            for (const denormalizedPtr of allDocumentsDenormalizedPointers) {
+                              if (typeof denormalizedPtr === 'string') {
+                                pointer[denormalizedPtr] = {};
+                              } else if (typeof denormalizedPtr === 'object') {
+                                Object.assign(pointer, denormalizedPtr);
+                              }
+                            }
+                            return pointer;
+                          };
```
eddeee888, line 336: Inlining `populateDocumentPointerMap` makes it a bit easier to read as we don't jump around the file. We could take this pure function and put it at the bottom of the file if needed.

packages/utils/plugins-helpers/src/types.ts
ikusakov2: 2 non-optional fields introduced. Any existing custom plugin or preset that constructs a DocumentFile manually will now have a TypeScript compile error. This is a breaking change, not a minor one (mentioned in the changeset)

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -308,41 +320,97 @@ export async function executeCodegen(
...
+                              return prev;
                             }
-                          });
 
-                          outputDocuments = result.documents;
+                            outputDocuments.push(file);
+                            processedFile[file.hash] = true;
+
+                            return prev;
+                          }, []);
```
ikusakov2: Looks like the `reduce` accumulator prev is never used — outputDocuments.push(file) writes to an external array. The return value of the entire reduce is discarded.
It works, by chance, but this is confusing. Maybe a plain for...of loop would be cleaner?
 
Additionally, if a document has hash: null (which hashDocument() can return when rawSDL and document are both absent), processedFile[null] will be "null" — meaning all null-hash documents get deduped down to 1, which may be wrong.

.changeset/social-worms-report.md
```diff
@@ -0,0 +1,14 @@
+---
+'@graphql-codegen/gql-tag-operations': minor
+'@graphql-codegen/visitor-plugin-common': minor
+'@graphql-codegen/typescript-operations': minor
+'@graphql-codegen/plugin-helpers': minor
+'@graphql-codegen/cli': minor
+'@graphql-codegen/client-preset': minor
+---
+
+Add support for `documentsReadOnly`
```
ikusakov2: It's the last chance to consider other names, such as: `externalDocuments`, `referenceDocuments`, 
`documentsReadOnly` has its clear benefits as well, so I don't mind keeping it.

packages/graphql-codegen-cli/src/load.ts
ikusakov2, line 73: The 2 new fields were added to `Types.DocumentFile`, not to `Source`, which makes is harder to see where the contract is established.

packages/graphql-codegen-cli/src/codegen.ts
eddeee888: > Looks like the reduce accumulator prev is never used — outputDocuments.push(file) writes to an external array. The return value of the entire reduce is discarded.
It works, by chance, but this is confusing. Maybe a plain for...of loop would be cleaner?

That's true! I think I was trying to reduce the result to `outputDocuments`, but it's accessible inside the loop anyways. I've made it for...of in https://github.com/dotansimha/graphql-code-generator/pull/10659/changes/7ec7e069a8c8d8af753631fdf5b1a0419b9fdcb2

> Additionally, if a document has hash: null (which hashDocument() can return when rawSDL and document are both absent), processedFile[null] will be "null" — meaning all null-hash documents get deduped down to 1, which may be wrong.

Huh, that's true, the typecheck in this repo is not strict enough. Something I'll need to configure after the major release.

I wonder when `null` would happen though 🤔 When Codegen is parsing an empty file?

packages/utils/plugins-helpers/src/types.ts
eddeee888: > Any existing custom plugin or preset that constructs a DocumentFile manually will now have a TypeScript compile error.

That's a good point, I can make this optional and treat no `type` as `standard`

.changeset/social-worms-report.md
eddeee888: The pro of `documentsReadOnly` is that it looks related to `documents`
The con of `documentsReadOnly` is that it doesn't sound like English.

`externalDocuments` sounds better. I'm happy to go with that if you are 🙂

packages/graphql-codegen-cli/src/load.ts
eddeee888: Sorry I don't fully understand this statement. However I've traced the CLI code to use the right type:
- before adding metadata (hash, type), the documents are `Source`
- after adding metadata the documents are `DocumentFile`

I feel this may not answer your concern. Could you please help me understand this a bit more?

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -308,41 +320,97 @@ export async function executeCodegen(
...
+                          for (const file of mergedDocuments) {
+                            if (processedFile[file.hash]) {
+                              continue;
                             }
-                          });
 
-                          outputDocuments = result.documents;
+                            outputDocuments.push(file);
+                            processedFile[file.hash] = true;
+                          }
```
eddeee888, line 418: I've considered forwarding `documentsReadOnly` as-is but hit a few annoying cases:

- documents are processed a lot after the initial loading stage [here](https://github.com/dotansimha/graphql-code-generator/blob/c1fc4ee929af3639640b7a7e51bb125bdc882e45/packages/graphql-codegen-core/src/codegen.ts#L32). If we kept documentsReadOnly separate, we'd need to run the same functions over them
- [Plugins use positional param](https://github.com/dotansimha/graphql-code-generator/blob/c1fc4ee929af3639640b7a7e51bb125bdc882e45/packages/graphql-codegen-core/src/execute-plugin.ts#L76-L85), which means `documentsReadOnly` need to be at the last position to avoid a breaking change. This works, but feels disconnected

For this reason, I feel it's easier if we merge them as early as possible, and run the same processing over both types, and plugins like typescript-operations can handle the filtering without having to access another param

.changeset/social-worms-report.md
ikusakov2: Sorry for bringing this one up!
It's really hard to decide! 

* `documentsReadOnly` has clear benefits. It is self-descriptive and it appears near `documents` (for code completion) - which is quite valuable as well. But it is not a proper English.

* `externalDocuments`: The feature is about documents that are loaded for reference but don't produce output. ReadOnly describes how the current output treats them, not a property of the documents themselves - which makes `documentsReadOnly` name slightly confusing. `externalDocuments` is a proper name - in terms of English and semantics, but it is not self-explanatory.

I guess a user of this new feature would have to go to the documentation in any case (either slightly confusing name or not informative enough name), so a better sounding name has a slight benefit, imho.

packages/graphql-codegen-cli/src/load.ts
ikusakov2: I was a bit worried that the `Types.DocumentFile[]` has 2 new properties (`hash` and `type`) and `Source` doesn't have them - which might mislead a user. 
But it is correct from the polymorphic world perspective. So - we can leave it as is.

.changeset/social-worms-report.md
eddeee888: Not at all! Naming is hard! Glad I can bounce ideas with you 🤝
Let's lock `externalDocuments` in. I like the reasoning here.

Code completion show `externalDocuments` as well if users start typing `documents` , so it's not bad in that regard.

PR #10728, opened 2026-04-16
Deduplicate external fragments in typescript-operations

Nurou:
This deduplicates `externalFragments` by name in `@graphql-codegen/typescript-operations` before they are passed to `TypeScriptDocumentsVisitor`.

Why:
- `near-operation-file` can supply repeated nested `externalFragments`
- `typescript-operations` currently forwards them as-is
- this can cause severe blow-ups when repeated fragment spreads are nested

This is related to dotansimha/graphql-code-generator-community#752.

Validation:
- added a regression test that asserts repeated `externalFragments` are deduped before visiting documents
- `yarn workspace @graphql-codegen/typescript-operations vitest run tests/ts-documents.externalFragments.spec.ts tests/ts-documents.externalDocuments.spec.ts`

Local benchmark from a downstream repo using `near-operation-file` + `typescript-operations`:
- before: ~248s
- after dedupe: ~2s

### DIFF
--- a/packages/plugins/typescript/operations/src/index.ts
+++ b/packages/plugins/typescript/operations/src/index.ts
@@ -57,6 +57,7 @@ export const plugin: PluginFunction<
   // For Fragment types to resolve correctly, we must get read all docs (`standard` and `external`)
   // Fragment types are usually (but not always) in `external` files in certain setup, like a monorepo.
   const allDocumentsAST = concatAST(parsedDocuments.all.documentNodes);
+  const fragmentNames = new Set<string>();
   const allFragments: LoadedFragment[] = [
     ...(
       allDocumentsAST.definitions.filter(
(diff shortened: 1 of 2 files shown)


eddeee888:
Hi @Nurou , I wonder if we should deduplicate the fragment in `near-operation-file-preset` instead or not? 🤔 
e.g. around [here](https://github.com/dotansimha/graphql-code-generator-community/blob/2b22acbb31e62fd4594db170c50f7aedcbac6cc4/packages/presets/near-operation-file/src/index.ts#L380)

Nurou:
Closing in favor of the source-level fix in https://github.com/dotansimha/graphql-code-generator-community/pull/1454

Nurou:
> Hi @Nurou , I wonder if we should deduplicate the fragment in `near-operation-file-preset` instead or not? 🤔 e.g. around [here](https://github.com/dotansimha/graphql-code-generator-community/blob/2b22acbb31e62fd4594db170c50f7aedcbac6cc4/packages/presets/near-operation-file/src/index.ts#L380)

Hey @eddeee888 and yes. That makes more sense as the duplication is in fact introduced there. So we'd benefit more by doing it upstream.

Issue #10725, opened 2026-04-14
CLI 6.3.0: near-operation-file preset silently skips output for files with duplicate content


stefanmaric:
## Description

Since `@graphql-codegen/cli` 6.3.0, the `near-operation-file` preset silently produces no output for `.graphql` files that share the same content as another file, even when they're in different directories and should each get their own generated `.d.ts`.

There are no warnings or errors. The second file simply gets no output.

## Reproduction

Given two identical `.graphql` files in different directories:

```
# src/components/A/foo.mutation.graphql
mutation foo($id: ID!) {
  foo(id: $id) { id }
}

# src/components/B/foo.mutation.graphql  (exact same content)
mutation foo($id: ID!) {
  foo(id: $id) { id }
}
```

With `near-operation-file` preset, only `src/components/A/foo.mutation.graphql.d.ts` is generated. `src/components/B/foo.mutation.graphql.d.ts` is never created.

## Cause

Two separate dedup mechanisms compound to cause this:

### 1\. CLI hash-based dedup (new in 6.3.0)

Introduced in [#10659](<https://github.com/dotansimha/graphql-code-generator/issues/10659>) (`e65d303`), the CLI deduplicates loaded documents by `file.hash` before passing them to the preset:

[https://github.com/dotansimha/graphql-code-generator/blob/e65d3036475cc1e3cb05a8367d30a5092b20080e/packages/graphql-codegen-cli/src/codegen.ts#L291-L299](<https://github.com/dotansimha/graphql-code-generator/blob/e65d3036475cc1e3cb05a8367d30a5092b20080e/packages/graphql-codegen-cli/src/codegen.ts#L291-L299>)

```js
const processedFile: Record<string, boolean> = {};
for (const file of mergedDocuments) {
  if (processedFile[file.hash]) {
    continue;
  }
  outputDocuments.push(file);
  processedFile[file.hash] = true;
}
```

When two files have the same content, they get the same hash. The second file is dropped entirely—the preset never sees it.

### 2\. Core in-place definition mutation (pre-existing)

The `validateDuplicateDocuments` function in `@graphql-codegen/core` mutates `file.document.definitions` in-place, removing definitions that share a name and content with a previously-seen definition. Even if both files reach the preset, the second file's definitions are stripped, resulting in empty output.

## Expected behavior

Each `.graphql` source file should produce its own output file when using `near-operation-file`, regardless of whether another file has the same content. The hash-based dedup should key on file location, not just content.

[...]

PR #10737, opened 2026-04-18
Improve document dedupe logic

eddeee888:
## Description

This PR fixes `documents` vs `externalDocuments` dedupe logic.
Previously, we only checked for file hash which means different files with the same content will be deduped and won't be passed on to preset or plugins.

Let's reserve the existing behaviour for preset/plugin integration, such as `near-operation-file`, for now to avoid making it a breaking change.

Related https://github.com/dotansimha/graphql-code-generator/issues/10725

## Type of change


packages/graphql-codegen-cli/tests/codegen.spec.ts
```diff
@@ -1487,167 +1487,6 @@ describe('Codegen Executor', () => {
     });
   });
 
-  describe('externalDocuments', () => {
```
eddeee888, line 1490: This is moved to `codegen.config.externalDocuments.spec.ts` to keep things isolated and easier to read.

Issue #10776, opened 2026-05-05
enumType "const" using type prefix/suffixes incorrectly.


AndrewIngram:
### Which packages are impacted by your issue?

@graphql-codegen/visitor-plugin-common

### Describe the bug

When using enumType: 'const' in combination with typesPrefix, the generated enum keys are incorrect

https://stackblitz.com/edit/github-qydcbfph?file=types.ts

Root cause appears to be in `convert-schema-enum-to-declaration-block-string`, change:

```ts
const optionName = makeValidEnumIdentifier(convertName({
    options: {
        typesPrefix: naming.options.typesPrefix,
        typesSuffix: naming.options.typesPrefix,
    },
```

to:
```ts
const optionName = makeValidEnumIdentifier(convertName({
    options: {
        useTypesPrefix: false,
        typesPrefix: naming.options.typesPrefix,
        typesSuffix: naming.options.typesSuffix,
    },
```

### Your Example Website or App

https://stackblitz.com/edit/github-qydcbfph?file=types.ts

### Steps to Reproduce the Bug or Issue

Use enumType: `const` and set a `typesPrefix` value

### Expected behavior

In this example, i'd expect the generated enum to look like this:

```ts
export const GQStatus = {
  Active: 'Active',
  Inactive: 'Inactive'
} as const;
```

But it's actually:
```ts
export const GQStatus = {
  GQActiveGQ: 'Active',
  GQInactiveGQ: 'Inactive'
} as const;
```

### Platform

- OS: any
- NodeJS: any
- `graphql` version: [e.g. 16.13.2]
- `@graphql-codegen/*` version(s): 6.0.0

### Codegen Config File

import { CodegenConfig } from '@graphql-codegen/cli';

const config: CodegenConfig = {
  schema: 'schema.graphql',
  documents: 'document.graphql',
  generates: {
    'types.ts': { plugins: ['typescript-operations'] },
  },
  config: {
    enumType: 'const',
    typesPrefix: 'GQ',
  },
};

export default config;

eddeee888:
This is fixed and released in https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1778079904444

Thanks for raising the issue! Please let me know if this doesn't fix your problem

PR #10781, opened 2026-05-06
Fix `const` enum type prefix/suffix misconfiguration

eddeee888:
## Description

This is a bug that happened during the refactoring enum prefix/suffix when `typesPrefix` and `typesSuffix` is used with `enumType=const` specifically. In this case, `typesPrefix` is accidentally used in the `typesSuffix` slot. 

This PR fixes the issue and add tests for relevant cases.

Related https://github.com/dotansimha/graphql-code-generator/issues/10776

## Type of change

### DIFF
--- a/packages/plugins/typescript/operations/tests/ts-documents.standalone.enum.spec.ts
+++ b/packages/plugins/typescript/operations/tests/ts-documents.standalone.enum.spec.ts
@@ -164,6 +164,72 @@ describe('TypeScript Operations Plugin - Enum', () => {
     validateTs(result, undefined, undefined, undefined, undefined, true);
   });
 
+  it('handles `const` enum with prefix and suffix', async () => {
+    const schema = buildSchema(/* GraphQL */ `
+      type Query {
+        me: User
+      }
+
+      type User {
+        id: ID!
+        name: String!
+        role: UserRole!
+        createdAt: DateTime!
+      }
+
+      enum UserRole {
+        A_B_C
+        X_Y_Z
+        _TEST
+        My_Value
+        _123
+      }
+
+      scalar DateTime
+    `);
+    const document = parse(/* GraphQL */ `
+      query Me($role: UserRole!) {
+        me {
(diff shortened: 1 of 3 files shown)


packages/plugins/typescript/operations/tests/ts-documents.standalone.enum.spec.ts
```diff
@@ -103,6 +103,65 @@ describe('TypeScript Operations Plugin - Enum', () => {
     validateTs(result, undefined, undefined, undefined, undefined, true);
   });
 
+  it('handles `native-numeric` enum with types prefix and suffix', async () => {
```
eddeee888, line 106: Added this test to ensure the behaviour is captured, even if it's not related to this PR

packages/plugins/typescript/operations/tests/ts-documents.standalone.enum.spec.ts
```diff
@@ -103,6 +103,65 @@ describe('TypeScript Operations Plugin - Enum', () => {
...
+    ]);
+
+    expect(result).toMatchInlineSnapshot(`
+      "/** Internal type. DO NOT USE DIRECTLY. */
+      type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
+      /** Internal type. DO NOT USE DIRECTLY. */
+      export type Incremental<T> = T | { [P in keyof T]?: P extends ' $fragmentName' | '__typename' ? T[P] : never };
+      export enum AA_UserRole_ZZ {
+        Admin_ZZ = 0,
+        Customer_ZZ = 1
```
eddeee888, line 150: Currently, the `typesPrefix` is not applied to enumValues e.g. this is `Admin_ZZ` and not `AA_Admin_ZZ`

I remember this is the previous behaviour. So I'm adding test here to be explicit to capture the current behaviour 

(especially when it feels wrong, and should be fixed )

Issue #10787, opened 2026-05-08
Option to disable legacy Federation v1 injection for Federation v2 schemas


robertobadalamenti:
### Is your feature request related to a problem? Please describe.

When using Apollo Federation v2, the CLI automatically injects `scalar _FieldSet` and Federation directives.
Since Federation v2 schemas typically import these via `@link`, the automatic injection causes "Invalid definition" errors. 
There is currently no way to disable this behavior while keeping federation: true.

**Error example:**
`AggregateGraphQLError: The schema is not a valid GraphQL schema.. Caused by:
Invalid definition for directive "@key": argument "fields" should have type "federation__FieldSet!" but found type "_FieldSet!"`

**Reproduction:**
[https://github.com/robertobadalamenti/codegen-federationv2-bug](https://github.com/robertobadalamenti/codegen-federationv2-bug)

### Describe the solution you'd like

I would like to introduce a new configuration flag, disableFederationDirectiveAndScalarInjection (defaulting to false for backward compatibility).

When set to true, this flag will prevent the automatic injection of the _FieldSet scalar and Federation directives, allowing the schema to rely solely on the definitions imported via @link in Federation v2 environments.

### Describe alternatives you've considered

I considered making the generator "smart" enough to auto-detect Federation v2 schemas and disable the injection automatically. However, this could lead to unexpected breaking changes or edge cases in complex setups.
I believe an explicit configuration flag is a safer and more predictable approach for now, especially until a full, native Federation v2 support is addressed across the entire codebase.

### Any additional important details?

The proposed flag specifically targets the `scalar _FieldSet` and the Federation directives (like `@key`, `@external`, etc.) that are currently hardcoded for injection.

I have already prepared a Pull Request including the implementation, unit tests, and documentation updates. I'm available to submit it for your review and to discuss any necessary adjustments.

robertobadalamenti:
Hi @eddeee888, did you have a chance to look at PR #10788? Thanks.

eddeee888:
Apologies for the delay @robertobadalamenti ! This is definitely my next focus. 
I'm currently looking at a few ops thing for the repo which is required for releasing

eddeee888:
This is released here: https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1779891963597

Thanks for the fix @robertobadalamenti !

PR #10788, opened 2026-05-08
feat(cli): allow disabling legacy federation injection for FederationV2 schemas

robertobadalamenti:
## Description

This PR introduces the disableFederationDirectiveAndScalarInjection flag to the codegen core.

Currently, when federation: true is set, the generator automatically injects legacy Federation v1 definitions (like scalar _FieldSet and directives @key, @shareable, etc.). In Apollo Federation v2, these are typically imported via the @link directive, causing "Invalid definition" errors during validation.

This flag allows users to opt-out of the automatic injection while keeping federation support active.

Related #10787

## Type of change

https://github.com/robertobadalamenti/codegen-federationv2-bug

## How Has This Been Tested?

I have added unit tests in packages/graphql-codegen-core/tests/codegen.spec.ts to verify that:

- The default behavior remains unchanged (injection still happens if the flag is not set).
- When the flag is true, no legacy federation definitions are added to the schema.
- The flag correctly interacts with existing federation settings.

**Test Environment**:

- OS: macOS
- `@graphql-codegen/cli`:
- NodeJS: v24.15.0

## Further comments

The solution was implemented at the core level to ensure it works across all plugins that rely on the central codegen logic. I considered automatic detection of Federation v2, but an explicit flag was chosen to avoid breaking changes and provide more predictable control to the user.


eddeee888:
Thank you @robertobadalamenti !
I feel this is a good quick option for now to support federation v2 👍 

Side note: It's a bit odd that this is happening in the core codegen package, so I'll have a think how to decouple it later.

Issue #10587, opened 2026-02-03
[typescript-resolvers] typesPrefix not applied to FederationTypes reference in   FederationReferenceTypes


dilaroga:
### Which packages are impacted by your issue?

@graphql-codegen/typescript-resolvers

### Describe the bug

  When using typesPrefix: "I" (or any prefix) with federation: true, the generated
  FederationReferenceTypes type references FederationTypes without the configured prefix,
  causing a TypeScript compilation error.

  The generated code defines IFederationTypes (with prefix) but references it as
  FederationTypes (without prefix) inside IFederationReferenceTypes.

### Your Example Website or App

N/A

### Steps to Reproduce the Bug or Issue

```yml
codegen.yml:
schema: src/api/graphql/schema.graphql
generates:
  src/api/graphql/types.ts:
    plugins:
      - typescript
      - typescript-resolvers
    config:
      typesPrefix: I
      federation: true

```
  schema.graphql (minimal example with federation):
```graphql
type Query {
  _service: _Service!
}

type _Service {
  sdl: String
}

type Channel @key(fields: "id") {
  id: ID!
}
```
Run codegen and compile with TypeScript.

Actual behavior:

  The generated code uses FederationTypes (without prefix) in the reference:
```graphql
export type IFederationTypes = ResolversObject<{
  Channel: IChannel;
}>;

export type IFederationReferenceTypes = ResolversObject<{
  Channel:
    ( { __typename: 'Channel' }
    & GraphQLRecursivePick<FederationTypes['Channel'], {"id":true}> );  // ❌ Missing "I" prefix
}>;
```
TypeScript error:
  error TS2552: Cannot find name 'FederationTypes'. Did you mean 'IFederationTypes'?

### Expected behavior

The generated code should use IFederationTypes consistently:
```graphql
  export type IFederationTypes = ResolversObject<{
    Channel: IChannel;
  }>;

  export type IFederationReferenceTypes = ResolversObject<{
    Channel:
      ( { __typename: 'Channel' }
      & GraphQLRecursivePick<IFederationTypes['Channel'], {"id":true}> );
  }>;
```

### Platform

OS: macOS / Linux
@graphql-codegen/cli: 6.1.1
@graphql-codegen/typescript-resolvers: 5.0.0 - 5.1.5 (all 5.x versions affected)
Node.js: 20.x / 22.x

### Codegen Config File

```yml
schema: src/api/graphql/schema.graphql
generates:
[...]

PR #10797, opened 2026-05-10
CODEGEN-907 [typescript-resolvers] Fix Federation types not apply prefix and suffix

eddeee888:
## Description

Previously, `FederationTypes` did not apply `typesPrefix` and `typesSuffix` correctly, mostly due to where the generation of this type happens (outside of the Visitor). 

Now, copy the `convertName` functionality for the federation type generation as well. Note that we are still duplicating the `convertName` functionality in two places, meaning the logic may drift in the future. However, convert logic hasn't changed for awhile, so we can do this to fix the issue, and think about the ideal state later.

Related https://github.com/dotansimha/graphql-code-generator/issues/10587

## Type of change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can
reproduce. Please also list any relevant details for your test configuration

- [x] Unit test

### DIFF
--- a/packages/plugins/typescript/resolvers/src/index.ts
+++ b/packages/plugins/typescript/resolvers/src/index.ts
@@ -6,7 +6,12 @@ import {
   PluginFunction,
   Types,
 } from '@graphql-codegen/plugin-helpers';
-import { parseMapper, type RootResolver } from '@graphql-codegen/visitor-plugin-common';
+import {
+  convertFactory,
+  convertName as convertNameUtil,
+  parseMapper,
+  type RootResolver,
+} from '@graphql-codegen/visitor-plugin-common';
 import { TypeScriptResolversPluginConfig } from './config.js';
 import { TypeScriptResolversVisitor } from './visitor.js';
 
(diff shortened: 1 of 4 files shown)


packages/plugins/typescript/resolvers/tests/ts-resolvers.federation.spec.ts
```diff
@@ -1006,6 +1007,179 @@ describe('TypeScript Resolvers Plugin + Apollo Federation', () => {
           },
         }
       `);
-    });
+  });
+});
```
eddeee888, line 1011: I moved this outside to avoid nested describes, which is hard to read and generally can lead to unnecessary shared setup

packages/plugins/typescript/resolvers/src/index.ts
```diff
@@ -88,7 +93,19 @@ export type Resolver${capitalizedDirectiveName}WithResolve<TResult, TParent, TCo
   }
 
   let { transformedSchema, federationMeta } = config.federation
-    ? addFederationReferencesToSchema(schema)
+    ? (() => {
+        const baseConvert = convertFactory(config);
```
eddeee888, line 97: Every time we need to call `convertName` outside of a visitor, it's a hassle because of the factory (which I just saw already supports prefix and suffix) and using `convertName` util.

Would be good to find an ideal way to do this without soon.

PR #10806, opened 2026-05-12
Fix Array used instead of ReadonlyArray for input type when immutableTypes config is on

vinassefranche:
I did not create an issue before this as it was easier to demo the issue with the P.R. Tell me if you think I should still open an issue.

## Description

When `immutableTypes` is set to `true` in the config, the generated types are supposed to use `ReadonlyArray` instead of `Array`. This works well for most cases but for some inputs, `Array` was still used.
This pull request fixes this issue, which is, if I'm not wrong, a regression.

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as
      expected)

## How Has This Been Tested?

There was already a test for this which was expecting `Array`. I updated it in this P.R.

## Checklist:

- [x] I have followed the
      [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the
      style guidelines of this project


eddeee888 (approved):
Thank you!

PR #10818, opened 2026-05-14
Migrate to pnpm

eddeee888:
## Description

Take 2 of https://github.com/dotansimha/graphql-code-generator/pull/9314

- follow the steps of the first PR
- ...
- disable a few tests [here](https://github.com/dotansimha/graphql-code-generator/pull/10818#discussion_r3251964756), will come back next


packages/presets/client/tests/client-preset.spec.ts
eddeee888, line 1: Temporarily replaced `validateTs` with inline snapshot test to change to pnpm. Otherwise it's too hard.
I'll follow up with fixes using `FIXME(pnpm-update):` markers

package.json
```diff
@@ -52,19 +41,40 @@
...
+    "@graphql-codegen/cli": "workspace:*",
+    "@graphql-codegen/client-preset": "workspace:*",
+    "@graphql-codegen/flow": "3.0.1",
+    "@graphql-codegen/flow-resolvers": "3.0.2",
+    "@graphql-codegen/graphql-modules-preset": "workspace:*",
+    "@graphql-codegen/typed-document-node": "workspace:*",
+    "@graphql-codegen/typescript": "workspace:*",
+    "@graphql-codegen/typescript-graphql-files-modules": "4.0.1",
+    "@graphql-codegen/typescript-operations": "workspace:*",
+    "@graphql-codegen/typescript-resolvers": "workspace:*",
```
eddeee888, line 54: This is required because `dev-test` is not a workspace project, i.e. the root project uses these packages for `dev-test`

I'll break `dev-test` to its own project next.

website/src/lib/transform.ts
```diff
@@ -16,7 +16,6 @@ export function transformDocs() {
     skipLibCheck: true,
     allowSyntheticDefaultImports: true,
     types: ['node'],
-    ignoreDeprecations: '6.0',
```
eddeee888, line 19: `typescript-json-schema` is using TS 5, so `ignoreDeprecations: '6.0'` doesn't work

This means we have 2 versions of TS in this codebase, however, it's not causing issues so I'll leave it like that to (1) avoid having to force resolution and (2) forcing TS version _may_ cause issues as well.

package.json
```diff
@@ -2,45 +2,34 @@
...
-  "workspaces": [
-    "packages/*",
-    "packages/utils/*",
-    "packages/plugins/typescript/*",
-    "packages/plugins/other/*",
-    "packages/presets/*",
-    "website",
-    "examples/**/*",
-    "dev-test-apollo-tooling"
-  ],
```
eddeee888, line 14: Moved to `pnpm-workspace.yaml`

package.json
```diff
@@ -2,45 +2,34 @@
...
-    "postinstall": "patch-package && husky install",
-    "prebuild": "rimraf dist/ .bob/ tsconfig.tsbuildinfo",
-    "prerelease": "yarn build",
     "prettier": "prettier --cache --write --list-different .",
     "prettier:check": "prettier --cache --check .",
     "rebuild": "bob build --incremental",
+    "prerelease": "pnpm build",
     "release": "changeset publish",
     "test": "vitest",
-    "test-and-build": "yarn build && yarn test",
```
eddeee888, line 40: We've never used this 😬 so removing for less config

PR #10833, opened 2026-05-16
Tidy up dependencies

eddeee888:
We recently migrated to pnpm https://github.com/dotansimha/graphql-code-generator/pull/10818
However, that was the "80%" PR with some cleanups to do.

This PR:
- cleans up incorrect deps (extraneous or in the wrong places)
- cleans up patches
- pin deps in examples packages for explicitness


package.json
eddeee888, line 1: - `memfs`: moved to CLI project because it's only used there
- `patch-package`: removed, not used
- `svelte`: removed, not used
- `ts-node`: removed, not used at root. Projects use `tsx` these days
- `tsx`: moved to projects that need them

PR #10835, opened 2026-05-17
Fix renovates config, root package.json commands

eddeee888:
## Description

This PR does a few cleanups:

- Fix duplicates in renovate.json
- Split `examples` and `dev-tests`  in root package.json command to be explicit
- Rename root package.json commands to group them for ease of read
- Remove `watch-build` root package.json command, as it uses `npx` and brings in untracked `tsc-watch` dep


package.json
```diff
@@ -10,28 +10,27 @@
     "prebuild": "rimraf dist/ .bob/ tsconfig.tsbuildinfo",
     "build": "bob build",
     "postbuild": "pnpm fix-bins",
-    "ci:lint": "eslint --cache --output-file eslint_report.json --format json .",
```
eddeee888, line 13: Renaming `ci:lint` to `lint:ci` to have both commands near each other in this `scripts` list

PR #10848, opened 2026-05-25
Correctly set dep of client-preset, fix publishing

eddeee888:
## Description

- Correctly set deps pinning of client-preset
- Also test publishing with pnpm


packages/graphql-codegen-cli/package.json
```diff
@@ -77,7 +77,7 @@
     "@babel/generator": "^7.18.13",
     "@babel/template": "^7.18.10",
     "@babel/types": "^7.18.13",
-    "@graphql-codegen/client-preset": "^6.0.0",
+    "@graphql-codegen/client-preset": "workspace:^",
```
eddeee888, line 80: Image

Confirmed `workspace:^` is working as intended

PR #10850, opened 2026-05-27
Fix CI, release-stable step

eddeee888:
## Description

We are getting errors: https://github.com/dotansimha/graphql-code-generator/actions/runs/26510692106
This is because `buildScript` is not a valid input. This is because when `release` is used, it triggers `prerelease` which  calls `build`


eddeee888:
Examples tests are failing, will fix in another PR

PR #10856, opened 2026-05-27
Fix Examples CI step: target non-rust examples with pnpm filter

eddeee888:
## Description

Examples step in CI currently gets stuck after pnpm update:
Screenshot 2026-05-28 at 1 30 18 am

We are using a custom script for `examples` projects. This PR fixes it by using pnpm filter


scripts/print-example-ci-command.js
eddeee888, line 1: Good bye custom command. You served us well 🫡  

This is now replaced by the native `pnpm filter=` command

Issue #10585, opened 2026-01-26
graphql-codegen with graphql > 17.0.0.alpha.7


Fyro-Ing:
### Which packages are impacted by your issue?

@graphql-codegen/visitor-plugin-common

### Describe the bug

I use graphql 17.alpha.2, and i try to upgrade with apollo server 5 & graphql17.alpha9 (@defer and @ stream)

graphql-codegen is ok until 17.0.0.alpha.7 but with alpha8 or alpha9, there is an error when codegen instrospect @deprecated directive
https://github.com/dotansimha/graphql-code-generator/blob/62c76188d4e660efa50dcf486921804adc70e1b5/packages/plugins/other/visitor-plugin-common/src/base-types-visitor.ts#L1055

node.directives is undefined, on alpha7, it's an empty array

Minimal reproduction repository :
https://github.com/Fyro-Ing/graphql17.alpha9

### Your Example Website or App

https://github.com/Fyro-Ing/graphql17.alpha9

### Steps to Reproduce the Bug or Issue

npm run generate

### Expected behavior

graphql-codegen generate files

### Platform

- OS: [Linux]
- NodeJS: [e.g. 24]
- `graphql` version: [17.0.0-alpha.8]
- `@graphql-codegen/*` version(s): [6.1.1]

### Codegen Config File

schema: './src/graphql/schema.graphql'
generates:
  ./src/graphql/schema.graphql.ts:
    preset: '@eddeee888/gcg-typescript-resolver-files'

Fyro-Ing:
directives is optional, and other piece of codes check nullable so I supposed fix as this

```
const deprecationDirective = node.directives?.find(v => v.name.value === 'deprecated'); 
```

Fyro-Ing:
build local with graphql > 17.0.0.alpha.7 failed with

> packages/utils/plugins-helpers/src/federation.ts(246,11): error TS2418: Type of computed property's value is '{ type: GraphQLInterfaceType<any, any>; }', which is not assignable to type 'GraphQLFieldNormalizedConfig<any, any, any>'.\n" +                                                                                                                                                                                              
>     "  Type '{ type: GraphQLInterfaceType<any, any>; }' is missing the following properties from type 'GraphQLFieldNormalizedConfig<any, any, any>': args, extensions\n" +
[...]

PR #10866, opened 2026-06-02
CODEGEN905 - Add GraphQL 17 support

eddeee888:
## Description

This PR prepares for GraphQL 17 release: https://github.com/graphql/graphql-js/issues/4786

Also related https://github.com/dotansimha/graphql-code-generator/issues/10585

## Type of change

## How Has This Been Tested?

- [x] Unit test


klarstrup (approved):
LGTM!

vitest.config.mts
eddeee888, line 1: This (and other `vitest.config` files) is changed to `.mts` since we use ESM syntax AND the CLI treats the file as ESM anyways.

Without this, it'd give errors about `import.meta` usage

mikevoets:
Thanks so much for working on this! 🙏 We're trying to upgrade to `graphql@17.0.1` and `@graphql-codegen/cli` is one of the two things gating it for us (the stable `7.1.3` peer range still tops out at `^16`). We can see graphql-17 support has made it into the `alpha` tag from this PR.

Is there anything blocking this from landing in a stable release, or any way we can help (testing, repro cases)? Really appreciate the effort here.

eddeee888:
Apologies for the delay everyone, we were making sure an underlying library is v17-compatible as well.
This will be released today.

Issue #10840, opened 2026-05-19
Check default values at the end, not at the beginning for multi-project GraphQL config


Diluka:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

There is a logic issue where default values are checked too early in the process. The default value should be checked last, but in the current implementation, it is checked first, which may cause errors or unexpected results.

### Your Example Website or App

-

### Steps to Reproduce the Bug or Issue

1. Set up codegen with multiple projects (using `projects` field).
2. At least one project uses a default value in configuration.
3. Run codegen and check generated output. The default value is applied earlier than expected.

### Expected behavior

Default values should only be used after all other conditions are checked – not as an initial shortcut. The current logic uses defaults too soon.

### Platform

- OS: Any
- NodeJS: Any
- graphql version: Any
- @graphql-codegen/* version(s): Any

### Additional context

Related PR: https://github.com/dotansimha/graphql-code-generator/pull/10335

eddeee888:
Thank you for the contribution @Diluka !

The fix has been released here: https://github.com/dotansimha/graphql-code-generator/releases/tag/release-1781617333413

PR #10869, opened 2026-06-16
Set up graphql config dev test

eddeee888:
## Description

This PR sets up dev-test for graphql-config but purposely fails it so we can be sure the implementation in https://github.com/dotansimha/graphql-code-generator/issues/10840 works.

Related https://github.com/dotansimha/graphql-code-generator/issues/10840

## Type of change

- [x] Test setup


dev-test/graphql-config/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "dev-test-graphql-config",
+  "version": "0.0.1",
+  "description": "Test graphql-config integration",
+  "private": true,
+  "scripts": {
+    "generate:cjs": "pnpm generate:cjs:single && pnpm generate:cjs:multi && pnpm generate:cjs:multi:project1",
+    "generate:cjs:multi": "node ../../packages/graphql-codegen-cli/dist/cjs/bin.js --config multi/.graphqlrc.yml || [ $? -eq 1 ]",
```
eddeee888, line 8: This is a test to assert failure because there is no "default" project.
This will expectedly fail with status 1, but we make it 0 to pass

Issue #10867, opened 2026-06-05
`extractAllFieldsToTypes` with `@defer` produces "Duplicate identifier" types


dklymenk:
### Which packages are impacted by your issue?

@graphql-codegen/visitor-plugin-common

### Describe the bug

Adding @defer around a field or a fragment in my codebase with `extractAllFieldsToTypes: true` results in generated types that have duplicate identifiers.

```graphql
fragment UserEmail on User {
    email
}

query user {
    user(id: 1) {
        id
        username
        ...UserEmail @defer
    }
}
```

```ts
export type UserEmailFragment = { email: string };

export type UserQuery_user_User = { id: string, username: string };

export type UserQuery_user_User = { email: string } | { email?: never };

export type UserQuery_Query = { user: UserQuery_user_User & UserQuery_user_User };
```

> Error: Duplicate identifier 'UserQuery_user_User'.

### Your Example Website or App

https://github.com/dklymenk/graphql-code-generator-issue-sandbox-template

### Steps to Reproduce the Bug or Issue

1. Clone https://github.com/dklymenk/graphql-code-generator-issue-sandbox-template
2. `npm i`
3. `npm run generate`
4. `git status` should report no changes
5. `types.ts` has `UserQuery_user_User` duplicate identifiers
6. Doing `... @defer { email }` instead of `...UserEmail @defer` produces the same duplicated identifiers upon regeneration

### Expected behavior

Code generator should produce unique identifiers for every type.

Maybe the deferred fields should be used in the type name `UserQuery_user_User_email`? This may not look/work ok for blocks of multiple fields that are deferred, but it certainly would work and scale well for fragments - `UserQuery_user_User_UserEmailFragment`.

Something simple like `UserQuery_user_User_deferred` would not be enough as you might have more than one deferred block on `UserQuery_user_User`.

### Platform

- OS: macOS
- NodeJS: 24.13.0
- `graphql` version: 16.14.1
- `@graphql-codegen/cli` version: 7.1.2
- `@graphql-codegen/typescript` version: 6.0.2
- `@graphql-codegen/typescript-operations version: 6.0.3

### Codegen Config File

```ts
import { CodegenConfig } from "@graphql-codegen/cli";

const config: CodegenConfig = {
  schema: "schema.graphql",
  documents: "document.graphql",
  generates: {
    "types.ts": {
      plugins: ["typescript", "typescript-operations"],
      config: {
        extractAllFieldsToTypes: true,
      },
    },
[...]

PR #10897, opened 2026-07-14
CODEGEN-935 - Fix `extractAllFieldsToTypes=true` with `@defer` causing duplicated types to be generated

eddeee888:
## Description

This PR fixes a scenario where `extractAllFieldsToTypes=true` with `@defer` are used together, it may generate duplicated types

Related https://github.com/dotansimha/graphql-code-generator/issues/10867

## Type of change

### DIFF
--- a/packages/plugins/typescript/operations/tests/extract-all-types-with-defer.spec.ts
+++ b/packages/plugins/typescript/operations/tests/extract-all-types-with-defer.spec.ts
@@ -1,0 +1,61 @@
+import { buildSchema, parse } from 'graphql';
+import { mergeOutputs } from '@graphql-codegen/plugin-helpers';
+import { validateTs } from '@graphql-codegen/testing';
+import { plugin } from '../src/index.js';
+
+describe('extractAllFieldsToTypes: true with @defer', () => {
+  it('should extract types from queries', async () => {
+    const schema = buildSchema(/* GraphQL */ `
+      type Query {
+        user(id: ID!): User
+      }
+
+      type User {
+        id: ID!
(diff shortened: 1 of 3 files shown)


packages/plugins/typescript/operations/tests/extract-all-types-with-defer.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(result).toMatchInlineSnapshot(`
+      "/** Internal type. DO NOT USE DIRECTLY. */
+      type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
+      /** Internal type. DO NOT USE DIRECTLY. */
+      export type Incremental<T> = T | { [P in keyof T]?: P extends ' $fragmentName' | '__typename' ? T[P] : never };
+      export type UserEmailFragment = { email: string };
+
+      export type UserQuery_user_User_0 = { id: string, username: string };
+
+      export type UserQuery_user_User_1 = { email: string } | { email?: never };
```
dklymenk, line 47: Should these have the `/** Internal type. DO NOT USE DIRECTLY. */` comment just like those other internal types above?

eddeee888: This is a great question!
Last time we took away some internal types, then lots of users were unhappy about the change in a non-major release.
I'll definitely review these types and mark them as internal appropriately.

If there are strong arguments about keeping them public, and a way to make them unique, that could be an option too 🙂

PR #10898, opened 2026-07-15
fix(visitor-plugin-common): keep deferred fields optional with overlapping named @defer spreads

redox:
## Summary
- Named `@defer`/`@stream` fragment spreads were typed as a single union of all fragment fields. When a field like `id` was also selected outside the spread, intersecting with the required outer field eliminated the absent branch and made deferred-only fields required.
- Expand incremental named spreads into per-selection nodes (same shape as inline `@defer`) so each deferred field stays its own optional union. Fragment masking (`inlineFragmentTypes: "mask"`) is unchanged and still uses `Incremental<>` fragment refs.

## Test plan
- [x] Added regression test for overlapping `id` + named `@defer` (fails before fix, passes after)
- [x] Updated related `@defer` / `@skip`+`@defer` snapshots for per-field unions
- [x] `vitest` incremental delivery + skip/include defer suites green
- [x] Client-preset `@defer` tests green

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/defer-interface-inline-spread.md
+++ b/.changeset/defer-interface-inline-spread.md
@@ -1,6 +1,0 @@
----
-'@graphql-codegen/visitor-plugin-common': patch
-'@graphql-codegen/typescript-operations': patch
----
-
-Keep `@defer` fields optional on interface/union selections, surface nested `@defer`/`@stream` through inlined fragment spreads, and restore trailing `_` on implementing-type aliases when `omitOperationSuffix` is enabled
--- a/.changeset/defer-interface-inline-spread.md
+++ b/.changeset/defer-interface-inline-spread.md
@@ -1,0 +1,6 @@
+---
+'@graphql-codegen/visitor-plugin-common': patch
+'@graphql-codegen/typescript-operations': patch
+---
+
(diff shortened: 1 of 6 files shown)


redox:
@eddeee888 most of this work was driven by a wrongly typed fragment on our private repository; I'm not sure I get why so many changes were required but I appreciated reviewing the tests the AI wrote. Let me know if you have any opinion on this.

eddeee888:
Hi @redox ,

Apologies, I couldn't get to this sooner. 
We recently fixed an issue with `@defer` [here](https://github.com/dotansimha/graphql-code-generator/pull/10899). I wonder if the latest versions [here](https://github.com/dotansimha/graphql-code-generator/releases#release-release-1785915616164) already fixes the issue? 🤔

redox:
> Hi @redox ,
> 
> Apologies, I couldn't get to this sooner.

no worries!

> We recently fixed an issue with `@defer` [here](https://github.com/dotansimha/graphql-code-generator/pull/10899). I wonder if the latest versions [here](https://github.com/dotansimha/graphql-code-generator/releases#release-release-1785915616164) already fixes the issue? 🤔

No I tested the latest version and it's not fixing this overlapping issue

Issue #10881, opened 2026-06-25
typescript-operations: spread-level @include fields silently dropped from operation type (present on fragment type) with inlineFragmentTypes: 'inline'


ItaiYosephi:
**Which packages are impacted?**

- `@graphql-codegen/typescript-operations` `6.0.4`
- `@graphql-codegen/visitor-plugin-common` `7.1.1`
- `@graphql-codegen/typescript` `6.0.2`
- `graphql` `16.14.2`

## Describe the bug

When `@include`/`@skip` is placed on a **fragment spread** (`...SomeFragment @include(if: $x)`) inside a named fragment, the conditional fields are typed correctly (as optional) on the **fragment** type, but are **completely dropped** from any **operation** that spreads that fragment — under the default `inlineFragmentTypes: 'inline'`.

It is specific to **spread-level** directives. A **field-level** `@include` (e.g. `bio @include(if: $x)`) is correctly preserved as optional on both the fragment type **and** the operation type. So within the very same selection, codegen marks one conditional field optional while silently omitting another.

The dropped fields are not made optional — they are absent from the operation type entirely. Consuming code that reads them via `obj.field` then fails with `Property 'field' does not exist`, or (after a `'field' in obj` check) gets `unknown`, because the key is not declared on the operation type at all.

(The same drop happens for `@defer` on a fragment spread / inline fragment, but `@include`/`@skip` is the far more common case.)

## Reproduction

`schema.graphql`:

```graphql
type Query {
  user: User
}
type User {
  id: ID!
  name: String!
  bio: String!
  email: String!
}
```

`operations.graphql`:

```graphql
fragment EmailFields on User {
  email
}

fragment UserFields on User {
  id
  name
  bio @include(if: $withBio)                # FIELD-level @include  -> preserved
  ...EmailFields @include(if: $withEmail)   # SPREAD-level @include -> DROPPED
}

query GetUser($withBio: Boolean!, $withEmail: Boolean!) {
  user {
    ...UserFields
  }
}
```

`codegen.yml` (default config — `inlineFragmentTypes` left at its `'inline'` default):

```yaml
schema: ./schema.graphql
documents: ./operations.graphql
generates:
  ./out.ts:
    plugins:
      - typescript
      - typescript-operations
```

## Expected behaviour

The operation type should carry the same conditional fields as the fragment type — exactly as it already does for the field-level `bio?`:

```ts
export type UserFieldsFragment =
[...]

PR #10899, opened 2026-07-18
CODEGEN-940 Fix conditional directive (`@skip` and `@include`) incorrectly drop fields in Result when used  on Inline Fragment / Fragment Spread

eddeee888:
## Description

This PR fixes an issue where conditional directives (`@skip` and `@include`) are ignored in the Result field.
This happened because conditional directives were split [here](selectionNodesByTypeNameConditional) and weren't handle consistently

Related https://github.com/dotansimha/graphql-code-generator/issues/10881

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
+++ b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
@@ -872,6 +872,9 @@ export class SelectionSetToObject<
         );
       }
 
+      // #region Recursively find all fields and nested fields of the SelectionSet
+      // And by pushing the fields to `selectionNodes`
+      // The for loop continues to build SelectionSet & nested SelectionSet exhaustively
       if (
         parentSchemaType.name === selectionNode.onType ||
         parentSchemaType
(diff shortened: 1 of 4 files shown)


packages/plugins/typescript/operations/tests/ts-documents.skip-include-directives.spec.ts
```diff
@@ -1109,12 +1265,12 @@ describe('TypeScript Operations Plugin - @include and @skip with @defer', () =>
       "export type UserSkipQueryVariables = Exact<{ [key: string]: never; }>;
 
 
-      export type UserSkipQuery = { user: { id: string } & { name?: string, niName?: string } & { age?: number, createdAt?: string } & ({ age: number, createdAt: string } | { age?: never, createdAt?: never }) | null };
+      export type UserSkipQuery = { user: { id: string } & { name?: string, niName?: string } & { age?: number, createdAt?: string } & ({ age?: number, createdAt?: string } | { age?: never, createdAt?: never }) | null };
```
eddeee888: This type change actually doesn't have any effect on the final merged type.

Here are the scenarios at runtime:
1. When `@include(if: false)`, the fields are not available i.e. `{ }`. This is captured in the `{ age?: number, createdAt?: string }` group where each field is `undefined`
2. When `@include(if: true)`:
   a. the fields are not available if they are deferred
   b. the fields are available if they are not deferred. HOWEVER, the fields could be `undefined` because it's used in conjunction with `@skip`.

packages/plugins/typescript/operations/tests/ts-documents.skip-include-directives.spec.ts
```diff
@@ -1065,7 +1221,7 @@ describe('TypeScript Operations Plugin - @include and @skip with @defer', () =>
         id: ID!
         name: String!
         nickName: String!
-        age: Int!
+        age: Int
```
eddeee888, line 1224: Making this nullable so we can see the difference between non-nullable and nullable fields when used with these directives

packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts
```diff
@@ -934,36 +949,46 @@ export class SelectionSetToObject<
...
+        isConditional:
+          hasConditionalDirectives(field.directives) ||
+          hasConditionalDirectives(field.fragmentDirectives),
+      })),
+      options.unsetTypes,
+    );
+    const transformedLinkFields = this._processor.transformLinkFields(
+      linkFields,
+      options.unsetTypes,
+    );
```
eddeee888, line 985: Drive-by splitting this to make it easier to debug in the future.

packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -26,10 +26,16 @@ import {
...
 import { parseMapper } from './mappers.js';
 import { DEFAULT_SCALARS } from './scalars.js';
-import { LoadedFragment, NormalizedScalarsMap, ParsedScalarsMap, ScalarsMap } from './types.js';
+import type { EnrichedFieldNode } from './selection-set-to-object.js';
+import type {
+  LoadedFragment,
+  NormalizedScalarsMap,
+  ParsedScalarsMap,
+  ScalarsMap,
+} from './types.js';
```
eddeee888, line 38: Drive-by importing as `type` as TS best practice

Issue #10896, opened 2026-07-13
`@include`/`@skip` on a fragment spread disables fragment masking (fields are inlined instead of ` $fragmentRefs`)


izumin5210:
### Which packages are impacted by your issue?

@graphql-codegen/visitor-plugin-common, @graphql-codegen/client-preset, @graphql-codegen/typescript-operations

### Describe the bug

With `inlineFragmentTypes: 'mask'` (the client preset default), putting `@include` or `@skip` on a **fragment spread** disables fragment masking for that spread: the fragment's fields are inlined into the parent operation type as optional fields, and the `' $fragmentRefs'` reference disappears.

As a result, passing the parent object to a component that accepts `FragmentType<typeof XFragment>` becomes a type error, and `useFragment` can no longer unmask it. Codebases relying on masked spreads are forced to either restructure their queries or work around it at the type level (casts / `@ts-expect-error`).

The `@defer` path is not affected: a spread with `@defer` keeps masking and emits `' $fragmentRefs'?: { 'XFragment': Incremental<XFragment> }`. So conditional directives and incremental directives — which are handled by adjacent code paths — behave asymmetrically.

This appears to be an accidental regression of the conditional-directives rework rather than an intentional change:

* the mask branch exists in the incremental (`@defer`) path (`selection-set-to-object.ts`, from [#9196](<https://github.com/dotansimha/graphql-code-generator/issues/9196>)) but is missing from the conditional path introduced by [#10645](<https://github.com/dotansimha/graphql-code-generator/issues/10645>) / [#10646](<https://github.com/dotansimha/graphql-code-generator/issues/10646>) (merged into v6 via [#10496](<https://github.com/dotansimha/graphql-code-generator/issues/10496>));
* the dedicated test suite (`ts-documents.skip-include-directives.spec.ts`) has no coverage for `inlineFragmentTypes: 'mask'`;
* neither the changelog nor the v5→v6 migration guide mentions the behavior change.

### Your Example Website or App

N/A (minimal reproduction below)

### Steps to Reproduce the Bug or Issue

1. Generate types with the client preset (fragment masking on by default) for:

```graphql
# schema
type Query {
  user: User
}

type User {
  id: ID!
  nicknames: [String!]
}
```

```graphql
# operation
query GetUser($withNicknames: Boolean!) {
  user {
    id
    ...UserNicknames @include(if: $withNicknames)
  }
}

fragment UserNicknames on User {
  nicknames
}
```

[...]

PR #10904, opened 2026-07-27
CODEGEN-947 - Fix client-preset fragment masking when conditional directives are used

eddeee888:
## Description

This PR fixes the issue where fragment spreads with conditional directives (`@include` or `@skip`)  do not get masked correctly

Related https://github.com/dotansimha/graphql-code-generator/issues/10896

## Type of change

## How Has This Been Tested?

- [x] Unit test


dotansimha (approved):

packages/presets/client/src/fragment-masking-plugin.ts
```diff
@@ -23,6 +23,12 @@ export function makeFragmentData<
 const defaultUnmaskFunctionName = 'useFragment';
 
 const createUnmaskFunctionTypeDefinitions = (unmaskFunctionName = defaultUnmaskFunctionName) => [
+  `// return partial if \`fragmentType\` is partial e.g. because of conditional directives
```
sandhose, line 26: I *think* the ordering here matters? For some reason, all my `useFragment`s started returning `Partial<TType>` instead of `TType`. This is the case because `T` is valid for `Partial<T>` (but not the other way around), and typescript overloads will use the first matching?

eddeee888: Thanks @sandhose , I'll take a look

Issue #10901, opened 2026-07-21
__TypeKind missing from v6 client-preset output of introspection query


markdascher:
### Which packages are impacted by your issue?

@graphql-codegen/typescript-operations, @graphql-codegen/client-preset

### Describe the bug

As described in #10479, going from client-preset v5 to v6 removes a bunch of unnecessary schema types. This is good.

However, there seems to be some collateral damage. When operating on an introspection query, codegen now produces invalid typescript that references a `__TypeKind` enum without defining it anywhere. v5 may have been unintentionally working because the `typescript` plugin would define `__TypeKind` whether we needed it or not. But now v6 omits it even if it's needed.

I don't know if `__TypeKind` is the only case like this, but it's the only one I've run into.

Similar to #1395, a workaround is to add something like this to the codegen output:

```js
import { TypeKind as __TypeKind } from 'graphql';
```

I imagine that manually re-adding the `typescript` plugin would work too, but that would bring in too many unnecessary types along with it.

### Your Example Website or App

https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/typescript/tests/typescript.spec.ts

### Steps to Reproduce the Bug or Issue

The query from [typescript.spec.ts](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/typescript/tests/typescript.spec.ts) is a good enough example:

```gql
query Info {
  __type(name: "Post") {
    name
    fields {
      name
      type {
        name
        kind
      }
    }
  }
}
```

Run codegen on that using the client preset, and it'll produce output that fails to compile.

```
Cannot find name '__TypeKind'.
```

### Expected behavior

The codegen output should define the `__TypeKind` enum if its own output references such a type.

If `__TypeKind` isn't referenced in the output, then the codegen output should continue to omit the `__TypeKind` enum, as it does already.

### Platform

Pretty sure I'm on the latest of everything:

@graphql-codegen/add: 7.1.0
@graphql-codegen/cli: 7.2.0
@graphql-codegen/client-preset: 6.1.0
@graphql-codegen/core: 6.2.0
@graphql-codegen/gql-tag-operations: 6.1.0
@graphql-codegen/plugin-helpers: 7.1.0
@graphql-codegen/schema-ast: 6.1.0
@graphql-codegen/typed-document-node: 7.1.0
@graphql-codegen/typescript: 6.1.0
@graphql-codegen/typescript-operations: 6.1.1
[...]

PR #10905, opened 2026-07-28
CODEGEN-952 - Fix missing introspection types when queried

eddeee888:
## Description

Fix missing import of introspection types from `graphql` when used.

Note that this approach is different from `typescript` plugin's where it recursively generates all introspection types, regardless of whether it's used or not, for a couple of benefits:
- less generated code, because we import rather than copying the types
- don't have to rely on `TypeScriptVisitor` from `typescript` plugin, ensuring we avoid bad dependencies like previously

Related https://github.com/dotansimha/graphql-code-generator/issues/10901

## Type of change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can
reproduce. Please also list any relevant details for your test configuration

- [x] Unit test


packages/plugins/other/visitor-plugin-common/src/selection-set-processor/pre-resolve-types.ts
```diff
@@ -15,6 +20,7 @@ export class PreResolveTypesProcessor extends BaseSelectionSetProcessor<Selectio
       {
         type,
         name,
+        isIntrospectionType: false,
```
eddeee888: `__typename` field is never `introspectionType`

packages/plugins/other/visitor-plugin-common/src/selection-set-processor/pre-resolve-types.ts
```diff
@@ -81,10 +90,12 @@ export class PreResolveTypesProcessor extends BaseSelectionSetProcessor<Selectio
         return {
           name,
           type: `'${schemaType.name}'`,
+          isIntrospectionType: false,
         };
```
eddeee888, line 84: `__typename` field is never `introspectionType`

packages/plugins/other/visitor-plugin-common/src/selection-set-processor/pre-resolve-types.ts
```diff
@@ -124,6 +137,7 @@ export class PreResolveTypesProcessor extends BaseSelectionSetProcessor<Selectio
     return fields.map(field => ({
       name: field.alias || field.name,
       type: unsetTypes ? 'never' : field.selectionSet,
+      isIntrospectionType: false,
```
eddeee888: `linkField` shouldn't be introspection types

Issue #9477, opened 2023-05-30
invalid esm import file url under windows (`ERR_UNSUPPORTED_ESM_URL_SCHEME`)


gibso:
### Which packages are impacted by your issue?

@graphql-codegen/cli, @graphql-codegen/typescript-resolvers, @graphql-codegen/typescript

### Describe the bug

We are using codegen to generate types for our graphql schema with @graphql-codegen/cli, @graphql-codegen/typescript and @graphql-codegen/typescript-resolvers 
Our codegen config references a relative schema file 
```typescript
const config: CodegenConfig = {
  ...
  schema: "dist/api/schema.js",
  ...
};
```

Furthermore we use the `graphql-codegen-esm` binary to execute our code, since we are using ESM in our project.
Executing codegen works fine under Linux, but under Windows we get the message

```bash
[FAILED]         Unable to load from file "C:/Users/xxxxxx/dev_win/xxxxx/dist/api/schema.js": 
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file and data are supported by the default ESM loader. 
On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

I do not know how and where the relative path to that file is parsed from the config into an absolute file. But on windows this results in a path beginning with "C:/...". This is not compatible with ESM imports though, it needs to be `file:///c:/x/y/z instead of c:/x/y/z` (see nodejs/node#31710). When I manipulate the code that imports this file in `node_modules\@graphql-tools\code-file-loader\esm\load-from-module.js` to append `file:///` to the path, it actually works fine under windows.

Thank you!

### Your Example Website or App

not-available

### Steps to Reproduce the Bug or Issue

1. Create a project with esm and codegen
2. reference as schema a relative file in the configuration
3. run codegen (under windows)

### Expected behavior

I expect the code generation to run, but instead I receive an error.

### Platform

- OS: Windows
- NodeJS: v18.16.0
- `graphql` version: 16.6.0
- "@graphql-codegen/add": 5.0.0
- "@graphql-codegen/cli": 4.0.0
- "@graphql-codegen/typescript": 4.0.0
- "@graphql-codegen/typescript-resolvers": 4.0.0

### Codegen Config File

import { CodegenConfig } from "@graphql-codegen/cli";

const config: CodegenConfig = {
  overwrite: true,
  schema: "dist/api/schema.js",
  generates: {
    "src/api/generated/types.ts": {
[...]

Issue #8887, opened 2023-02-01
`afterAllFileWrite` hook fails on Windows when near-operation-file preset is used


jet2jet:
### Which packages are impacted by your issue?

@graphql-codegen/cli

### Describe the bug

When using `afterAllFileWrite` hook with prettier or etc. in Windows, the invalid file path is passed to the tool and causing error.

### Your Example Website or App

https://github.com/jet2jet/graphql-code-generator-issue-hooks-failed-on-windows

### Steps to Reproduce the Bug or Issue

1. Checkout the example repository
2. Run `npm run codegen` in Windows (not in WSL)

### Expected behavior

No error has occurred (prettier success)

### Screenshots or Videos

Output is below:

```
F:\Test\gql-codegen-test\graphql-code-generator-issue-hooks-failed-on-windows>npm run codegen

> graphql-codegen-issue-template@1.0.0 codegen
> graphql-codegen --config ./codegen.ts

✔ Parse Configuration
✔ Generate outputs
Error: Command failed: prettier --write types.ts F\:/Test/gql-codegen-test/graphql-code-generator-issue-hooks-failed-on-windows/document.generated.ts
[error] No files matching the pattern were found: "F\:/Test/gql-codegen-test/graphql-code-generator-issue-hooks-failed-on-windows/document.generated.ts".

    at ChildProcess.exithandler (child_process.js:383:12)
    at ChildProcess.emit (events.js:400:28)
    at ChildProcess.emit (domain.js:475:12)
    at maybeClose (internal/child_process.js:1058:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:293:5) {
  killed: false,
  code: 2,
  signal: null,
  cmd: 'prettier --write types.ts F\\:/Test/gql-codegen-test/graphql-code-generator-issue-hooks-failed-on-windows/document.generated.ts'
}
```

### Platform

- OS: Windows
- NodeJS: 14.18.2
- `graphql` version: 16.2.0
- `@graphql-codegen/*` version(s): 2.13.7

### Codegen Config File

(see the repository)

### Additional context

When the file path includes spaces, the argument passed to prettier is enclosed with single quotations, which are invalid in Windows.

Cellule:
For what it's worth, I pinpointed the problem to this line

[...]

Issue #7699, opened 2022-03-28
Codegen generating inconsistent files styling in Windows VS Mac/Linux


boi87:
**Describe the bug**

Hi, I'm having a strange issue while using codegen. When I generate code, my resulting file has a different "styling" then the one in the codebase. 

For some reason, after running `yarn codegen`, some of my `export type XYZ` statements are moved and grouped together on top of the file.
I'm the only one having this issue in my company, and I am also the only dev on a Windows machine - all my colleagues are using Mac/Linux.

Our build pipelines generated file is consistent with the one of my colleagues, and therefore generating conflicts (and failing the pipeline) of all my PRs. 

This is a snippet from my `api-types` file changes after I run `yarn codegen`: as you can see a lot of `export type` statements were moved together to be at the top of the file and they were removed from their original positions further down in the code.

![image](https://user-images.githubusercontent.com/25410752/160376961-0997597f-d89f-4bd5-9b45-06d281ea95ed.png)

Example of a couple of statements that were moved to the top:

![image](https://user-images.githubusercontent.com/25410752/160377303-3db5223a-4bee-48c4-b1c7-dc1245f1971a.png)

**To Reproduce**

1. I checkout my company's `master` branch.
2. I run `yarn codegen`. 

Can't share my `schema` unfortunately 😢 .

6. My `graphql-codegen.yml` config file:

```
overwrite: true
schema:
  - '../pkg/graphql/uk/schema/gql/*.graphql'
  - '../pkg/approval/api/graphql/*.graphql'
documents:
  - './src/**/*.ts'
  - './src/**/*.tsx'
generates:
  src/api-types.tsx:
    config:
      withComponent: true
      withHOC: false
      enumValues:
        Gender: ./customTypes#Gender
        PersonalTitle: ./customTypes#PersonalTitle
      apolloReactHooksImportFrom: ./components/GraphQL/hooks
      scalars:
        Time: string
        TaxCode: string
        Hours: number

    plugins:
      - add:
          content: '/* eslint-disable */'
      - 'typescript'
      - 'typescript-operations'
      - 'typescript-react-apollo'

```

**Expected behavior**
3. Expected result: no code changes.
4. Actual result: the snippets above describe a part of it. No actual code generation changes, but styling/sorting differences happen.

**Environment:**

- OS: Windows 11 (was also happening on Windows 10)
```
    "@graphql-codegen/add": "^3.1.1",
    "@graphql-codegen/cli": "2.6.2",
[...]

PR #10908, opened 2026-08-01
fix: run the windows-latest CI legs on Windows

gaoflow:
The `test` and `test-watcher` jobs in `.github/workflows/main.yml` each keep a single `windows-latest` leg via `include:`, but that leg has never run on Windows:

- the include entry uses `node-version: 22` while the axis is `node_version`, so the windows combination is created with an empty `matrix.node_version` (visible as the double space in the job name);
- `runs-on` is hardcoded to `ubuntu-latest`, so the leg labelled `(windows-latest)` executes on Ubuntu with the runner's default Node.

The latest run on master (30454416097) still reports both `(windows-latest)` jobs with `labels: ["ubuntu-latest"]`. The include entry has carried the wrong key since it was added in 2021-11 (4c42b3bc, "run tests on windows (#7048)").

Changes: consume `matrix.os` in `runs-on` for both jobs; fix the include key to `node_version` so the Windows leg uses the pinned Node; add `scripts/check-workflows.js` + a `workflow-check` job that assert every matrix job consumes `matrix.os` and every `include` key matches a declared axis (actionlint flags neither fault, so the class needs an explicit check).

Verified locally: the new check fails on the previous file with the four violations (two per job) and passes after the change; the workflows still pass actionlint.

This is the first time the unit suite runs on a real Windows runner, so any Windows-only failure would be a pre-existing issue surfaced by this fix (see #9526, #7699, #8887, #9477), not a regression from it.


eddeee888:
Hi @gaoflow !
Thank you for finding the issue, and clearly explain the expected failure.

Hope you don't mind I've added this [commit](https://github.com/dotansimha/graphql-code-generator/pull/10908/commits/af0065156000daaa26c61b5741e2aac832657d33) to revert the workflows check step. I recognise that it's good to have tests, but I'm on the fence atm because it is specifically testing for the interface of GitHub config, which could change and may create confusing errors.

If it becomes a frequent issue, we'll come back to review the workflow test approach.

PR #10921, opened 2026-08-17
Add overwrite.updateExistingFiles and overwrite.removeStaleFiles

eddeee888:
## Description

Whilst working on Server Preset **next version**, I noticed that resolver files may get removed in very specific conditions:
- Run in watch mode
- Server Preset returns a new file to Codegen to create (as part of `buildGeneratesSection`). At this point, Codegen tracks this file.
- Watch mode is triggered
- For any reason, Server Preset stops returning the said file as part of `buildGeneratesSection`. This is a deliberate move because sending files triggers filesystem write, which would be very slow for big codebases.
- The file is seen as stale by Codegen
- Codegen's `overwrite: true` config is used to remove stale files, so the resolver file is removed.

Problem: `overwrite: true` is also used to update exiting files. Server Preset needs to be able to update existing files to add resolvers/comments

For this reason, I'm extending the existing `overwrite` config to be more granular:

- `overwrite.updateExistingFiles`: Whether updating existing files is allowed
- `overwrite.removeStaleFiles`: Whether Codegen should remove stale files (files that are tracked in one run, but not the next) during watch mode

## Type of change

## TODO

- [ ] Add watcher test
- [ ] Update docs

### DIFF
--- a/packages/graphql-codegen-cli/tests/watcher.run.spec.ts
+++ b/packages/graphql-codegen-cli/tests/watcher.run.spec.ts
@@ -1,8 +1,11 @@
-import { mkdirSync, mkdtempSync, writeFileSync } from 'fs';
+import { existsSync, mkdirSync, mkdtempSync, writeFileSync } from 'fs';
 import * as path from 'path';
 import type { Mock } from 'vitest';
 import { CodegenContext } from '../src/config.js';
-import { createWatcher } from '../src/utils/watcher.js';
+import { generate } from '../src/generate-and-save.js';
+import * as watcherModule from '../src/utils/watcher.js';
+
+const createWatcherSpy = vi.spyOn(watcherModule, 'createWatcher');
 
 /**
  * waitForNextEvent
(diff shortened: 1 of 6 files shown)


packages/graphql-codegen-cli/tests/watcher.run.spec.ts
```diff
@@ -1,8 +1,11 @@
-import { mkdirSync, mkdtempSync, writeFileSync } from 'fs';
+import { existsSync, mkdirSync, mkdtempSync, writeFileSync } from 'fs';
 import * as path from 'path';
 import type { Mock } from 'vitest';
 import { CodegenContext } from '../src/config.js';
-import { createWatcher } from '../src/utils/watcher.js';
+import { generate } from '../src/generate-and-save.js';
+import * as watcherModule from '../src/utils/watcher.js';
+
+const createWatcherSpy = vi.spyOn(watcherModule, 'createWatcher');
```
eddeee888, line 8: New testing approach uses `generate` which is the end-to-end test behaviour.
However, because `generate` calls `createWatcher`, we cannot get access to the `stopWatching` to stop the test.

Adding a spy is a trick that allows us to get the `stopWatching` function from the return type.

packages/graphql-codegen-cli/tests/watcher.run.spec.ts
```diff
@@ -137,3 +140,176 @@ describe('Watch runs', () => {
...
+
+describe('Watch runs - overwrite.removeStaleFiles', () => {
+  const runWatchAndGetStopWatching = async (
+    codegenContext: ConstructorParameters<typeof CodegenContext>[0],
+  ) => {
+    const context = new CodegenContext(codegenContext);
+    const runningWatcher = generate(context);
+    await waitForNextEvent();
+
+    const { stopWatching } = createWatcherSpy.mock.results.at(-1)!.value as ReturnType<
```
eddeee888, line 152: TIL, `createWatcherSpy.mock.results.at(-1)!.value` means getting the last result returned by `createWatch`

PR #10930, opened 2026-09-01
Fix `overwrite` being ignored for preset-based `generates` outputs

eddeee888:
## Description

Fixes `overwrite` being silently ignored for **preset-based** `generates` outputs.

A `generates` entry that uses a preset and sets `overwrite` — e.g.
`overwrite: { removeStaleFiles: false }` — had that setting dropped entirely. In watch
mode its generated files were still deleted as "stale". This was originally hit via
`@eddeee888/gcg-typescript-resolver-files` (the server preset), which hard-codes
`overwrite.removeStaleFiles: false` specifically to stop watch mode deleting resolver files.
(`updateExistingFiles` was dropped identically; it just happened to coincide with the
fallback default, so only `removeStaleFiles` was observable.)

### Root cause

`normalizeOverwriteConfig` in `generate-and-save.ts` resolved `overwrite` per generated
file by looking the file's path up in `config.generates`. Two independent assumptions
break for a preset output:

1. **Wrong lookup key.** A preset's `generates` entry is keyed by its `baseOutputDir`, not
   by any generated file's path — so the lookup never matches. (Presets can also emit files
   outside that dir, so path-prefix matching wouldn't be a valid fix either.)
2. **`isConfiguredOutput` required a `plugins` key**, which preset entries don't have — the
   preset supplies plugins itself.

Both cases fell through to the global `config.overwrite` (default `true` → `removeStaleFiles: true`).

### Fix

Don't look it up by path — carry it on the file, exactly as the codebase already does for
`hooks` and `contentComparison`:

- `Types.FileOutput` gains `overwrite?: boolean | Partial<NormalizedOverwriteOption>`.
- `codegen.ts` sets `overwrite: outputConfig.overwrite` on each pushed file. Intentionally
  **not** `outputArgs.overwrite ?? outputConfig.overwrite` (unlike `contentComparison`) —
  that would grant presets a new per-file capability from `buildGeneratesSection` that this
  bug doesn't need.
- `normalizeOverwriteConfig` now takes the two `overwrite` values directly
  (`configOverwrite`, `fileOverwrite`) and resolves `fileOverwrite ?? configOverwrite ?? true`
  (`??`, not `||`, is load-bearing — an explicit `overwrite: false` must survive). The
  now-unused `isConfiguredOutput` is removed.
- Stale-file tracking stores the `{ filename, overwrite }` projection so a file that
  disappears is still judged by the `overwrite` of the entry that produced it — including
  after its `generates` entry has been removed from config.

This keeps plain-plugin `overwrite` working exactly as before (the same value is now carried
on the file instead of re-looked-up by path), and additionally honors it for preset outputs.

## Type of change

## How Has This Been Tested?

- [x] Unit test
- [x] Alpha version test

## Other changes

Drive-by: the pre-existing `contentComparison` watch tests were switched from `@graphql-codegen/add` to the same inline-plugin stub the new tests use,

### DIFF
--- a/packages/graphql-codegen-cli/tests/watcher.run.spec.ts
+++ b/packages/graphql-codegen-cli/tests/watcher.run.spec.ts
@@ -9,7 +9,6 @@ import {
 } from 'fs';
 import * as path from 'path';
 import type { Mock } from 'vitest';
-import * as addPlugin from '@graphql-codegen/add';
 import type { Types } from '@graphql-codegen/plugin-helpers';
 import { CodegenContext } from '../src/config.js';
 import { generate } from '../src/generate-and-save.js';
@@ -613,8 +612,8 @@ describe('Watch runs - externally modified output files', () => {
                   schemaAst: options.schemaAst,
                   documents: [],
                   config: {},
-                  pluginMap: { add: addPlugin },
-                  plugins: [{ add: { content: 'Default Content' } }],
+                  pluginMap: { inline: { plugin: () => 'Default Content' } },
+                  plugins: [{ inline: {} }],
                   contentComparison: 'disk',
                 },
               ],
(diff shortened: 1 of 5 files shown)


packages/graphql-codegen-cli/tests/watcher.run.spec.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from 'fs';
 import * as path from 'path';
 import type { Mock } from 'vitest';
-import * as addPlugin from '@graphql-codegen/add';
```
eddeee888, line 12: Drive-by removing `addPlugin` because we can use a simple inline version

PR #10931, opened 2026-09-01
fix: patch bob-the-bundler so pnpm build actually produces dist on Windows

eddeee888:
## Description

Root-caused and fixed the Windows `pnpm build` failure: **every** package's `dist/` was silently coming out empty (not just `@graphql-codegen/cli` — that package's `postbuild` script was just the first thing to touch `dist` and crash, which masked that nothing had actually been built at all).

**Root cause:** `bob-the-bundler`'s `getWorkspacePackagePaths` builds glob patterns with `path.join(workspacePattern, 'package.json')`. `path.join` uses the native OS separator, so on Windows this turns a pattern like `packages/*` into `packages\*\package.json` before handing it to `globby`/`fast-glob`. But `fast-glob` patterns must use `/` on every platform — backslash is its escape character, not a directory separator — so every workspace glob silently matches **zero** packages on Windows. TypeScript still compiles everything correctly into `.bob/`, but bob's per-package copy step into `dist/` then iterates an empty list and does nothing, while `bob build` still exits `0` (an empty `Promise.all` isn't an error).

Confirmed directly rather than by inference:
- Instrumented `bob build` to log `packageInfoList.length` → `0` on Windows.
- Isolated repro: `path.join('packages/*', 'package.json')` → `0` matches via `globby`; the posix-joined equivalent → matches correctly.

**Fix:** a `pnpm patch` on `bob-the-bundler@7.0.1` changing that one call to `path.posix.join`. This is upstream third-party code (not part of this repo), so it's patched via `pnpm patch` rather than edited directly; happy to also file it against `kamilkisiela/bob`.

This supersedes the earlier approach in this PR (defensive fallbacks in `inject-cli-version.mjs`/`fix-bin.js` that reconstructed missing files from `.bob`). Those treated a symptom on one package; the actual bug meant **no** package was ever built on Windows, so this PR now fixes the cause instead.

## Type of change

## How Has This Been Tested?

Verified end-to-end on Windows with a full clean install to match how CI installs:

```
rm -rf node_modules && pnpm install && pnpm build
```

- All 18 buildable packages print `✔ Built ...` (the 19th, `client-preset-swc-plugin`, is intentionally build-skipped per its own config).
- `dist/` is populated for every package, including `@graphql-codegen/cli`.
- The built CLI runs correctly: `node packages/graphql-codegen-cli/dist/cjs/bin.js --version` prints the correct version.


eddeee888:
Status update on Windows CI for this PR.

**Fixed and confirmed on real Windows CI (this PR's own runs):**
1. `bob build`'s copy step (`.bob` → `dist/`) intermittently drops a file on Windows without failing. `scripts/inject-cli-version.mjs`/`fix-bin.js` now recover `version.js`/`bin.js` from `.bob` when that happens — confirmed firing and recovering correctly on this PR's Windows runs.
2. Watch mode's `ignore` glob patterns (passed to `@parcel/watcher`) were built with `path.relative()`/`path.join()`, which produce backslash paths on Windows — these never matched anything, so the CLI's own generated output was never excluded from its own file-watcher. Fixed in `watcher.ts` (+ a matching bug in the test helper's own path conversion). `tests/watcher.patterns.spec.ts` (14 tests) now passes cleanly on both Windows legs — zero failures from this class of bug as of the latest run.

**Still red, unrelated to this PR's changes:**
- `tests/integration.spec.ts` (in `@graphql-codegen/graphql-modules-preset`): a schema glob (`'./tests/test-files/modules/*/types/*.graphql'`, already POSIX-written) matches 0 files on Windows instead of 5. Not the same backslash-path bug — root cause not yet identified.
- `tests/watcher.run.spec.ts` (CLI package): initial watch-mode runs never write their output files on Windows (`existsSync` false right after the "initial run" comment, before any watcher event is even involved).

Both of these are pre-existing latent Windows incompatibilities that were invisible until this PR fixed the Build step — they've never had a chance to run on Windows before now (windows-latest silently ran on Ubuntu until #10908, then Build crashed every run until this PR). They're a separate investigation from what this PR set out to fix.

---

patches/bob-the-bundler@7.0.1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/dist/utils/get-workspace-package-paths.js b/dist/utils/get-workspace-package-paths.js
+index 7510c02d0ff5d8aaee00366c809001e8599fdb0b..b88c94857f37d6adf6a9006de0d975b1bc452cb0 100644
+--- a/dist/utils/get-workspace-package-paths.js
++++ b/dist/utils/get-workspace-package-paths.js
+@@ -5,7 +5,7 @@ export async function getWorkspacePackagePaths(workspaces, cwd = process.cwd())
+     const packageJSONPaths = await globby(workspaces
+         /** We are only interested in workspaces that are packages (for now.) */
+         .filter(workspacePattern => workspacePattern.startsWith('packages/'))
+-        .map(workspacePattern => path.join(workspacePattern, 'package.json')), {
++        .map(workspacePattern => path.posix.join(workspacePattern, 'package.json')), {
```
eddeee888, line 10: I considered bringing this to bob but it's an ok local fix for now.

PR #10936, opened 2026-09-06
Fix Windows watch-mode ignore globs and rimraf test-cleanup glob handling

eddeee888:
Extracted out of #10935 to keep it focused: Windows-specific watch-mode and test-utility fixes, plus their test coverage.

## Fix 1: watch-mode ignore glob patterns use `\` on Windows

`createWatcher` builds the `ignore` glob patterns handed to `@parcel/watcher` with `path.join()` / `path.sep`, which use `\` on Windows. `@parcel/watcher` expects glob-style ignore patterns, which are always forward-slash regardless of platform — so on Windows those patterns never matched, and generated output files were watched (and could re-trigger builds) instead of being ignored.

[watcher.ts](packages/graphql-codegen-cli/src/utils/watcher.ts) now always joins ignore-pattern segments with `/` instead of `path.join`/`path.sep`.

## Fix 2: `rimraf.sync()` rejects its own glob pattern on Windows

The test suite's `TempDir.clean()` helper passes a glob-style cleanup pattern to `rimraf.sync()`, which on Windows rejects it as containing illegal path characters unless `{ glob: true }` is explicitly passed.

[tests/utils.ts](packages/graphql-codegen-cli/tests/utils.ts) now passes `{ glob: true }`. This is test-only — not part of the published package — bundled here because it was needed to get the suite green on Windows alongside fix 1.

## Changed files

- [watcher.ts](packages/graphql-codegen-cli/src/utils/watcher.ts) — fix 1
- [assert-watcher-build-triggers.ts](packages/graphql-codegen-cli/tests/watcher-test-helpers/assert-watcher-build-triggers.ts) — test helper normalizes to forward-slash before comparing (fix 1's coverage)
- [tests/utils.ts](packages/graphql-codegen-cli/tests/utils.ts) — fix 2
- [generate-and-save.spec.ts](packages/graphql-codegen-cli/tests/generate-and-save.spec.ts) — carried over from the source branch; note this file's diff also includes some pre-existing Windows-only assertion branches unrelated to either fix above (multi-line `[FAILED]` prefix handling, source-path forward-slash normalization in error snapshots) that were needed to keep the suite green on Windows

Verified on Windows: `watcher.patterns.spec.ts` + `watcher.run.spec.ts` + `generate-and-save.spec.ts` all pass with just these files applied on top of plain `master`, and fail (9 tests) without them.

### DIFF
--- a/packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
+++ b/packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
@@ -1,4 +1,5 @@
-import { dirname, join } from 'path';
+import { platform } from 'os';
+import { dirname, join, sep } from 'path';
 import logSymbols from 'log-symbols';
 import { Types } from '@graphql-codegen/plugin-helpers';
 import '@graphql-codegen/testing';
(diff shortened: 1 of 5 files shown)


packages/graphql-codegen-cli/tests/generate-and-save.spec.ts
```diff
@@ -401,8 +402,34 @@ describe('generate-and-save', () => {
           false,
         );
       } catch {
-        const cwd = process.cwd(); // cwd is different for every machine, remember to replace local path with this after updating snapshot
-        expect(outputErrorSpy.mock.calls[0][0]).toMatchInlineSnapshot(`
+        // cwd is different for every machine, remember to replace local path with this after updating snapshot.
+        // GraphQL source names are always forward-slash, regardless of platform, so normalize to match.
+        const cwd = process.cwd().split(sep).join('/');
+        if (platform() === 'win32') {
```
eddeee888, line 408: Windows output a slightly different version compared to others, so we should do integration test like this, as much as possible, to avoid unexpected changes

Issue #10940, opened 2026-09-08
Out of memory generating types with `near-operation-file` + `typescript-operations`


jakubwawrzyczek:
### Which packages are impacted by your issue?

@graphql-codegen/visitor-plugin-common

### Describe the bug

We tried to upgrade `visitor-plugin-common` from `5.8.0` to `7.2.5`. After the upgrade, codegen runs out of memory and never finishes.

Our config has four outputs in which three of them still work. The one that breaks is using `near-operation-file` and `typescript-operations`.

Increasing memory does not help:
- 4 GB  ->  crashes after  75s
- 8 GB  ->  crashes after 112s

It used all 8 GB, so it is not a matter of raising the limit. The garbage collector was busy about 90% of the time and could not free anything, so something is being held onto.

## I looked into it and found two things

_I did this investigation with an AI agent doing the profiling and the patching. The measurements below are real, taken from heap snapshots and allocation profiles of our own repo, and I verified each change by re-running codegen. But I do not know this codebase well myself, so if you ask follow-up questions I may need a little time to come back with a proper answer._

### 1. The cache key is the entire list of field paths

The cache uses a label to look things up, and that label is every field name in the selection set glued into one string. For our schema a single label is 168 MB, and the labels are kept until the run ends, so memory fills up with labels.

In `transformSelectionSet`, [selection-set-to-object.ts#L1110C1-L1134C68](https://github.com/dotansimha/graphql-code-generator/blob/cf205e01ac5fe320f01e3cab00987d65620f1a4b/packages/plugins/other/visitor-plugin-common/src/selection-set-to-object.ts#L1110C1-L1134C68):

```typescript
const fieldSelections = [...getFieldNames({ selections, loadedFragments })].sort();
const cacheHashKey = `${fieldSelections.join(',')} @ ${possibleTypes.join('|')}`;
objMap.set(cacheHashKey, [result.mergedTypeString, fieldName]);
```

The key is every field path in the selection set glued together with commas, and it stays in processor.typeCache until the run ends. In our project the three biggest keys are 168 MB, 166 MB and 154 MB. They are strings.

We changed the key to a hash and took heap snapshots before and after, at the same memory limit:

`strings over 16KB:   517 MB in 261 strings   ->   30 MB in 298 strings`

So the key really was the problem for those 517 MB. The hash keeps the cache working:

```typescript
const h = createHash('sha1');
for (const f of fieldSelections) { h.update(f); h.update(','); }
[...]

PR #10944, opened 2026-09-12
[visitor-plugin-common] fix: memoize fragment field names in getFieldNames to stop exponential blow-up (#10940)

eddeee888:
## Description

This PR fixes one of the two root causes reported in #10940 (out-of-memory generating types with `near-operation-file` + `typescript-operations` on schemas with deeply nested, widely reused fragments): `getFieldNames()` in `visitor-plugin-common` recomputed a fragment's field names from scratch on every `FRAGMENT_SPREAD`, instead of computing them once per fragment and reusing them. When fragments are nested inside other reused fragments (a "diamond" shape), that work doubles at every level, turning a document with N fragments into O(2^N) work instead of O(N) — the same class of bug as [graphql-code-generator-community#752](https://github.com/dotansimha/graphql-code-generator-community/issues/752).

A fragment's own field names don't depend on where it's spread from, only on its own body, so they can be computed once and reused. This is scoped to `getFieldNames()` itself in `visitor-plugin-common` — it doesn't touch the run-wide `typeCache` or the giant joined-string cache key in `transformSelectionSet` (the other root cause described in #10940), which is left for a follow-up.

Related # (issue): #10940

## What's changed

- `packages/plugins/other/visitor-plugin-common/src/utils.ts`: `getFieldNames()` now takes an optional `fragmentFieldNamesCache` (a `Map<string, ReadonlySet<string>>`), threaded through the recursion. On a `FRAGMENT_SPREAD`, it computes that fragment's field names *relative to its own root* once, caches them by fragment name, and re-prefixes them with the current `parentName` on every reuse — so a fragment spread at different nesting depths within the same call still gets the correct fully-qualified paths.
- The checkpoint test from the previous commit (`get-field-names-perf.spec.ts`) is unchanged — it's the acceptance criterion for this fix.
- Added a changeset (`patch` on `@graphql-codegen/visitor-plugin-common`).

## Type of change

## How Has This Been Tested?

- [x] The checkpoint test (`getFieldNames (issue #10940) > does not re-walk a fragment subtree once per spread`) now passes: 42 `Set.add` calls instead of 1,048,576 for a 20-level diamond fragment chain.
- [x] Full `visitor-plugin-common` test suite (51 tests) passes.
- [x] Downstream consumers that exercise this code path — `typescript-operations`, `typescript`, `near-operation-file-preset`, `client-preset` (440 tests total) — all still pass.
- [x] `tsc --noEmit` on the whole repo is clean.

**Test Environment**:

- OS: Linux
- `@graphql-codegen/visitor-plugin-common`: current `master`
- NodeJS: 22

## Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works — *the checkpoint test from the previous commit now passes*

## Further comments

Root cause #1 from #10940 (the selection-set-to-object.ts cache key built from the full joined field-path list, measured at up to 168MB per key on the reporter's schema) is intentionally out of scope here — it's a separate, additive concern from the exponential re-walk fixed in this PR.

https://claude.ai/code/session_01RmAsddNmmJrhGmTSa1euU4


packages/plugins/other/visitor-plugin-common/src/utils.ts
```diff
@@ -682,19 +690,30 @@ export const getFieldNames = ({
...
+              .filter(def => def.name === fragmentName)
+              .flatMap(s => s.node.selectionSet.selections),
+            loadedFragments,
+            fragmentFieldNamesCache,
+          });
+          fragmentFieldNamesCache.set(fragmentName, relativeFieldNames);
+        }
+        for (const relativeFieldName of relativeFieldNames) {
+          fieldNames.add(parentName ? `${parentName}.${relativeFieldName}` : relativeFieldName);
+        }
```
eddeee888, line 720: What's this for? Why we need it?

eddeee888: The `relativeFieldNames` in the cache are computed once relative to the fragment's own root (`parentName ''`), so they can be reused no matter where the fragment is spread from. But different spread sites for the same fragment can have different `parentName`s, so on every cache hit we still need to re-join each cached relative name onto *this* call's `parentName` to get the correct fully-qualified path — that's what this loop does. Added a comment above it in 538068d to make that explicit.

---

eddeee888:
Checking with user for more info

Issue #10922, opened 2026-08-19
`typescript-operations` emits an unused schema-types import when fragment types are referenced


wassim-mirus:
### Which packages are impacted by your issue?

`@graphql-codegen/typescript-operations`

### Describe the bug

With `importSchemaTypesFrom` and `inlineFragmentTypes: 'combine'`, a document that only *spreads* a
fragment gets an `import type * as Types` it never references. Under `noUnusedLocals` that fails the
build.

## Reproduction

```bash
npm install
npm run generate
npm run typecheck
```

```
src/book.generated.ts(4,1): error TS6133: 'Types' is declared but its value is never read.
```

Two fragments, in separate files. Only `Category` selects the enum field:

```graphql
# src/category.graphql
fragment Category on Category { id kind }   # kind: CategoryKind!

# src/book.graphql
fragment Book on Book { id category { ...Category } }
```

`src/category.generated.ts` — import used, correct:

```ts
import type * as Types from '../types';

export type CategoryFragment = { id: string, kind: Types.CategoryKind };
```

`src/book.generated.ts` — import emitted, never referenced:

```ts
import { CategoryFragment } from './category.generated';
import type * as Types from '../types';   // <-- unused

export type BookFragment = { id: string, category: CategoryFragment };
```

## Expected

No `Types` import in `src/book.generated.ts`, since nothing in the file refers to it.

## Cause

`TypeScriptDocumentsVisitor.getExternalSchemaTypeImports()`
(`packages/plugins/typescript/operations/src/visitor.ts`) gates the import on `_usedSchemaTypes`:

```ts
const hasTypesToImport =
  Object.values(this._usedSchemaTypes).filter(
    value => value.type === 'GraphQLEnumType' || value.type === 'GraphQLInputObjectType',
  ).length > 0;

if (!hasTypesToImport) {
  return [];
}
```

`CategoryKind` is used by the document, so this is `true` for both files. But *used by the document*
and *named by the generated file* are different, and only the second justifies an import.
`_usedSchemaTypes` cannot answer the second on its own: it also decides whether enum and input
definitions are generated locally, which must account for types reached through fragments.

Only `inlineFragmentTypes` decides which of the two applies:

| `inlineFragmentTypes` | `BookFragment` | names `Types`? | `tsc` |
| --- | --- | --- | --- |
| `'inline'` (default) | `{ category: { id: string, kind: Types.CategoryKind } }` | yes | passes |
[...]

PR #10953, opened 2026-09-13
[typescript-operations] test: reproduce #10922 — unused schema-types import with combine fragments (failing)

eddeee888:
## Description

This is a **checkpoint PR**: it adds a failing test that proves the bug reported in
#10922 is real. It does not fix anything.

`typescript-operations`, when configured with both `importSchemaTypesFrom` and
`inlineFragmentTypes: 'combine'`, emits an unused `import type * as Types from '...'`
in generated files whose document only spreads an external fragment (and never names
a schema type itself). Under `noUnusedLocals`/`noUnusedImports` this fails the
TypeScript build.

The root cause (per the issue) is in
`TypeScriptDocumentsVisitor.getExternalSchemaTypeImports()`
(`packages/plugins/typescript/operations/src/visitor.ts`): the import is gated on
`_usedSchemaTypes`, which tracks whether a schema type is *used by the document*
(including through fragments, since it also drives local enum/input generation) —
not whether the *generated file itself* ends up naming the `Types` namespace. Those
two only diverge when `inlineFragmentTypes` is `'combine'` (or `'mask'`), because
only then does a spread fragment collapse to a bare `FooFragment` reference instead
of inlining the referenced schema types.

Related #10922

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — *test only, no fix yet*

Reporter's reproduction: https://stackblitz.com/edit/gdtmg1pc

## How Has This Been Tested?

Added `packages/plugins/typescript/operations/tests/ts-documents.import-schema-types-combine.spec.ts`,
mirroring the issue's repro: a `Category` fragment that selects an enum field, and a
`Book` fragment (external-fragment style, own file) that only spreads `...Category`.
With `importSchemaTypesFrom` + `inlineFragmentTypes: 'combine'`, the generated
`book.generated.ts` output should not import `Types` at all, since it only ever
references `CategoryFragment`.

This currently **fails** as expected:

```
AssertionError: expected 'import type * as Types from \'./types…' not to contain 'Types'

- Expected
+ Received

- Types
+ import type * as Types from './types';
+
+
+ /** Internal type. DO NOT USE DIRECTLY. */
+ export type Incremental<T> = T | { [P in keyof T]?: P extends ' $fragmentName' | '__typename' ? T[P] : never };
+ export type BookFragment = { id: string, category: CategoryFragment };
```

All 242 other existing tests in the `typescript-operations` package still pass.

- [x] Test A: `ts-documents.import-schema-types-combine.spec.ts` (new, failing — this is the point)

**Test Environment**:

- OS: Linux
- `@graphql-codegen/typescript-operations`: 6.1.6 (as reported)
- NodeJS: 22

## Checklist:

- [x] I have added tests that prove the bug described in #10922 is reproducible
- [ ] This PR does not need to merge or go green on its own — a follow-up PR builds
      the actual fix directly on top of this commit

## Further comments

This is intentionally left red. Nothing here should be "fixed" by skipping or
loosening the test — the next step is a separate PR that makes
`getExternalSchemaTypeImports()` (or equivalent) decide the import based on whether
the *generated file* itself names a schema type, not just whether the *document*
uses one.

https://claude.ai/code/session_01EefzkxyCDaB4zSdjEuGHou

---
_Generated by [Claude Code](https://claude.ai/code/session_01EefzkxyCDaB4zSdjEuGHou)_


packages/plugins/typescript/operations/tests/ts-documents.import-schema-types-combine.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        category {
+          ...Category
+        }
+      }
+    `);
+
+    const categoryFragmentDef = categoryFragmentDocument.definitions[0];
+    if (categoryFragmentDef.kind !== 'FragmentDefinition') {
+      throw new Error('Expected fragment definition');
+    }
```
eddeee888, line 52: Why do this?

packages/plugins/typescript/operations/tests/ts-documents.import-schema-types-combine.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      namespacedImportName: 'Types',
+    };
+
+    // category.generated.ts: selects `kind` directly, so it correctly imports `Types`
+    // and references `Types.CategoryKind`.
+    const categoryResult = mergeOutputs([
+      await plugin(schema, [{ document: categoryFragmentDocument }], config, {
+        outputFile: './category.generated.ts',
+      }),
+    ]);
```
eddeee888, line 69: Why not put both documents together?

packages/plugins/typescript/operations/tests/ts-documents.import-schema-types-combine.spec.ts
eddeee888: That's a type-narrowing guard, not new invention — it mirrors the same pattern already used a few lines above it in `ts-documents.external-fragments-enum.spec.ts` in this same directory.

`categoryFragmentDocument.definitions[0]` is typed as `DefinitionNode`, but `externalFragments[].node` on the plugin config requires `FragmentDefinitionNode`. Without the runtime `kind` check, `categoryFragmentDef` can't be passed as `node` below without a cast. The check both satisfies TS and fails loudly (rather than misconfiguring the repro silently) if `parse()` ever returns something unexpected.

Happy to switch it to a cast if you'd rather not have the extra guard — let me know which you prefer.

---

packages/plugins/typescript/operations/tests/ts-documents.import-schema-types-combine.spec.ts
eddeee888: Intentional, and actually load-bearing for reproducing the bug: this is modeling two separate generated files, the way `near-operation-file-preset` actually produces them — one `plugin()` call per source document, where `book.generated.ts` only learns about the `Category` fragment through `externalFragments` (as a fragment defined and compiled *elsewhere*), not by having both documents in the same document set.

If I merged `categoryFragmentDocument` and `bookFragmentDocument` into a single `plugin()` call, the two fragments would be seen together as one document, `_usedSchemaTypes` would be computed once for that combined set, and the enum usage from `Category` would "leak" into the same `_usedSchemaTypes` — which is exactly the assumption the bug report says is the correct one at the *document* level (it's used by the document), but wrong at the *generated file* level. The whole bug only shows up when the file that actually gets written (`book.generated.ts`) never itself names the enum — which requires it to be generated from a document that only contains the spread, with the enum-selecting fragment external to it.

So the split into two `plugin()` calls with `isExternal: true` isn't incidental — it's what makes this reproduce `near-operation-file-preset`'s actual per-file generation instead of a single merged-file case where the bug doesn't occur.

---

eddeee888:
Moved to https://github.com/dotansimha/graphql-code-generator/pull/10954

PR #10956, opened 2026-09-18
[graphql-codegen-cli] fix: resolve dynamic plugin/preset imports in ESM, including on Windows

eddeee888:
## Description

This PR fixes `makeDefaultLoader`'s ESM branch resolving bare module specifiers from the CLI package's own location instead of the consuming project's, so a plugin/preset only loaded if it happened to be reachable from the CLI's own `node_modules`.

- `codegen.ts`: `makeDefaultLoader` now always resolves `mod` via `relativeRequire.resolve(mod)` for both module systems, then wraps the resolved path in `pathToFileURL(resolved).href` before the ESM `import()` call. **Without the URL wrapping, a raw Windows path (`C:\...`) is misread as a `c:` protocol scheme and throws `ERR_UNSUPPORTED_ESM_URL_SCHEME`.**
- `.changeset/cli-esm-plugin-import-resolution.md`: patch changeset for `@graphql-codegen/cli`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] #10957's checkpoint passes: `dev-test/general` and `dev-test/apollo-tooling` can both find their own dependencies again, producing byte-identical output on `ubuntu-latest`, `windows-latest`, and `cjs`.

**Known, separate failure — not fixed by this PR:** `dev-test/apollo-tooling`'s `start` still fails past the point this PR's bug caused, on a pre-existing, unrelated bug in `near-operation-file-preset`'s own document resolution (filed as #10970). Not part of the `dev-tests` CI job either way.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

## Further comments

Builds directly on #10957 (the `windows-latest` CI leg + the `eddeee888:oss:issue-verify` checkpoint). This is the `eddeee888:oss:issue-fix` half. #10970 tracks the separate issue this uncovered.

PR #10957, opened 2026-09-19
[graphql-codegen-cli] test: add windows-latest coverage to dev-tests, then verify it catches a real ESM plugin-loader bug

eddeee888:
## Description

This PR adds `windows-latest` coverage to the `dev-tests` CI job, then uses it — together with the existing `ubuntu-latest` leg — to prove a real bug in the CLI's ESM plugin loader. This is an `eddeee888:oss:issue-verify` checkpoint: **intentionally left failing**.

- `.github/workflows/main.yml`: adds `windows-latest` to the `dev-tests` matrix and forces `shell: bash` on its Test step. **`windows-latest`'s default `pwsh` doesn't abort a multi-line `run:` block on a failing command the way bash's `-e` does**, so a failing `generate:esm` was previously masked by a later `git diff --exit-code` still reporting success.
- `packages/graphql-codegen-cli/package.json`: removes every plugin/preset devDependency that existed purely as a workaround for `makeDefaultLoader`'s ESM branch resolving bare specifiers from the CLI's own `node_modules` instead of the consuming project's — `typescript-operations`, `typescript-resolvers`, `typed-document-node`, `graphql-modules-preset`, `add`, `flow`, `flow-resolvers`, `typescript-graphql-files-modules`, and `near-operation-file-preset` (9 total, mixing `workspace:*` links and pinned external versions). **`typescript` stays, since it's also `require()`'d directly in this package's own test suite.**

## Type of change

- [x] This change requires a documentation update — N/A (CI + checkpoint only, no fix)

## How Has This Been Tested?

- [ ] `dev-tests` (esm) — intentionally failing: `dev-test/general` can no longer find most of its plugins, on both `ubuntu-latest` and the `windows-latest` leg this PR adds.
- [ ] `dev-test/apollo-tooling` — intentionally failing, verified separately (`pnpm --filter=dev-test-apollo-tooling start`): it isn't wired into the `dev-tests` job's generate checks, so this doesn't show up in that CI leg.
- [x] `dev-tests` (cjs) — stays green throughout, isolating the failure to the ESM-specific resolution bug.

## Checklist:

## Further comments

The fix lives in #10956, built on top of this branch, marked `eddeee888:oss:issue-fix`.


eddeee888:
Verification confirmed on `7cdff27`: `Validating dev-tests (windows-latest, esm)` fails with the expected error —

```
Unable to load template plugin matching 'typescript-operations'.
Reason:
  Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
```

— while `ubuntu-latest, esm` and everything else on the run is green. This is the intended, expected outcome: this branch is intentionally built on the pre-Windows-fix intermediate commit (`2e525a9` from #10956) specifically so this new CI leg has something real to fail against.

One thing worth flagging for anyone reviewing this: the first version of this leg (commit `e65dab0`) actually reproduced the same failure in its logs but the job still reported `success`, because `windows-latest` defaults multi-line `run:` blocks to `pwsh`, which doesn't stop the script on a failing external command the way bash's default `-e` does — execution silently continued to `git diff --exit-code`, found nothing to report, and that became the step's result. `7cdff27` adds `shell: bash` to fix that masking, independent of the ESM bug itself.

**This branch should not be merged.** The only change that should land for real is the `.github/workflows/main.yml` diff (`windows-latest` leg + `shell: bash`), cherry-picked onto #10956 (or `master` once that merges), where it's expected to pass against the already-fixed `codegen.ts`. Closing this once that's done.

---

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -31,18 +30,7 @@ const makeDefaultLoader = (from: string) => {
...
-        ? /**
-           * For ESM we currently have no "resolve path" solution
-           * as import.meta is unavailable in a CommonJS context
-           * and furthermore unavailable in stable Node.js.
-           **/
-          // FIXME(pnpm-update): this causes dev-test devDeps to be brought into CLI's package.json, which is not ideal.
-          // Note that `relativeRequire.resolve(mod)` seems to work correctly for ESM as well.
-          mod
-        : relativeRequire.resolve(mod)
-    );
```
eddeee888, line 45: Revert this. I want to see if windows-latest fails.

.github/workflows/main.yml
```diff
@@ -50,12 +50,20 @@ jobs:
 
   dev-tests:
     name: Validating dev-tests
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-latest]
```
eddeee888: Run windows-latest for both cjs and esm

.github/workflows/main.yml
```diff
@@ -76,6 +84,12 @@ jobs:
         env:
           CI: true
       - name: Test dev-tests ${{matrix.method}}
+        # windows-latest defaults multi-line `run:` blocks to pwsh, which does not
+        # stop the script when an earlier command exits non-zero (unlike bash's
+        # default -e). That silently turned a failing `generate:esm` into a green
+        # step here, because `git diff` then found nothing to report. Forcing bash
+        # (bundled via Git for Windows) makes this fail the way it does on Linux.
+        shell: bash
```
eddeee888, line 86: Why don't we have this for other windows tests?

eddeee888: Good question. The other windows-latest jobs in this workflow (`test`, `test-watcher`) don't hit this masking issue because their `run:` step is a single command (`run: pnpm test --exclude=...`). The bug here specifically needs *multiple sequential commands in one `run:` block*: pwsh (the windows-latest default shell) doesn't abort on a failing native command mid-script the way bash's `-e` does, so it keeps going to the next line. For a single-command step there's no "next line" to mask the failure — whatever that one command's exit code is becomes the step's result. `dev-tests` was the one job with two commands chained (`generate` then `git diff --exit-code`), which is exactly what let the first one's failure hide behind the second one's success.

That said, it's fragile to rely on "this step only has one line today" — anyone adding a second command to `test`/`test-watcher` later would silently reintroduce the same masking. Want me to add `shell: bash` to those steps too as a defensive default, even though nothing there is currently broken?

---

eddeee888:
Results for `a0019b0` (codegen.ts fully reverted to master's original pre-#10956 state; `windows-latest` now runs both `cjs`/`esm`):

- **`windows-latest, esm`: still fails** — but for a different reason than the intermediate-fix commit. Looking at the actual log, `dev-test/general`'s ESM generation now **succeeds** (no `ERR_UNSUPPORTED_ESM_URL_SCHEME`-style error) — confirming the original bare-`import(mod)` code doesn't hit that particular failure mode, since a bare package specifier never gets URL-parsed. The job still goes red because `dev-test/graphql-config`'s `generate:esm:multi` script fails separately:
  ```
  $ node ../../packages/graphql-codegen-cli/dist/esm/bin.js --config multi/.graphqlrc.yml || [ $? -eq 1 ]
  ✖ Parse Configuration [FAILED: Invalid Codegen Configuration! ...]
  [: invalid integer '$?'
  ```
  That script intentionally expects this invocation to fail with exit code 1 (multi-project config without `--project` should be rejected) and uses `$?`/`[ ]` shell-test syntax to check it — which only works under a POSIX shell. `shell: bash` on the *workflow* step doesn't help here because this line runs inside a *nested* `pnpm` script (`generate:esm` → `generate:esm:multi`), and pnpm's own script runner still defaults to a non-POSIX shell for `package.json` scripts. `generate:cjs:multi` has the identical pattern and would hit the same issue — its job was cancelled by matrix `fail-fast` before it got there.

- **`windows-latest, cjs`: cancelled** (matrix default `fail-fast: true` killed it once `esm` failed). No independent data point there, but it very likely would've hit the same `graphql-config` issue.

So: the original resolution-root bug itself is confirmed **not** to be the thing failing dev-tests' own generation on Windows — consistent with why nobody caught it via this suite before. But there's now a third, independent, pre-existing Windows-portability bug in `dev-test/graphql-config`'s package.json scripts (the `|| [ $? -eq 1 ]` pattern) that blocks a `windows-latest` `dev-tests` leg from ever going green at all, on any codegen.ts state — that needs a cross-platform fix (e.g. a small Node helper instead of shell `$?`/`[ ]` syntax) before this CI leg can land for real.

---

.github/workflows/main.yml
eddeee888: Current setting is ok. Resolved.

.github/workflows/main.yml
```diff
@@ -50,9 +50,14 @@ jobs:
...
   dev-tests:
     name: Validating dev-tests
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        # windows-latest runs both methods: dynamic plugin loading goes through a
+        # real `import()` in the ESM build, whose loader is sensitive to path/URL
+        # handling in ways `require()` (used by the CJS build) is not. See #10935
+        # and the fix in #10956 for a concrete regression this would catch.
```
eddeee888: Remove comment, having this in PR is ok

PR #10959, opened 2026-09-19
fix(cli): route preset-resolution errors into ctx.errors so allowPartialOutputs works

eddeee888:
## Description

`dev-test/general/codegen.ts` configures three outputs with `preset: 'client'`, but `dev-test/general/package.json` never lists `@graphql-codegen/client-preset` as a devDependency. Every `dev-test:generate` run has been failing to resolve that preset ("Unable to find preset matching 'client'") — visible as a red `✖ FAILED` line in the Listr output — yet CI has never caught it, on any OS.

Root cause, in `packages/graphql-codegen-cli/src/codegen.ts`: `executeCodegen()` only reports failure based on `ctx.errors`, which is populated by `wrapTask()`. `getPresetByName(...)` was called before any `wrapTask()`-wrapped step, so when it throws, Listr shows the failure but it never reaches `ctx.errors` — `executeCodegen()` returns `error: null`, and the CLI exits 0 regardless of `allowPartialOutputs` (which defaults to `false` specifically to prevent exactly this: "one failed `generates` means no output is written to files").

This PR wraps the preset lookup in a try/catch that pushes into `ctx.errors`, the same way every other step in this task already does — so `allowPartialOutputs`'s existing default now actually applies to preset-resolution failures too. No CI-side workaround needed; this fixes it for every user of the CLI, not just this one dev-test.

*(Earlier version of this PR added a CI-side grep check as a workaround before this real fix was written — see comment history.)*

**Now based on #10960** — the `issue-verify`-style checkpoint PR that added a deterministic failing unit test for this exact bug. Stacking this fix on top of it proves the fix directly: the checkpoint test, red on #10960 alone, goes green once this fix lands on top.

Related # (issue): none filed — Relates to #10960 (checkpoint test this fixes) and #10957 (the unrelated Windows CI gap where this was originally surfaced).

## Type of change

## What

- `packages/graphql-codegen-cli/src/codegen.ts`: wrap the `getPresetByName(...)` call in try/catch, pushing any error into `ctx.errors` before rethrowing, with a short comment explaining why that push has to happen before the rethrow (otherwise the CLI reports success even though the terminal printed the error).
- `.changeset/cli-preset-resolution-errors-fail-build.md`: patch changeset for `@graphql-codegen/cli`.

## How Has This Been Tested?

- [x] #10960's checkpoint test (`generate-and-save.spec.ts`, `config.allowPartialOutputs`): red on its own branch, green on this PR's head (`pnpm --filter="@graphql-codegen/cli" test generate-and-save` → 24 passed, 0 failed) — confirms the fix directly, no manual repro needed.
- [x] Isolated repro: a codegen config pointing `preset` at a package that doesn't exist. Before this fix, the CLI exited `0`; after, it exits `1` — verified for both the `esm` and `cjs` builds.
- [x] `pnpm --filter="@graphql-codegen/cli" test`: 170 passed, 2 skipped, 5 failed — same 5 pre-existing, unrelated snapshot failures (a terminal-glyph rendering difference in `@inquirer/testing`, confirmed present on unmodified `master` too), no new regressions.
- [x] Real CI on this PR: all `Unit Test on Node X (os) and GraphQL vY` jobs green, confirming the checkpoint test passes on the actual pipeline, not just locally.

**Test Environment**:

- OS: ubuntu-latest (GitHub-hosted runner) and local
- `@graphql-codegen/cli`: this fix
- NodeJS: 24

## Checklist:

- [ ] Everything else: N/A — no user-facing docs to update for this bug fix

## Further comments

This fixes the CLI's error propagation, not `dev-test/general`'s missing `@graphql-codegen/client-preset` devDependency — once this merges, that fixture's own `Validating dev-tests` job should go red (correctly) until someone adds that dependency. That's a separate, follow-up fix.

---

https://claude.ai/code/session_01P2vvyEtEqahXy3gCBAJBTT


eddeee888:
Confirmed on `411c983`: `Validating dev-tests (cjs)` fails as designed —

```
dev-test/graphql-config generate:cjs: ✔ Generate outputs
dev-test/graphql-config generate:cjs: Done
##[error]dev-test generation reported an unresolved preset (e.g. missing @graphql-codegen/client-preset devDependency) but the CLI still exited 0
##[error]Process completed with exit code 1.
```

`pnpm run dev-test:generate:cjs` itself exited 0 (both `dev-test/general` and `dev-test/graphql-config` report clean success) — the new grep check is what caught the swallowed "Unable to find preset matching" failure and forced the step red. This is the expected, intended result: it proves the gap was real on `master` as-is, on ubuntu-latest, with no Windows or ESM-fix involvement at all.

`esm` was cancelled by matrix fail-fast (no `fail-fast: false` on this branch — that's a separate, unrelated change on #10957 if useful here too).

This PR does what it set out to do: the test setup now fails where it silently didn't before. The actual fixes — adding `@graphql-codegen/client-preset` to `dev-test/general`'s devDependencies, and/or making `codegen.ts` route preset-resolution errors into `ctx.errors` — are follow-ups, not part of this PR's scope.

---

---

packages/graphql-codegen-cli/src/codegen.ts
```diff
@@ -240,11 +240,17 @@ export async function executeCodegen(
...
-                  : null;
+                let preset: Types.OutputPreset | null;
+                try {
+                  preset = hasPreset
+                    ? typeof outputConfig.preset === 'string'
+                      ? await getPresetByName(outputConfig.preset, makeDefaultLoader(context.cwd))
+                      : outputConfig.preset
+                    : null;
+                } catch (error: any) {
+                  ctx.errors.push(error);
```
eddeee888, line 254: Add a 1 or 2 line concise comment here saying:
- `ctx.errors` is what used to report errors, so errors need to be added, before throwing
- Otherwise CLI thinks it's successful, even if the terminal prints errors

eddeee888: Added:
```ts
// ctx.errors is what determines CLI success/failure, so it must be pushed here
// before rethrowing — otherwise the CLI reports success even though the
// terminal printed this error.
```
Pushed as `ddc130e`. `generate-and-save.spec.ts` still 24/24.

---

---

eddeee888:
Failing is expected, will fix it in next PR

PR #10965, opened 2026-09-19
[graphql-codegen-cli] test: reproduce hooks.ts POSIX-only shell quoting breaking on Windows (failing)

eddeee888:
## Description

This is a checkpoint, not a fix — it proves finding 4 from #10957 is real, with a deterministic unit test.

Root cause: in `packages/graphql-codegen-cli/src/hooks.ts`, `executeHooks()` quotes hook-script arguments via the `shell-quote` package's `quote()`, which produces POSIX-style quoting unconditionally (wraps any argument containing a backslash — i.e. any Windows path — in single quotes). That command string is handed to `execShellCommand()`, which calls Node's `child_process.exec()` with no explicit `shell` option, so on Windows it defaults to `cmd.exe`. `cmd.exe` doesn't strip single quotes, so any lifecycle hook receiving a Windows-shaped path argument gets the literal quote characters baked into it.

Confirmed live on real CI: `dev-test/general/codegen.ts` configures `hooks: { afterAllFileWrite: ['prettier --write'] }`, and on `windows-latest` (both `cjs` and `esm` legs, PR #10957 commit `17553f5` and PR #10956 commit `833aaae`) it fails with:
```
Error: Command failed: prettier --write ./test-schema/resolvers-types.ts ...
[error] No files matching the pattern were found: "'D:\a\graphql-code-generator\graphql-code-generator\dev-test\general\modules\types.ts'".
[ELIFECYCLE] Command failed with exit code 1.
```
(5 occurrences, one per path under `dev-test/general/modules/`) — [job log](https://github.com/dotansimha/graphql-code-generator/actions/runs/35432724450/job/105870054876).

Related # (issue): none filed — Relates to #10957 (where this was surfaced) and #10956 (where it also reproduces). This is a pre-existing bug in `hooks.ts`, unrelated to either PR's own change.

## Type of change

- [x] Bug fix — this is a checkpoint test only, not a fix.

## What

- `packages/graphql-codegen-cli/tests/hooks.spec.ts` (new): two platform-gated `describe` blocks, each skipped via `describe.skipIf(...)` on the actual `platform()` from Node's `os` module, asserting the full command string handed to `child_process.exec()` via `toMatchInlineSnapshot`:
  - `lifecycleHooks - win32`: snapshot intentionally locks in the current *buggy* output — the Windows-shaped path wrapped in POSIX single quotes — with a `FIXME:` comment stating exactly what it should become once the platform-conditional fix lands.
  - `lifecycleHooks - posix`: snapshot locks in the existing, correct POSIX quoting behavior (a path containing a space gets wrapped in single quotes), so a platform-conditional fix for the `win32` case can't regress it.

`packages/graphql-codegen-cli`'s unit tests already run on a real `windows-latest` CI runner (the `test` job's matrix in `.github/workflows/main.yml`), so the `win32` describe gets exercised against real `cmd.exe` quoting behavior there — a more authentic reproduction than simulating the platform.

## How Has This Been Tested?

- [x] Locally (this sandbox is Linux): the `posix` describe runs and passes; the `win32` describe is skipped — expected, since this sandbox isn't Windows.
- [x] Confirmed the `win32` snapshot was captured for the right reason: temporarily forced `process.platform` to `'win32'` locally, scoped to just this test file, to populate the inline snapshot from the actual (buggy) `exec()` call — `"prettier --write 'D:\...\types.ts'"`, the same single-quote wrapping seen in the real Windows CI failure.
- [x] Confirmed this checkpoint on real `windows-latest` CI: the job fails exactly on `lifecycleHooks - win32 > does not wrap a path argument in POSIX shell quoting` while every `ubuntu-latest` leg (running the `posix` describe) passes.

**Test Environment**:

- Local (Linux) for the `posix` describe and the temporary-simulation sanity check above; real `windows-latest` CI for the `win32` verification
- NodeJS: 24

## Checklist:

- [ ] Everything else: N/A for this checkpoint-only PR

## Further comments

Leave this failing — don't merge it as its own fix. The actual fix is a platform-conditional quoting change in `hooks.ts` (chosen approach: quote per `process.platform`/`os.platform()`, avoiding a new runtime dependency on bash being installed on end-user machines), now on #10966 (stacked on this branch). Once that lands, the `windows-latest` unit-test job should go green — the `FIXME:` comment above the `win32` snapshot states exactly what the updated snapshot should be.

One nuance worth flagging: the `win32` test's assertion uses `toMatchInlineSnapshot` with a snapshot captured *from the buggy output itself* (per earlier review feedback) — so this checkpoint reads as **passing** locally and on `ubuntu-latest`, not failing; it documents the bug via the snapshot + `FIXME:` comment rather than via a failing assertion. It genuinely fails on the real `windows-latest` CI leg for that reason above.


packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    Object.defineProperty(process, 'platform', { value: originalPlatform });
+  });
+
+  it('does not wrap a Windows-shaped path argument in POSIX shell quoting on win32', async () => {
+    // Forcing `process.platform` here means this test reproduces (and will later verify
+    // the fix for) Windows-only behavior regardless of which OS actually runs it --
+    // this sandbox has no windows-latest runner available, and the real bug was only
+    // ever observed on GitHub's windows-latest: this is the same mechanism, exercised
+    // deterministically instead of depending on that specific CI environment.
+    Object.defineProperty(process, 'platform', { value: 'win32' });
```
eddeee888: Split this into 2 describes:

```
describe('lifecycleHooks - win32', () => {}
describe('lifecycleHooks - <replace with the other one>', () => {}
```

And use `os.platform()` check to conditionally run the appropriate `describe`

eddeee888: Done — split into `describe('lifecycleHooks - win32', ...)` and `describe('lifecycleHooks - posix', ...)`, each gated with `describe.skipIf(...)` on the real `platform()` from `os` (not a simulated/forced one).

This is actually better than what I had: `packages/graphql-codegen-cli`'s unit tests already run on a real `windows-latest` CI runner (the `test` job's matrix), so the `win32` describe now gets exercised against real `cmd.exe` quoting behavior there instead of a simulation. Also added a `posix`-side assertion locking in the existing correct quoting behavior, so a platform-conditional fix for `win32` can't regress it.

Pushed as `9b05ae6`.


---

packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+      'D:\\a\\graphql-code-generator\\graphql-code-generator\\dev-test\\general\\modules\\types.ts';
+
+    await lifecycleHooks({ afterAllFileWrite: ['prettier --write'] }).afterAllFileWrite([
+      windowsPath,
+    ]);
+
+    expect(exec).toHaveBeenCalledTimes(1);
+    const [cmd] = vi.mocked(exec).mock.calls[0];
+
+    expect(cmd).not.toContain(`'${windowsPath}'`);
```
eddeee888: Use `toMatchInlineSnapshot` for both tests

eddeee888: Done — both tests now use `toMatchInlineSnapshot` on the full command string.

The `win32` snapshot intentionally locks in the current *buggy* output (path wrapped in POSIX single quotes) — that's what keeps this checkpoint red on real `windows-latest` CI. Added a `FIXME:` comment above it noting exactly what the snapshot should become once the platform-conditional fix lands (no surrounding quotes), so whoever updates it doesn't have to guess.

Pushed as `a081f16`.


---

packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+}));
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
+
+// packages/graphql-codegen-cli's unit tests run on a real `windows-latest` CI runner
+// (see the `test` job's matrix in .github/workflows/main.yml), so gating on the actual
+// `platform()` here exercises real `cmd.exe` quoting behavior there instead of
+// simulating it.
```
eddeee888: Remove, unnecessary comment

packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { exec } from 'child_process';
+import { platform } from 'os';
+import { lifecycleHooks } from '../src/hooks.js';
+
+vi.mock('child_process', () => ({
+  exec: vi.fn((cmd: string, _options: unknown, callback: (...args: any[]) => void) => {
+    callback(null, '', '');
+  }),
+}));
+
```
eddeee888, line 10: Do a `mockExec = vi.mocked(exec)` and use it below

packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+  });
+});
+
+describe.skipIf(platform() === 'win32')('lifecycleHooks - posix', () => {
+  it('wraps a path argument containing special characters in POSIX shell quoting', async () => {
+    // Locks in the existing, correct behavior on POSIX platforms, so a platform-
+    // conditional fix for the win32 case above can't accidentally regress this.
+    // `shell-quote`'s `quote()` only wraps arguments that actually need it -- a space
+    // is enough to trigger that on POSIX, the same way a backslash does for win32 above.
```
eddeee888: Unncessary comment

packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+describe.skipIf(platform() !== 'win32')('lifecycleHooks - win32', () => {
+  it('does not wrap a path argument in POSIX shell quoting', async () => {
+    // A path shaped like what Windows actually reports (backslash-separated, drive
+    // letter). `shell-quote`'s `quote()` wraps any argument containing a backslash in
+    // single quotes, unconditional of platform -- but `child_process.exec`'s default
+    // shell on Windows (cmd.exe) does not strip single quotes, so the hook script
+    // receives the literal quote characters baked into its argument. This is exactly
+    // what breaks `dev-test/general`'s `hooks: { afterAllFileWrite: ['prettier --write'] }`
+    // on `windows-latest` CI:
+    // https://github.com/dotansimha/graphql-code-generator/actions/runs/35432724450/job/105870054876
```
eddeee888: Bro. Make this concise

packages/graphql-codegen-cli/tests/hooks.spec.ts
eddeee888: Removed. Pushed as `611390a`.


---

packages/graphql-codegen-cli/tests/hooks.spec.ts
eddeee888: Done — `const mockExec = vi.mocked(exec)` right after the mock, used at both call sites. Pushed as `611390a`.


---

packages/graphql-codegen-cli/tests/hooks.spec.ts
eddeee888: Removed. Pushed as `611390a`.


---

packages/graphql-codegen-cli/tests/hooks.spec.ts
eddeee888: Trimmed to one line: "cmd.exe (Windows' default shell) doesn't strip POSIX single quotes." Pushed as `611390a`.


---

packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    // cmd.exe (Windows' default shell) doesn't strip POSIX single quotes.
+    const windowsPath =
+      'D:\\a\\graphql-code-generator\\graphql-code-generator\\dev-test\\general\\modules\\types.ts';
+
+    await lifecycleHooks({ afterAllFileWrite: ['prettier --write'] }).afterAllFileWrite([
+      windowsPath,
+    ]);
+
+    expect(mockExec).toHaveBeenCalledTimes(1);
+    const [cmd] = mockExec.mock.calls[0];
```
eddeee888: Just do `mockExec.mock.calls[0][0]` here and below

packages/graphql-codegen-cli/tests/hooks.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
+
+describe.skipIf(platform() !== 'win32')('lifecycleHooks - win32', () => {
+  it('does not wrap a path argument in POSIX shell quoting', async () => {
+    // cmd.exe (Windows' default shell) doesn't strip POSIX single quotes.
+    const windowsPath =
+      'D:\\a\\graphql-code-generator\\graphql-code-generator\\dev-test\\general\\modules\\types.ts';
```
eddeee888: - Bring this inline. Same for the other test.
- Expand this comment a bit:
```
// cmd.exe (Windows' default shell) doesn't strip POSIX single quotes.
// <insert why that's important>
```

packages/graphql-codegen-cli/tests/hooks.spec.ts
eddeee888: Done — `mockExec.mock.calls[0][0]` used directly in both assertions, no more destructured `cmd`. Pushed as `212348e`.


---

packages/graphql-codegen-cli/tests/hooks.spec.ts
eddeee888: Done — both path strings are inlined directly into `afterAllFileWrite([...])`, and the comment now explains why the quote-stripping gap matters:

> cmd.exe (Windows' default shell) doesn't strip POSIX single quotes, so a quoted path is passed to the hook script with the quote characters still attached, and it fails to find the file at that (now-wrong) path.

Pushed as `212348e`.


---

packages/graphql-codegen-cli/src/hooks.ts
```diff
@@ -15,6 +15,18 @@ const DEFAULT_HOOKS: Types.LifecycleHooksDefinition = {
   beforeAllFileWrite: [],
 };
 
+// shell-quote's `quote()` always produces POSIX-style single-quoting, but
+// `execShellCommand` below runs through `cmd.exe` on Windows (Node's
+// `child_process.exec` default shell there), which doesn't understand single
+// quotes and leaves them in literally. Quote per-platform instead: cmd.exe's
+// own convention is double-quoting only the arguments that need it.
+function quoteShellArgs(args: string[]): string {
+  if (process.platform === 'win32') {
```
eddeee888: Use `platform`  from `os`

eddeee888: Done — imports `platform` from `os` and uses `platform() === 'win32'`, matching how `hooks.spec.ts` already checks it. Pushed as `7fa05a6`.


---

PR #10966, opened 2026-09-19
[graphql-codegen-cli] fix: quote lifecycle hook arguments per-platform on Windows

eddeee888:
## Description

This PR fixes lifecycle hook scripts (e.g. `hooks: { afterAllFileWrite: ['prettier --write'] } }`) failing on Windows whenever the file paths passed to them contain a backslash or other POSIX shell-special character — which, on Windows, is essentially always (every path has backslashes).

Root cause: in `packages/graphql-codegen-cli/src/hooks.ts`, `executeHooks()` quotes hook-script arguments via the `shell-quote` package's `quote()`, which produces POSIX-style quoting unconditionally (wraps any argument containing a backslash — i.e. any Windows path — in single quotes). That command string is handed to `execShellCommand()`, which calls Node's `child_process.exec()` with no explicit `shell` option, so on Windows it defaults to `cmd.exe`. `cmd.exe` doesn't strip single quotes, so any lifecycle hook receiving a Windows-shaped path argument gets the literal quote characters baked into it, and fails to find the file at that (now-wrong) path.

This PR quotes per `os.platform()` instead: POSIX quoting (`shell-quote`'s `quote()`) is unchanged everywhere else, and on `win32`, arguments are wrapped in double quotes — `cmd.exe`'s own convention — only when they actually contain a character that needs it.

Related # (issue): none filed — Relates to #10965 (the `eddeee888:oss:issue-verify` checkpoint this PR builds directly on top of) and #10957 (where the bug was first surfaced).

## Type of change

## What

- `packages/graphql-codegen-cli/src/hooks.ts`: new `quoteShellArgs()` helper — POSIX quoting (`quote(args)`) unchanged on non-`win32`; on `win32`, double-quotes only the arguments that contain a character needing it (space, `&`, `|`, `<`, `>`, `^`, `"`), matching `cmd.exe`'s own quoting convention instead of a POSIX one it doesn't understand.
- `packages/graphql-codegen-cli/tests/hooks.spec.ts`: the `win32` checkpoint test (from #10965's `eddeee888:oss:issue-verify` commit) now asserts the corrected, unquoted output instead of the previously-locked-in buggy one; gains a second case covering a path that does need quoting (contains a space), to verify the new double-quoting branch actually engages when needed.
- `.changeset/cli-hooks-windows-shell-quoting.md`: patch changeset for `@graphql-codegen/cli`.

## How Has This Been Tested?

- [x] Locally, scoped to `hooks.spec.ts`, with `process.platform` temporarily forced to `'win32'`: re-ran the checkpoint test against the *unfixed* code first to confirm it currently passes with the old buggy snapshot (see #10965 for why that's expected — the snapshot itself locks in the buggy value), then against the fix — the test correctly starts failing (snapshot mismatch: actual output no longer has the surrounding quotes), confirming the fix changes behavior. Updated the snapshot to the new, correct value.
- [x] Added and populated a second `win32` test case (a path containing a space) the same way, confirming the new double-quote-wrapping branch produces `"C:\Program Files\...\types.ts"` correctly.
- [x] `pnpm --filter="@graphql-codegen/cli" test hooks.spec`: 1 passed, 2 skipped locally (both `win32` cases correctly skip on this Linux sandbox; `posix` passes).
- [x] `pnpm --filter="@graphql-codegen/cli" test` (full package suite): same result as on unmodified `master` — 172 passed, 5 failed (pre-existing, unrelated `init.spec.ts` terminal-glyph snapshot differences, confirmed present with this fix's changes stashed out), 4 skipped. No new failures.
- [x] Confirmed on real `windows-latest` CI (`Unit Test on Node 22 (windows-latest) and GraphQL v16`): `hooks.spec.ts (3 tests | 1 skipped)`, both `win32` cases pass. Full suite on that run: 903 passed, 5 skipped, 0 failed.

**Test Environment**:

- Local (Linux) plus temporary `process.platform` forcing, scoped to just this test file, for the `win32` verification above; real `windows-latest` CI for the final confirmation
- NodeJS: 24

## Checklist:

- [ ] Everything else: N/A for this fix

## Further comments

Stacked on #10965 (the `eddeee888:oss:issue-verify` checkpoint) to prove the fix directly: the checkpoint test, passing-but-buggy on #10965 alone (see that PR's "Further comments" for the `toMatchInlineSnapshot` nuance), asserts the genuinely correct behavior once this fix lands on top — confirmed on real Windows CI above, not just reasoning.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P2vvyEtEqahXy3gCBAJBTT)_


packages/graphql-codegen-cli/src/hooks.ts
```diff
@@ -15,6 +16,18 @@ const DEFAULT_HOOKS: Types.LifecycleHooksDefinition = {
...
 };
 
+// shell-quote's `quote()` always produces POSIX-style single-quoting, but
+// `execShellCommand` below runs through `cmd.exe` on Windows (Node's
+// `child_process.exec` default shell there), which doesn't understand single
+// quotes and leaves them in literally. Quote per-platform instead: cmd.exe's
+// own convention is double-quoting only the arguments that need it.
+function quoteShellArgs(args: string[]): string {
+  if (platform() === 'win32') {
+    return args.map(arg => (/[\s&|<>^"]/.test(arg) ? `"${arg}"` : arg)).join(' ');
```
eddeee888, line 30: Put up to 3 lines of code comment for this for what this regex is doing.
With example scenarios

eddeee888: Done — 3-line comment with example scenarios:

```
// Only wrap an argument that contains a character cmd.exe treats specially --
// whitespace, pipe/redirection, caret, or a literal quote. E.g. `C:\Program
// Files\x.ts` becomes `"C:\Program Files\x.ts"`, but `D:\a\x.ts` is left bare.
```

Pushed as `49e4201`.


---

packages/graphql-codegen-cli/src/hooks.ts
```diff
@@ -15,6 +16,21 @@ const DEFAULT_HOOKS: Types.LifecycleHooksDefinition = {
...
+// shell-quote's `quote()` always produces POSIX-style single-quoting, but
+// `execShellCommand` below runs through `cmd.exe` on Windows (Node's
+// `child_process.exec` default shell there), which doesn't understand single
+// quotes and leaves them in literally. Quote per-platform instead: cmd.exe's
+// own convention is double-quoting only the arguments that need it.
+function quoteShellArgs(args: string[]): string {
+  if (platform() === 'win32') {
+    // Only wrap an argument that contains a character cmd.exe treats specially --
+    // whitespace, pipe/redirection, caret, or a literal quote. E.g. `C:\Program
+    // Files\x.ts` becomes `"C:\Program Files\x.ts"`, but `D:\a\x.ts` is left bare.
```
eddeee888:
```suggestion
    // Only wrap an argument that contains a character cmd.exe treats specially --
    // whitespace, pipe/redirection, caret, or a literal quote. e.g. 
    // - `C:\Program Files\x.ts` becomes `"C:\Program Files\x.ts"`
    // - `D:\a\x.ts` is left bare.
```

eddeee888: Applied as suggested. Pushed as `92e9833`.


---

PR #10955, opened 2026-09-18
Reorganize dev-test-apollo-tooling into dev-test directory

eddeee888:
## Description

This PR reorganizes the `dev-test-apollo-tooling` directory structure by moving it into the `dev-test` directory as `dev-test/apollo-tooling`. This consolidates all dev-test packages under a single parent directory for better organization and consistency.

The changes include:
- Moving `dev-test-apollo-tooling/` → `dev-test/apollo-tooling/`
- Updating workspace configuration to reflect the new structure
- Updating relative import paths in vitest config
- Updating prettier ignore patterns

Related # (issue)

## Type of change

- [x] This change requires a documentation update (workspace structure)

## How Has This Been Tested?

The workspace configuration has been updated to include the new path under `dev-test/*`, which will be picked up by the existing monorepo tooling. The vitest configuration has been updated with correct relative paths to the shared config.

**Test Environment**:
- Workspace configuration validation
- Path resolution in vitest config

## Checklist:

- [x] I have followed the CONTRIBUTING doc and the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

This is a straightforward directory reorganization to improve project structure consistency. All dev-test packages are now co-located under the `dev-test/` directory, making the monorepo layout more intuitive and maintainable.

https://claude.ai/code/session_01AFw7gorux9HixUPKqgNFrc

PR #10969, opened 2026-09-19
test: reproduce ESM plugin loader failing to resolve from consumer project

eddeee888:
## Description

This PR is a checkpoint: it adds two real, end-to-end failing CI checks that reproduce a bug in `makeDefaultLoader`'s ESM branch (`packages/graphql-codegen-cli/src/codegen.ts`), split out of #10956 so the repro and the fix can be reviewed separately. No synthetic/unit test — both checkpoints fail for real in `dev-tests`.

`makeDefaultLoader`'s ESM branch hands the bare module specifier straight to `import(mod)` instead of resolving it relative to the consuming project first — the way the CJS branch already does via `relativeRequire.resolve(mod)`. **A plugin/preset only loads if it happens to be reachable from the CLI package's own `node_modules`, not guaranteed for a real consumer project's dependency layout.**

Two devDependencies were added to the CLI package purely as workarounds for this bug — this PR removes each one to prove it, against two different dependency shapes:

- `@graphql-codegen/near-operation-file-preset` (pinned external version, added in #10955) — `dev-test/apollo-tooling` can no longer find its own preset.
- `@graphql-codegen/typescript-resolvers` (`workspace:*` link) — `dev-test/general` can no longer find it either, across most of its generate targets (picked over the more obvious `@graphql-codegen/typescript` since that one is also `require()`'d directly in this package's own `codegen.config.externalDocuments.spec.ts`, which would've muddied the checkpoint with an unrelated test failure).

This PR's own `chore(dev-test): wire apollo-tooling into dev-tests generate checks` commit (needed so the first point above actually fails in CI, rather than being silently skipped) started out split into #10971 too — but that PR has since narrowed to just a stale-output refresh, based on `master` directly, so it no longer shares content with this branch. The CI-wiring stays here.

## Type of change

- [x] This change requires a documentation update — N/A (checkpoint only, no fix)

## How Has This Been Tested?

- [ ] `dev-tests` (esm) — intentionally failing, reproduces the bug for real: `pnpm run dev-test:generate:esm` now fails for `dev-test/apollo-tooling` with `Unable to find preset matching 'near-operation-file'` and for `dev-test/general` with `Unable to find template plugin matching 'typescript-resolvers'`.

## Checklist:

## Further comments

Stacked on #10957 to keep this on the same base as #10956, which now builds on top of this PR as the fix.


eddeee888:
Closing as superseded — #10956 was rebased directly onto #10957's current tip and now carries the same checkpoint commits (`chore(dev-test): wire apollo-tooling into dev-tests generate checks`, both devDependency-removal checkpoints) followed by the fix and changeset, all in one line. Merging #10957 → #10956 gets everything this PR had, so there's no need to keep this open as a separate, now out-of-sync branch.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SV6N43kNobK5CczJHujxT3)_

---

Issue #10970, opened 2026-09-19
near-operation-file-preset: "Unable to find parent type" for dev-test/apollo-tooling's Component.ts


eddeee888:
## Context

While wiring `dev-test/apollo-tooling` into the `dev-tests` CI generate checks (#10969 / #10956, splitting an ESM plugin-loader fix into verify + fix PRs), this package's `start`/`generate:esm` script had apparently never been run in CI before — it had no `generate`/`generate:esm` script, so `pnpm --filter="dev-test*" generate:esm` silently skipped it.

## Problem

Once wired in and run for real (`pnpm --filter=dev-test-apollo-tooling generate:esm`, i.e. `tsx cli/index.ts`), codegen fails on `src/Component.ts`:

```
Error: Unable to validate GraphQL document!

    File .../dev-test/apollo-tooling/src/Component.ts caused error:
    Unable to find parent type! Please make sure you operation passes validation

    at .../node_modules/@graphql-codegen/near-operation-file-preset/cjs/resolve-document-imports.js:68:19
    at Array.map (<anonymous>)
    at resolveDocumentImports (.../resolve-document-imports.js:22:10)
    at Object.buildGeneratesSection (.../near-operation-file-preset/cjs/index.js:32:82)
```

This reproduces consistently and is unrelated to the ESM loader bug #10956 fixes — it only surfaced because that fix (or, equivalently, temporarily removing the CLI's `near-operation-file-preset` devDependency workaround) is the first thing that's ever actually exercised this codegen run for real.

## What's been ruled out

Checked whether this is a duplicate-`graphql`-package-instance issue (a common cause of confusing "type not found" errors in the GraphQL ecosystem) — it isn't: the whole monorepo resolves to a single `graphql@17.0.2` in the pnpm store (`ls node_modules/.pnpm | grep '^graphql@'` returns exactly one entry). So the root cause is still open.

## Reproduction

```bash
pnpm build
pnpm --filter=dev-test-apollo-tooling start
```

(as of this writing, on `master`, since `dev-test/apollo-tooling` isn't yet part of `dev-test:generate:esm`)

## Suggested next step

Root-cause why `near-operation-file-preset`'s `resolveDocumentImports` can't find the parent type for `Component.ts`'s operations against `dev-test/apollo-tooling/schema.graphql` — likely worth a minimal repro isolating just that preset + that schema/document pair outside the full CLI pipeline.

eddeee888:
[...]

PR #10973, opened 2026-09-19
[dev-test-apollo-tooling] fix: run under node instead of tsx, and wire into the dev-tests CI job (#10970)

eddeee888:
## Description

Relates to #10970. Also closes the dev-tests coverage gap noted in #10969.

This PR fixes `dev-test/apollo-tooling` by running it under plain `node` instead of `tsx`, and wires the package into the `dev-tests` CI job that had been silently skipping it. #10970 **is not a codegen bug** — it is a `tsx` dual-package hazard — so nothing in this repo's source changes, only how this dev-test package runs itself.

### Why #10970 happened

Under `tsx`, `import 'graphql'` resolves to `graphql/index.mjs` while a CJS `require('graphql')` resolves to `graphql/index.js`, so **two distinct `graphql@17.0.2` module instances coexist in one process**. Plain Node dedupes those two entry points via the `module-sync` export condition; tsx does not.

The CJS-only `near-operation-file-preset` then calls `isUsingTypes()` from its own CJS copy of `@graphql-codegen/plugin-helpers`, against a schema built by the ESM instance:

1. `isNonNullType()` is an `instanceof` check, so it returns `false` for a genuine `GraphQLNonNull` from the other realm.
2. `getBaseType()` therefore returns the still-wrapped type, whose `.name` is `undefined`.
3. `schema.getType(undefined)` yields `undefined`, which lands on the visitor's type stack.
4. The next nested selection set reads that entry and throws `Unable to find parent type!`.

Instrumented output from the failing run, confirming steps 1–3 on `Query.superUser`:

```
field= superUser | field.type= SuperUser! | ctor= GraphQLNonNull
| isNonNull(cjs)= false | baseType.name= undefined | getType(baseName)= undefined
```

The duplicate-`graphql` theory in #10970 was right in spirit but checked the wrong level: there is exactly one `graphql@17.0.2` in the pnpm store, but **that single package ships both a CJS and an ESM entry point**, and tsx loads both.

Two corrections to the issue's framing, both confirmed by running it:

- Not specific to `Component.ts` or its external fragment — that file is **simply first in glob order**, and the `HelperFields` fragment is already inlined by the time the preset runs (`externalFragments` is empty). Any query nested two or more levels deep hits it.
- It never blocked wiring the package into CI — the job runs `node .../dist/{cjs,esm}/bin.js`, and **both lanes pass**.

### What changed

- `start`, `start:debug` and `start:verbose` run `node cli/index.ts` instead of `tsx cli/index.ts`. Node 22.18+ strips TypeScript natively, so **tsx bought nothing here and was the sole cause of #10970**; it is dropped from `devDependencies`.
- Adds `codegen.ts` plus `generate`, `generate:{cjs,esm}` and `watch*` scripts, matching how the rest of `dev-test/*` is wired. The package **previously had no `generate` script**, so `pnpm --filter="dev-test*" generate:esm` silently skipped it.
- Adds `cli/config.ts` holding `buildCodegenConfig()`, shared by the programmatic `start` entry point and the new config file so the two **cannot drift apart**. `codegen.ts` no longer reaches through `cli/index.ts` to get the config.
- Relative imports within the package now carry their real `.ts` extension, with `allowImportingTsExtensions` + `noEmit` added to its tsconfig. **Node's native type stripping, unlike jiti or tsx, does not rewrite `./x.js` to `./x.ts`**, so `node cli/index.ts` cannot resolve a `./config.js` specifier.
- Adds the missing `@graphql-codegen/add` dependency — the config uses the `add` plugin but **only ever resolved it transitively through the preset**, which #10956's resolution fix exposed.

No changeset: only a private `dev-test` package changes, so nothing ships.

## Type of change

## How Has This Been Tested?

- [x] dev-tests generate checks, cjs and esm — apollo-tooling is now genuinely included rather than skipped, and the **regenerated output is byte-identical to what is committed**, so the job's `git diff --exit-code` assertion passes
- [x] Lint, Prettier and typecheck — no new `tsc` errors; **the four it does report are all pre-existing on `master`** (`src/Component.ts`, `vitest.config.ts`, `@wry/caches`)
- [x] Frozen-lockfile install — passes without rewriting the lockfile, and **the lockfile diff is 6 lines**, confined to this package's importer block
- [x] `pnpm --filter=dev-test-apollo-tooling start` → exit 0, **run manually because no CI job covers the `start` script** (it previously failed with `Unable to find parent type`)

**Test Environment**:

- OS: Linux
- `@graphql-codegen/cli`: `7.4.1` (workspace), `@graphql-codegen/near-operation-file-preset`: `5.2.1` (npm), `graphql`: `17.0.2`
- NodeJS: v22.22.2

## Further comments

Worth noting for anyone who hits this class of error again: `resolveDocumentImports` wraps the inner error and drops its stack, which is most of why this was hard to root-cause from the report alone. The same tsx realm split would affect any ESM codegen config run through tsx that loads a CJS plugin or preset, so it is worth knowing about beyond this package.

Also spotted while here. **Not fixed in this PR.** `dev-test/apollo-tooling/tests/Component.test.ts` asserts `roleType: UserManagerRoleType;` while the committed generated output uses a comma separator (changed in #10971). Nothing runs that suite today — the root `vitest.config.mts` projects glob covers `packages/**` and `examples/**` but not `dev-test/**` — so it is latent rather than failing.

https://claude.ai/code/session_012mFLsRYRvZveLKSAuDAyru


dev-test/apollo-tooling/cli/index.ts
```diff
@@ -30,9 +30,14 @@ const GRAPHQL_CODEGEN_CONFIG = {
...
-  const cwd = process.cwd();
-
+/**
+ * Shared by both entry points into this package's codegen:
+ * - `main()` below, i.e. the programmatic `start` script, which is what the real
+ *   apollo-tooling setup this package mimics does.
+ * - `codegen.ts`, which the `generate:{cjs,esm}` scripts feed to the CLI binary so
+ *   the dev-tests CI job covers this package the same way it covers the others.
+ */
+export const buildCodegenConfig = (cwd: string): Types.Config => {
```
eddeee888: Extract buildCodegenConfig fn to another file to be shared

eddeee888: Done in 33cf865e5 — `buildCodegenConfig` (and the constants it uses) now live in `dev-test/apollo-tooling/cli/config.ts`, imported by both `cli/index.ts` and `codegen.ts`.

A side benefit: `codegen.ts` no longer reaches through `cli/index.ts`, so feeding the config to the CLI binary no longer transitively imports `generate` from `@graphql-codegen/cli`. `cli/index.ts` is now just the programmatic `start` entry point.

One wrinkle worth flagging, since it shaped the change: relative imports in this package now carry their real `.ts` extension. Node's native type stripping — unlike jiti or tsx — does not rewrite `./x.js` to `./x.ts`, so `node cli/index.ts` failed to resolve a `./config.js` specifier. The package's `tsconfig.json` gains `allowImportingTsExtensions` with `noEmit` to match, which is accurate for a dev fixture that is only ever typechecked and run directly, never built.

Verified: `start`, `generate:cjs` and `generate:esm` all exit 0, both full `dev-test:generate:{cjs,esm}` sweeps pass, the regenerated output stays byte-identical to what is committed, and `tsc` reports no new errors.

---

